PR #12, opened 2024-10-28
Provisionally comment out --mount while deploying to Heroku

zerolab:
Temporarily disable BuildKit cache as Heroku doesn't support it

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,12 +32,16 @@
 
 FROM python:3.12-slim AS base
 
+ARG DOCKER_BUILDKIT
+
 WORKDIR /app
 
 # Install common OS-level dependencies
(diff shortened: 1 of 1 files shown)


ababic:
@zerolab Are there any implications for local dev here?

zerolab:
@ababic rebuilds take longer. No difference from the Alpha approach

ababic (approved):

RealOrangeOne (approved):

zerolab:
erm, Jake, do ignore that last request. I removed dissemination/cms from reviewers and instead it logged that I explicitly requested you 🤦🏼

MebinAbraham (approved):

PR #13, opened 2024-10-28
Further Dockerfile tweaks for Heroku build

zerolab:
Follow up to #12. It turns out Heroku doesn't support HEREDOC either

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,18 +32,17 @@
 
 FROM python:3.12-slim AS base
 
-ARG DOCKER_BUILDKIT
-
 WORKDIR /app
 
 # Install common OS-level dependencies
-# TODO: when moving to ONS infrastructure, replace RUN with
+# TODO: when moving to ONS infrastructure, replace RUN
 # RUN --mount=type=cache,target=/var/lib/apt/lists,sharing=locked \
 #     --mount=type=cache,target=/var/cache/apt,sharing=locked \
 #     <<<EOF
-RUN <<EOF
-    apt --quiet --yes update
-    apt --quiet --yes install --no-install-recommends \
+#     commands
+# EOF
+RUN apt --quiet --yes update \
+    && apt --quiet --yes install --no-install-recommends \
         build-essential \
         curl \
         libpq-dev \
@@ -52,7 +51,6 @@ RUN <<EOF
         unzip \
         gettext \
     && apt --quiet --yes autoremove
-EOF
 
 
 # Create an unprivileged user and virtual environment for the app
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -52,33 +51,37 @@ RUN <<EOF
...
-    # Create the unprivileged user and group. If you have issues with file
-    # ownership, you may need to adjust the UID and GID build args to match your
-    # local user.
-    groupadd --gid $GID $USERNAME
-    useradd --gid $GID --uid $UID --create-home $USERNAME
-    python -m venv --upgrade-deps $VIRTUAL_ENV
-    chown -R $UID:$GID /app $VIRTUAL_ENV
-EOF
+# TODO: when moving to ONS infrastructure, replace the RUN command with
+# RUN <<<OEF
```
ababic, line 62: Do we have a superfluous `<` here?

Dockerfile
```diff
@@ -32,18 +32,17 @@
 
 FROM python:3.12-slim AS base
 
-ARG DOCKER_BUILDKIT
-
 WORKDIR /app
 
 # Install common OS-level dependencies
-# TODO: when moving to ONS infrastructure, replace RUN with
+# TODO: when moving to ONS infrastructure, replace RUN
```
ababic, line 38: Was this intentional?

Dockerfile
```diff
@@ -103,11 +107,11 @@ USER $USERNAME
 COPY pyproject.toml poetry.lock ./
 # TODO: when moving to ONS infrastructure, replace RUN with
 # RUN --mount=type=cache,target=/home/$USERNAME/.cache/,uid=$UID,gid=$GID \
-#     <<EOF
-RUN <<EOF
-    # Install the production dependencies
-    poetry install --no-root --without dev
-EOF
+#     <<< EOF
```
ababic, line 110: Do we have a superfluous `<` here?

Dockerfile
RealOrangeOne:
```suggestion
# RUN <<EOF
```

Dockerfile
RealOrangeOne:
```suggestion
#     << EOF
```

Dockerfile
zerolab: it was, but on reflection I should revert it

Dockerfile
zerolab: Old habits die hard. 🦅 👁️

zerolab:
@ababic @RealOrangeOne I've addressed your queries/suggestions

ababic (approved):

MebinAbraham (approved):

PR #15, opened 2024-10-31
Add support for RDS DB with read replica

RealOrangeOne:

- https://jira.ons.gov.uk/browse/CMS-214
- https://jira.ons.gov.uk/browse/CMS-217

#### IAM authentication

In production, rather than a username and password, authentication to the database will be done using IAM. This means there are no credentials to leak or rotate. 

When a connection is opened, a signed token is generated and used for authentication. This token is only valid for 15 minutes. The complexities of this are handled by an external library.

Connecting the application to IAM itself is done by the cluster, hence no additional configuration. 

#### Multiple database connections

In production, multiple database connections will be used: one for writes and one for reads. The write connection points to a single database server, whilst reads are load balanced between a set of servers ("read replicas").

It's up to the application to decide whether the write or read connections is used, which is what the database router handles. Django internally identifies whether a query is a "read" or a "write", and uses the router to identify the connection. Most of the time, there's no need for a developer to worry about this.

Specifically:

- Writes should always be handled by the write server (read replica connections are read-only)
- Reads should _usually_ be handled by the read replicas
- If there is an open transaction, reads should be handled by the write connection. This ensures that whilst a transaction is open, the application gets a consistent view of state (since data held in the transaction can't be seen by the read connection).

Note: AWS is responsible for balancing traffic across the replicas - this is not the responsibility of the application.

For consistency, local development also uses multiple connections handled by the same router, however these point to the same server, and there's no restriction that the read connection only handles reads. 

### How to review

This shouldn't make a difference to local development.

Relevant Django documentation: https://docs.djangoproject.com/en/5.1/topics/db/multi-db/

### Follow-up Actions

There are discussions happening about being able to handle a completely read-only connection. "read replica" functionality is separate.

### DIFF
--- a/cms/core/db_router.py
+++ b/cms/core/db_router.py
@@ -1,3 +1,5 @@
+from typing import Any, Optional
+
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.db import DEFAULT_DB_ALIAS, router, transaction
@@ -18,7 +20,7 @@ class ReadReplicaRouter:  # pylint: disable=unused-argument,protected-access
         if self.REPLICA_DB_ALIAS not in settings.DATABASES:  # pragma: no cover
             raise ImproperlyConfigured("Read replica is not configured.")
 
-    def db_for_read(self, model: type[Model], **hints: dict) -> str | None:
+    def db_for_read(self, model: type[Model], **hints: Any) -> Optional[str]:
         """Which database should be used for read queries?"""
         # If the write database is in a (uncommitted) transaction,
         # a subsequent SELECT (or other read query) may return inconsistent data.
@@ -32,13 +34,13 @@ class ReadReplicaRouter:  # pylint: disable=unused-argument,protected-access
 
         return self.REPLICA_DB_ALIAS
 
-    def db_for_write(self, model: type[Model], **hints: dict) -> str | None:
+    def db_for_write(self, model: type[Model], **hints: Any) -> Optional[str]:
         """Which database should be used for write queries?"""
         # This should always be the "default" database, since the replica
         # doesn't allow writes.
         return DEFAULT_DB_ALIAS
 
-    def allow_relation(self, obj1: Model, obj2: Model, **hints: dict) -> bool | None:
+    def allow_relation(self, obj1: Model, obj2: Model, **hints: Any) -> Optional[bool]:
         """Are two objects allowed to be related?"""
         # If both instances are in the same database (or its replica), allow relations
         if obj1._state.db in self.REPLICA_DBS and obj2._state.db in self.REPLICA_DBS:
@@ -47,7 +49,7 @@ class ReadReplicaRouter:  # pylint: disable=unused-argument,protected-access
         # No preference
         return None
 
-    def allow_migrate(self, db: str, app_label: str, model_name: str | None = None, **hints: dict) -> bool:
+    def allow_migrate(self, db: str, app_label: str, model_name: Optional[str] = None, **hints: Any) -> bool:
         """Can migrations be run for the app on the database?"""
         # Don't allow migrations to run against the replica (they would fail anyway)
         return db != self.REPLICA_DB_ALIAS
--- a/cms/core/db_router.py
+++ b/cms/core/db_router.py
@@ -1,0 +1,54 @@
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+from django.db import DEFAULT_DB_ALIAS, router, transaction
+from django.db.models import Model
+
+
+class ReadReplicaRouter:  # pylint: disable=unused-argument,protected-access
+    """A database router which routes reads to the read replica."""
+
+    REPLICA_DB_ALIAS = "read_replica"
+
+    REPLICA_DBS = frozenset({REPLICA_DB_ALIAS, DEFAULT_DB_ALIAS})
+
+    def __init__(self) -> None:
+        if self.REPLICA_DB_ALIAS not in settings.DATABASES:
+            raise ImproperlyConfigured("Read replica is not configured.")
+
+    def db_for_read(self, model: type[Model], **hints: dict) -> str | None:
+        """Which database should be used for read queries?"""
+        # If the write database is in a (uncommitted) transaction,
+        # a subsequent SELECT (or other read query) may return inconsistent data.
+        # In this case, use the write connection for reads, with the aim of
+        # improved consistency.
+        write_db = router.db_for_write(model, **hints)
+
+        if not transaction.get_autocommit(using=write_db):
+            # In a transaction, use the write database
+            return write_db
+
+        return self.REPLICA_DB_ALIAS
+
+    def db_for_write(self, model: type[Model], **hints: dict) -> str | None:
+        """Which database should be used for write queries?"""
+        # This should always be the "default" database, since the replica
+        # doesn't allow writes.
+        return DEFAULT_DB_ALIAS
+
+    def allow_relation(self, obj1: Model, obj2: Model, **hints: dict) -> bool | None:
+        """Are two objects allowed to be related?"""
+        # If both instances are in the same database (or its replica), allow relations
+        if obj1._state.db in self.REPLICA_DBS and obj2._state.db in self.REPLICA_DBS:
+            return True
+
+        # No preference
+        return None
+
+    def allow_migrate(self, db: str, app_label: str, model_name: str | None = None, **hints: dict) -> bool | None:
+        """Can migrations be run for the app on the database?"""
+        # Don't allow migrations to run against the replica (they would fail anyway)
+        if db == self.REPLICA_DB_ALIAS:
+            return False
+
+        # No preference
+        return None
--- a/cms/core/db_router.py
+++ b/cms/core/db_router.py
@@ -24,7 +24,7 @@ class ReadReplicaRouter:  # pylint: disable=unused-argument,protected-access
             raise ImproperlyConfigured("Read replica is not configured.")
 
     def db_for_read(self, model: type[Model], **hints: Any) -> Optional[str]:
-        """Which database should be used for read queries?"""
+        """Determine which database should be used for read queries."""
         # If the write database is in a (uncommitted) transaction,
         # a subsequent SELECT (or other read query) may return inconsistent data.
         # In this case, use the write connection for reads, with the aim of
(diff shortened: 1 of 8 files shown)


cms/core/db_router.py
```diff
@@ -0,0 +1,50 @@
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+from django.db import DEFAULT_DB_ALIAS, router, transaction
+from django.db.models import Model
+
+
+class ReadReplicaRouter:  # pylint: disable=unused-argument,protected-access
```
zerolab, line 7: suggestion: link to the Django docs. e.g. https://docs.djangoproject.com/en/5.1/topics/db/multi-db/#automatic-database-routing

cms/settings/base.py
```diff
@@ -224,7 +254,7 @@
 # Search
 # https://docs.wagtail.io/en/latest/topics/search/backends.html
 
-WAGTAILSEARCH_BACKENDS = {"default": {"BACKEND": "wagtail.search.backends.database"}}
+WAGTAILSEARCH_BACKENDS = {"default": {"BACKEND": "cms.core.search.ONSPostgresSearchBackend"}}
```
zerolab, line 257:
```suggestion
# TODO: revert to using "wagtail.search.backends.database" when
# https://github.com/wagtail/wagtail/pull/12508 is fixed and released.
WAGTAILSEARCH_BACKENDS = {"default": {"BACKEND": "cms.core.search.ONSPostgresSearchBackend"}}
```

zerolab (approved):

cms/settings/base.py
```diff
@@ -1,9 +1,12 @@
 """Django settings for ons project."""
 
+# pylint: disable=fixme
```
MebinAbraham, line 3: I would disable this globally.

cms/settings/base.py
```diff
@@ -170,12 +173,41 @@
...
+    DATABASES = {
+        "default": cast(
+            dj_database_url.DBConfig,
+            {
+                "ENGINE": "django_iam_dbauth.aws.postgresql",
+                "NAME": env["PG_DB_DATABASE"],
+                "USER": env["PG_DB_USER"],
+                "HOST": env["PG_DB_ADDR"],
+                "PORT": env["PG_DB_PORT"],
+                "CONN_MAX_AGE": 870,  # Just under 15 minutes, to match password expiry
```
MebinAbraham, line 188: Nit: Let's make this a config with a default.

RealOrangeOne:
Putting this PR on hold whilst the `pytest` issues are resolved in a separate PR

AdamHawtin (commented):
When developing off this branch locally I'm hitting an issue when I try to migrate in the local docker container. 

When running `dj migrate` it errors out with

```
...
  File "/app/cms/core/db_router.py", line 19, in __init__
    raise ImproperlyConfigured("Read replica is not configured.")
django.core.exceptions.ImproperlyConfigured: Read replica is not configured.
```

Is there some change required in the dev or docker compose config?

cms/settings/base.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/15/commits/3c4ca21b9d3d3b2f288b219f8686cce81aacd78a

cms/settings/base.py
```diff
@@ -170,12 +173,41 @@
...
+    if "PG_DB_READ_ADDR" in env:
+        DATABASES["read_replica"] = {**deepcopy(DATABASES["default"]), "HOST": env["PG_DB_READ_ADDR"]}
+
+else:
+    DATABASES = {
+        "default": dj_database_url.config(conn_max_age=600, default="postgres:///ons"),
+    }
+
+    if "READ_REPLICA_DATABASE_URL" in env:
+        DATABASES["read_replica"] = dj_database_url.config(env="READ_REPLICA_DATABASE_URL", conn_max_age=600)
```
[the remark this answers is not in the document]
RealOrangeOne: Definitely https://github.com/ONSdigital/dis-wagtail/pull/15/commits/3c4ca21b9d3d3b2f288b219f8686cce81aacd78a

cms/settings/base.py
RealOrangeOne: Looks like someone beat me to it.

RealOrangeOne:
> When developing off this branch locally I'm hitting an issue when I try to migrate in the local docker container.
> 
> When running `dj migrate` it errors out with
> 
> ```
> ...
>   File "/app/cms/core/db_router.py", line 19, in __init__
>     raise ImproperlyConfigured("Read replica is not configured.")
> django.core.exceptions.ImproperlyConfigured: Read replica is not configured.
> ```
> 
> Is there some change required in the dev or docker compose config?

@AdamHawtin I've just rebased this against `main` - do you still get the error? There should always be a `read_replica` connection configured - this check is just a safety net.

cms/settings/base.py
```diff
@@ -179,12 +180,45 @@
...
+        "default": dj_database_url.config(conn_max_age=db_conn_max_age, default="postgres:///ons"),
+    }
+
+    if "READ_REPLICA_DATABASE_URL" in env:
+        DATABASES["read_replica"] = dj_database_url.config(
+            env="READ_REPLICA_DATABASE_URL", conn_max_age=db_conn_max_age
+        )
+
+if "read_replica" not in DATABASES:
+    DATABASES["read_replica"] = deepcopy(DATABASES["default"])
```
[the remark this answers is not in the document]
RealOrangeOne: We don't want a separate DB for the read replica, because reads need to go to the same database as writes.

I don't think there'll be any downsides, besides more database connection handling. Isolated connections to the same DB should be functionally the same as the production read replica setup anyway.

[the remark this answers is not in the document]
zerolab: Sounds good to me. I think what would help once the infra is set up is to have a page in the docs outlining the read/write replicas.

cms/core/db_router.py
```diff
@@ -0,0 +1,53 @@
...
+
+    REPLICA_DB_ALIAS = "read_replica"
+
+    REPLICA_DBS = frozenset({REPLICA_DB_ALIAS, DEFAULT_DB_ALIAS})
+
+    def __init__(self) -> None:
+        if self.REPLICA_DB_ALIAS not in settings.DATABASES:  # pragma: no cover
+            raise ImproperlyConfigured("Read replica is not configured.")
+
+    def db_for_read(self, model: type[Model], **hints: dict) -> str | None:
```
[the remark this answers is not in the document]
RealOrangeOne: As it turns out, `**kwargs` hints are intended to be the type of each object (eg `dict[str, T]`), so typing as `dict` isn't valid either. Had to use explicit `Any`.

cms/core/db_router.py
```diff
@@ -0,0 +1,55 @@
+from typing import Any, Optional
+
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+from django.db import DEFAULT_DB_ALIAS, router, transaction
+from django.db.models import Model
+
+
+class ReadReplicaRouter:  # pylint: disable=unused-argument,protected-access
+    """A database router which routes read queries to a read replica.
```
MebinAbraham, line 10:
```suggestion
    """A database router for directing queries between the default database and a read replica.

    This router routes read queries to a read replica while ensuring write queries
    and migrations are directed to the default database.
```

cms/core/db_router.py
```diff
@@ -0,0 +1,55 @@
...
+    REPLICA_DB_ALIAS = "read_replica"
+
+    REPLICA_DBS = frozenset({REPLICA_DB_ALIAS, DEFAULT_DB_ALIAS})
+
+    def __init__(self) -> None:
+        if self.REPLICA_DB_ALIAS not in settings.DATABASES:  # pragma: no cover
+            raise ImproperlyConfigured("Read replica is not configured.")
+
+    def db_for_read(self, model: type[Model], **hints: Any) -> Optional[str]:
+        """Which database should be used for read queries?"""
```
MebinAbraham, line 24: Let's make the doc strings a bit clearer.
```suggestion
        """Determines which database should be used for read operations.

        By default, read queries are routed to the read replica. However, if a transaction
        is active on the default database, reads are directed to the default database to
        ensure consistency.
```

cms/core/db_router.py
```diff
@@ -0,0 +1,55 @@
...
+        write_db = router.db_for_write(model, **hints)
+
+        if not transaction.get_autocommit(using=write_db):
+            # In a transaction, use the write database
+            return write_db
+
+        return self.REPLICA_DB_ALIAS
+
+    def db_for_write(self, model: type[Model], **hints: Any) -> Optional[str]:
+        """Which database should be used for write queries?"""
```
MebinAbraham, line 38:
```suggestion
        """Determines which database should be used for write operations.

        All write queries are routed to the default database as the read replica does not
        support write operations.
        """
```

cms/core/db_router.py
```diff
@@ -0,0 +1,55 @@
...
+        # This should always be the "default" database, since the replica
+        # doesn't allow writes.
+        return DEFAULT_DB_ALIAS
+
+    def allow_relation(self, obj1: Model, obj2: Model, **hints: Any) -> Optional[bool]:
+        """Are two objects allowed to be related?"""
```
MebinAbraham, line 44:
```suggestion
        """Determines if a relation is allowed between two objects.

        Relations are allowed if both objects belong to the same database or its replica.
        """
```

cms/core/db_router.py
```diff
@@ -0,0 +1,55 @@
...
+        # If both instances are in the same database (or its replica), allow relations
+        if obj1._state.db in self.REPLICA_DBS and obj2._state.db in self.REPLICA_DBS:
+            return True
+
+        # No preference
+        return None
+
+    def allow_migrate(self, db: str, app_label: str, model_name: Optional[str] = None, **hints: Any) -> bool:
+        """Can migrations be run for the app on the database?"""
```
MebinAbraham, line 53:
```suggestion
        """Determines if migrations are allowed on the specified database.

        Migrations are not allowed on the read replica, as it is intended for read-only operations.
        """
```

cms/core/wagtail_search.py
```diff
@@ -0,0 +1,12 @@
+from wagtail.search.backends.database.postgres.postgres import Index, PostgresSearchBackend
+
+
+class ONSPostgresSearchBackend(PostgresSearchBackend):
+    """A custom search backend which ensures the index uses the correct database backend.
+
+    A work-around until https://github.com/wagtail/wagtail/pull/12508 ships.
```
MebinAbraham, line 7: I suggest adding this as a ``TODO:`

MebinAbraham (commented):
Code looks fine, unable to test for 'real' atm.

cms/core/db_router.py
MebinAbraham: Yup, that's fine, but let's not use 'questions' for docstring as it won't be clearer to all levels. It should be descriptive or imperative style. So in this cause I would still suggest we use `Determines which database should be used for read operations.`

sanjeevz3009 (commented):
Reviewed =) Have no comments. Some of the code here overlapped with some things I was looking into to come up with the solution for the BDD UI test suite issues, particularly the code in test_db_router.py tests interestingly.

sanjeevz3009 (approved):
LGTM!

PR #16, opened 2024-11-01
Release calendar pages

zerolab:
https://jira.ons.gov.uk/browse/CMS-46 and https://jira.ons.gov.uk/browse/CMS-80

- Adds the `ContactDetails` snippet
- add the release calendar index and page model and tests
- plugs in the templates
- adds test coverage
- expands the core tests

Screencast: https://www.loom.com/share/5a7d8a9ca004448981526056e28ab5cd
Dev walkthrough: https://www.loom.com/share/39f7bde66cdc48b9872fa3d5cace22d6


cms/jinja2/templates/components/streamfield/release_date_change_block.html
```diff
@@ -0,0 +1,6 @@
+<li class="ons-list__item">
+    <h3 class="ons-u-mb-no ons-u-fs-r--b">{{ _("Previous date") }}</h3>
+    <p>{{ value.previous_date|date("j F Y g:ia") }}</p>
```
RealOrangeOne, line 3: Suggestion: Should this format be defined globally, since I suspect all will be shown in a standard way

cms/jinja2/templates/pages/release_calendar/release_calendar_page.html
```diff
@@ -0,0 +1,145 @@
...
+                <span class="ons-u-nowrap">{{ page.release_date|date("j F Y g:ia") or page.release_date_text }}</span>
+            </div>
+            {% if page.next_release_date or page.next_release_text %}
+                <div class="ons-grid__col ons-col-8@m ons-col-12@s ons-u-p-no@xxs@m">
+                    <span class="ons-u-fs-r--b">{{ _("Next release") }}:</span>
+                    <span class="ons-u-nowrap">{{ page.next_release_date|date("j F Y g:ia") or page.next_release_text }}</span>
+                </div>
+            {% endif %}
+
+            <p></p>{# added empty p for some margin #}
```
RealOrangeOne, line 34: Nit: This feels like a hack - is there a better way to add some margin? Perhaps with CSS or a `<br>`?

cms/jinja2/templates/pages/release_calendar/release_calendar_page.html
```diff
@@ -0,0 +1,145 @@
...
+        {% endif %}
+        {% block release_meta %}
+            <div class="ons-grid__col ons-grid__col--sticky@m ons-col-4@m ons-col-12@s ons-u-p-no">
+                <span class="ons-u-fs-r--b">{{ _("Release date") }}:</span>
+                <span class="ons-u-nowrap">{{ page.release_date|date("j F Y g:ia") or page.release_date_text }}</span>
+            </div>
+            {% if page.next_release_date or page.next_release_text %}
+                <div class="ons-grid__col ons-col-8@m ons-col-12@s ons-u-p-no@xxs@m">
+                    <span class="ons-u-fs-r--b">{{ _("Next release") }}:</span>
+                    <span class="ons-u-nowrap">{{ page.next_release_date|date("j F Y g:ia") or page.next_release_text }}</span>
```
RealOrangeOne, line 30: Question: How does `|date` handle receiving `None`? Might this need to be expanded to an `if`?

cms/release_calendar/migrations/0002_create_releasecalendarindex.py
```diff
@@ -0,0 +1,47 @@
...
+        draft_title="Release calendar",
+        live=True,
+        first_published_at=now,
+        last_published_at=now,
+        slug="releases",
+        path="%s00%02d" % (home_page.path, home_page.numchild + 1),
+        content_type=index_content_type,
+        depth=home_page.depth + 1,
+        url_path=f"{home_page.url_path}releases/",
+        locale=Locale.objects.first(),  # cannot use Locale.get_default(), but since we only have one, this works
```
RealOrangeOne, line 32: Nit: Worth using `.get` to enforce there's only one?
```suggestion
        locale=Locale.objects.get(),  # cannot use Locale.get_default(), but since we only have one, this works
```

cms/release_calendar/models.py
```diff
@@ -0,0 +1,196 @@
...
+
+    # Fields
+    status = models.CharField(choices=ReleaseStatus.choices, default=ReleaseStatus.PROVISIONAL, max_length=32)
+    summary = RichTextField(features=settings.RICH_TEXT_BASIC)
+
+    release_date = models.DateTimeField(
+        blank=True, null=True, help_text=_("Required once the release has been confirmed.")
+    )
+    release_date_text = models.CharField(
+        max_length=50, blank=True, help_text=_("Format: 'Month YYYY', or 'Month YYYY to Month YYYY'.")
```
RealOrangeOne, line 56: Question: Should this format be validated?

cms/release_calendar/tests/test_models.py
```diff
@@ -0,0 +1,187 @@
...
+    release_calendar_page.is_census = True
+    release_calendar_page.is_accredited = False
+
+    assert release_calendar_page.table_of_contents == expected_with_census_or_accredited
+    assert release_calendar_page.get_context(request)["table_of_contents"] == expected_with_census_or_accredited
+
+    del release_calendar_page.table_of_contents  # clear the cached property
+    release_calendar_page.is_census = False
+    release_calendar_page.is_accredited = True
+
```
RealOrangeOne, line 86: Suggestion: This feels like it's testing a number of different conditions in the same test. It's probably worth sticking to ["Arrange Act Assert"](https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/) where possible.

cms/settings/base.py
```diff
@@ -708,8 +710,8 @@
 # We normally don't want editors to use the images
 # in the rich text editor, for example.
 # They should use the image stream block instead
-RICH_TEXT_BASIC = ["bold", "italic", "link", "ol", "ul", "document-link"]
-RICH_TEXT_FULL = ["h3", "h4", *RICH_TEXT_BASIC]
+RICH_TEXT_BASIC = ["bold", "italic", "link", "ol", "ul"]
+RICH_TEXT_FULL = ["h3", "h4", *RICH_TEXT_BASIC, "document-link"]
```
RealOrangeOne, line 714: Question: Why are these separate settings rather than using the editors setting below? Saves needing to import them around.

conftest.py
```diff
@@ -29,3 +33,31 @@ def enable_console_logging():
...
+
+@pytest.fixture()
+def root_page() -> Page:
+    """Returns the root page. Useful for adding pages in tests."""
+    return Page.objects.filter(depth=1).get()
+
+
+@pytest.fixture()
+def home_page() -> HomePage:
+    """Returns the home page."""
```
RealOrangeOne, line 46:
```suggestion
    """Returns the home page, which is created via a migration."""
```

conftest.py
```diff
@@ -29,3 +33,31 @@ def enable_console_logging():
...
+    return HomePage.objects.first()
+
+
+@pytest.fixture()
+def release_calendar_index() -> ReleaseCalendarIndex:
+    """Returns the release calendar index page, which is created via a migration."""
+    return ReleaseCalendarIndex.objects.first()
```
RealOrangeOne, line 53: Thought: Currently there's no constraint that only one of these exists.

- Should there be a constraint, or
- Should the fixture be a factory instead?

conftest.py
```diff
@@ -29,3 +33,31 @@ def enable_console_logging():
...
+@pytest.fixture()
+def root_page() -> Page:
+    """Returns the root page. Useful for adding pages in tests."""
+    return Page.objects.filter(depth=1).get()
+
+
+@pytest.fixture()
+def home_page() -> HomePage:
+    """Returns the home page."""
+    return HomePage.objects.first()
```
RealOrangeOne, line 47: Nit: Since there's only ever meant to be one homepage
```suggestion
    return HomePage.objects.get()
```

cms/jinja2/templates/pages/release_calendar/release_calendar_page.html
zerolab: I am not doing all the front-end here. We have FE tasks for heroes

conftest.py
zerolab: But we won't, since we will have translations

conftest.py
zerolab: There's one index per homepage.

cms/settings/base.py
zerolab: We have two types of editors - basic for things like notices, summary etc. And full (for paragraph).

We can swap them around.. ie make basic the default and use the full one where needed

cms/release_calendar/migrations/0002_create_releasecalendarindex.py
zerolab: leaving this so things don't break when add i18n

cms/jinja2/templates/pages/release_calendar/release_calendar_page.html
zerolab: `|date` returns an empty string for `None`

cms/release_calendar/models.py
zerolab: good shout.

cms/jinja2/templates/components/streamfield/release_date_change_block.html
zerolab: was changed to `|date("DATETIME_FORMAT")`

cms/release_calendar/models.py
```diff
@@ -0,0 +1,201 @@
...
+    )
+    next_release_date = models.DateTimeField(blank=True, null=True)
+    next_release_text = models.CharField(
+        max_length=255, blank=True, help_text=_("Formats needed: 'DD Month YYYY Time' or 'To be confirmed'.")
+    )
```
MebinAbraham, line 64: We should make these mutually exclusive. It is confusing that both can be populated.

cms/release_calendar/models.py
```diff
@@ -0,0 +1,201 @@
...
+    # Fields
+    status = models.CharField(choices=ReleaseStatus.choices, default=ReleaseStatus.PROVISIONAL, max_length=32)
+    summary = RichTextField(features=settings.RICH_TEXT_BASIC)
+
+    release_date = models.DateTimeField(
+        blank=True, null=True, help_text=_("Required once the release has been confirmed.")
+    )
+    release_date_text = models.CharField(
+        max_length=50, blank=True, help_text=_("Format: 'Month YYYY', or 'Month YYYY to Month YYYY'.")
+    )
```
MebinAbraham, line 60: We should make these mutually exclusive. It is confusing that both can be populated.

cms/release_calendar/models.py
```diff
@@ -0,0 +1,201 @@
...
+                items += block.block.to_table_of_contents_items(block.value)
+
+        if self.status in NON_PROVISIONAL_STATUSES and self.changes_to_release_date:
+            items += [{"url": "#changes-to-release-date", "text": _("Changes to this release date")}]
+
+        if self.status == ReleaseStatus.PUBLISHED and self.contact_details_id:
+            items += [{"url": "#contact-details", "text": _("Contact details")}]
+
+        if self.is_accredited or self.is_census:
+            items += [{"url": "#about-the-data", "text": _("About the data")}]
```
MebinAbraham, line 192: I am wondering if there is a nicer way to manage the ID and title. At the moment, we'd have to remember to update the title and the section ID in the HTML template if these were to change and vice versa.

cms/release_calendar/forms.py
```diff
@@ -0,0 +1,82 @@
...
+    """Convenience administrative form for easier calendar page data validation."""
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+
+        if self.instance.status != ReleaseStatus.PROVISIONAL:
+            # Once a release calendar page is confirmed, it cannot go back to provisional
+            self.fields["status"].choices = NON_PROVISIONAL_STATUS_CHOICES
+
+        if self.instance.status == ReleaseStatus.PUBLISHED:
```
MebinAbraham, line 21: Q: Is there any reason why equality is chosen over identity for enums? I expected `self.instance.status is ReleaseStatus.PUBLISHED`

cms/release_calendar/forms.py
```diff
@@ -0,0 +1,82 @@
...
+                    raise ValidationError({"release_date_text": _("The end month must be after the start month.")})
+
+        except ValueError:
+            raise ValidationError(
+                {
+                    "release_date_text": _(
+                        "The release date text must be in the 'Month YYYY' or 'Month YYYY to Month YYYY' format."
+                    )
+                }
+            ) from None
```
MebinAbraham, line 82: As-is, this would not work for localised dates, i.e. if they input a Welsh date, this will always raise an error.

cms/settings/formats/en_GB/formats.py
```diff
@@ -0,0 +1,6 @@
+# 1 November 2025
+# as opposed to "j M Y" which is '1 Oct 2024'
+DATE_FORMAT = "j F Y"
+# 1 November 2024, 1p.m.
+# as opposed to "j M Y, P" which is '25 Oct 2006, 2:30 p.m.'
+DATETIME_FORMAT = "j F Y g:ia"
```
MebinAbraham, line 6: Minor, this doesn't match the live site exactly; we should check if that is a problem. Ours would be `a.m.` and p.m.` vs live site which is `am` and`pm`

cms/release_calendar/models.py
```diff
@@ -0,0 +1,201 @@
...
+    max_count_per_parent = 1
+
+
+class ReleasePageRelatedLink(LinkFieldsMixin, Orderable):
+    """Related links. e.g. https://www.ons.gov.uk/releases/welshlanguagecensus2021inwales."""
+
+    parent = ParentalKey("ReleaseCalendarPage", related_name="related_links", on_delete=models.CASCADE)
+
+
+class ReleaseCalendarPage(BasePage):  # type: ignore[django-manager-missing]
```
MebinAbraham, line 43: Currently, this page can be deleted; we should prevent that. They should be able to cancel but not delete.

cms/release_calendar/forms.py
MebinAbraham: Ah I though it resolved back to enums not strings. 👍

cms/settings/formats/en_GB/formats.py
zerolab: This has to do with how the Django date filter work - https://docs.djangoproject.com/en/5.1/ref/templates/builtins/#std-templatefilter-date

If it is a problem, we could look at providing a custom one.

cms/release_calendar/models.py
zerolab: How is 461eb36?

cms/release_calendar/forms.py
zerolab: I've explicitly made this to only kick in for English. 2 options:
- extend it now to cover localized dates. Couldn't think of anything elegant off the top of my head
- add as a task to the multi-language feature

Which one would you prefer?

cms/release_calendar/models.py
zerolab: There's a balance between directly slugified titles vs something more controlled. I tweaked this a bit in 3cda0ed, what do you think?

cms/release_calendar/models.py
zerolab: Done in 37e56d9

cms/release_calendar/models.py
zerolab: Done in 37e56d9

cms/jinja2/templates/pages/release_calendar/release_calendar_page.html
```diff
@@ -0,0 +1,126 @@
...
+                            <p>{{ page.contact_details.name }}</p>
+                            <h3 class="ons-u-mb-no ons-u-fs-r--b">Email</h3>
+                            <p><a href="mailto:{{ page.contact_details.email }}">{{ page.contact_details.email }}</a></p>
+                            {% if page.contact_details.phone %}
+                                <h3 class="ons-u-mb-no ons-u-fs-r--b">Phone</h3>
+                                <p><a href="tel:{{ page.contact_details.phone|urlencode() }}">{{ page.contact_details.phone }}</a></p>
+                            {% endif %}
+                        </div>
+                    </section>
+                {% endif %}
```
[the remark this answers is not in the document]
zerolab: it was somewhat intentional (as the contact details block is usually last), but updated to match live in f77c19f. Good shout, thanks

cms/release_calendar/models.py
```diff
@@ -0,0 +1,201 @@
...
+
+    parent_page_types: ClassVar[list[str]] = ["home.HomePage"]
+    subpage_types: ClassVar[list[str]] = ["ReleaseCalendarPage"]
+    max_count_per_parent = 1
+
+
+class ReleasePageRelatedLink(LinkFieldsMixin, Orderable):
+    """Related links. e.g. https://www.ons.gov.uk/releases/welshlanguagecensus2021inwales."""
+
+    parent = ParentalKey("ReleaseCalendarPage", related_name="related_links", on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
zerolab: The choices was a StreamField block or using [inline models](https://docs.wagtail.org/en/stable/topics/pages.html#inline-models).
I think for this purpose inline models fits a bit better, but happy to change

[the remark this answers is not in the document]
zerolab: Discussed at the tech session, will change to StreamField.

cms/settings/formats/en_GB/formats.py
MebinAbraham: As discussed, let keep as-is and revisit if it becomes a problem.

cms/release_calendar/models.py
MebinAbraham: That's great; I recommend we do the same for the "Release calendar index", which was programmatically created.
I haven't tested Move, but I suspect Move will enforce the subpage/parent page types list.

cms/release_calendar/forms.py
MebinAbraham: Let's `add as a task to the multi-language feature` but worth adding a TODO comment.

cms/release_calendar/models.py
MebinAbraham: That will do, will come back to this if it becomes a problem in future development.

cms/release_calendar/forms.py
```diff
@@ -0,0 +1,91 @@
...
+            self.fields["release_date"].disabled = True
+
+    def clean(self) -> dict:
+        """Validate the submitted release calendar data."""
+        cleaned_data: dict = super().clean()
+
+        status = cleaned_data["status"]
+
+        if status == ReleaseStatus.CANCELLED and not cleaned_data["notice"]:
+            raise ValidationError({"notice": _("The notice field is required when the release is cancelled")})
```
MebinAbraham, line 32: General point: some error messages are being marked for localisation, and some aren't. I suggest we mark all and be consistent.

cms/release_calendar/forms.py
```diff
@@ -0,0 +1,91 @@
...
+                            "If a confirmed calendar entry needs to be rescheduled, "
+                            "the 'Changes to release date' field must be filled out."
+                        )
+                    }
+                )
+
+            if len(self.instance.changes_to_release_date) > len(cleaned_data["changes_to_release_date"]):
+                raise ValidationError(
+                    {"changes_to_release_date": _("You cannot remove entries from the 'Changes to release date'.")}
+                )
```
MebinAbraham, line 58: Is this needed? Could there not be a genuine reason to remove it?

If we are keeping it, I think it needs some changes. Right now, if I add a change notice and then save as draft, I can't remove the notice. So anything "saved", even in the draft, gets enforced as final even though it was never published. It would be reasonable if already published content wasn't removable.

cms/release_calendar/models.py
```diff
@@ -0,0 +1,207 @@
...
+    release_date = models.DateTimeField(
+        blank=True, null=True, help_text=_("Required once the release has been confirmed.")
+    )
+    release_date_text = models.CharField(
+        max_length=50, blank=True, help_text=_("Format: 'Month YYYY', or 'Month YYYY to Month YYYY'.")
+    )
+    next_release_date = models.DateTimeField(blank=True, null=True)
+    next_release_text = models.CharField(
+        max_length=255, blank=True, help_text=_("Formats: 'DD Month YYYY Time' or 'To be confirmed'.")
+    )
```
MebinAbraham, line 59: Non-blocker: If we have both date types, it makes sense to enforce that the next release date much be greater than release date.

cms/core/blocks/markup.py
```diff
@@ -39,3 +44,63 @@ class QuoteBlock(blocks.StructBlock):
     class Meta:  # pylint: disable=missing-class-docstring,too-few-public-methods
         icon = "openquote"
         template = "templates/components/streamfield/quote_block.html"
+
+
+class BasicTableBlock(WagtailTableBlock):
```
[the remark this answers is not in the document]
zerolab: added in b8864f2

cms/release_calendar/models.py
zerolab: Move indeed enforces parent/subpage types

cms/release_calendar/forms.py
zerolab: will remove all together. it was not in the requirements, and I thought it would be a nice touch, but it can be confusing. You're right.

cms/release_calendar/wagtail_hooks.py
```diff
@@ -0,0 +1,31 @@
...
+
+    if page.specific_class == ReleaseCalendarPage:
+        messages.warning(
+            request, _("Release Calendar pages cannot be deleted. You can mark them as cancelled instead.")
+        )
+        return redirect("wagtailadmin_pages:edit", page.pk)
+
+    if page.specific_class == ReleaseCalendarIndex:
+        messages.warning(request, _("The Release Calendar index cannot be deleted."))
+        return redirect("wagtailadmin_pages:edit", page.pk)
```
MebinAbraham, line 29: Non-blocker: I'm not convinced we need to redirect to edit. Made sense for CalendarPage as you can mark as "Cancelled".

MebinAbraham (approved):
👍

cms/release_calendar/wagtail_hooks.py
zerolab: Happy to redirect somewhere else. e.g. to the explorer under its parent. or the main dashboard. The question what makes most sense?

zerolab: Updated in ef92b16
Also, improved the experience so you're redirected directly

MebinAbraham (approved):

sanjeevz3009 (commented):
Reviewed!

PR #17, opened 2024-11-01
Update to Wagtail 6.3

zerolab:
This PR builds on #16 and upgrades the build to Wagtail 6.3 (the new LTS)

Reference:
- https://docs.wagtail.org/en/stable/releases/6.3.html
- https://wagtail.org/blog/wagtail-63/


conftest.py
```diff
@@ -61,3 +70,10 @@ def release_calendar_page(release_calendar_index) -> ReleaseCalendarPage:  # pyl
     page.save()
 
     return page
+
+
+@pytest.fixture()
+def document():
+    """Creates a test document."""
+    file = ContentFile(b"A boring example document", name="file.txt")
```
RealOrangeOne, line 78: Nit: Text files are text
```suggestion
    file = ContentFile("A boring example document", name="file.txt")
```

conftest.py
```diff
@@ -2,13 +2,22 @@
...
 from cms.home.models import HomePage
 from cms.release_calendar.models import ReleaseCalendarIndex, ReleaseCalendarPage
 
 
+@pytest.fixture(scope="session", autouse=True)
+def _custom_settings(tmpdir_factory):
+    overrides = override_settings(MEDIA_ROOT=str(tmpdir_factory.mktemp("media")))
+    overrides.enable()
+
+
```
RealOrangeOne, line 20: Question: is it possible to do this as a context manager? So the settings are properly disabled

```suggestion
@pytest.fixture(scope="session", autouse=True)
def _custom_settings(tmpdir_factory):
    with override_settings(MEDIA_ROOT=str(tmpdir_factory.mktemp("media"))):
        yield


```

RealOrangeOne (approved):

conftest.py
zerolab: pushed these to #16. thanks, this works

MebinAbraham (approved):

PR #18, opened 2024-11-05
Replace `pytest` with Django's built-in runner

RealOrangeOne:

As discovered as part of #15, `pytest-django` doesn't work well with multiple databases (added in #15). It's possible to work around this "not working", however these bodges affect reliability of local testing, and aren't worth the risk.

Instead, this PR ports to using Django's built-in test runner (based more closely on `unittest`), which has full support for multiple databases.

Coverage is performed manually using `coverage run` and `pytest-xdist` is replaced by Django's built-in `--parallel`.

#### Alternatives

I tried using Django's test classes but with `pytest` as the runner, and this had the same issues. I also tried removing `pytest-django` whilst still using Django's test classes - which also failed.

This is because the issues are with `pytest-django` itself, and how it sets up the test environment, as opposed to how the tests are executed. Working around the issues of using `pytest` without `pytest-django` would likely involve partially implementing `pytest-django`.

### How to review

There should be no functional difference with this PR. Tests should still run as they did before (minus `pytest`).

### Follow-up Actions

This will require a few small changes to #15 to remove the workarounds for `pytest-django`.

Any open PRs will need their tests rewriting before merge, else they won't be run.

Upcoming functional tests were intending to also use `pytest`, however I'm going to spike using Django's test runner for those in a separate PR.


cms/settings/base.py
```diff
@@ -276,7 +276,7 @@
 }
 
 # Place static files that need a specific URL (such as robots.txt and favicon.ico) in the "public" folder
-WHITENOISE_ROOT = BASE_DIR / "public"
+# WHITENOISE_ROOT = BASE_DIR / "public"
```
RealOrangeOne, line 279: This is commented out because we're not using the functionality. When we need it, we can easily add it back. Commenting it out removes a warning when running tests.

zerolab: let's remove it all together. We can re-add, if and when we need it

cms/core/tests/test_context_processors.py
```diff
@@ -1,33 +1,41 @@
...
 
 from cms.core.context_processors import global_vars
 from cms.core.models import Tracking
 
-pytestmark = pytest.mark.django_db
 
+class ContextProcessorTestCase(TestCase):
+    """Tests for context processors."""
+    def setUp(self):
+        request_factory = RequestFactory()
```
zerolab, line 11: nit: why not set this up in `setUpTestData`?
Granted we have a couple of tests, but my preference for anything that can be global to go to `setUpTestData`

cms/core/tests/test_views.py
```diff
@@ -1,14 +1,18 @@
...
 
 
-@pytest.mark.django_db
-def test_csrf_token_mismatch_logs_an_error(csrf_check_client, caplog, enable_console_logging):  # pylint: disable=unused-argument
-    """Check that the custom csrf error view logs CSRF failures."""
-    csrf_check_client.cookies["csrftoken"] = "wrong"
+class CSRFTestCase(TestCase):
+    """Tests for CSRF enforcement."""
+    def setUp(self):
+        self.client = Client(enforce_csrf_checks=True)
```
zerolab, line 9: suggestion/nit: move to `setUpTestData`

Makefile
```diff
@@ -52,7 +51,7 @@ lint-frontend:  ## Run front-end linters
 
 .PHONY: test
 test:  ## Run the tests and check coverage.
-	poetry run pytest -n auto --ds=cms.settings.test --cov=cms --cov-report term-missing
+	poetry run coverage run ./manage.py test --parallel --settings=cms.settings.test
```
zerolab, line 54:
```suggestion
	poetry run coverage run ./manage.py test --parallel --settings=cms.settings.test
	poetry run coverage report
```

also we may need something like
```
branch = true
parallel = true
concurrency = multiprocessing, thread
``` 
in the coverage run configuration to better work with parallel runs. https://coverage.readthedocs.io/en/7.6.4/config.html#run-parallel

+ `python run coverage combine` before `poetry run coverage report`

zerolab (changes requested):
Left a few notes.
The main one is around the coverage config and reporting on when running `make test`

Makefile
zerolab: https://stackoverflow.com/questions/42721896/coverage-in-parallel-for-django-tests has some notes too

cms/core/tests/test_context_processors.py
RealOrangeOne: Whilst the factory isn't really mutable, the request definitely could be (eg site caches), so i think it makes sense to do it per test. Separating the factory is possibly, but probably more noise than it's worth.

MebinAbraham (approved):

cms/core/tests/test_views.py
RealOrangeOne: Django's `TestCase` creates the client for each case, so I think the same should be done here.

Makefile
RealOrangeOne: Parallel and sequential test runs are now reporting the same coverage numbers :partying_face:

zerolab (approved):

PR #19, opened 2024-11-05
Add Cookie banner and Google Tag Manager snippet

kacperpONS:

We wanted to add the DS cookies banner as part of our base template. 

- Logic for handling the language of the banner
Even though multi-language isn't yet supported, I added placeholder logic for displaying the Banner in Welsh if the LANGUAGE_CODE is set to cy. 

- Updated tests
I've replaced the existing tests with the versions that Hannah added for the November prototype (See: https://github.com/ONSdigital/dis-wagtail-alpha/pull/112). 

### How to review

1. Spin up the project locally
1. Make sure that the tests pass
1. Go to localhost:8000 in Incognito and make sure that the banner is displayed
1. Change the `LANGUAGE_CODE` in settings to "cy" and confirm that the banner is displayed in Welsh

You can also test that the GTM snippet works correctly by going to https://tagmanager.google.com/, creating an account and retrieving a GTM Container ID and passing it in the Docker-compose file. 

### Follow-up Actions

### DIFF
--- a/cms/jinja2/templates/base.html
+++ b/cms/jinja2/templates/base.html
@@ -1,4 +1,5 @@
 {% extends "layout/_template.njk" %}
+{% from "components/cookies-banner/_macro.njk" import onsCookiesBanner %}
 
 {% set page_title %}
     {% if current_site and page.pk == current_site.root_page.pk and current_site.site_name %}{{ current_site.site_name }} | {% endif %}{% block title %}{% if page.seo_title %}{{ page.seo_title }}{% else %}{{ page.title}}{% endif %}{% endblock %}{% block title_suffix %}{% if current_site and page.pk != current_site.root_page.pk and current_site.site_name %} | {{ current_site.site_name }}{% endif %}{% endblock %}
@@ -56,10 +57,52 @@
     }
 }
 %}
-{# fmt:on #}
+
 {% block head %}
-    <link rel="stylesheet" href="{{ static('css/main.css') }}">
-{% endblock %}
+
+  <!-- Google Tag Manager -->
+  <script>
+
+    function loadGTM(){
+      (function(w,d,s,l,i){
+        w[l]=w[l]||[];
+        w[l].push({
+          'gtm.start': new Date().getTime(),
+          event:'gtm.js'
+        });
+        var f=d.getElementsByTagName(s)[0],
+        j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';
+        j.async=true;
+        j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;
+        f.parentNode.insertBefore(j,f);
+      })
+      (window,document,'script','dataLayer', "{{ GOOGLE_TAG_MANAGER_CONTAINER_ID }}");
+    }
+
+    function isUsageCookieAllowed(){
+      var re = /^(.*)?\s*'usage':true\s*[^;]+(.*)?$/;
+      return document.cookie.match(re);
+    }
+
+    if (isUsageCookieAllowed()) {
+        loadGTM();
+    }
+  </script>
+  <!-- End Google Tag Manager -->
+
+{% endblock  %}
+
+
+{% block bodyStart %}
+  <!-- Google Tag Manager (noscript) -->
+  <noscript>
+    <iframe src="https://www.googletagmanager.com/ns.html?id={{ GOOGLE_TAG_MANAGER_CONTAINER_ID }}"
+    height="0" width="0" style="display:none;visibility:hidden">
+    </iframe>
+  </noscript>
+  <!-- End Google Tag Manager (noscript) -->
+{% endblock  %}
+
 
 {% block preMain %}
     {% if not IS_EXTERNAL_ENV %}
@@ -69,4 +112,22 @@
 
 {% block scripts %}
     <script src="{{ static('js/main.js') }}"></script>
+{% endblock %}
+
+{% block preHeader %}
+    {#
+      NOTE:
+      the "settingsLinkTextURL" setting currently doesn't match what's in the guidelines
+      that's because we're on version 70.0.17 and it has been renamed to "settingsLinkUrl" in 72.0.0
+      See: https://github.com/ONSdigital/design-system/pull/3300
+    #}
+    {{
+        onsCookiesBanner({
+            'lang': "en" if LANGUAGE_CODE == "en-gb" else "cy",
+            'serviceName': COOKIE_BANNER_SERVICE_NAME,
+            'settingsLinkText': 'Manage settings',
+            'settingsLinkTextURL': MANAGE_COOKIE_SETTINGS_URL,
+        })
+    }}
+
 {% endblock %}
--- a/cms/core/context_processors.py
+++ b/cms/core/context_processors.py
@@ -12,7 +12,12 @@ def global_vars(request: "HttpRequest") -> dict[str, str | bool | None]:
     """Set our global variables to use in templates."""
     tracking = Tracking.for_request(request)
     return {
-        "GOOGLE_TAG_MANAGER_ID": getattr(tracking, "google_tag_manager_id", None),
+        "GOOGLE_TAG_MANAGER_CONTAINER_ID": settings.GOOGLE_TAG_MANAGER_CONTAINER_ID,
+        
+        # Cookie banner settings
+        "ONS_COOKIE_BANNER_SERVICE_NAME": settings.ONS_COOKIE_BANNER_SERVICE_NAME,
+        "MANAGE_COOKIE_SETTINGS_URL": settings.MANAGE_COOKIE_SETTINGS_URL,
+        
         "SEO_NOINDEX": settings.SEO_NOINDEX,
         "LANGUAGE_CODE": settings.LANGUAGE_CODE,
         "IS_EXTERNAL_ENV": settings.IS_EXTERNAL_ENV,
--- a/cms/core/models/settings.py
+++ b/cms/core/models/settings.py
@@ -73,5 +73,5 @@ class Tracking(BaseSiteSetting):
     google_tag_manager_id = models.CharField(
         max_length=255,
         blank=True,
-        help_text="Your Google Tag Manager ID",
+        help_text="Your Google Tag Manager Container ID",
     )
--- a/cms/core/tests/test_context_processors.py
+++ b/cms/core/tests/test_context_processors.py
@@ -9,13 +9,9 @@ pytestmark = pytest.mark.django_db
 
 def test_when_no_tracking_settings_defined(rf):
     """Check the global vars include sensible defaults when no Tracking settings defined."""
-    request = rf.get("/")
-    assert global_vars(request) == {
-        "GOOGLE_TAG_MANAGER_ID": "",
-        "SEO_NOINDEX": False,
-        "LANGUAGE_CODE": "en-gb",
-        "IS_EXTERNAL_ENV": False,
-    }
+    request = RequestFactory().get("/")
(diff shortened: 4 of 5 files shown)


ababic:
Hi @kacperpONS. I think you might find (as I did for the prototype) that using https://www.ons.gov.uk/cookies as the management URL will produce some strange inconsistencies, especially in environments that aren't using the `ons.gov.uk` domain. Essentially, changes made on https://www.ons.gov.uk/cookies might not 'filter through' to the active environment, whereas clicking the buttons in the banner directly will.

On the prototype, we had to add our own version of the https://www.ons.gov.uk/cookies page, following the advice in the design system. You might find that you also need the URL and view changes from https://github.com/ONSdigital/dis-wagtail-alpha/pull/112 too.

kacperpONS:
Replying to @ababic, we've decided that for the time being the cookies page will not reside in Wagtail and we'll point to the existing Cookies page on the ONS domain. This is noted on the card in Jira [CMS-189](https://jira.ons.gov.uk/browse/CMS-189)

The risk that Andy mentions is known and that has been confirmed at today's stand up meeting. 👍

cms/jinja2/templates/base.html
```diff
@@ -70,3 +115,21 @@
...
+      See: https://github.com/ONSdigital/design-system/pull/3300
+    #}
+    {{
+    onsCookiesBanner({
+    'lang': "cy" if LANGUAGE_CODE == "cy" else "en",
+    'serviceName': COOKIE_BANNER_SERVICE_NAME,
+    'settingsLinkText': 'Manage settings',
+    'settingsLinkTextURL': MANAGE_COOKIE_SETTINGS_URL,
+    })
+    }}
```
zerolab, line 133:
```suggestion
    {#
      NOTE:
      the "settingsLinkTextURL" setting currently doesn't match what's in the guidelines
      that's because we're on version 70.0.17 and it has been renamed to "settingsLinkUrl" in 72.0.0
      See: https://github.com/ONSdigital/design-system/pull/3300
    #}
    {# fmt:off #}
    {{
        onsCookiesBanner({
            'lang': "cy" if LANGUAGE_CODE == "cy" else "en",
            'serviceName': COOKIE_BANNER_SERVICE_NAME,
            'settingsLinkText': 'Manage settings',
            'settingsLinkTextURL': MANAGE_COOKIE_SETTINGS_URL,
        })
    }}
    {# fmt:on #}
```

cms/jinja2/templates/base.html
```diff
@@ -56,10 +57,54 @@
...
+                    'gtm.start': new Date().getTime(),
+                    event:'gtm.js'
+                });
+                var f=d.getElementsByTagName(s)[0],
+                    j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';
+                j.async=true;
+                j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;
+                f.parentNode.insertBefore(j,f);
+            })
+            (window,document,'script','dataLayer', "{{ GOOGLE_TAG_MANAGER_CONTAINER_ID }}");
```
zerolab, line 81: it is a good idea to [escape](https://docs.djangoproject.com/en/5.1/ref/templates/builtins/#std-templatefilter-escape) these values when added like that.

```suggestion
            (window,document,'script','dataLayer', "{{ GOOGLE_TAG_MANAGER_CONTAINER_ID|escape }}");
```

cms/jinja2/templates/base.html
```diff
@@ -56,10 +57,54 @@
...
+    </script>
+  <!-- End Google Tag Manager -->
+
+{% endblock  %}
+
+
+{% block bodyStart %}
+  <!-- Google Tag Manager (noscript) -->
+    <noscript>
+        <iframe src="https://www.googletagmanager.com/ns.html?id={{ GOOGLE_TAG_MANAGER_CONTAINER_ID }}"
```
zerolab, line 101:
```suggestion
        <iframe src="https://www.googletagmanager.com/ns.html?id={{ GOOGLE_TAG_MANAGER_CONTAINER_ID|escape }}"
```

cms/jinja2/templates/base.html
```diff
@@ -56,10 +57,54 @@
     }
 }
 %}
+
 {# fmt:on #}
 {% block head %}
     <link rel="stylesheet" href="{{ static('css/main.css') }}">
-{% endblock %}
+
+  <!-- Google Tag Manager -->
```
zerolab, line 65: question: do we need the HTML comments?

cms/jinja2/templates/base.html
```diff
@@ -56,10 +57,54 @@
...
+    <script>
```
zerolab, line 66: suggestion: wrap this and the other related bits in a `{% if GOOGLE_TAG_MANAGER_CONTAINER_ID %}` conditional. While not necessary, I think it is good practice, especially if one prefers to turn this off during local development

zerolab (changes requested):
Good work @kacperpONS.
Left a few notes and suggestions

cms/jinja2/templates/base.html
ababic: I might add to that and suggest `{% if GOOGLE_TAG_MANAGER_CONTAINER_ID and not request.is_preview %}`

zerolab: That is a great shout, Andy. 👍🏼

cms/jinja2/templates/base.html
helenb: Saw this in passing - noting that we will need to update the design system to the latest version (and probably again once the new components are in). Not relevant to this PR but thought I would mention it.

cms/jinja2/templates/base.html
kacperpONS: Assumption made here: I'm guessing you're meaning that the banner should not be displayed in the preview mode so I'm wrapping it in the conditional too. Correct me if I'm wrong 🫡

cms/jinja2/templates/base.html
kacperpONS: I've removed these as now with the conditional it should be clear from the context what that is related to. 👍

cms/jinja2/templates/base.html
```diff
@@ -56,10 +57,53 @@
...
+
+
+{% block bodyStart %}
+  <!-- Google Tag Manager (noscript) -->
+    <noscript>
+        <iframe src="https://www.googletagmanager.com/ns.html?id={{ GOOGLE_TAG_MANAGER_CONTAINER_ID|escape }}"
+                height="0" width="0" style="display:none;visibility:hidden">
+        </iframe>
+    </noscript>
+  <!-- End Google Tag Manager (noscript) -->
```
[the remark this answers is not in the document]
kacperpONS: Then I'd say it **must** be removed 😉

cms/jinja2/templates/base.html
```diff
@@ -56,10 +57,42 @@
...
+            }
+
+            function isUsageCookieAllowed(){
+                var re = /^(.*)?\s*'usage':true\s*[^;]+(.*)?$/;
+                return document.cookie.match(re);
+            }
+
+            if (isUsageCookieAllowed()) {
+                loadGTM();
+            }
```
MebinAbraham, line 90: To verify this actually works, we should add a small integration test.

- Set cookie with name 'ons_cookie_policy' and value which contains ''usage':true'
- Assert in the head 'https://www.googletagmanager.com' and the `GOOGLE_TAG_MANAGER_CONTAINER_ID` is present.

kacperpONS:
^ Comment to the above commit: 
I've just spotted that the because I've used a custom value for the `settingsLinkText` the custom text ("Manage settings")on the Welsh version of the banner was still appearing in English. 
I've removed that field completely because it wasn't a requirement on this card (I've accidentally copied that from the Nov prototype version). We can still add it back at some point but that would need to be complemented by a translation for it to work as expected.

cms/core/tests/test_views.py
```diff
@@ -18,3 +20,19 @@ def test_csrf_token_mismatch_logs_an_error(self):
...
 
         self.assertIn("CSRF Failure: CSRF cookie", logs.output[0])
+
+
+class TestCookiesBannerTestCase(TestCase):
+    """Test for the cookie banner functionality."""
+
+    def setUp(self):
+        self.client = Client()
+
```
zerolab, line 30:
```suggestion
```

This is not needed. Django's `TestCase` provides a client which is pretty much the duplicate of this.

The reason the test case above defines it separately is that it needs to enforces CSRF checks, which effectively makes it a custom client

cms/core/tests/test_views.py
```diff
@@ -17,3 +20,16 @@ def test_csrf_token_mismatch_logs_an_error(self):
...
 
         self.assertIn("CSRF Failure: CSRF cookie", logs.output[0])
+
+
+class TestCookiesBannerTestCase(TestCase):
+    """Test for the cookie banner functionality."""
+
+    def test_google_tag_manager_script_present_when_tracking_consent_given(self):
+        """Check that the Google Tag Manager script is present on the page when the user agrees to tracking."""
+        self.client.cookies = SimpleCookie({"ons_cookie_policy": "'usage':true"})
```
zerolab, line 30: suggestion:
```suggestion
        self.client.cookies["ons_cookie_policy"] = "'usage':true"
```

because `self.client.cookies` is already a `SimpleCookie` morsel ([ref](https://docs.djangoproject.com/en/5.1/topics/testing/tools/#django.test.Client.cookies)), which acts like a dict. Doing it this way ensures we don't accidentally overwrite anything else, and it should not break in the future.

zerolab (approved):
One non-blocking suggestion from me.
Otherwise LGTM

sanjeevz3009 (approved):
Pulled down the branch, tested the features locally and the tests all passes. LGTM!

Minor: The branch is out of date with main.

cms/core/tests/test_views.py
kacperpONS: Good shout, I'll do that and also change the cookie value value to dict rather than a string.

kacperpONS:
@sanjeevz3009 @zerolab can I get a re-review plz 🙏

cms/core/tests/test_views.py
```diff
@@ -17,3 +19,16 @@ def test_csrf_token_mismatch_logs_an_error(self):
...
+    """Tests for the cookie banner functionality."""
+
+    def test_google_tag_manager_script_present_when_tracking_consent_given(self):
+        """Check that the Google Tag Manager script is present on the page when the user agrees to tracking."""
+        self.client.cookies["ons_cookie_policy"] = {"usage": True}
+
+        response = self.client.get("/")
+
+        self.assertIn("https://www.googletagmanager.com/gtm.js?id=", response.rendered_content)
+        self.assertIn(settings.GOOGLE_TAG_MANAGER_CONTAINER_ID, response.rendered_content)
```
MebinAbraham, line 34: Actually, this approach results in a false positive (this will always pass regardless of the cookie), as the GTM script is created via JS, which integration tests don’t execute. Therefore, we can’t reliably verify the presence of the GTM ID here. I recommend removing this test for now, as we can add a functional test using Playwright to reliably test this.

kacperpONS: After conversation with @MebinAbraham we've decided to keep the test and change its purpose. 
Rather than an integration test checking the GTM/banner functionality, it'll now be a unit test doing a simple check that the GTM ID and URL are present on the page.
The functional tests for GTM/banner will have to follow once we get them set up. These will check the logic for cookie selection and whether the GTM script is present on the page.

MebinAbraham (approved):
👍

zerolab (approved):
:shipit:

PR #23, opened 2024-11-11
Update CODEOWNERS to be .github only

zerolab:

Switched to .github folder only. Should avoid unnecessary noise while preserving code ownership.

This follows the pattern used in other DIS repositories

### How to review

Check that is looks OK and in line with approach to codeowners for multiple teams

### Follow-up Actions

N/A


zerolab:
Closing as this approach is intended for shared repos

PR #31, opened 2024-11-18
Introduce Functional Tests using Playwright and Behave

AdamHawtin:

This PR adds functional tests for the app, written in the Behave BDD framework and using Playwright to test browser interactions.

### How to review

- Try running the functional tests in your development environment, following the new instructions in the README

- Review the new documentation. Does it provide enough guidance for you to be happy adding new test features/scenarios as part of future work? Do the suggested code standards and patterns make sense? These are intended to be a starting point, based on my experience with large test suites in Behave, but they should be reviewed agreed as a team, not just taken as written. 

- Try running the tests in your IDE or editor of choice. Is it possible to easily run individual scenarios or debug them? Certain functionality might require installing extensions.

- Check the custom fixture code for test setup and teardown, does it all make sense and is it sufficiently documented and understandable?

- Check the features and scenarios. These are my initial guesses at some suitable scenarios, they are not driven by any explicit requirements from those features. I am perfectly happy to modify them, or even tear them up and do it differently if people take issue, I just needed some example features in place to demonstrate the code structure.

### Follow-up Actions

We still need to agree how we will utilise this method on testing in further development. I would suggest the test scenarios need to be agreed at refinement time for future incoming tickets, but this still needs wider discussion and agreement around how these new tests fit into our definitions or ready and done.

We will also need to dicuss and add tasks for backfilling tests on existing features.

Topics to Revisit:
- Potential to use `serialized_rollback` instead of our hand-rolled DSLR based database clean/restore between tests. Can we get it working and to a satisfactory speed?
- Building reusable page classes for consistent navigation in playwright
- Consider setting `IS_EXTERNAL_ENV` to `true` when testing the external site, to better replicate a production setting
- Pursuing multi-browser testing

### Known Issues

- The initial scenarios currently only reliably pass in the `chromium` browser. We have agreed to address this in future work.


functional_tests/README.md
```diff
@@ -0,0 +1,274 @@
...
+1. This should leave you with a zip file for each failed scenario
+1. Open the traces zip files one at a time using the [Playwright Trace Viewer](https://playwright.dev/docs/trace-viewer)
+
+You should then be able to step through the failed tests and get a better idea of the state and cause of the failure.
+
+## Test Data Setup
+
+Some tests may require objects to be set up in the database, such as a user or set of pages that the feature relies
+upon. For this, we can use [Factory Boy](https://factoryboy.readthedocs.io/en/stable/orms.html#django) to see data
+directly into the database.
```
sanjeevz3009, line 146: `see data`. I think you meant `seed data` =)

functional_tests/behave_fixtures.py
```diff
@@ -0,0 +1,87 @@
...
+    # Set up Django, and it's test runner to give us a clean test database and initialised/set up Django resources
+    django.setup()
+    context.test_runner = DiscoverRunner()
+    context.test_runner.setup_test_environment()
+    context.old_db_config = context.test_runner.setup_databases()
+
+    # At this point, the Django test runner has initialised the test database and run migrations
+    # Initialize DSLR and take a snapshot of the clean, post migration test database
+    initialize_dslr()
+    context.clean_snapshot_name = "clean_snapshot"
```
sanjeevz3009, line 31: Shouldn't the comment and the function call be named `initialise_dslr`? 

Are you trying to be consistent with `dslr_settings.initialize(url=db_conn_str, debug=False)` code?

functional_tests/behave_fixtures.py
```diff
@@ -0,0 +1,87 @@
...
+    context.test_case.tearDown()
+    restore_snapshot(find_snapshot(context.clean_snapshot_name))
+    context.test_case_class.tearDownClass()
+    del context.test_case_class
+    del context.test_case
+
+
+def initialize_dslr() -> None:
+    """Initialize DSLR to point to the Django default database."""
+    db_config = settings.DATABASES["default"]
```
sanjeevz3009, line 82: Why does this function contain a `z`.

Is it to be consistent with `dslr_settings.initialize(url=db_conn_str, debug=False)` code?

sanjeevz3009 (commented):
Reviewed every file. It looks good to me. Pulled down the branch and tested everything locally and things are working as it should.

Left some minor comments =)

sanjeevz3009:
@MebinAbraham I know we had this conversation at various stages. But we need to end the docstring madness 😭

I'll show you Mebin in another comment and tag you in.

functional_tests/steps/cms_login.py
```diff
@@ -0,0 +1,43 @@
...
+
+@when("the user navigates to the beta CMS admin page")  # pylint: disable=E1102
+def cms_admin_navigates_to_beta_homepage(context: Context) -> None:
+    """Navigate to admin login page."""
+    context.page.goto(f"{context.base_url}/admin/login/")
+
+
+@when("they enter a their valid username and password and click login")  # pylint: disable=E1102
+def enter_a_valid_username_and_password_and_sign_in(context: Context) -> None:
+    """Enter the username and password and click login."""
```
sanjeevz3009, line 22: @MebinAbraham @AdamHawtin  An example of the docstrings madness all the step definitions are like this, since the linter (I think ruff) enforces this.

The given step decorator has "they enter a their valid username and password and click login".

The function is named "enter_a_valid_username_and_password_and_sign_in".

Then we have a docstring that says "Enter the username and password and click login".

This is literally weird, and soon this will swirl into madness.

Not sure what everyone else thinks.

functional_tests/step_helpers/utilities.py
```diff
@@ -0,0 +1,9 @@
+def str_to_bool(bool_string: str) -> bool:
+    """Takes a string argument which indicates a boolean, and returns the corresponding boolean value.
+    raises ValueError if input string is not one of the recognized boolean like values.
+    """
```
sanjeevz3009, line 4: Nit: `recognized`. I think it should be `recognised`.

functional_tests/README.md
```diff
@@ -0,0 +1,274 @@
...
+
+Factories that use Django models rely on Django being initialised. This causes issues if they are imported at a module
+level in the tests, as those imports then get run before Django has initialised, so can fail with a error like:
+
+```
+Models aren't loaded yet.
+```
+
+To work around this, the Factory classes must be imported within the step functions, so that
+
```
sanjeevz3009, line 158: `so that`? Unfinished sentence?

functional_tests/behave_fixtures.py
AdamHawtin: I believe historically it's been a UK vs American English divide, but these days either spelling is considered accepted. I don't have strong feelings beyond maintaining consistency and within our code base, so I defaulted to the same spelling as they use in the DSLR module itself.

functional_tests/README.md
AdamHawtin: This issue was resolved in a different way, so this will be removed entirely.

functional_tests/steps/cms_login.py
MebinAbraham: The requirement for all functions needing docstring has been disabled in the Bundles PR.

zerolab (commented):
So much good stuff in here 💪🏼 
Left a few suggestions.

nehakerung:
LGTM - just pulled down the branch and followed readme. Worked as intended.

.github/workflows/ci.yml
```diff
@@ -135,3 +135,65 @@ jobs:
...
+          - chromium
+    needs:
+      - lint
+      - lint-front-end
+      - compile_static
+
+    env:
+      DJANGO_SETTINGS_MODULE: cms.settings.functional_test
+      SECRET_KEY: fake_secret_key_to_run_tests # pragma: allowlist secret
+      BROWSER: ${{ matrix.browser }}
```
zerolab, line 153:
```suggestion
      PLAYWRIGHT_BROWSER: ${{ matrix.browser }}
```

.github/workflows/ci.yml
```diff
@@ -135,3 +135,68 @@ jobs:
...
+        run: make functional-tests-up
+
+      - name: Fetch Design System templates
+        run: make load-design-system-templates
+
+      - name: Collect Static Files
+        run: poetry run ./manage.py collectstatic --verbosity 0 --noinput --clear
+
+      - name: Prepare Database
+        run: poetry run ./manage.py migrate --noinput
```
zerolab, line 189: so this is failing for me locally

<details>
<summary>shell outputs</summary>

```shell
~ ❯❯❯ PLAYWRIGHT_HEADLESS=False PLAYWRIGHT_SLOW_MO=1000 make functional-tests
docker compose -f functional_tests/docker-compose.yml up -d
[+] Running 4/4
 ✔ Network functional_tests_default    Created                                                                            0.1s
 ✔ Volume "functional_tests_pgdata"    Created                                                                            0.0s
 ✔ Container functional_tests-db-1     Started                                                                            0.4s
 ✔ Container functional_tests-redis-1  Started                                                                            0.4s
poetry run behave functional_tests
/Users/dan/Sites/ons/.venv/lib/python3.12/site-packages/behave/matchers.py:267: SyntaxWarning: invalid escape sequence '\d'
  """Registers a custom type that will be available to "parse"
/Users/dan/Sites/ons/.venv/lib/python3.12/site-packages/behave/formatter/ansi_escapes.py:57: SyntaxWarning: invalid escape sequence '\['
  _ANSI_ESCAPE_PATTERN = re.compile(u"\x1b\[\d+[mA]", re.UNICODE)
Creating test database for alias 'default'...
/Users/dan/Sites/ons/.venv/lib/python3.12/site-packages/django/db/backends/postgresql/base.py:512: RuntimeWarning: Normally Django will use a connection to the 'postgres' database to avoid running initialization queries against the production database when it's not needed (for example, when running tests). Django was unable to create a connection to the 'postgres' database and will use the first PostgreSQL database instead.
  warnings.warn(
HOOK-ERROR in before_all: OperationalError: connection failed: connection to server at "127.0.0.1", port 15432 failed: server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
HOOK-ERROR in after_all: AttributeError: 'Context' object has no attribute 'playwright_trace'

ABORTED: By user.
0 features passed, 0 failed, 0 skipped, 4 untested
0 scenarios passed, 0 failed, 0 skipped, 4 untested
0 steps passed, 0 failed, 0 skipped, 0 undefined, 22 untested
Took 0m0.000s
make: *** [functional-tests-run] Error 1
(.venv)
# dan at tbx-m1p-zero.local in ~/Sites/ons on git:playwright-functional-tests-CMS-198 ● [15:14:07]
~ ❯❯❯ PLAYWRIGHT_HEADLESS=False PLAYWRIGHT_SLOW_MO=1000 make functional-tests
docker compose -f functional_tests/docker-compose.yml up -d
[+] Running 2/0
 ✔ Container functional_tests-redis-1  Running                                                                            0.0s
 ✔ Container functional_tests-db-1     Running                                                                            0.0s
poetry run behave functional_tests
Creating test database for alias 'default'...
HOOK-ERROR in before_all: ProgrammingError: relation "auth_group" does not exist
LINE 1: ...URSOR WITH HOLD FOR SELECT "auth_group"."id" FROM "auth_grou...
                                                             ^
HOOK-ERROR in after_all: AttributeError: 'Context' object has no attribute 'playwright_trace'

ABORTED: By user.
0 features passed, 0 failed, 0 skipped, 4 untested
0 scenarios passed, 0 failed, 0 skipped, 4 untested
0 steps passed, 0 failed, 0 skipped, 0 undefined, 22 untested
Took 0m0.000s
make: *** [functional-tests-run] Error 1
(.venv)
```

</details>

and if I change `functional-tests-run` to

```makefile
functional-tests-run:  ## Run the functional tests
	poetry run ./manage.py collectstatic --verbosity 0 --noinput --clear
	poetry run ./manage.py migrate --noinput
	poetry run behave functional_tests
```

it fails with no connection to the DB as I expect I'd need to massage `.env` to point to the right funcational db and redis

It may be worth adding a bash script that wraps it all (i.e. sets the right env vars inline before calling `behave functional_tests`

zerolab: yep, changing to
```
functional-tests-run:  ## Run the functional tests
	poetry run ./manage.py collectstatic --verbosity 0 --noinput --clear
	DATABASE_URL=postgres://ons:pass@localhost:15432/postgres REDIS_URL=redis://localhost:16379 poetry run ./manage.py migrate --noinput
	DATABASE_URL=postgres://ons:pass@localhost:15432/postgres REDIS_URL=redis://localhost:16379 poetry run behave functional_tests
```

makes it work

AdamHawtin: Yes, I realised after splitting out the dev app from the docker compose that the test DB relies on having the main default DB structure there to copy, rather than creating it entirely from scratch itself.

You shouldn't need to pass the config through to the run tests command though, as it should be hardcoded to use the correct settings module in the behave `environment.py`, I'll investigate today and get it more robust in local environments.

functional_tests/docker-compose-dev-app.yml
```diff
@@ -0,0 +1,40 @@
+services:
```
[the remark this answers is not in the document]
AdamHawtin: Nice suggestion, I've been able to remove pretty much all the duplication 👍

.github/workflows/ci.yml
zerolab: this works now. thanks!

functional_tests/steps/home_page.py
```diff
@@ -0,0 +1,17 @@
...
+from behave.runner import Context
+from playwright.sync_api import expect
+
+
+@when("An external user navigates to the ONS beta site homepage")  # pylint: disable=E1102
+def external_user_navigates_to_beta_homepage(context: Context) -> None:
+    context.page.goto(context.base_url)
+
+
+@then("they can see the beta homepage")  # pylint: disable=E1102
```
zerolab, line 11:
```suggestion
@then("they can see the beta homepage")  # pylint: disable=not-callable
```

imho this is easier to grok and doesn't require one to look for the codes

functional_tests/steps/release_page.py
```diff
@@ -0,0 +1,54 @@
+from behave import then, when  # pylint: disable=E0611
```
zerolab, line 1:
```suggestion
from behave import then, when  # pylint: disable=no-name-in-module 
```

zerolab (approved):
LGTM.

I was in two minds whether we want `functional_tests/README.md` to be a page in docs, but this can be separate change

functional_tests/steps/home_page.py
```diff
@@ -1,9 +1,9 @@
-from behave import then, when  # pylint: disable=E0611
+from behave import then, when  # pylint: disable=no-name-in-module
```
MebinAbraham, line 1: Have you tried this? https://pypi.org/project/pylint-behave/

AdamHawtin: I'll look into it. It looks like the issue is already fixed in the Behave source, but the next release that'll include it has been in development for a few years already, so no indication when it'll be available. Hopefully the plugin will stop us having to ignore everywhere 👍

AdamHawtin: Tried the plugin, it does deal with the Behave false warnings, but it also caused new false warnings from a different, unrelated file, so I won't introduce it for now, perhaps can be revisited in a future iteration.

zerolab (approved):

zerolab (approved):

Makefile
```diff
@@ -152,3 +152,27 @@ runserver: ## Run the Django application locally
 
 .PHONY: dev-init
 dev-init: load-design-system-templates collectstatic makemigrations migrate createsuperuser ## Run the pre-run setup scripts
+
+.PHONY: functional-tests-up
+functional-tests-up:  ## Start the functional tests docker compose dependencies
+	docker compose -f functional_tests/docker-compose.yml up -d
+
+.PHONY: functional-tests-up-dev
+functional-tests-up-dev:  ## Start the functional tests docker compose dependencies and dev app
```
MebinAbraham, line 161: Nit: `dev-up` to be consistent with the existing ones.

MebinAbraham, line 161: I think these are the wrong way around. `up` should be all, and `dev` should be backing services.

Makefile
```diff
@@ -152,3 +152,27 @@ runserver: ## Run the Django application locally
...
+	docker compose -f functional_tests/docker-compose-dev-app.yml up -d
+
+.PHONY: functional-tests-down
+functional-tests-down:  ## Stop the functional tests docker compose dependencies
+	docker compose -f functional_tests/docker-compose-dev-app.yml down
+
+.PHONY: functional-tests-run
+functional-tests-run: collectstatic  ## Run the functional tests
```
MebinAbraham, line 169: This also requires the backing services to be up, so chain the call?. DS templates has to be loaded as well, it would be good for commands to not assume they would be running against a containerised app.

Makefile
```diff
@@ -152,3 +152,27 @@ runserver: ## Run the Django application locally
...
+	DJANGO_SETTINGS_MODULE=cms.settings.functional_test poetry run ./manage.py migrate --noinput
```
MebinAbraham, line 170: Might not be clear to all, but, this will only work if you comment out conflicting env vars in `.env` when running locally.

MebinAbraham, line 170: As above, use Make commands where possible, so this could be replaced by:
```suggestion
	DJANGO_SETTINGS_MODULE=cms.settings.functional_test make migrate
```

Makefile
```diff
@@ -152,3 +152,27 @@ runserver: ## Run the Django application locally
...
+	poetry run behave functional_tests
+
+.PHONY: functional-tests
+functional-tests: functional-tests-up functional-tests-run functional-tests-down  ## Run the functional tests with backing services (all in one)
+
+.PHONY: playwright-install
+playwright-install:  ## Install Playwright dependencies
+	poetry run playwright install --with-deps
```
MebinAbraham, line 178: Suggestion: Add this after the `make install-dev` call so all dependencies are installed in 1 place?

functional_tests/docker-compose-dev-app.yml
```diff
@@ -0,0 +1,38 @@
...
+    environment:
+      DJANGO_SETTINGS_MODULE: cms.settings.functional_test
+    ports: !override
+      - 18000:8000 # runserver
+    volumes:
+      # Mount the repo at /app/
+      - ../:/app/
+    command: >
+      sh -c "poetry install &&
+             django-admin collectstatic --verbosity 0 --noinput --clear &&
```
MebinAbraham, line 18: Are these two needed?. These are done when the docker image is built?.

functional_tests/README.md
```diff
@@ -0,0 +1,303 @@
...
+To replicate the clean test database environment, you may want to completely clean your database.
+
+You can use the django admin `reset_db` to completely clear the database, then re-run migrations to re-initialise the
+tables and seeded data. Ensure you include the `DJANGO_SETTINGS_MODULE` environment variable, to avoid accidentally
+wiping your normal local development environment.
+
+```shell
+DJANGO_SETTINGS_MODULE=cms.settings.functional_test poetry run python manage.py reset_db
+DJANGO_SETTINGS_MODULE=cms.settings.functional_test poetry run python manage.py migrate
```
MebinAbraham, line 89:
```suggestion
DJANGO_SETTINGS_MODULE=cms.settings.functional_test poetry run python manage.py reset_db
DJANGO_SETTINGS_MODULE=cms.settings.functional_test make migrate
```

Could do a make command for reset_db but not needed.

functional_tests/README.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+Then for logging into the CMS, create a superuser with
+
+```shell
+DJANGO_SETTINGS_MODULE=cms.settings.functional_test poetry run python manage.py createsuperuser
```
MebinAbraham, line 95:
```suggestion
DJANGO_SETTINGS_MODULE=cms.settings.functional_test make createsuperuser
```

MebinAbraham (commented):
This is great work! Docs are well written also 🎉 

Just some suggestions from me, otherwise, LGTM.

functional_tests/behave_fixtures.py
sanjeevz3009: Related to the previous comment, so resolving this one as well.

functional_tests/behave_fixtures.py
sanjeevz3009: The second part, yes thought so. Thanks!

sanjeevz3009 (approved):
Reviewed the latest code changes and tested everything locally. All is good when it comes to running tests =)

.github/workflows/ci.yml
```diff
@@ -137,3 +137,68 @@ jobs:
...
```
RealOrangeOne, line 191: Question: Is the need to run this command related to the bug I found? If so I'd recommend linking it so it's obvious it can be removed.

Since it looks like the test runner uses the test database instance, there _shouldn't_  be a need to run the real migrations

functional_tests/README.md
```diff
@@ -0,0 +1,303 @@
...
+complex data than single strings or the other basic types supported by the default parser, we
+use [custom registered types](https://behave.readthedocs.io/en/stable/api.html#behave.register_type). These are
+registered in the [environment.py](environment.py) so they are available to all steps.
+
+## How the Tests Work
+
+### Django Test Runner and Test Case
+
+Due to issues with the Django `TransactionTestCase` which prevent us using the built-in database teardown/setup
+in between scenarios, we have implemented our own database snapshot and restore pattern between tests. We still make use
```
RealOrangeOne, line 246: Question: What is this issue?

You can use [`serialized_rollback`](https://docs.djangoproject.com/en/5.1/topics/testing/overview/#rollback-emulation) to ensure data created in migrations is loaded between tests.

Or is the need more about having a live server to test against?

functional_tests/README.md
```diff
@@ -0,0 +1,303 @@
...
+[`LiveServerTestCase`](https://docs.djangoproject.com/en/stable/topics/testing/tools/#django.test.LiveServerTestCase)
+to perform the initial database setup and run a live server on a random port for the tests.
+
+### Database Snapshot and Restore
+
+We are using [DSLR](https://pypi.org/project/dslr/) for fast database snapshots and restores.
+
+After we have used the Django test runner to set up the test database and initialise it by running migrations, we take a
+DSLR snapshot of this clean, initial state. In the test case fixture, post test, we then restore the clean snapshot,
+ensuring each test gets a clean, migrated database, isolated from other tests.
```
RealOrangeOne, line 257: Question: Django should be handling this itself with its test classes, using a test connection? Is that not working?

functional_tests/README.md
```diff
@@ -0,0 +1,303 @@
...
+Live server testing is accomplished with a fixture, which under the hood uses a Django `LiveServerTestCase`. This
+inherits from
+the [
+`TransactionTestCase`](https://docs.djangoproject.com/en/stable/topics/testing/tools/#django.test.TransactionTestCase),
+which causes us serious compatibility issues, as it uses an isolated test database and flushes all data in between
+tests. We have migrations which seed critical data rows, so a flush operation breaks the app.
+
+We tried various workarounds such as using a fixture file to restore the data, but this runs into Wagtail issues, and
+even if it worked it would be non-ideal as that fixture file would have to be kept up to date and recreated when any new
+seeded data is added.
```
RealOrangeOne, line 292: There's now a custom `TransactionTestCase` in the repo which should fix this :slightly_smiling_face: Hopefully, that means we can use one of these libraries again :crossed_fingers:

functional_tests/steps/home_page.py
```diff
@@ -0,0 +1,17 @@
...
+from behave.runner import Context
+from playwright.sync_api import expect
+
+
+@when("An external user navigates to the ONS beta site homepage")  # pylint: disable=not-callable
```
RealOrangeOne, line 6: Issue: "external" here should probably set `IS_EXTERNAL_ENV=True`, so that it's using the external environment.

Although, that might work better as a `@given`

functional_tests/README.md
zerolab: I believe there were several. off the top of my head:

1. the setup tear down was rather slow (and there was some quirk with behave)
2. The `Locale` issue which is now covered by https://github.com/ONSdigital/dis-wagtail/blob/92ed3b95e77995e28edd923859b23f01fc7b1832/cms/core/tests/__init__.py#L33

RealOrangeOne: Yeah the performance isn't something which I've solved - Django warns that the slowdown can be as much as 3x with `serialized_rollback`. But slightly slower tests may be worth the significantly reduced complexity.

functional_tests/steps/home_page.py
AdamHawtin: How does this variable work exactly? Does it have to be set at app startup time or can it be toggled at any point?

Makefile
AdamHawtin: In this case I thought this way made more sense, since the app is only needed if the developer wants it, for manually interacting with in development, but I can change it for consistency.

AdamHawtin: My reasoning for doing it this way round was making the targets more discoverable and consistent with shell auto-complete (e.g. if you hit tab on `make functionial-tests-up` then it can suggest this as a variant), but I can change them for consistency if we prefer the convention on the other new targets.

functional_tests/docker-compose-dev-app.yml
```diff
@@ -0,0 +1,38 @@
+services:
```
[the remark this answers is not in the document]
AdamHawtin: Naming them that way around here feels odd to me, as the dockerized app instance is specifically only required for manually developing against, not for running the tests. Not a huge sticking point, but I prefer keeping the name indicative, since the pattern of use here is different, I think.

Makefile
AdamHawtin: I avoided this because some existing CI jobs call `make install-dev` and I don't want to burden them with downloading browsers they don't need.

functional_tests/docker-compose-dev-app.yml
AdamHawtin: They probably shouldn't be needed, but I had issues with dependencies and statics files not being properly initialised locally and running them here was the only way to resolve. I'll check again, and if it's still an issue I'll add a comment explaining.

cms/settings/functional_test.py
```diff
@@ -0,0 +1,10 @@
+import copy
+
+from .test import *  # noqa: F403  # pylint: disable=wildcard-import,unused-wildcard-import
+
+DATABASES = {
+    "default": dj_database_url.config(default="postgres://ons:ons@localhost:15432/ons"),  # noqa: F405
+}
+DATABASES["read_replica"] = copy.deepcopy(DATABASES["default"])
+
+REDIS_URL = "redis://localhost:16379"
```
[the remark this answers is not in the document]
AdamHawtin: They're using different ports to match the functional tests docker compose configuration.

.github/workflows/ci.yml
AdamHawtin: Possibly, I believed the way it worked was the test runner copies the configured default database structure so pre-running migrations was required, but it might be related to the bug.

functional_tests/README.md
AdamHawtin: There is more detail about this is in the [why-arent-we-using-existing-django-testing-modules section](https://github.com/ONSdigital/dis-wagtail/blob/308d61bb4c32593d2fa446f2cd633f15520c97de/functional_tests/README.md#why-arent-we-using-existing-django-testing-modules). TL;DR, we're still using the test runner to set up the test database initially, but the between test teardown was breaking things, removing the required data which is seeded in migrations.

functional_tests/README.md
AdamHawtin: This might allow us to use the native behaviour, but we would need to craft a version of `LiveServerTestCase` which inherits from the customised `TransactionTestCase`. 

Although the current solution is also a bit custom, it is actually quite simple in implementation, requiring less than 10 SLOC to work and has proven to be very fast, so I would suggest keeping to this solution initially, and we can separately investigate if replacing it with a customised `LiveServerTestCase` is preferable in terms of understandability and performance down the line?

functional_tests/README.md
zerolab: My preference is to get this in, then revisit. 
Worth keeping in mind that the slowness will increase as we add tests.

functional_tests/README.md
RealOrangeOne: I suspect the solution is https://github.com/ONSdigital/dis-wagtail/pull/31#discussion_r1875705404, but as Dan mentions it's probably one to revisit.

functional_tests/steps/home_page.py
RealOrangeOne: It should be settable at any point, most of the handling is in a database router which can handle it dynamically.

functional_tests/README.md
AdamHawtin: Yes, I think a combination of `serialised_rollback` and the fix in https://github.com/wagtail/wagtail/pull/12649 would resolve it. I think the core issue was the combination of data flushing and the Wagtail not being able to cope with restoring from scratch with fixtures. However with a quick play copying the solution in our `TransactionTestCase` I still wasn't able to get it passing locally.

I would agree with Dan that we should continue with the current solution and revisit.

functional_tests/docker-compose-dev-app.yml
```diff
@@ -0,0 +1,38 @@
...
+    ports: !override
+      - 18000:8000 # runserver
+    volumes:
+      # Mount the repo at /app/
+      - ../:/app/
+    command: >
+      sh -c "poetry install &&
+             django-admin collectstatic --verbosity 0 --noinput --clear &&
+             django-admin migrate --noinput &&
+             django-admin runserver 0.0.0.0:8000"
```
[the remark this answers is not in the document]
AdamHawtin: These are here specifically to be run within the container as I was having issues with things not being initialised properly when they are only run from the outside environment.

Makefile
AdamHawtin: The `functional-tests` target combines the setup and test run, the purpose of this command is to allow the dev to easily run the tests without the backing containers being torn down post-test.

Makefile
AdamHawtin: Realising it should be passed in with the `--settings` option

AdamHawtin: I'm leaving this as a direct call so that I can supply the `--settings` option.

.github/workflows/ci.yml
AdamHawtin: It looks like it is related, so I'm adding those comments

functional_tests/docker-compose-dev-app.yml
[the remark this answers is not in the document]
MebinAbraham: If removing causes issues, we should investigate why we would need them (can be separate).

Makefile
MebinAbraham: I don't disagree, there was a [convo](https://github.com/ONSdigital/dis-wagtail/pull/38#discussion_r1857286036) on this PR about it. I don't mind either if we are consistent.

Makefile
MebinAbraham: Right I see, then in that case, we just need to ensure DS templates are loaded.

Makefile
MebinAbraham: If that your preference over env var?. It is set as env var elsewhere?.

Makefile
MebinAbraham: Fair point, I'm happy to leave as is.

functional_tests/README.md
MebinAbraham: Agreed, let's revisit if it becomes a problem.

Makefile
AdamHawtin: I'll leave it for now as the Makefile is starting to get a bit bloated and it's easy enough to change down the line if we decide we want to.

Makefile
AdamHawtin: Just checked the docs and the command line flag has highest order of precedence, so for cases like this where we wouldn't want someone's accidentally left over `.env` settings to override it, it would be the best choice. I'll update the other uses to match 👍

Makefile
AdamHawtin: I've renamed to be consistent for now 👍

.github/workflows/ci.yml
```diff
@@ -147,3 +147,53 @@ jobs:
...
+      - name: Install dependencies
+        run: make install-dev
+
+      - name: Playwright Install
+        run: poetry run python -m playwright install --with-deps ${{ matrix.browser }}
+
+      - uses: actions/download-artifact@v4
+        with:
+          name: static
+          path: cms/static_compiled/
```
MebinAbraham, line 189: Why is this needed?. The docker container builds the static assets already?

MebinAbraham (approved):

sanjeevz3009 (approved):
Tested it everything once again after the recent changes. All is good =)

.github/workflows/ci.yml
AdamHawtin: The app container is never being built in the CI functional tests, the app is only run through the `LiveServerTestCase`.

functional_tests/steps/home_page.py
RealOrangeOne: So long as we find a solution for it, I think that's fine.

With the isolation of the `given` blocks, I'm not sure how to handle this either, but it ought to be a solved problem.

functional_tests/README.md
RealOrangeOne: This is true, however it's a fairly complex set of 10 lines. 

Perhaps one to address once the various fixes are included in Django and Wagtail.

RealOrangeOne (approved):
Assuming the various "revisit later" tasks are tracked somewhere, I think this is fine as a first pass to get something in and working.

MebinAbraham (approved):

zerolab:
🎉

PR #35, opened 2024-11-21
Add readiness probe

RealOrangeOne:

https://jira.ons.gov.uk/browse/CMS-215

When running in Kubernetes, it needs a way to identify when the application is "ready" to serve requests. This is done using a [readiness probe](https://kubernetes.io/docs/concepts/configuration/liveness-readiness-startup-probes/#readiness-probe).

There are other probes available, which I've not implemented:

- "Liveness" probe probably isn't useful to us right now. Application processes are cycled regularly, so shouldn't result in deadlocks. Monitoring external infrastructure here (eg database connectivity) could result in death spirals on external issues.
- "Startup" probe may potentially be useful to allow the application to be considered "ready" to serve requests sooner, but since the behaviour would be the same as this readiness probe, it can use the same endpoint.

### How to review

There are no visible changes, just a new endpoint.

The endpoint has intentionally been designed to be as lightweight as possible. No database queries, no templates, just a functional view.

I've added a set of "internal" URLs under the `/-/` prefix. The intention is that these URLs are not externally accessible, and exist only for internal use (monitoring, readiness probes etc). In future, these will be blocked from public use by the infrastructure (somewhere).

### Follow-up Actions

Configuring the endpoint to be used is out of scope of this PR.


cms/core/tests/test_views.py
```diff
@@ -32,3 +33,25 @@ def test_google_tag_manager_script_present(self):
...
+        self.assertEqual(response.content, b"")
+        self.assertEqual(response.templates, [])
+
+    def test_head(self):
+        """Check the ready endpoint works with a HEAD request."""
+        response = self.client.head(self.url)
+
+        self.assertEqual(response.status_code, 204)
+        self.assertEqual(response.content, b"")
+        self.assertEqual(response.templates, [])
```
zerolab, line 57: question: seeing that the tests look at GET/HEAD, should the view forbid all the other methods?

cms/urls.py
```diff
@@ -11,15 +11,19 @@
...
 if TYPE_CHECKING:
     from django.urls import URLPattern, URLResolver
 
+# Internal URLs are not intended for public use.
+internal_urlpatterns = [path("readiness/", core_views.ready, name="readiness")]
 
 # Private URLs are not meant to be cached.
 private_urlpatterns = [
     path("documents/", include(wagtaildocs_urls)),
+    path("-/", include((internal_urlpatterns, "internal"))),
```
zerolab, line 26: question: will this restricted to internal IPs only?

cms/core/tests/test_views.py
RealOrangeOne: I explicitly want `GET` and `HEAD`, since that's what the probes and/or a test `curl` will use. I don't really care about other methods.

We could restrict the method, but I thought it's not worth the extra complexity (sure, there's a decorator), or miniscule performance overhead (premature optimisation, I know).

cms/urls.py
RealOrangeOne: In future, I'd like it to be, yes, if not further. Having the namespace here ready for those kinds of URLs makes it easier to add them in future. It might be YAGNI, but it's not much extra overhead at the moment.

zerolab (approved):

cms/core/tests/test_views.py
MebinAbraham: From a security perspective, we should avoid leaving flows untested. We should restrict to GET/HEAD if that's all we care about.

cms/urls.py
MebinAbraham: My browser doesn't seem to like this?.
Why not just use `internal`, it's explicit. We need to, at some point, do a mapping to ensure our URLs don't overlap with existing URLs on the site.

MebinAbraham (approved):

zerolab (approved):

PR #37, opened 2024-11-21
Connect to elasticache using IAM

RealOrangeOne:

https://jira.ons.gov.uk/browse/CMS-216

For security reasons, it's preferred to connect to Elasticache using IAM, rather than fixed credentials. This PR adds a custom credential provider to defer to IAM, based on [an example from the `redis-py` docs](https://redis.readthedocs.io/en/stable/examples/connection_examples.html#Connecting-to-a-redis-instance-with-ElastiCache-IAM-credential-provider). The notable difference is that rather than using an external cache library, I've reused Django's cache with an in-memory backend.

### How to review

Unfortunately, it's not really possible to fully test this, since it requires external infrastructure. Parts could be mocked out with [`moto`](https://pypi.org/project/moto/), but it's probably not worth it.

### Follow-up Actions

Configuring the infrastructure to require IAM authentication is out of scope for this PR.


cms/settings/base.py
```diff
@@ -196,38 +197,51 @@
...
     }
 
-    CACHES = {
-        "default": {
-            "BACKEND": "django_redis.cache.RedisCache",
-            "LOCATION": REDIS_URL,
-            "OPTIONS": redis_options,
-        }
+elif elasticache_addr := env.get("ELASTICACHE_ADDR"):
+    port = env["ELASTICACHE_PORT"]
```
zerolab, line 225: nitpick: there is no checking for this and cluster name/region.
I know this won't be a problem per se as it will be diligently Terraformed, but still.

cms/core/cache.py
```diff
@@ -36,3 +42,61 @@ def get_default_cache_control_decorator() -> Callable:
...
+            self.region,
+            "elasticache",
+            "v4",
+            session.get_credentials(),
+            session.get_component("event_emitter"),
+        )
+
+        self.cache_key = f"elasticache_{user}_{cluster_name}_{region}"
+
+        self.connection_url = urlunparse(
```
zerolab, line 73: question: the redis example caches the connection URL under `get_credentials`. Is this used anywhere else? If not, why not keep it out of init?

cms/settings/base.py
```diff
@@ -196,38 +197,51 @@
 #
 # Do not use the same Redis instance for other things like Celery!
 
-# Prefer the TLS connection URL over non
-REDIS_URL = env.get("REDIS_TLS_URL", env.get("REDIS_URL"))
+CACHES: dict = {"memory": {"BACKEND": "django.core.cache.backends.locmem.LocMemCache", "LOCATION": "memory"}}
```
zerolab, line 200: suggestion:

```suggestion
# The memory cache will be used by ElastiCacheIAMCredentialProvider when configured
CACHES: dict = {"memory": {"BACKEND": "django.core.cache.backends.locmem.LocMemCache", "LOCATION": "memory"}}
```

or something along those lines.

zerolab (changes requested):

cms/core/cache.py
```diff
@@ -36,3 +42,61 @@ def get_default_cache_control_decorator() -> Callable:
     """
     cache_control_kwargs = get_default_cache_control_kwargs()
     return cache_control(**cache_control_kwargs)
+
+
+class ElastiCacheIAMCredentialProvider(redis.CredentialProvider):
```
MebinAbraham, line 47: We will want tests using mocks for this class.

cms/core/cache.py
RealOrangeOne: I moved it to the constructor so it's only generated once, since it doesn't change per call or based on any parameters.

cms/settings/base.py
RealOrangeOne: I don't think that's useful. The memory cache can be used by anything which needs it (sure, it might not be that many things in reality), but putting the comment there means it's more likely to get out of date.

zerolab: I suppose this will be mitigated when we have the tests for the new credential provider. What I want to avoid is someone removing it down the line only to discover it explodes

cms/settings/base.py
zerolab: Fair enough

cms/settings/base.py
RealOrangeOne: Yep let's solve this with tests instead!

cms/core/cache.py
```diff
@@ -36,3 +42,60 @@ def get_default_cache_control_decorator() -> Callable:
...
+                netloc=self.cluster_name,
+                path="/",
+                query=urlencode({"Action": "connect", "User": self.user}),
+                fragment="",
+            )
+        )
+
+    def get_credentials(self) -> tuple[str, str]:
+        """Get credentials from IAM."""
+        if (signed_url := caches["memory"].get(self.cache_key)) is None:
```
MebinAbraham, line 85: We will have various other uses for in-memory cache. What is your preference? Historically, we have used TTL cache from cachetools; given we have a local memory cache, it seems reasonable to use it. It would be nice to have a decorator to handle the getting/setting to cache using our mem cache.

MebinAbraham (approved):
Unable to test but LGTM.

cms/core/cache.py
```diff
@@ -36,3 +42,60 @@ def get_default_cache_control_decorator() -> Callable:
...
+            self.region,
+            "elasticache",
+            "v4",
+            session.get_credentials(),
+            session.get_component("event_emitter"),
+        )
+
+        self.cache_key = f"elasticache_{user}_{cluster_name}_{region}"
+
+        self.connection_url = urlunsplit(
```
MebinAbraham, line 73: Minor, let's make any attributes not needed to external callers private.

zerolab (approved):

cms/core/cache.py
RealOrangeOne: In some cases, `functools.lru_cache` is generally better. We could add a decorator around this, but in most cases we'd probably want the cache centralised through `redis`, rather than in memory.

cms/core/cache.py
RealOrangeOne: Python doesn't really have private attributes. Prefixing with understores might hint it, but then requires a lot of ignores as part of tests. Only `get_credentials` is part of the public API.

cms/settings/base.py
```diff
@@ -196,38 +197,51 @@
...
 
-if REDIS_URL:
-    connection_pool_kwargs: dict[str, str | None] = {}
+redis_options = {
+    "IGNORE_EXCEPTIONS": True,
+    "SOCKET_CONNECT_TIMEOUT": 2,  # seconds
+    "SOCKET_TIMEOUT": 2,  # seconds
```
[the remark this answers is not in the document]
RealOrangeOne: The timeouts are simply to ensure that if there are network issues with the cache, skip it, as opposed to waiting, since it'd replace the benefits of having a cache in the first place.

cms/core/cache.py
MebinAbraham: I know but we have linters that do enforce it...
But not a blocker.

cms/core/cache.py
MebinAbraham: I mean specially for cases where memory is sufficient as we don't want a network call and are happy with cache per process like here.

cms/core/cache.py
AdamHawtin: I would also advocate for marking things as "private" in cases like this. While not enforced, it's useful to communicate the usage intent of the attributes in a way that can't be missed by future developers, and can be caught in linting.

cms/core/cache.py
RealOrangeOne: Yeah `lru_cache` doesn't take a TTL :( 

I've added https://github.com/peterbe/django-cache-memoize, and used that for the memory cache.

MebinAbraham (approved):

sanjeevz3009 (commented):
Reviewed! No comments!

PR #38, opened 2024-11-22
Run CMS locally with supporting services in Docker 

kacperpONS:

Added a way to run Wagtail outside of Docker with the supporting services still running in containers. 

Changes made:
- added convenience make commands
- added an example `.env` file for the local config 
- added a new docker-compose for just the supporting services
- added documentation in the README

### How to review

Run the steps under **Running Wagtail locally with supporting services in Docker** and make sure that they work as expected. 

### Follow-up Actions

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,8 +119,8 @@ Follow these steps to set up and run the project using Docker.
 
     Once the containers are running, you need to manually start Django from within the web container. This allows for running both the Django server and any additional background services (e.g., schedulers).
 
-    ⚠️ WARNING  
-    The `honcho` command will pick up your local mounted `.env` file when running via `docker-compose`. Ensure that you comment out any variables in the `.env` file which might cause clashes in the container context as they will take precedence when running `honcho start`.
+    > ⚠️ WARNING  
+    >  The `honcho` command will pick up your local mounted `.env` file when running via `docker-compose`. Ensure that you comment out any variables in the `.env` file which might cause clashes in the container context as they will take precedence when running `honcho start`.
 
     ```bash
     # Start both Django and the scheduler using Honcho
@@ -136,7 +136,7 @@ You can then access the admin at `http://0.0.0.0:8000/admin/` or `http://localho
 
 You can also run the main application locally with the supporting backend services such as the Postgres and Redis running in Docker. This can be useful when you want to make changes that require the app to be restarted in order to be picked up.
 
-For this method you can specify the runtime configuration either in your IDE (for PyCharm see [here](https://www.jetbrains.com/help/pycharm/run-debug-configuration.html#createExplicitly)), or copy the `.env-dev-example` and rename it to `.env` which will allow Django to pick up the config.
+For this method you can specify the runtime configuration either in your IDE (for PyCharm see [here](https://www.jetbrains.com/help/pycharm/run-debug-configuration.html#createExplicitly)), or copy the `.development.env` and rename it to `.env` which will allow Django to pick up the config.
 
 Once you create the `.env` file, and you'd like to switch back to running the application in a container, the `.env` file will be accessible inside the containers and it will be picked up by the `honcho` command. In order to avoid conflicts you may need to comment out some variables (such as `DATABASE_URL` and `REDIS_URL`) in the `.env` file.
 


README.md
```diff
@@ -126,6 +126,44 @@ Follow these steps to set up and run the project using Docker.
 
 You can then access the admin at `http://0.0.0.0:8000/admin/` or `http://localhost:8000/admin/`.
 
+#### Running Wagtail locally with supporting services in Docker
+
+You can also run the main application locally with the supporting backend services such as the Postgres and Redis running in Docker. This can be useful when you want to make changes that require the app to be frequently restarted to be picked up.
+
+In order tu run it:
```
nehakerung, line 133: to instead of tu?

docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
       context: .
       target: dev
     env_file:
-      - path: .env
+      - path: .docker-compose-env
```
RealOrangeOne, line 7: Question: Why does this need to be a different `.env` file?

kacperpONS: The initial idea here was to allow to distinguish between the local and containerised config files and allow them to run with different settings. On second thought it's probably mostly pointless as you could just overwrite the setting in the `environment` block.

Makefile
```diff
@@ -85,8 +85,8 @@ docker-build: load-design-system-templates  ## Build Docker container
 	docker compose pull
 	docker compose build
 
-.PHONY: docker-start
-docker-start:  ## Start Docker containers
+.PHONY: docker-up
+docker-up:  ## Start Docker containers
```
nehakerung, line 89: Is it necessary to remove docker-start as it's still part of README instructions on Using Docker - Build and Start the Container?

kacperpONS: Well spotted! It's not actually removed, I've only renamed the make command so that it's consistent with the Docker command it uses. I forgot to rename it in the README so will do that 👍

Makefile
```diff
@@ -85,8 +85,8 @@ docker-build: load-design-system-templates  ## Build Docker container
 	docker compose pull
 	docker compose build
 
-.PHONY: docker-start
-docker-start:  ## Start Docker containers
+.PHONY: docker-up
```
sanjeevz3009, line 88: You haven't updated the README to reflect this change: `README line 95`. It still says `make docker-start`, it should be `make docker-up` now due to this change.

sanjeevz3009 (changes requested):

Makefile
sanjeevz3009: Make sure to update the README to reflect this change as well! I have requested changes for this =)

Makefile
MebinAbraham: I propose using `compose` instead of `docker` in our command names. Starting with an action word makes commands clearer and more intuitive. For instance, `make compose-up` / `make compose-up-dev` is more descriptive than `make docker-up` / `make docker-dev-up`. The term `docker-up` could be misinterpreted as starting the Docker daemon itself, whereas compose better conveys the intent of using Docker Compose. In past projects, we’ve used compose, and developers found it clearer and more meaningful.

If retaining the term docker is important, we could consider `docker-compose-up` for clarity, though shorter commands are generally preferable if they remain meaningful. See what the team thinks and go with the consensus.

---

Also, I recommend ensuring commands follow consistent prefixes. This approach improves shell autocomplete functionality by grouping related commands, aligning with common patterns for shell commands. For example, using `makemigrations` and `makemigrations-check` is more intuitive than `makemigrations` and `check-makemigrations`. This patterns allows developers to easily discover and understand available commands.

Lastly, I suggest dropping `make` from command names like `makemigrations`. Typing `make makemigrations` feels redundant and less natural. Instead, aliases like `make migrations` and `make migrate` are more straightforward and intuitive. While the command names don’t need to perfectly mirror the underlying operations, they should be clear and easy to use. Again, see what the team prefers and go with the consensus.

docker-compose-dev.yml
```diff
@@ -0,0 +1,29 @@
...
+    ports:
+      - 5432:5432
+
+  redis:
+    image: redis:7.2
+    expose:
+      - 6379
+    attach: false
+    ports:
+      - 6379:6379
```
MebinAbraham, line 26: If you change the env vars to use localhost, you don't need both `expose` and `ports`. It can be misleading to define both.

.env-dev-example
```diff
@@ -0,0 +1,4 @@
+DJANGO_SETTINGS_MODULE=cms.settings.dev
```
MebinAbraham, line 1: We should keep the extension as `.env` so I suggest renaming to `.development.env`. It is not just an example per-se, this can be used as is for development as is.

Makefile
```diff
@@ -100,3 +100,43 @@ docker-shell:  ## SSH into Docker container
...
+.PHONY: check-makemigrations
+check-makemigrations: ## Convenience alias to run makemigrations in venv
+	poetry run python ./manage.py makemigrations --check
+
+.PHONY: makemigrations
+makemigrations: ## Convenience alias to run makemigrations in venv
+	poetry run python ./manage.py makemigrations
+
+.PHONY: collectstatic
+collectstatic:  ## Convenience alias to run collectstatic in venv
```
MebinAbraham, line 129: We should replace these with comments that describe what the command does. All make targets are convenience aliases so this doesn't add much value.

README.md
```diff
@@ -266,13 +304,13 @@ Below are the commands you will most commonly use, note that these have to be ru
...
-poetry run django-admin makemigrations --check
+make check-makemigrations
 
 # Generate migrations
-poetry run django-admin makemigrations
+make makemigrations
 
 # Apply migrations. Needed if new migrations have been generated (either by you, or via upstream code)
-poetry run django-admin migrate
+make migrate
```
MebinAbraham, line 313: I suggest replacing all the `dj` alias commands like `djrun` to use the makefile target instead so we have a consistent commands to run regardless of whether we are running via docker or locally.

MebinAbraham (commented):
Good work, left some initial comments.

Makefile
kacperpONS: I agree with the idea to replace `docker` with `compose` in the Make commands where appropriate.

I think that it makes more sense to have the prefix of the dev commands as `compose-dev...` so in this example it'd be `make compose-dev-up` rather than `make compose-up-dev`. 
This would mean that the shell autocomplete would be performed for all the `compose-dev-...` commands and print all the available dev commands rather than run for `compose-up-...` which would only print the dev and non-dev variations of the command.

kacperpONS: Re. the last paragraph, personally I'd still stick to using `makemigrations` rather than `migrations` as name of the make command.

Semantically the `make` command is completely separate to the "make" in `makemigations` - it doesn't replace it and hence we shouldn't try to trade clarity for conciseness.

Secondly, in your comment you highlight that using verbs in the make commands makes them more intuitive and easier to read. I think that's the most convincing argument for me here.

Thirdly IMO `migrations` and `migrate` would be dangerously close to each other - at first sight you're not able to tell what each one is actually doing. 

Also, for all the other all the other Make recipes we use exactly the same naming to `django-admin` commands. I think it's beneficial to keep it consistent.

I suggest that the reviewers react to this comment to show their preference:
👍 - rename it to `make migrate`
👎  - keep it as `make makemigrations`

MebinAbraham: > I agree with the idea to replace `docker` with `compose` in the Make commands where appropriate.
> 
> I think that it makes more sense to have the prefix of the dev commands as `compose-dev...` so in this example it'd be `make compose-dev-up` rather than `make compose-up-dev`. This would mean that the shell autocomplete would be performed for all the `compose-dev-...` commands and print all the available dev commands rather than run for `compose-up-...` which would only print the dev and non-dev variations of the command.

LGTM.

MebinAbraham: > Re. the last paragraph, personally I'd still stick to using `makemigrations` rather than `migrations` as name of the make command.
> 
> Semantically the `make` command is completely separate to the "make" in `makemigations` - it doesn't replace it and hence we shouldn't try to trade clarity for conciseness.
> 
> Secondly, in your comment you highlight that using verbs in the make commands makes them more intuitive and easier to read. I think that's the most convincing argument for me here.
> 
> Thirdly IMO `migrations` and `migrate` would be dangerously close to each other - at first sight you're not able to tell what each one is actually doing.
> 
> Also, for all the other all the other Make recipes we use exactly the same naming to `django-admin` commands. I think it's beneficial to keep it consistent.
> 
> I suggest that the reviewers react to this comment to show their preference: 👍 - rename it to `make migrate` 👎 - keep it as `make makemigrations`

Yup, I don't have a strong view, happy with either, the main ask is that, it is clear to devs so go with the consensus.

README.md
```diff
@@ -126,6 +126,44 @@ Follow these steps to set up and run the project using Docker.
 
 You can then access the admin at `http://0.0.0.0:8000/admin/` or `http://localhost:8000/admin/`.
 
+#### Running Wagtail locally with supporting services in Docker
```
sanjeevz3009, line 129: Add this to table of contents please =)

Makefile
kacperpONS: Fixed the typo 👍

sanjeevz3009 (approved):
Left some minor comments and suggestions. Pulled down the branch, tested everything locally and all seems to be good. LGTM!

README.md
kacperpONS: 🫡

README.md
```diff
@@ -126,6 +126,44 @@ Follow these steps to set up and run the project using Docker.
...
+4. Run the Django server locally via your IDE or with the following command:
+
+```
+make runserver
+```
+
+You can specify the runtime configuration either in your IDE (for PyCharm see [here](https://www.jetbrains.com/help/pycharm/run-debug-configuration.html#createExplicitly)), or copy the `.env-dev-example` and rename it to `.env` which will allow Django to pick up the config.
+
+> [!NOTE]
+> When running the application in a virtual environment via Poetry the `.env` file will not be picked up automatically. For this to work you'll need to install the [poetry-plugin-dotenv](https://github.com/pivoshenko/poetry-plugin-dotenv). However if you installed Poetry with `brew` rather than `pip` that currently isn't going to work (see the [issue](https://github.com/pivoshenko/poetry-plugin-dotenv/issues/327)) and you'll need to install an older and seemingly no longer maintained [poetry-dotenv-plugin](https://github.com/mpeteuil/poetry-dotenv-plugin).
```
sanjeevz3009, line 165: Nit: This maybe helpful for other devs. Maybe you can add a line about this in the `README` file.

I have installed Poetry using [the official installer](https://python-poetry.org/docs/#installing-with-the-official-installer). I think the recommended way is Pipx or using the official installer.

Not sure installing Poetry using pip is a good idea. https://github.com/orgs/python-poetry/discussions/8687

Since I used the official installer, I didn't have to do any of the steps in the `README` =)

sanjeevz3009: Also wouldn't this be a dev dependency? If we truly need this, which I didn't need due to installing it using the official installer.

docker-compose-dev.yml
kacperpONS: I think that the right way to do here is to remove the `expose` bit and leave the `ports`.

If I understand it correctly the `expose` is analogous to the [EXPOSE in a Dockerfile](https://docs.docker.com/reference/dockerfile/#expose) in that:
> "The EXPOSE instruction doesn't actually publish the port. It functions as a type of documentation between the person who builds the image and the person who runs the container, about which ports are intended to be published." 

and that the actual port forwarding needs to be set up as a setting when running `docker run` or in the `ports` bit in a compose file.

An experiment confirmed that in fact the `expose` is not necessary so I'm going to remove it.

README.md
sanjeevz3009: I really don't have a strong preference haha.

I mean I agree with the approach that @MebinAbraham is suggesting but when it comes to removing them from `/.docker/bashrc.sh` it makes to get rid of it, unless there is a solid reason to keep those still.

zerolab: The `dj` aliases are something we've been using at Torchbox as a shortcut, so not wedded to it

Makefile
```diff
@@ -80,23 +80,72 @@ megalint:  ## Run the mega-linter.
 load-design-system-templates:  ## Load the design system templates
 	./scripts/load-design-system-templates.sh $(DESIGN_SYSTEM_VERSION)
 
-.PHONY: docker-build
-docker-build: load-design-system-templates  ## Build Docker container
-	docker compose pull
+# Docker and docker compose make commands
+
+.PHONY: compose-build
```
zerolab, line 85: one for a follow-up PR: I think we're at a point where we want to have a single `.PHONY` declaration at the top of the file rather thant at the top of each command.

zerolab (commented):
Lots of good work here.
Left a few notes and a suggestion for fixing a typo.

Makefile
kacperpONS: Yeah I was thinking the same considering we now how a Makefile that's 150 lines long.. 👀 
However the way we do it now makes it easier to maintain as you can quickly add/ confirm that the recipe is marked as PHONY.
So I guess that's up for debate - let's see how the current way works for us and we might change it in the future.

README.md
```diff
@@ -121,11 +125,49 @@ Follow these steps to set up and run the project using Docker.
...
+make compose-dev-up
+```
+
+3. Run the following pre-run steps:
+
+```
+make makemigrations
+make collectstatic
+make migrate
+make createsuperuser
```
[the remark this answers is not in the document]
kacperpONS: Yeah, another thing I wanted to add, but I was concerned that the Makefile is getting bloated 😄 

If we're adding this, what do you think about also adding a makefile recipe for the [one liner](https://github.com.mcas.ms/torchbox/wagtailmedia/blob/main/tox.ini#L77) to create an admin user that you've linked at the tech session last week? That would make the `make dev-init` command even easier to run as it won't stop waiting for input.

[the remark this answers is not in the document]
zerolab: we spoke in Slack. yes to adding it to `make dev-init`

nehakerung:
LGTM - Pulled down the branch, tested locally and all is well. Readme was good to follow.

README.md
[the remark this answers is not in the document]
kacperpONS: We've talked about it and I'll add it add it as part of `make dev-init` rather than adding a new `make` recipe.

zerolab (approved):

Makefile
```diff
@@ -80,23 +80,76 @@ megalint:  ## Run the mega-linter.
...
+createsuperuser: ## Create a super user 
+	poetry run python ./manage.py createsuperuser
+
+.PHONY: runserver
+runserver: ## Run the Django application locally	
+	poetry run python ./manage.py runserver
+
+.PHONY: dev-init 
+dev-init: makemigrations collectstatic migrate ## Run the pre-run setup scripts
+	poetry run python ./manage.py shell -c "from cms.users.models import User;(not User.objects.filter(username='admin').exists()) and User.objects.create_superuser('admin', 'super@example.com', 'changeme')"
```
RealOrangeOne, line 155: Question: Is there a reason to do this over `createsuperuser`?

RealOrangeOne (approved):

kacperpONS:
For people who already have given this PR an approval:

I've noticed that if you were to follow the steps in the README about adding a `.env` file in your codebase, the old way of running everything in Docker would stop working. 
This is caused by the fact that when you run the `honcho start` command inside the `web` container, `honcho` will pick up the local `.env` file, because it's [mounted](https://github.com/ONSdigital/dis-wagtail/blob/6d9d6b501baa45ab974efced5d7bea52b7cad579/docker-compose.yml#L26) to the `web` container when it runs via `docker-compose`. 
This would then lead to the `DATABASE_URL` and `REDIS_URL` env variables to be set to localhost, rather than to their containers.

I've talked to @zerolab and we've decided that in order to save time rather than try to work around this issue, we'll put a note about that in the README.

README.md
```diff
@@ -86,46 +87,87 @@ Ensure you have the following installed:
...
+```
+
+4. Run the Django server locally via your IDE or with the following command:
+
+```
+make runserver
+```
+
+You can specify the runtime configuration either in your IDE (for PyCharm see [here](https://www.jetbrains.com/help/pycharm/run-debug-configuration.html#createExplicitly)), or copy the `.env-dev-example` and rename it to `.env` which will allow Django to pick up the config.
+Note that once you create the `.env` file, and you'd like to switch back to running the application in a container with `make compose-up`, the `.env` file will be accessible inside the containers and it will be picked up by the `honcho` command. In order to avoid conflicts you should comment out the `DATABASE_URL` and `REDIS_URL` variables in the `.env` file.
```
RealOrangeOne, line 166:
```suggestion
Note that once you create the `.env` file, and you'd like to switch back to running the application in a container with `make compose-up`, the `.env` file will be accessible inside the containers and it will be picked up by the `honcho` command. In order to avoid conflicts you may need to comment out some variables (such as `DATABASE_URL` and `REDIS_URL`).
```

RealOrangeOne (approved):

zerolab (approved):

README.md
```diff
@@ -86,46 +87,87 @@ Ensure you have the following installed:
...
-    dj createsuperuser
+    make createsuperuser
     ```
 
 3. **Start Django Inside the Container**
 
     Once the containers are running, you need to manually start Django from within the web container. This allows for running both the Django server and any additional background services (e.g., schedulers).
 
+> [!WARNING]  
+> The `honcho` command will pick up your local mounted `.env` file when running via `docker-compose`. Ensure that you comment out any local variables which might cause clashes in the container context as they will take precedence when running `honcho start`.
```
MebinAbraham, line 123: This breaks the indentation and doesn't escape the markdown. I suggest just using a warning emoji, as it doesn't look like you can use indented panels.
image

README.md
```diff
@@ -86,46 +87,87 @@ Ensure you have the following installed:
...
-1. **Build and Start the Container**
+1. **Build and Start the Containers**
 
     ```bash
-    # build the container
+    # pull the supporting containers
+    make compose-pull
+
+    # build the main application image
     make docker-build
```
MebinAbraham, line 97:
```suggestion
    make compose-build
```

README.md
```diff
@@ -86,46 +87,87 @@ Ensure you have the following installed:
...
+make dev-init
+```
+
+4. Run the Django server locally via your IDE or with the following command:
+
+```
+make runserver
+```
+
+You can specify the runtime configuration either in your IDE (for PyCharm see [here](https://www.jetbrains.com/help/pycharm/run-debug-configuration.html#createExplicitly)), or copy the `.env-dev-example` and rename it to `.env` which will allow Django to pick up the config.
```
MebinAbraham, line 165: `.env-dev-example` has been renamed to `.development.env`

Makefile
```diff
@@ -80,23 +80,76 @@ megalint:  ## Run the mega-linter.
...
+.PHONY: createsuperuser
+createsuperuser: ## Create a super user 
+	poetry run python ./manage.py createsuperuser
+
+.PHONY: runserver
+runserver: ## Run the Django application locally	
+	poetry run python ./manage.py runserver
+
+.PHONY: dev-init 
+dev-init: makemigrations collectstatic migrate ## Run the pre-run setup scripts
```
MebinAbraham, line 154: Minor, let's make collectstatic either first or last so the migration command outputs are together and not separated by verbose output by `collectionstatic`

MebinAbraham, line 154: This could also include `load-design-system-templates`

README.md
```diff
@@ -86,46 +87,87 @@ Ensure you have the following installed:
...
+2. Start the supporting services in Docker.
+
+```
+make compose-dev-up
+```
+
+3. Run the below command to run the necessary pre-run steps:
+
+```
+make dev-init
```
MebinAbraham, line 156: Either `load-design-system-templates` needs to be added to `dev-init` or needs to be added as a new step.

README.md
```diff
@@ -86,46 +87,87 @@ Ensure you have the following installed:
...
+Note that once you create the `.env` file, and you'd like to switch back to running the application in a container with `make compose-up`, the `.env` file will be accessible inside the containers and it will be picked up by the `honcho` command. In order to avoid conflicts you should comment out the `DATABASE_URL` and `REDIS_URL` variables in the `.env` file.
+
+> [!NOTE]
+> When running the application in a virtual environment via Poetry the `.env` file will not be picked up automatically. For this to work you'll need to install the [poetry-plugin-dotenv](https://github.com/pivoshenko/poetry-plugin-dotenv). However if you installed Poetry with `brew` rather than `pip` that currently isn't going to work (see the [issue](https://github.com/pivoshenko/poetry-plugin-dotenv/issues/327)) and you'll need to install an older and seemingly no longer maintained [poetry-dotenv-plugin](https://github.com/mpeteuil/poetry-dotenv-plugin).
```
MebinAbraham: We should move both of these up as they are required before the app can run. We should inform the users of anything they need before they need it. These are essentially prerequisites.

Makefile
```diff
@@ -80,23 +80,76 @@ megalint:  ## Run the mega-linter.
...
+collectstatic:  ## Collect static files from all Django apps
+	poetry run python ./manage.py collectstatic --verbosity 0 --noinput --clear
+	
+.PHONY: migrate
+migrate: ## Apply the database migrations
+	poetry run python ./manage.py migrate 
+
+.PHONY: createsuperuser
+createsuperuser: ## Create a super user 
+	poetry run python ./manage.py createsuperuser
```
MebinAbraham, line 147: Is there any reason not to use the auto-create user py script here and reference it in `dev-init`? I don't really see an occasion where I need to manually create a user. Additional users can also be created via the UI?.

README.md
MebinAbraham, line 156: This should mention that an admin user will be created with user `admin` and pass `changeme`; otherwise, they won't know unless they dig around.

MebinAbraham (changes requested):

README.md
kacperpONS: Added to `dev-init`

Makefile
kacperpONS: It also tends to fail often so I'll put it right after `load-design-system-templates` so it fails early in the chain.

Makefile
kacperpONS: Yeah, I don't see why it might be useful, so I'll remove it 👍

README.md
```diff
@@ -86,46 +87,87 @@ Ensure you have the following installed:
...
+3. Run the below command to run the necessary pre-run steps:
+
+```
+make dev-init
+```
+
+4. Run the Django server locally via your IDE or with the following command:
+
+```
+make runserver
```
nehakerung, line 162: The current README has `dj runserver 0.0.0.0:8000` so should it not be `make runserver 0.0.0.0:8000` ? I don't suppose this will have a large effect but thought it be better to point out.

zerolab: The default is 127.0.0.1:8000 [ref](https://docs.djangoproject.com/en/5.1/ref/django-admin/#runserver), so this is fine

README.md
```diff
@@ -86,46 +87,95 @@ Ensure you have the following installed:
...
     # Create a superuser for accessing the admin interface
-    dj createsuperuser
+    make createsuperuser
     ```
 
 3. **Start Django Inside the Container**
 
     Once the containers are running, you need to manually start Django from within the web container. This allows for running both the Django server and any additional background services (e.g., schedulers).
 
+    ⚠️ WARNING
```
MebinAbraham, line 122: You can still use the panel.
```suggestion
    > ⚠️ WARNING  
```

README.md
MebinAbraham: This is still outstanding.

MebinAbraham (approved):

README.md
kacperpONS: I swear I've changed that 🤔 Done now!

README.md
kacperpONS: Done, `make format` enforces that we put a chevron before the text on line 123 as well

PR #41, opened 2024-11-26
Main Menu

sanjeevz3009:

Add a navigation app in Wagtail to manage the main menu, allowing for structured, customisable navigation.

### How to review

- Spin up the ONS BETA Web App Locally as normal.
- Snippets -> Main Menu -> Add main menu.
- Fill in the menu content -> After saving every time you should see the preview on the preview panel.
- Go to Settings -> Navigation Settings ->  Choose main menu -> Choose the menu shown so it can be displayed on the website.

Loom walkthrough of the main menu: [Main Menu Features Loom Walkthrough](https://www.loom.com/share/8d9f8f883a194448ab125332e80547f3?sid=01b18433-87d8-40f9-8796-b8c8d973ddd4)

Credit: @zerolab for the help and guidance.

### Follow-up Actions

Fix column alignment issues (can be seen during the loom video): https://github.com/ONSdigital/dis-wagtail/pull/63


cms/navigation/models.py
```diff
@@ -0,0 +1,115 @@
...
+        return value
+
+    class Meta:
+        icon = "star"
+        label = "Highlight"
+
+
+# Section StructBlock for columns
+class SectionBlock(StructBlock):
+    section_link = LinkBlock(help_text="Main link for this section (Theme pages or external URLs).")
```
zerolab, line 51: this may need a separate block definition as the requirement says the PageChooserBlock should be limited to Theme page ([Wagtail docs](https://docs.wagtail.org/en/stable/reference/streamfield/blocks.html#wagtail.blocks.PageChooserBlock))

cms/navigation/models.py
```diff
@@ -0,0 +1,115 @@
...
+        FieldPanel("highlights"),
+        FieldPanel("columns"),
+    ]
+
+    def __str__(self):
+        return "Main Menu"
+
+    class Meta:
+        verbose_name = "Main Menu"
+        verbose_name_plural = "Main Menus"
```
zerolab, line 94: These should not be needed. See https://docs.djangoproject.com/en/5.1/ref/models/options/#verbose-name

Howeve, if you want to make it translateable, then please follow https://docs.djangoproject.com/en/5.1/topics/i18n/translation/#model-verbose-names-values

zerolab (commented):
I know this is a draft. Thought I'd leave a few notes to boot

cms/navigation/models.py
```diff
@@ -0,0 +1,132 @@
...
+
+# Custom LinkBlock to support both pages and URLs
+class BaseLinkBlock(StructBlock):
+    page = PageChooserBlock(required=False)
+    url = URLBlock(required=False, label="URL")
+    title = CharBlock(required=False, help_text="Optional. Displayed as the link text.")
+
+    def clean(self, value: dict) -> dict:
+        value = super().clean(value)
+        if not value.get("page") and not value.get("url"):
```
helenb, line 21: Some validation is needed when there is both a url and a page added, as this should not be possible.

cms/navigation/models.py
```diff
@@ -0,0 +1,132 @@
...
+class TopicLinkBlock(BaseLinkBlock):
+    page = PageChooserBlock(required=False, page_type="topics.TopicPage")
+
+    class Meta:
+        icon = "link"
+        label = "Topic Link"
+
+
+class HighlightsBlock(BaseLinkBlock):
+    description = CharBlock(required=True, max_length=50, help_text="E.g., It's never been more important")
```
helenb, line 52: Suggestion: use 'View our latest and upcoming releases.' as the example help text as this is perhaps more meaningful to an editor, whereas the current example is very specific to a survey link. (Note that this example includes a full stop as Amelia confirmed these descriptions need a full stop for the prototype)

cms/navigation/models.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def clean(self, value: dict) -> dict:
+        value = super().clean(value)
+        if not value.get("page") and not value.get("url"):
+            raise ValidationError("Either a page or a URL must be provided.")
```
helenb, line 57: As above we need to add validation for the case when both are added.

helenb (commented):
Thanks @sanjeevz3009 I have tested locally and it all functions well. A couple of small comments on the validation and example help text.

sanjeevz3009:
> Thanks @sanjeevz3009 I have tested locally and it all functions well. A couple of small comments on the validation and example help text.

@helenb Will push my latest code up as I haven't done that. It should be way better =) Might resolve some of the issues raised by you. Thanks Helen, will sort out any issues not addressed already locally and then push it up.

cms/navigation/models.py
sanjeevz3009: Done!

cms/navigation/models.py
sanjeevz3009: Done!

cms/navigation/models.py
```diff
@@ -0,0 +1,115 @@
...
+from wagtail.admin.panels import FieldPanel
+from wagtail.blocks import CharBlock, ListBlock, PageChooserBlock, StructBlock, URLBlock
+from wagtail.contrib.settings.models import BaseSiteSetting, register_setting
+from wagtail.snippets.models import register_snippet
+
+from cms.core.fields import StreamField
+
+
+# Custom LinkBlock to support both pages and URLs
+class LinkBlock(StructBlock):
```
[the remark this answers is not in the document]
sanjeevz3009: Done

cms/navigation/models.py
sanjeevz3009: Done!

cms/navigation/models.py
sanjeevz3009: Done

helenb:
I don't think the menu is using the styling quite as expected - compare my test with the menu settings (first screenshot) with what is in the prototype (second screenshot). I'm happy to pick that up as a follow-up task.

Screenshot 2024-12-09 at 15 26 12

Screenshot 2024-12-09 at 15 27 37

sanjeevz3009:
@helenb 

Thanks for testing this. I fixed the link issues.

Now adding main menu content and it being usable on the menu on the website should be all good.

Screenshot 2024-12-09 at 18 27 21

So now we are just left with alignment front-end issues.

- Where if you have only 2 columns the alignment looks weird and goes out of place slightly compared to 3 columns.
- If you have 3 sections and the 1st one has 5 topics and then the rest of the sections have 3 topics then the the section alignment for that column goes off as well.
- So there are similar issues like the above.

Would be amazing if you could follow that up on a task. Happy to demo the front-end styling issues to you.

Thanks =)

cms/navigation/models.py
sanjeevz3009: Removed and gettext_lazy added.

cms/navigation/models.py
sanjeevz3009: Done =)

cms/navigation/models.py
```diff
@@ -0,0 +1,102 @@
...
+
+if TYPE_CHECKING:
+    from django.http import HttpRequest
+
+
+class ThemeLinkBlock(LinkBlock):
+    page = PageChooserBlock(required=False, page_type="themes.ThemePage")
+
+    class Meta:
+        label = "Theme Link"
```
zerolab, line 21: as discussed, for consistency, we need to mark for translation any strings that will be displayed to the user

zerolab (commented):
A few more notes

cms/navigation/models.py
sanjeevz3009: Done!

cms/navigation/models.py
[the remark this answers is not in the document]
sanjeevz3009: The code no longer exists!

helenb:
> @helenb
> 
> Thanks for testing this. I fixed the link issues.
> 
> As previously for highlights when a user selects a Wagtail page it didn't turn it into a link. Now it does. Same for the topic page.
> 
> Also fixed some other edge cases I missed out on previously, so now all should be good on that side.
> 
> Let me know if there are any further issues.
> 
> Now adding main menu content and it being usable on the website menu should be all good.
> 
> Screenshot 2024-12-09 at 18 31 58
> 
> So now we are just left with front-end alignment issues.
> 
> * Where if you have only 2 columns the alignment looks weird and goes out of place slightly compared to 3 columns.
> * If you have 3 sections and the 1st one has 5 topics and then the rest of the sections have 3 topics then the the section alignment for that column goes off as well.
> * So there are similar issues like the above.
> 
> Would be amazing if you could follow that up on a task. Happy to demo the front-end styling/ alignment issues to you.
> 
> Thanks =)

Thanks @sanjeevz3009 - once you've merged this PR, you could re-assign the ticket to me to do the front-end tweaks.

cms/jinja2/templates/base.html
```diff
@@ -35,6 +35,59 @@
     {%- endwith -%}
 {%- endset -%}
 
+{% set navigation_settings = settings.navigation.NavigationSettings %}
+
+{% set keyLinksList = [] %}
+
+{% if navigation_settings.main_menu and navigation_settings.main_menu.highlights %}
```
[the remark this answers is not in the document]
zerolab: @sanjeevz3009 don't forget about this one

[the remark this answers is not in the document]
zerolab: discussed. let's do a follow up PR for this (and ticket)

cms/jinja2/templates/base.html
```diff
@@ -35,6 +35,59 @@
...
+    {% for highlight in navigation_settings.main_menu.highlights %}
+        {% if highlight.value.page %}
+            {% set _ = keyLinksList.append({
+                'text': highlight.value.title if highlight.value.title else highlight.value.page.title,
+                'description': highlight.value.description,
+                'url': highlight.value.page.url,
```
zerolab, line 48: replaces such usages with `pageurl(highlight.value.page)`

cms/jinja2/templates/base.html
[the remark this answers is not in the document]
sanjeevz3009: Will create a card for this!

cms/navigation/models.py
```diff
@@ -0,0 +1,133 @@
...
+        blank=True,
+        max_num=3,
+        help_text=_("Up to 3 columns. Each column contains sections with links."),
+    )
+
+    panels: ClassVar[list] = [
+        FieldPanel("highlights"),
+        FieldPanel("columns"),
+    ]
+    max_num = 3
```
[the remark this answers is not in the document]
sanjeevz3009: Removed!

cms/jinja2/templates/base.html
sanjeevz3009: Done!

cms/navigation/wagtail_hooks.py
```diff
@@ -0,0 +1,32 @@
...
+    from cms.users.models import User
+
+
+class NoAddModelPermissionPolicy(ModelPermissionPolicy):
+    """Model permission that doesn't allow creating more than one main menu instance."""
+
+    def user_has_permission(self, user: Optional["User"] = None, action: str | None = None) -> bool:
+        if action == "add" and MainMenu.objects.exists():
+            return False
+        return user is not None and user.has_perm(self._get_permission_name(action))
```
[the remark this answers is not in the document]
sanjeevz3009: This doesn't work without passing in `action` as a parameter too. I have made the change but passed in `action` as well.  `return super().user_has_permission(user)`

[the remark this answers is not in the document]
zerolab: yes, indeed!

cms/jinja2/templates/base.html
```diff
@@ -35,6 +35,59 @@
...
+            {% set sectionData = {
+                'text': section.section_link.title or (section.section_link.page and section.section_link.page.title) or '',
+                'url': section.section_link.external_url or (section.section_link.page and pageurl(section.section_link.page)) or False,
+                'children': []
+            } %}
+            
+            {% for link in section.links %}
+                {% set _ = sectionData.children.append({
+                    'text': link.title or (link.page and link.page.title) or '',
+                    'url': link.external_url or (link.page and pageurl(link.page)) or False,
```
zerolab, line 79: surely this should be `or ''` ?

cms/jinja2/templates/base.html
```diff
@@ -35,6 +35,59 @@
     {%- endwith -%}
 {%- endset -%}
 
+{% set navigation_settings = settings.navigation.NavigationSettings %}
+
+{% set keyLinksList = [] %}
+
+{% if navigation_settings.main_menu and navigation_settings.main_menu.highlights %}
+    {% for highlight in navigation_settings.main_menu.highlights %}
+        {% if highlight.value.page %}
```
zerolab, line 44:
```suggestion
        {% if highlight.value.page and highlight.value.page.live %}
```

so we don't inadvertently display a non-published page.

this would need to be reflected below too (for sections)

cms/jinja2/templates/base.html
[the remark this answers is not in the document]
zerolab: resolve this once you have the card created

zerolab (changes requested):

cms/navigation/wagtail_hooks.py
```diff
@@ -0,0 +1,32 @@
...
+from .models import MainMenu
+
+if TYPE_CHECKING:
+    from cms.users.models import User
+
+
+class NoAddModelPermissionPolicy(ModelPermissionPolicy):
+    """Model permission that doesn't allow creating more than one main menu instance."""
+
+    def user_has_permission(self, user: Optional["User"] = None, action: str | None = None) -> bool:
```
zerolab, line 16:
```suggestion
    def user_has_permission(self, user: "User", action: str) -> bool:
```
as neither `user` nor `action` is optional.

cms/navigation/wagtail_hooks.py
```diff
@@ -0,0 +1,32 @@
...
+        if action == "add" and MainMenu.objects.exists():
+            return False
+        return super().user_has_permission(user, action)
```
zerolab, line 19:
```suggestion
        has_permission: bool = super().user_has_permission(user, action)
        return has_permission
```

should help with `Any from function declared to return "bool"  [no-any-return]`

sanjeevz3009: Done!

cms/navigation/wagtail_hooks.py
sanjeevz3009: Done!

cms/jinja2/templates/base.html
sanjeevz3009: Done!

cms/jinja2/templates/base.html
sanjeevz3009: Done!

cms/jinja2/templates/base.html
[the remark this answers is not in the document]
zerolab: https://jira.ons.gov.uk/browse/CMS-298

nehakerung (commented):
Functionally it works but there's no working validation to force URLs/links to be unique (which is required on the ticket)
image

cms/navigation/models.py
```diff
@@ -0,0 +1,107 @@
...
+
+class ColumnBlock(StructBlock):
+    sections = ListBlock(SectionBlock(), label="Sections", max_num=MAIN_MENU_MAX_NUM_SECTIONS)
+
+    class Meta:
+        icon = "list-ul"
+        label = _("Column")
+
+
+class MainMenu(DraftStateMixin, RevisionMixin, PreviewableMixin, models.Model):
```
zerolab, line 68: I see `MainMenu` now inherits from `RevisionMixin`. Could you please read https://docs.wagtail.org/en/stable/topics/snippets/features.html#saving-revisions-of-snippets and update accordingly?

cms/navigation/models.py
```diff
@@ -0,0 +1,107 @@
...
+from wagtail.models import DraftStateMixin, PreviewableMixin, RevisionMixin
+
+from cms.core.blocks.base import LinkBlock
+from cms.core.fields import StreamField
+from cms.core.models import BaseSiteSetting
+
+MAIN_MENU_MAX_NUM_HIGHLIGHTS = 3
+MAIN_MENU_MAX_NUM_COLUMNS = 3
+MAIN_MENU_MAX_NUM_SECTIONS = 3
+MAIN_MENU_MAX_NUM_TOPIC_LINKS = 15
```
zerolab, line 17: To be honest, I am don't think these add much value.

cms/navigation/models.py
```diff
@@ -0,0 +1,107 @@
...
+        icon = "folder"
+        label = _("Section")
+
+
+class ColumnBlock(StructBlock):
+    sections = ListBlock(SectionBlock(), label="Sections", max_num=MAIN_MENU_MAX_NUM_SECTIONS)
+
+    class Meta:
+        icon = "list-ul"
+        label = _("Column")
```
zerolab, line 65: suggestion: move these to `blocks.py`.
it makes for a cleaner models file, especially that there's a the footer menu feature coming up

cms/navigation/wagtail_hooks.py
```diff
@@ -0,0 +1,33 @@
...
+from .models import MainMenu
+
+if TYPE_CHECKING:
+    from cms.users.models import User
+
+
+class NoAddModelPermissionPolicy(ModelPermissionPolicy):
+    """Model permission that doesn't allow creating more than one main menu instance."""
+
+    def user_has_permission(self, user: "User", action: str) -> bool:
```
zerolab, line 16: needs a couple of tests (both can be with a superuser):
1. main menu add view can be accessed, if a main menu doesn't exist
2. you get redirected to the Wagtail dashboard if you try to the main menu add url when a main menu exists.

The URLs:
- main menu add url would be `reverse("wagtailsnippets_navigation_mainmenu:add")`
- Wagtail dashboard `reverse("wagtailadmin_home")`


as a separate note, snippet URLs follow the 

`wagtailsnippets_{snippet._meta.app_label}_{snippet._meta.model_name}` pattern for the prefix, then `:add`, `:edit`, `:delete` for the suffix when used with `reverse()`

cms/navigation/models.py
```diff
@@ -0,0 +1,107 @@
...
+    def get_preview_template(self, request: "HttpRequest", mode_name: str) -> str:
+        return "templates/components/navigation/main_menu_preview.html"
+
+    def __str__(self) -> str:
+        return "Main Menu"
+
+
+@register_setting(icon="list-ul")
+class NavigationSettings(BaseSiteSetting):
+    main_menu: models.ForeignKey = models.ForeignKey(
```
zerolab, line 96:
```suggestion
    main_menu = models.ForeignKey(
```

to placate mypy

zerolab (changes requested):

kacperpONS (commented):
I'm not sure if `blocks/base.py` is the best name for the new file that holds the `LinkBlock`, it'd suggest something like `blocks/links.py`.
But in all honesty I don't think I would move it at all - it'd still make sense for the `LinkBlock` to live in `blocks/related.py`.

sanjeevz3009:
> Functionally it works but there's no working validation to force URLs/links to be unique (which is required on the ticket) image

Good spot @nehakerung It's only for the headers but as we discussed, I'll look into this and clear some doubts around this requirement and resolve this.

sanjeevz3009:
> I'm not sure if `blocks/base.py` is the best name for the new file that holds the `LinkBlock`, it'd suggest something like `blocks/links.py`. But in all honesty I don't think I would move it at all - it'd still make sense for the `LinkBlock` to live in `blocks/related.py`.

The initial idea from @zerolab and myself is that in the future we will be adding more reusable blocks in the `blocks/base.py` file.

cms/navigation/models.py
sanjeevz3009: Done. I had rest of the code for it before and then not sure what iteration I lost it.

cms/navigation/models.py
sanjeevz3009: Done!

cms/navigation/models.py
sanjeevz3009: I thought this would be easy to follow and read (even though I know these numbers may not change often). I'm not too picky about this, so I'm happy to remove it.

cms/navigation/models.py
sanjeevz3009: Great suggestion. Done!

cms/navigation/wagtail_hooks.py
sanjeevz3009: Main menu viewset tests added now.

zerolab:
@ababic @AdamHawtin I've tagged you for review as I have been involved in some of the changes and want to keep impartial, and fresh eyes help. Thanks

helenb:
I've just done a functional test, and my only comment is that it would be better to just allow one set of main menu settings, and not have to manually select the menu once you've added it.

- Initially I get the option to add a main menu, and this opens in a new tab which is confusing
- After I save my main menu settings, I still have another step to select the main menu before it appears on the site
- After I save the main menu, there is an option to choose another main menu, but the only option is the one I've already created, and there no way to create one
- I can edit my main menu but not delete it in order to add a new one

I think it would be better if the `settings -> navigation` landed you straight in a screen with the main menu interface visible.

Update: after discussing this with Sanjeev I gather there are plans to improve this in future and Mebin has said it is OK for now.

sanjeevz3009:
> I've just done a functional test, and my only comment is that it would be better to just allow one set of main menu settings, and not have to manually select the menu once you've added it.
> 
> * Initially I get the option to add a main menu, and this opens in a new tab which is confusing
> * After I save my main menu settings, I still have another step to select the main menu before it appears on the site
> * After I save the main menu, there is an option to choose another main menu, but the only option is the one I've already created, and there no way to create one
> * I can edit my main menu but not delete it in order to add a new one
> 
> I think it would be better if the `settings -> navigation` landed you straight in a screen with the main menu interface visible.
> 
> Update: after discussing this with Sanjeev I gather there are plans to improve this in future and Mebin has said it is OK for now.

@helenb Thank you Helen for this. We did have a good discussion about this. Yes, we will be reiterating this in the future especially when we get to Wagtail workflows etc. Additionally, as we discussed there are more questions that need to be answered and decisions that need to be made around the main menu features from having multiple main menus or having a single one, if we do maybe make it so we don't have to choose that in the navigation settings like you mentioned and also this is something @AdamHawtin pointed out and something I had in my mind. In most of these cases, @MebinAbraham and I have discussed these caveats.

I guess for now this is good as a first iteration, as we already scope creeped quite a bit in some areas.

I have made some notes of the feedback from you and @AdamHawtin. I will have some discussions with @MebinAbraham in capturing some of this somewhere or creating cards etc if needed soon.

Thanks!

sanjeevz3009:
> > So now we are just left with front-end alignment issues.
> > 
> > * Where if you have only 2 columns the alignment looks weird and goes out of place slightly compared to 3 columns.
> > * If you have 3 sections and the 1st one has 5 topics and then the rest of the sections have 3 topics then the the section alignment for that column goes off as well.
> > * So there are similar issues like the above.
> > 
> > Would be amazing if you could follow that up on a task. Happy to demo the front-end styling/ alignment issues to you.
> > Thanks =)
> 
> Noting for completeness that these issues are addressed in #63

Thanks, Helen, tested this out and it works great, when I remember to do npm build on the branch I have your branch pulled into =)

nehakerung (approved):

AdamHawtin (approved):
Agree with the points @helenb raised, I found the workflow odd around being locked down to one main menu snippet, but then the settings menu still behaving as though there was a choice, but as a starting point to iterate on this, looks good, apart from small styling niggles noted the menu and editor options all work and look good. 

It seems like the two rough paths we could go down to improve this are either:
- Allow multiple different main menu snippets to be created, and allow users to then choose which is currently active.
- Stick to only allowing one main menu snippet which must be edited directly to make changes, but then also do the work to make this one main menu always active by default since it is the only choice.

We should probably also look at either removing the main menu drop down when no menu has been created, or populating some sensible defaults?

zerolab:
On having the navigation settings + menus as snippets:
- the idea was to keep the settings clean, and allow editing the two menus separately. e.g. have different levels of access. This also make the UI cleaner in that you don't have to handle lots and lots of items all in one go
- plus the original plan was to allow multiple main menu instances with one being in use, allowing for safe updates and prep ahead of time. This is now done via revisions.

cms/navigation/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class NavigationConfig(AppConfig):
+    default_auto_field = "django.db.models.BigAutoField"
```
kacperpONS, line 5:
```suggestion
    default_auto_field = "django.db.models.AutoField"
```
I think that can be safely changed as we don't expect more than [2147483647](https://docs.djangoproject.com/en/5.1/ref/models/fields/#integerfield) Main Menus 😄.

In fact we probably don't expect to see that much of anything else so we could set it globally with [DEFAULT_AUTO_FIELD](https://docs.djangoproject.com/en/5.1/ref/settings/#default-auto-field) instead of setting it per app. Not related to this ticket, but a potential improvement in the future?

kacperpONS (approved):

cms/jinja2/templates/base.html
```diff
@@ -35,7 +35,76 @@
...
+                {% endfor %}
+            {% endif %}
+
+            {% do columnData.linksList.append(sectionData) %}
+        {% endfor %}
+
+        {% do itemsList.append(columnData) %}
+    {% endfor %}
+{% endif %}
+
```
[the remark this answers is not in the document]
RealOrangeOne: Ignore - I see this is covered by https://github.com/ONSdigital/dis-wagtail/pull/85

cms/core/blocks/base.py
```diff
@@ -0,0 +1,82 @@
...
+        desc = self.get("description")
+        has_description = "description" in self
+
+        if external_url := self.get("external_url"):
+            value = {"url": external_url, "text": title}
+            if has_description:
+                value["description"] = desc
+
+        if (page := self.get("page")) and page.live:
+            value = {"url": page.url, "text": title or page.title}
```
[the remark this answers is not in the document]
RealOrangeOne: Ignore - I see this is covered by https://github.com/ONSdigital/dis-wagtail/pull/85/

cms/jinja2/templates/base.html
```diff
@@ -35,7 +35,76 @@
     {%- endwith -%}
 {%- endset -%}
 
-{# navlinks are temporarily hard-coded until the back-end menu code is merged #}
+{% set navigation_settings = settings.navigation.NavigationSettings %}
```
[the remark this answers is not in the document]
zerolab: yes. Let's split as a follow-up task

cms/jinja2/templates/base.html
```diff
@@ -56,194 +125,8 @@
...
-                                {
-                                    "text": "National accounts",
-                                    "url": False,
-                                }
-                            ]
-                        }
-                    ]
-                }
-            ]
+            "keyLinksList": keyLinksList,
```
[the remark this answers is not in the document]
zerolab: should be covered in #85 ?

cms/core/blocks/base.py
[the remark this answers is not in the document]
ababic: @sanjeevz3009 My personal preference would be to remove this custom value class and just override [`get_context()`](https://github.com/wagtail/wagtail/blob/main/wagtail/blocks/base.py#L225) method on the block to include the link there. That method should contain the current request in `context`, and `value` will be the value for the current block

cms/jinja2/templates/base.html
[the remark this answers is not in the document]
sanjeevz3009: Correct @zerolab @RealOrangeOne

cms/jinja2/templates/base.html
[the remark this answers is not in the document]
sanjeevz3009: Will create a ticket for this. Can do it as a follow-up.

cms/navigation/tests/factories.py
```diff
@@ -0,0 +1,68 @@
...
+from wagtail_factories import ListBlockFactory, PageChooserBlockFactory, StreamFieldFactory, StructBlockFactory
+
+from cms.core.tests.factories import LinkBlockFactory
+from cms.navigation.blocks import SectionBlock, ThemeLinkBlock, TopicLinkBlock
+from cms.navigation.models import ColumnBlock, MainMenu, NavigationSettings
+from cms.themes.tests.factories import ThemePageFactory
+from cms.topics.tests.factories import TopicPageFactory
+
+
+class ThemePageChooserFactory(PageChooserBlockFactory):
```
ababic, line 11: Non-blocking: I've personally always hated block factories. Historically, I've always recommend that devs just get comfortable with raw block data representations by defining real life objects and inspecting `object.streamfield_name.raw_data` in the shell to see what they look like. It's basically just a list of dicts in a very predictable structure, which you can provide as a value to a new object.

cms/navigation/wagtail_hooks.py
```diff
@@ -0,0 +1,33 @@
...
+        if action == "add" and MainMenu.objects.exists():
+            return False
+        has_permission: bool = super().user_has_permission(user, action)
+        return has_permission
+
+
+class MainMenuViewSet(SnippetViewSet):
```
ababic, line 23: Using the [Settings](https://docs.wagtail.org/en/v3.0.1/reference/contrib/settings.html) contrib app would negate the need for anything custom in this module... it only allows you to create one instance per site, and the UI is a little better, as it doesn't have the list view (which is kinda redundant)... Instead, if there are multiple sites, it just shows a 'site switcher' in the edit interface to switch to the entry for different sites.

zerolab: There are two considerations here:
1. we'd have potentially separate people managing the main and footer menus. (this could be satisfied with separate settings)
2. there is a need to preview the changes, and a push for a simple review workflow before the main menu (at a min) is updated.


Snippets cover both needs. And the navigation setting becomes a question of selecting the approviate instances.

cms/jinja2/templates/base.html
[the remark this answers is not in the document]
zerolab: https://jira.ons.gov.uk/browse/CMS-328

cms/navigation/wagtail_hooks.py
ababic: @zerolab Okay, good to know about the translatable aspect... that probably rules out settings. Is it worth thinking about multisite at this juncture, or do you think that's something to solve later if needed?

zerolab: I have been thinking about it, and `NavigationSettings` inherits from `BaseSiteSetting` so it should scale with it. At that point we can remove the artificial singleton `MainMenu` aspect

cms/navigation/models.py
```diff
@@ -0,0 +1,67 @@
...
+from cms.core.fields import StreamField
+from cms.core.models import BaseSiteSetting
+from cms.navigation.blocks import ColumnBlock, HighlightsBlock
+from cms.navigation.forms import MainMenuAdminForm
+
+if TYPE_CHECKING:
+    from django.http import HttpRequest
+
+
+class MainMenu(DraftStateMixin, RevisionMixin, PreviewableMixin, models.Model):
```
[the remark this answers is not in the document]
ababic: Requirement confirmed by @zerolab, so marking this as resolved

RealOrangeOne (approved):
Approved in the basis #85 and other future tickets will fix my outstanding comments.

cms/jinja2/templates/base.html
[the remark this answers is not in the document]
ababic: Tip: Include the `last_published_at` datetime in the cache key to have the cache clear automatically when a new version is published

cms/core/blocks/base.py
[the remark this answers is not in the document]
zerolab: Created https://jira.ons.gov.uk/browse/CMS-337 to look at the link blocks holistically

cms/navigation/tests/factories.py
zerolab: Let's follow this up.

zerolab: Sorry Andy, I've resolved to unblock merging.

PR #42, opened 2024-11-26
[POC] Secure media (simple version)

ababic:
## What is the context of this PR?

### Testing parenting of images when creating a NEW page

1. In Wagtail, add a new "Information Page" to somewhere in the page tree.
2. Use "Image upload test" for the title and "Test" for summary.
3. For the "Content" field, add an "Image" block and click "Choose an image" to open the chooser modal.
4. Click on the "Upload" tab in the modal to access the 'create' form.
5. Inspect the form code using dev tools and look for the hidden fields. You should see the following set of hidden field values (because the page doesn't have an ID yet, but we DO know the content-type):
    - `image-chooser-upload-parent_object_content_type` should have a value
    - `image-chooser-upload-parent_object_id` should NOT have a value 
    - `image-chooser-upload-parent_object_id_outstanding` should have the value `True`
6. Complete the form to upload a new image and click the "Upload" button.
7. In a new tab, go to the 'Images' library and edit the image you just added.
8. Inspect the edit form code using dev tools and look for the hidden fields. The field values should reflect those seen in modal form (The ID of the parent page is still not known, but the content type is).
9. Switch back to the original tab, then click "Save draft" at the bottom of the page.
10. Switch back to the "Edit image" tab and REFRESH the page.
11. Inspect the form code using dev tools and look for the hidden fields once more. Now that the ID of the parent page is known, the `parent_object_id` field should have a value reflective of the newly saved page, and `parent_object_id_outstanding` should now have a value of 'False'.

#### NOTES

When the ID of the parent is not yet known, media must be marked in such a way that keeps it 'private' in the immediate term, and the ID should be set once it is known. When media-containing pages are first saved, a signal handler looks for any 'unparented' media added within the last 1 hour by the same user, for a page with the same content type, and backfills the missing data.

#### POTENTIAL ISSUES

If the editor fails to hit "Save draft" within the 1-hour time-frame, and creates and saves a page of the same time within the same hour, that media could be parented to the wrong page! Thoughts on how to address this are welcome :) I don't think we can comb the reference index first time round, because it isn't populated until afterwards.

### Testing parenting of images when editing an existing page

1. In Wagtail, edit the draft page you created in the previous section.
2. For the "Content" field, add a new "Image" block and click "Choose an image" to open the chooser modal.
3. Click on the "Upload" tab in the chooser modal to access the 'create' form
4. Inspect the form code using dev tools and look for the hidden fields. You should see the following set of hidden field values:
    - `image-chooser-upload-parent_object_content_type` should have a value
    - `image-chooser-upload-parent_object_id` should match the current page ID
    - `image-chooser-upload-parent_object_id_outstanding` should have the value `False`
5. Complete the form to upload a new image and click the "Upload" button.
6. In a new tab, go to the 'Images' library and edit the image you just added.
7.  Inspect the edit form code using dev tools and look for the hidden fields. The field values should reflect those seen in modal form.

#### NOTES

When the ID of the parent known, media can be tied to the object immediately, allowing the 'privacy' to immediately reflect that of the current page (if the current page is live, the media can be public, otherwise, it is private). 

### Testing parenting of documents when creating a NEW page

1. In Wagtail, add a new "Information Page" to somewhere in the page tree.
2. Use "Document upload test" for the page title and "Test" for the summary.
3. For the "Content" field, add an "Documents" block, then click the "+" symbol to add a "Document", then "Choose a document" to open the chooser modal.
4. Click on the "Upload" tab in the chooser modal to access the 'create' form.
5. Inspect the form code using dev tools and look for the hidden fields. You should see the following set of hidden field values (because the page doesn't have an ID yet, but we DO know the content-type):
    - `document-chooser-upload-parent_object_content_type` should have a value
    - `document-chooser-upload-parent_object_id` should NOT have a value 
    - `document-chooser-upload-parent_object_id_outstanding` should have the value `True`
6. Complete the form to upload a new document and click the "Upload" button.
7. In a new tab, go to the 'Documents' library and edit the document you just added.
8. Inspect the edit form code using dev tools and look for the hidden fields. The field values should reflect those seen in modal form (The ID of the parent page is still not known, but the content type is).
9. Switch back to the original tab, then click "Save draft" at the bottom of the page.
10. Switch back to the "Edit document" tab and REFRESH the page.
11. Inspect the form code using dev tools and look for the hidden fields once more. Now that the ID of the parent page is known, the `parent_object_id` field should have a value reflective of the newly saved page, and `parent_object_id_outstanding` should now have a value of 'False'.

### Testing parenting of documents when editing an existing page

1. In Wagtail, edit the draft page you created in the previous section
2. For the "Content" field, use the "+" action to add a new "Document" to the existing "Documents" block and click "Choose a document" to open the chooser modal.
3. Click on the "Upload" tab in the modal to access the 'create' form
4. Inspect the form code using dev tools and look for the hidden fields. You should see the following set of hidden field values
    - `document-chooser-upload-parent_object_content_type` should have a value
    - `document-chooser-upload-parent_object_id` should match the current page ID
    - `document-chooser-upload-parent_object_id_outstanding` should have the value `False`
5. Complete the form to upload a new document and click the "Upload" button.
6. In a new tab, go to the 'Documents' library and edit the document you just added.
7.  Inspect the edit form code using dev tools and look for the hidden fields. The field values should reflect those seen in modal form.

### Testing image privacy for a draft page

1. In Wagtail, edit the draft "Image upload test" page created earlier and open the preview pane. 
2. Inspect the HTML where the image appears.
3. Because the page is still a draft, the `src` value of the rendition should be something like `/images/{secure key}/2/original/{image filename}`.
4. Open this URL in a new tab.
5. Sign out of Wagtail.
6. Switch to the 'Image URL' tab you opened in step 4.
7. Refresh the page. You should get a "403 Forbidden" response.
8. In a new tab, Sign back in to Wagtail.
9. Refresh the 'Image URL' tab once more, and you should see the generated rendition again.

#### NOTES

Images added to draft pages set their `is_private` field value to `True` on save. Where the storage backend supports it the original file AND the files for any renditions generated from it it will also be set to 'private', preventing both direct and indirect access. However, you won't be able to see this locally. Whilst an image is private, renditions are served by a 'Serve' view, which only responds to authenticated users with some level of permission for the image (by default, this is based on the 'collection' the image belongs to).

### Testing image privacy once a draft page has been published

1. In Wagtail, edit the draft "Image upload test" page again, and publish it.
2. In the green success message on the next page, click the "View live" button to view the live/public-facing version of the page.
3. Inspect the HTML for the image. The `src` value should look something like: `media/images/{image filename}.fill-446x390.format-webp-.original.png`.
4. Access Wagtail again edit the "Image upload test" page once more.
5. Open the preview pane and inspect the HTML for the image there too. The `src` value should look the same as it did on the live version.
6. At the bottom of the page, click the arrow next to the "Save draft" button to view the additional options, and click "Unpublish page".
7. On the confirmation screen, click "Yes, unpublish it" to confirm the action.
8. Edit the page again in Wagtail and look again at the preview pane.
9. Inspect the HTML for the image once more. The `src` value should now again look like: `/images/{secure key}/2/original/{image filename}`.

#### NOTES

When a page with child media is published, signal handlers detect this and trigger updates to any private media items, resulting in their `is_private` flag being set to `False`, and the permissions on files themselves being set to 'publically readable' (where the storage backend supports it).

When a live page is unpublished, signal handlers again detect this and trigger updates to any public media items, resulting in their `is_private` flag being set to `True`, and the permissions on the files themselves being set to 'private' again (where the storage backend supports it). To allow editors to still see the images in the preview pane, we rely again on the 'Serve' view to serve image renditions, so that only authenticated users with some level of permission on the image can see it.

#### POTENTIAL ISSUES

- It's currently possible for pages to reuse images uploaded for other pages, in which case, that media will only become public once the official parent page is published (potentially leaving a published page displaying a mix of 'public' and 'private' media). In reality, this is only likely to happen between pages that are part of the same 'bundle', so the pages (and by extension, their child media) should be published more-or-less simultaneously. However, we might want to think about solutions here (e.g. the media being 're-parented' to the page that was published first).
- We need to review this behaviour more closely in environments where edge-level caching is involved. It's likely that more work will be needed to submit purge requests for certain URLs.

## Next steps

- [ ] Update documentation
- [ ] Ensure `python manage.py retry_file_permission_set_attempts` is set to run on a cron every 10 minutes in all target environments


cms/private_media/storages.py
```diff
@@ -0,0 +1,41 @@
...
+from storages.backends.s3 import S3Storage
+
+if TYPE_CHECKING:
+    from django.db.models.fields.files import FieldFile
+
+
+logger = logging.getLogger(__name__)
+
+
+class PrivacySettingS3Storage(S3Storage):
```
ababic, line 14: @MebinAbraham NOTE: The ACL setting is now the responsibility of the 'Storage' class as outlined in requirements, and I've adjusted the language throughout to use 'set file permissions' instead of 'set file ACLs'. However, If we want to use other storage backends in future, they'll need to be subclassed and extended in a similar fashion.

.gitignore
```diff
@@ -182,8 +182,8 @@ copier_scripts/
 /cms/jinja2/layout
 /cms/static_compiled
 
-node_modules/
-static/
+node_modules/*
+static/*
```
ababic, line 186: NOTE: This was needed to allow adding of the modal override JS to the repo

cms/standard_pages/models.py
```diff
@@ -7,9 +7,10 @@
 from cms.core.blocks.stream_blocks import CoreStoryBlock
 from cms.core.fields import StreamField
 from cms.core.models import BasePage
+from cms.private_media.models import MediaParentMixin
 
 
-class InformationPage(BasePage):  # type: ignore[django-manager-missing]
+class InformationPage(MediaParentMixin, BasePage):  # type: ignore[django-manager-missing]
```
ababic, line 13: @MebinAbraham NOTE: This mixin must be used any model class that we want to be seen as a 'media parent'. That might also include snippets in future.

cms/private_media/signal_handlers.py
```diff
@@ -0,0 +1,140 @@
...
+            raise ImproperlyConfigured(
+                f"The default manager for {model_class.__name__} does not have a 'bulk_make_private' method."
+            )
+    return None
+
+
+def register_signal_handlers() -> None:
+    """Register signal handlers for models using the private media system."""
+    page_published.connect(publish_media_on_page_publish, dispatch_uid="publish_media")
+    page_unpublished.connect(unpublish_media_on_page_unpublish, dispatch_uid="unpublish_media")
```
ababic, line 135: NOTE: Wagtail also has non-page versions of these signals (`published` and `unpublished`, respectively) that we can use for parent 'Snippet' models. But, there aren't any existing snippets with media yet to try those out on.

cms/private_media/static/private_media/js/document-modal-overrides.js
```diff
@@ -0,0 +1,12 @@
+// eslint-disable-next-line
+class DocumentChooserModal extends window.DocumentChooserModal {
+  getURLParams(opts) {
+    const urlParams = super.getURLParams(opts);
+    // eslint-disable-next-line
```
helenb, line 5: Hey @ababic just looking at the js changes here as I was curious. I think these eslint disables might have been needed in the alpha repo due to a `prefer-destructuring` rule in the torchbox eslint setup, but in this repo we're inheriting the wagtail rules so you may find you don't need them. I can help resolve the issue if there is still an error when you remove them.

ababic: Thanks @helenb. I don't remember the error exactly, but was more to do with referencing window, which isn't defined within the script.

helenb: Ah you can avoid that by declaring it as a global - see https://eslint.org/docs/latest/rules/no-undef. But again I am not sure this applies in this build so you may just be able to remove the comments completely.

cms/standard_pages/models.py
zerolab: suggestion: given the steer was to always treat assets as private until the pages they are attached to are published, why not add it to `BasePage`?

ababic:
On review, I've come to a conclusion that:
- 'Media permissioning' (in Wagtail) and 'Media privacy' (what end-users can see) are two separate things, and this PR perhaps unnecessarily conflates the two things
- Whilst the parent -> child association is useful from the perspective of 'Media permissioning' (which editors have permission to see/use/edit it in Wagtail), I don't think it's something we need to tie directly to the 'privacy' of the image.
- When it comes to 'Media privacy' (what end-users can see), I think things can be simpler: The media is always 'private' on upload, and only becomes 'public' on page publish when the media is actually present in the revision being published.

ababic:
@MebinAbraham / @zerolab Will close this now, as it's superseded by #46 and #48

PR #43, opened 2024-11-26
Django Migration Linter Integration

sanjeevz3009:

### How to review

### Follow-up Actions


Makefile
```diff
@@ -50,6 +50,10 @@ lint-html:  ## Run HTML Linters
 lint-frontend:  ## Run front-end linters
 	npm run lint
 
+.PHONY: lint-migrations
+lint-migrations: ## Run django-migration-linter
+	python manage.py lintmigrations
```
zerolab, line 55:
```suggestion
	poetry run python manage.py lintmigrations
```

so it uses the virtual environment with all the installed dependencies. Should fix the `ModuleNotFoundError: No module named 'django'` error

.github/workflows/ci.yml
```diff
@@ -52,6 +52,25 @@ jobs:
...
+        uses: actions/setup-python@v5
+        with:
+          python-version-file: .python-version
+          cache: poetry
+
+      - name: Install dependencies
+        run: make install-dev
+
+      - name: Lint migrations
+        run: make lint-migrations
```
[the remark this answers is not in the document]
nehakerung: It needs to be done in docker-shell hence i put it in a separate job. Shall I implement this suggestion then change the README to just make lint should be done in the shell?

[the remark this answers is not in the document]
zerolab: Right, in that case this step needs to follow the logic of the test-python step (which also links up with the database)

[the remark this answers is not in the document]
zerolab: on second thought, considering that the `test-python` job has a check for migrations - https://github.com/ONSdigital/dis-wagtail/blob/6d9d6b5/.github/workflows/ci.yml#L125-L126 - perhaps we add this step there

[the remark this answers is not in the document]
nehakerung: Think that does make better sense so added.

zerolab:
There are conflict with the main branch and I suspect this may be the reason the CI is not starting. Best to get this branch up to date with `main`. Will require sorting out the conflicts in pyproject.toml, and poetry.lock. With the latter, I usually remove it and do `poetry lock`

scripts/lintmigrations_custom.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+# Extract the app label from the directory structure
+def get_app_labels():
+    labels = []
+    for root, dirs, _files in os.walk("../cms"):
+        if "migrations" in dirs:
+            app_label = root.split(os.sep)[-1]
+            labels.append(app_label)
+    return labels
```
zerolab, line 14:
```suggestion
from pathlib import Path


BASE_DIR = Path(__file__).parent.parent

# Extract the app label from the directory structure
def get_app_labels():
    return [
        path.parent.name
        for path in Path(BASE_DIR / "cms").glob("*/migrations")
    ]
```

and `import os` removed.

scripts/lintmigrations_custom.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+# Run the lintmigrations command for a given app_label
+def run_lintmigration(app_label):
+    try:
+        result = subprocess.run(  # noqa: S603
+            [sys.executable, "../manage.py", "lintmigrations", app_label],
```
zerolab, line 21:
```suggestion
            [sys.executable, BASE_DIR / "manage.py", "lintmigrations", app_label],
```

Makefile
```diff
@@ -50,6 +50,14 @@ lint-html:  ## Run HTML Linters
 lint-frontend:  ## Run front-end linters
 	npm run lint
 
+.PHONY: lint-migrations
+lint-migrations: ## Run django-migration-linter
+	poetry run python manage.py lintmigrations
+
+.PHONY: lint-migrations-custom
+lint-migrations-custom: ## Run django-migration-linter
+	cd scripts && poetry run python lintmigrations_custom.py
```
zerolab, line 59:
```suggestion
	poetry run python scripts/lintmigrations_custom.py
```

with the change suggestions in `scripts/lintmigrations_custom.py` below

zerolab (commented):
Left a couple of suggestions around the custom script

sanjeevz3009:
@MebinAbraham @zerolab @kacperpONS

Hello team,

Neha and I fixed the linting issues which was flagged by Ruff. Neha did some good investigations around it, to come up with potential solutions and fixes.

After I dived into it, I realised some of the issues involved security issues related to Bandit. These issues have been resolved now.

Got another issue where if you look at the failed CI runs: https://github.com/ONSdigital/dis-wagtail/actions/runs/12146805442/job/33871555246

Where it keeps saying `ModuleNotFoundError: No module named 'django_migration_linter'` I think this is due to maybe in the `Dockerfile` we have a command `RUN poetry install --no-root --without dev` which installs the packages (excluding dev). This means the `django-migration-linter` won't be installed. Which makes sense but in this case for some reason it's flagging it as an error not sure why in this exact case.

But this is a package where it's also required to be added to the `installed_apps` in our project, so that may be the reason. So for now we went with the approach of having the `django-migration-linter` in normal/ main dependencies instead of dev and this resolves the issue.

Let us know if this is okay or if there is another way of doing this/ that we should look into.

scripts/lintmigrations_custom.py
nehakerung: Updated 🙌

Makefile
nehakerung: We tried this before but we didn't have `BASE_DIR = Path(__file__).parent.parent` and `BASE_DIR / "manage.py"`😅- all working now! Thank you 🎉

RealOrangeOne:
It'd be much better to avoid installing it in the production image, since it's a linter.

The reason is because it's listed in `INSTALLED_APPS`. I'd suggest only adding it during development, as we do with [django-debug-toolbar](https://github.com/ONSdigital/dis-wagtail/blob/main/cms/settings/dev.py#L40).

scripts/lintmigrations_custom.py
```diff
@@ -0,0 +1,60 @@
...
+# Extract the app label from the directory structure
+def get_app_labels():
+    return [
+        path.parent.name
+        for path in Path(BASE_DIR / "cms").glob("*/migrations")
+    ]
+
+
+# Run the lintmigrations command for a given app_label
+def run_lintmigration(app_label):
```
RealOrangeOne, line 17: Suggestion: It looks like `lintmigrations` supports a `--include-apps` flag, which can be passed multiple times. That might be better to run than running the command in parallel.

scripts/lintmigrations_custom.py
RealOrangeOne, line 1: I'm not sure I see the intention behind this script? Is only including our first-party apps a significant improvement over just linting them all (minus the excluded ones)?

nehakerung: The script provides a friendlier output for the linter. I've tagged you on Slack in the discussion for this. 😃

scripts/lintmigrations_custom.py
sanjeevz3009: Regarding adding the linter only during development, will make that change. As just needed to make sure of some things/ it was a temporary thing we did to make things work =)

RealOrangeOne: I think that comment might be related to my other one. In this case, you're using `concurrent.futures` to run the command in parallel. If you instead pass `--include-apps`, you can avoid the concurrency and likely end up with a briefer output anyway.

The auto-discovery of modules would remain.

sanjeevz3009: Yes

kacperpONS: Yh, please ignore the comments from Friday afternoon 😄 

@RealOrangeOne, it seems that if you run the linter with either `--include-apps` or  `--exclude-apps`  it still prints the log lines for the apps it ignores, so our suggestions wouldn't give us what we want in terms of readability.

However - I think we can ignore all of that, because there's an option that allows you to skip the log lines based on their type. See my suggested change below and the option from the library's documentation:
<!DOCTYPE html>
--quiet or -q {ok,ignore,warning,error} | Suppress certain output messages, instead of writing them to stdout.
-- | --

Makefile
```diff
@@ -50,6 +50,10 @@ lint-html:  ## Run HTML Linters
 lint-frontend:  ## Run front-end linters
 	npm run lint
 
+.PHONY: lint-migrations
+lint-migrations: ## Run django-migration-linter
+	poetry run python scripts/lintmigrations_custom.py
```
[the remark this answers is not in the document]
sanjeevz3009: Main menu is complete to test visually but writing tests for it now. Will need to push the changes I made. Need write the PR description and record a loom. Hopefully it's up soon/ early tomorrow.

Just wanted to help Neha get this through and it's basically done. Just some preference discussions going on which she can take over soon!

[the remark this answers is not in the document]
kacperpONS: I'd still go for the built-in option that the tool comes with, but I won't block on 🖖

scripts/lintmigrations_custom.py
RealOrangeOne: Interesting. Using `-q ignore` ought to avoid the need for a custom script entirely, as it makes the output far more readable. If not, then using `--include-apps` should still avoid the threading complexity.

nehakerung: `poetry run python manage.py lintmigrations --quiet ignore ` - we can use this instead (this still uses `exclude_apps` in `dev.py`) and yes, would mean we don't need the script . The output being:
`(analysis, 0001_initial)... WARNING
        CREATE INDEX prolongs transaction, delaying lock release
(bundles, 0001_initial)... WARNING
        CREATE INDEX prolongs transaction, delaying lock release
(core, 0001_initial)... WARNING
......
(themes, 0001_initial)... WARNING
        CREATE INDEX prolongs transaction, delaying lock release
(topics, 0001_initial)... WARNING
        CREATE INDEX prolongs transaction, delaying lock release
(users, 0001_initial)... WARNING
        CREATE INDEX prolongs transaction, delaying lock release
*** Summary ***
Valid migrations: 1/206
Erroneous migrations: 3/206
Migrations with warnings: 13/206
Ignored migrations: 189/206
` 
As more migrations are made, this may not be an ideal output but shall I change the command to this?

RealOrangeOne: That sounds like it's going to be much easier to maintain. The list might get long, but it shows each migration as being valid, which is something. it might also be an idea to add "ok" to the quiet flag, so it only prints out the errors (making it even quieter).

cms/documents/migrations/0002_alter_customdocument_file_size.py
```diff
@@ -1,17 +0,0 @@
...
-        ("documents", "0001_initial"),
-    ]
-
-    operations = [
-        migrations.AlterField(
-            model_name="customdocument",
-            name="file_size",
-            field=models.PositiveBigIntegerField(editable=False, null=True),
-        ),
-    ]
```
zerolab, line 17: please do not delete migrations once they are on `main`. They will have been applied in different environments and you end with a dirty history

README.md
```diff
@@ -244,6 +244,26 @@ To lint the Python code, run:
...
+
+```bash
+make lint-migrations
+```
+
+To lint the django migration files and view the output in a more readable format:
+
+```bash
+make lint-migrations-custom
+```
```
kacperpONS, line 263: This can go too 😄

cms/documents/migrations/0002_alter_customdocument_file_size.py
nehakerung: ok 🫡 I did have to manually add it back as the main was merged in case that may be a problem in future.

README.md
nehakerung: added 🙏 [c1083d1](https://github.com/ONSdigital/dis-wagtail/pull/43/commits/c1083d15710f5abb98e8bbd2cf3803d04b24c1a7)

cms/documents/migrations/0001_initial.py
```diff
@@ -28,7 +28,7 @@ class Migration(migrations.Migration):
                 ("title", models.CharField(max_length=255)),
                 ("file", models.FileField()),
                 ("created_at", models.DateTimeField(auto_now_add=True)),
-                ("file_size", models.PositiveIntegerField(editable=False, null=True)),
```
zerolab, line 31: @nehakerung and undo this migration too :D

.github/workflows/ci.yml
```diff
@@ -131,6 +131,51 @@ jobs:
       - name: Test
         run: make test
 
+  test-lint-migrations:
```
RealOrangeOne, line 134: Question: Does this need to be its own job? Could it be added to either `test-python` or `lint`?

MebinAbraham: I suggest adding it as part of `lint`.

nehakerung: The settings is different (test vs dev) but updated to `run: DJANGO_SETTINGS_MODULE=cms.settings.dev make lint-migrations` in [b22b9c0](https://github.com/ONSdigital/dis-wagtail/pull/43/commits/b22b9c03f24fddd3794286fd6b1c504513cbbcd0) credits to @zerolab

cms/documents/migrations/0001_initial.py
nehakerung: 🫡 [dab2db0](https://github.com/ONSdigital/dis-wagtail/pull/43/commits/dab2db074966db2b9e38932202f054a1f4af3382)

cms/settings/dev.py
```diff
@@ -57,3 +57,19 @@
...
+        "wagtailcore",
+        "wagtailembeds",
+        "wagtailimages",
+        "wagtailadmin",
+        "wagtailsearch",
+        "wagtaildocs",
+        "wagtailredirects",
+        "wagtailusers",
+        "core",
+        "images",
```
MebinAbraham, line 73: We should disable specific migrations instead of the whole app, as we want to catch any new migrations that have genuine issues.

cms/home/migrations/0002_create_homepage.py
```diff
@@ -37,4 +37,4 @@ class Migration(migrations.Migration):
 
     dependencies = [("home", "0001_initial")]
 
-    operations = [migrations.RunPython(create_homepage)]
+    operations = [migrations.RunPython(create_homepage, migrations.RunPython.noop)]
```
MebinAbraham, line 40: We should revert this; any migrations that have already merged should not be modified.

cms/settings/dev.py
```diff
@@ -28,7 +28,7 @@
 # Enable Wagtail's style guide in Wagtail's settings menu.
 # http://docs.wagtail.io/en/stable/contributing/styleguide.html
 INSTALLED_APPS += ["wagtail.contrib.styleguide"]  # noqa: F405
-
+INSTALLED_APPS += ["django_migration_linter"]
```
MebinAbraham, line 31: Nit: Add them on like 30?
```py
INSTALLED_APPS += ["wagtail.contrib.styleguide", "django_migration_linter"]  # noqa: F405
```

cms/home/migrations/0002_create_homepage.py
zerolab: `migrations.RunPython.noop` returns `None`, which is the equivalent of not specifying it at all as the `reverse_code` kwarg defaults to `None`.

This makes the linter happy

cms/release_calendar/migrations/0002_create_releasecalendarindex.py
```diff
@@ -43,5 +43,5 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunPython(create_index_page),
+        migrations.RunPython(create_index_page, migrations.RunPython.noop),
```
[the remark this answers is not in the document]
nehakerung: So will be ignoring this based on the previous comment. Is that right?

cms/settings/dev.py
sanjeevz3009: I remember we did try this before. Tried again now but due to some things to do with how things get initialised, this doesn't work.

cms/settings/dev.py
sanjeevz3009: Neha and I completed this =) [ce114c7](https://github.com/ONSdigital/dis-wagtail/pull/43/commits/ce114c7ec9989047e5c39850b411dd9651c8901b)

cms/release_calendar/migrations/0002_create_releasecalendarindex.py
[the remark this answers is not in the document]
zerolab: I would say yes (ref: https://github.com/ONSdigital/dis-wagtail/pull/43/files#r1884052999)

kacperpONS (approved):

.docker/bashrc.sh
zerolab, line 1: question: why is the file permission change needed here?
it is best to revert the change

zerolab (commented):
One final question from me (sorry!)

.docker/bashrc.sh
nehakerung: 💀 Not sure how this changed but should be identical to the main branch.

nehakerung: So this file didn't change but it's not reflected on the PR for some reason? (It did show in my local code as well as the git code). To make it show, had to delete and re-add [92d2350](https://github.com/ONSdigital/dis-wagtail/pull/43/commits/92d2350b61a05e235a0358e4cbf6ea52ffd27f15) 🕺

zerolab: Can you run `git checkout main -- .docker/bashrc.sh`?

nehakerung: Update 🙏 - [145c672](https://github.com/ONSdigital/dis-wagtail/pull/43/commits/145c672c916208418ffa1a9c8826442d0eeddf0a)

zerolab (approved):

MebinAbraham (approved):
The script was a nice addition. Let's see how this goes. If we find it hard to use and read, it's worth revisiting.

PR #44, opened 2024-11-28
Support a mostly read-only database connection for external environment

RealOrangeOne:

https://jira.ons.gov.uk/browse/CMS-161
https://jira.ons.gov.uk/browse/CMS-218

For the external environment, the database connection will be read-only, apart from Wagtail's image rendition table.

This PR adds a few things to facilitate this:

- Add a database router to prevent writes to all tables except a small subset.
- Add a custom `BaseSiteSetting`, which ensure the setting can function if the connection is read-only, by falling back to the default.
    - This also prioritizes the read connection, as Django's `get_or_create` always uses a write connection (https://jira.ons.gov.uk/browse/CMS-218)
- Add tests for pages with `IS_EXTERNAL_ENV=True` to ensure they function in the external environment
- A custom `TransactionTestCase` to add extra assertions and work around a Wagtail bug with clearing tables. This case must be used instead of Django's `TransactionTestCase` (`TestCase` is fine).
- Use dummy cache in tests, to avoid issues counting querys
- Run tests with `--shuffle` to help avoid non-deterministic test issues.

There's also a change to CI to work around an issue which looks to be a bug in Django itself: https://code.djangoproject.com/ticket/35967

### How to review

Most of this PR is tests, so giving those a look is a good start. There shouldn't be any functional change to the UI or development workflow.

### Follow-up Actions

Restricting the connection in production is out of scope for this PR.


cms/settings/base.py
```diff
@@ -219,8 +219,7 @@
 if "read_replica" not in DATABASES:
     DATABASES["read_replica"] = deepcopy(DATABASES["default"])
 
-# Ensure the correct connection is used depending on the query
-DATABASE_ROUTERS = ["cms.core.db_router.ReadReplicaRouter"]
+DATABASE_ROUTERS = ["cms.core.db_router.ExternalEnvRouter", "cms.core.db_router.ReadReplicaRouter"]
```
zerolab, line 222: Note for other devs: https://docs.djangoproject.com/en/5.1/topics/db/multi-db/#using-routers explains how this works

zerolab (commented):
Left a suggestion (to apply globally), a question and a note for others.

cms/core/tests/test_settings.py
```diff
@@ -0,0 +1,67 @@
...
+            SocialMediaSettings.for_site(None)
+
+    def test_create_setting(self):
+        """Test creating a setting."""
+        self.assertEqual(SocialMediaSettings.objects.count(), 0)
+
+        with self.assertNumQueriesConnection(default=4, replica=1):
+            SocialMediaSettings.for_request(self.request)
+
+        # Explicitly use default connection to work around strange issue with Django
```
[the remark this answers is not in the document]
RealOrangeOne: The model just doesn't appear in the DB. If I leave out `.using`, there's a count of 0, even though there's a clear DB query run. 

It might be another strange bug in Django, but being explicit here made it go away so I left it.

cms/core/db_router.py
```diff
@@ -56,3 +49,33 @@ def allow_migrate(self, db: str, app_label: str, model_name: Optional[str] = Non
...
         return db != self.REPLICA_DB_ALIAS
+
+
+class ExternalEnvRouter:  # pylint: disable=unused-argument,protected-access
+    """A database router which prevents writes to certain models in the external environment.
+
+    In production, this will also be enforced at the database level.
+    """
+
+    WRITE_ALLOWED_MODELS: ClassVar[list[type[Model]]] = [get_image_model().get_rendition_model()]
```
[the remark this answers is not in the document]
RealOrangeOne: Good point, it'd be much easier to just hardcode the model to avoid the disconnect.

cms/settings/base.py
ababic: Q: Would we want this change in base settings, or perhaps just in `production.py`? My worry is that added complication isn't really needed for dev or test environments.

cms/core/db_router.py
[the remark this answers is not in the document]
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/44/commits/3bce4ab7c46913ee5028ca72156e5bf89ff4efbb

cms/core/tests/test_read_only_connection.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@override_settings(IS_EXTERNAL_ENV=True)
+class ReadOnlyConnectionTestCase(TestCase):
+    def test_cannot_write_to_disallowed_table(self):
+        """Test that disallowed models cannot be written to in external env."""
+        self.assertFalse(CustomImage.objects.exists())
+
+        with self.assertRaises(ConnectionDoesNotExist):
+            ImageFactory.create()
```
MebinAbraham, line 15: I'm not the biggest fan of this. It's misleading to have this error raised in production when it was because the writes were disabled. I would prefer if we could raise a custom exception to make the cause of the error clear.

cms/analysis/tests/test_models.py
```diff
@@ -31,6 +31,15 @@ def test_index_redirects_to_latest(self):
         response = self.client.get(self.series.url)
         self.assertRedirects(response, self.series.url + self.series.reverse_subpage("latest_release"))
 
+    def test_index_redirects_to_latest_in_external_env(self):
```
MebinAbraham, line 34: Q: Is there something we can do to ensure, all external pages render when external true vs devs having to remember to add a test case with external true as pages get added.

cms/core/tests/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+            num,
+            f"{len(queries)} queries executed, {num} expected\nCaptured queries were (in alphabetical order):\n"
+            + "\n".join(queries),
+        )
+
+    @contextmanager
+    def assertNumQueriesConnection(self, *, default=0, replica=0):  # pylint: disable=invalid-name
+        """Assert the number of queries per connection."""
+        with (
+            self.assertNumQueries(default, using="default"),
```
zerolab, line 64: nitpicky suggestion: now that we use `READ_REPLICA_DB_ALIAS`, it may be nice to use `DEFAULT_DB_ALIAS` instead of "default" in tests

cms/core/models/base.py
```diff
@@ -79,3 +86,35 @@ def has_ons_embed(self) -> bool:
...
+            # Explicitly call `.get` first to ensure the
+            # read connection is used.
+            return cast(Self, queryset.get(site=site))
+        except cls.DoesNotExist:
+            if settings.IS_EXTERNAL_ENV:
+                # In the external env, the database connection is read only,
+                # so just use the default values if the instance doesn't exist.
+                return cls(site=site)
+
+            instance, _created = queryset.get_or_create(site=site)
```
MebinAbraham, line 119: I was wondering why Django does not handle this for us by using read for `get` first, then `write`? But I saw this comment in the `get_or_create` method.
```py
# The get() needs to be targeted at the write database in order
# to avoid potential transaction consistency problems.
```
Could this not be a problem for us (edge case)?

cms/core/tests/test_read_only_connection.py
RealOrangeOne: This would happen anyway when someone tries to run a write query anyway, this just raises a similar error. Sadly it's not viable to raise an error in `db_for_write`, as that can be obtained even for read queries, just never used.

The alternative is subclassing the connection itself and checking the queries executed, but this feels better than that.

cms/analysis/tests/test_models.py
RealOrangeOne: Possibly, but it'd be quite brittle. I could try and discover pages, their factories and attempt to render them, but that would be quite brittle for more complex pages.

cms/core/models/base.py
RealOrangeOne: No, for us this is fine. `get_or_create` is itself internally consistent. If it doesn't exist, it attempts to create it. If it did infact exist, just read it again.

The reason for this override is to ensure that even the first `.get` uses the read connection, which `get_or_create` doesn't. Since 99% of the times this is run, there'll be an instance, it's worth the extra bit of code to ensure the replicas are used.

cms/core/tests/test_read_only_connection.py
RealOrangeOne: It looks like this may be doable with some of Django's database instrumentation APIs. I'll keep poking.

zerolab: I think the main thing we want to keep in mind is how far down the rabbit hole we're going (ie. "configure vs customise" 🙈 )

RealOrangeOne: Unfortunately, the instrumentation has no concept of models, without parsing the SQL, so this is the best way. I can use a cuustom connection to change the exception, but it's still an exception.

MebinAbraham: Ya, it would have been a nice to have but don't want to spend long on it, or want it to become very beskpoke.

MebinAbraham: > I can use a cuustom connection to change the exception, but it's still an exception.

That's fine, it was more to help with support, logging and alerting based on specific exceptions etc.

MebinAbraham (approved):

zerolab (approved):
I thought I approved earlier. It looks like I was commenting directly. :shipit:

sanjeevz3009 (commented):
Reviewed!

PR #46, opened 2024-11-29
Restrict media privacy until a referencing page is published

ababic:
## What is the context of this PR?

This is a follow-up to #42 that focusses on addressing solely public/private status of media, without attempting to tie objects to pages to aid with permission policy customisation.

It's fully functional and has lots of tests (which are worth checking out to get a feel for the intended behaviour).

### A brief summary:

- All media should be 'private' by default (regardless of upload route)
- Media is only made public when used within the content field of a page, and the revision featuring the media is published.
- If the referencing page is unpublished, any media referenced solely by that page is made private again (anything referenced by other 'live' pages remains public)

### When a media item's privacy changes:

- Where the storage backend supports it (S3), attempts are made to set file-level permissions to reflect the new privacy value.
- Where permissions for all a media item's file are updated successfully, the `file_permissions_last_set` timestamp is updated for the media item, and the object's `file_permissions_are_outdated()` method will return `False`.
- If any errors occurred during the update process, `file_permissions_last_set` is not updated, and the `file_permissions_are_outdated()` method will return `True`.
- Where front-end cache invalidation is configured, purge requests will also be triggered for all relevant 'serve' or direct file URLs

### Testing media privacy for a draft page

1. In Wagtail, create a new Information Page
2. Enter "Media upload test" for the title and "Test" for the summary
3. Add an "Image" block to the content, and use the chooser modal to upload a new image
4. Add a "Documents" block to the content, use the sub-block options to add a document, and use the chooser modal for that item to upload a new document
5. Open the preview pane for the page and inspect the HTML where the blocks are rendered.
7. The `src` value of the image rendition should look something like `/images/{secure key}/2/original/{image filename}`, which means it's being served by the 'serve' view.
8. The `href` value of the document link should look something like `/documents/`
9. Open the document link URL and  image `src` URL in new tabs (they should still work).
10. Sign out of Wagtail.
11.  Refresh the the tabs opened in step 9 - You should now get a "403 Forbidden" response for both.
12. In a new tab, Sign back in to Wagtail.
13. Refresh the tabs once more, and they should work again

### Testing media privacy after a page has been published

1. Publish the "Media upload test" page created in the above steps
2. In the green success message on the next page, click the "View live" button to view the live/public-facing version of the page.
3. Inspect the HTML for the image. The `src` value should have changed to something like: `media/images/{image filename}.fill-446x390.format-webp-.original.png`.
5. Access Wagtail again edit the "Media upload test" page once more.
6. Open the preview pane and inspect the HTML for the image there too. The `src` value should look the same as it did on the live version.
7. At the bottom of the page, click the arrow next to the "Save draft" button to view the additional options, and click "Unpublish page".
8. On the confirmation screen, click "Yes, unpublish it" to confirm the action.
9. Edit the page again in Wagtail and look again at the preview pane.
10. Inspect the HTML for the image once more. The `src` value should now again look like: `/images/{secure key}/2/original/{image filename}`.

### Testing image privacy once a draft page has been published

1. In Wagtail, edit the draft "Image upload test" page again, and publish it.
2. In the green success message on the next page, click the "View live" button to view the live/public-facing version of the page.
3. Inspect the HTML for the image. The `src` value should look something like: `media/images/{image filename}.fill-446x390.format-webp-.original.png`.
4. Access Wagtail again edit the "Image upload test" page once more.
5. Open the preview pane and inspect the HTML for the image there too. The `src` value should look the same as it did on the live version.
6. At the bottom of the page, click the arrow next to the "Save draft" button to view the additional options, and click "Unpublish page".
7. On the confirmation screen, click "Yes, unpublish it" to confirm the action.
8. Edit the page again in Wagtail and look again at the preview pane.
9. Inspect the HTML for the image once more. The `src` value should now again look like: `/images/{secure key}/2/original/{image filename}`.

### Follow-up Actions

- [x] Update documentation
- [ ] Ensure `PRIVATE_MEDIA_PERMISSION_SETTING_MAX_WORKERS` is set appropriately for all target environments
- [ ] Ensure `python manage.py retry_file_permission_set_attempts` is set to run on a cron every 10 minutes in all target environments


cms/documents/migrations/0004_alter_customdocument__privacy.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 5.1.3 on 2024-11-29 16:20
```
zerolab, line 1: any chance of merging the migrations?

ababic: Not really. Any existing documents and images are added with an initial `_privacy` value of 'public', because it's likely already referenced by pages, and it's too late to hook into 'page_published' signals to change that.

The second migration for each app changes the default value to 'private', so that newly added documents and images receive that status.

zerolab: I was looking at the volume of images/docs on dev/staging and it super low (so the question made sense to me 🙈 ), but I agree this is the proper way when there is content.

docs/custom-features/media_privacy.md
```diff
@@ -0,0 +1,34 @@
...
+
+1. The privacy of an image or document cannot be changed manually via the Wagtail UI.
+2. Images and documents are private by default, and will only be served to authenticated users with some level of permission on the media item whilst they have this privacy status (so previewing from Wagtail should still work as expected).
+3. Where the storage backend supports it, files for private images and documents (and any renditions created from the original image) will have their access permissions updated at the storage level, preventing direct file access.
+4. Images and documents only become public when they are referenced by blocks in page content, and that page is published.
+5. When a page is unpublished, any public images or documents referenced solely by that page are made private again (references from other draft pages are ignored, as the media can easily be made public again when the draft page is published).
+
+## Use of Wagtail's reference index
+
+Wagtail's reference index is used to track when images and documents (and other objects) are referenced by pages and other objects.
```
zerolab, line 14: suggestion:

```suggestion
Wagtail's [reference index](https://docs.wagtail.org/en/stable/advanced_topics/reference_index.html) is used to track when images and documents (and other objects) are referenced by pages and other objects.
```

docs/custom-features/media_privacy.md
```diff
@@ -0,0 +1,34 @@
...
+
+Population of the reference index is automatically triggered by the `post_save` signal for all registered models.
```
zerolab, line 16: how about

```suggestion
The reference index is automatically populated via the `post_save` signal for all registered models.
```

?

docs/custom-features/media_privacy.md
```diff
@@ -0,0 +1,34 @@
...
+When publishing or unpublishing a page (or other type of 'publishable' object), object-level changes are saved before any `published`, `page_published`, `unpublished` or `page_unpublished` signals are emitted. So, as long as we use these signals to review the privacy
+of referenced media, we can be sure that the reference index will be up-to-date.
+
+## Handling of file permission setting errors
+
+Whilst file-permission setting requests are quite reliable in S3, they can fail occasionally (as with any web service).
+
+Failed attempts are tracked using a `file_permissions_last_set` timestamp for each media item, which is only updated if all update requests for the media item were successful.
+
+The `retry_file_permission_set_attempts` management command seeks out any media items with an outdated timestamp and attempts to update the files again.
```
zerolab, line 27: Note to self: we will need a page in docs that lists the management commands and how often they run. One for a follow up PR

cms/private_media/views.py
```diff
@@ -0,0 +1,34 @@
...
+        user = self.request.user
+        if image.is_private and (
+            not user.is_authenticated
+            or not permission_policy.user_has_any_permission_for_instance(user, ["choose", "add", "change"], image)
+        ):
+            raise PermissionDenied
+
+        # Serve the file until it is no longer private, or file permissions
+        # have been set successfully
+        response: HttpResponse = super().serve(rendition)
```
zerolab, line 33: thought: could we serve this without any cache. It is for authenticated users with permissions after all. This way we'd only need to purge `rendition.file.url`s should anything change for the public files.

cms/private_media/storages.py
```diff
@@ -0,0 +1,52 @@
...
+
+if TYPE_CHECKING:
+    from django.db.models.fields.files import FieldFile
+
+
+logger = logging.getLogger(__name__)
+
+
+class PrivacySettingS3Storage(S3Storage):
+    # pylint: disable=W0223
```
zerolab, line 16: had to look it up. how about
```suggestion
    # pylint: disable=abstract-method
```

cms/private_media/storages.py
```diff
@@ -0,0 +1,52 @@
...
+            obj_acl.put(ACL=acl_name)
+        except ClientError as e:
+            logger.warning("Failed to set ACL for %s: %s", file.name, repr(e))
+            return False
+
+        logger.info("ACL set successfully for %s", file.name)
+        return True
+
+
+class DummyPrivacySettingFileSystemStorage(FileSystemStorage):
```
zerolab, line 45:
```suggestion
class DummyPrivacySettingFileSystemStorage(FileSystemStorage):
    """Dummy storage class for use in tests."""
```

cms/private_media/signal_handlers.py
```diff
@@ -0,0 +1,88 @@
...
+    for model_class in get_private_media_models():
+        model_ct = ContentType.objects.get_for_model(model_class)
+        referenced_pks = (
+            ReferenceIndex.get_references_for_object(instance)
+            .filter(to_content_type=model_ct)
+            .annotate(int_object_id=Cast("to_object_id", output_field=IntegerField()))
+            .values_list("int_object_id", flat=True)
+            .distinct()
+        )
+        queryset = model_class.objects.filter(pk__in=referenced_pks)
```
zerolab, line 32: question: should this limit to instances where `_privacy=Privacy.PRIVATE`?
i.e. if referencing an already public image, do we need to process it again? Or better said, only bulk make public any items that are private, or public but outdated?

cms/private_media/models/images.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+        return "wagtail-rendition-" + "-".join(
+            [
+                str(image.id),
+                image.file_hash,
+                str(image.privacy),
+                filter_cache_key,
+                filter_spec,
+            ]
+        )
```
zerolab, line 91: q: could this use the core `construct_cache_key` and append the privacy suffix?

cms/private_media/views.py
zerolab: ah, but the manager `bulk_set_privacy` invalidates FE cache. Still, we could eliminate `serve_url_batch.purge()` from that 🤔

cms/private_media/storages.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def make_private(self, file: "FieldFile") -> bool:
+        """Make the provided file private in S3."""
+        return self._set_file_acl(file, self.private_acl_name)
+
+    def make_public(self, file: "FieldFile") -> bool:
+        """Make the provided file publically readable in S3."""
+        return self._set_file_acl(file, self.public_acl_name)
+
+    def _set_file_acl(self, file: "FieldFile", acl_name: str) -> bool:
```
zerolab, line 28: this shows as not covered by tests. It would be worth using moto with the s3 extra (we already have moto[iam]) to cover this as well

zerolab (commented):
Loads of goodies here 👏🏼 
Did a first pass review with questions, suggestions (and couldn't help it, a few nitpicks)

cms/private_media/views.py
ababic: > thought: could we serve this without any cache. It is for authenticated users with permissions after all. This way we'd only need to purge rendition.file.urls should anything change for the public files

I'm open to this. This approach is certainly a good idea whilst the media is still private. Once public, caching is probably more desirable, but 99% of the time, the image serve view should no longer receive hits, because the `src` value on images should use the direct file URL.

cms/private_media/signal_handlers.py
ababic: This is something already checked by `bulk_make_private` and `bulk_make_public` before they make and updates, but it wasteful to fetch object that we know we don't need, so will apply!

docs/custom-features/media_privacy.md
ababic: Applied in [04e31c2](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/04e31c2223dbc6004a1397f19ddaadee04101b93)

docs/custom-features/media_privacy.md
ababic: Applied with slight wording change in [04e31c2](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/04e31c2223dbc6004a1397f19ddaadee04101b93)

cms/private_media/storages.py
ababic: Addressed in [ab6a27d](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/ab6a27dc13b51e1360124de6c8ea07d86dc40576)

cms/private_media/storages.py
ababic: Addressed in [7f7cb4f](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/7f7cb4f469bb2e0ae0e2bdf00ae33c99107d9b6b)

cms/private_media/views.py
ababic: Implemented in [19dde02](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/19dde02607d9cae65100c8ceac05407b035ffc65)

cms/private_media/signal_handlers.py
ababic: Addressed in [f7c46e0](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/f7c46e0ac723804ca31151133198afaf9757b467)

cms/private_media/models/images.py
ababic: I tried, but super() doesn't work for `staticmethods` unfortunately (only @classmethods and instance methods).

cms/private_media/storages.py
zerolab: Last comment on issue 3446 on https://github.com/getmoto/moto/issues/ has an example. Not doing the direct link as I don't want the GH cross-linking

cc @MebinAbraham what do you think?

cms/private_media/models/images.py
zerolab: ah, but of course. the beauty and curse of static methods

cms/private_media/storages.py
ababic: Added something working in [712a337](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/712a3372a1cdec7c87ce40b86b9b620f154c5448) using regular mocking

ababic: Found a way to get `mock_aws` to work, but it certainly isn't pretty: [67a4cec](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/67a4ceca17af9d3d69f0a0aaf44403ac1b03dac0)

cms/private_media/tests/test_images.py
```diff
@@ -0,0 +1,369 @@
...
+            self.assertFalse(image.file_permissions_are_outdated())
+
+        # Another attempt should result in no updates
+        with self.assertNumQueries(1):
+            # Query summary:
+            # 1. to fetch the images
+            self.assertEqual(self.model.objects.bulk_make_private(self.model.objects.all()), 0)
+
+
+class TestImageServeView(TestCase):
```
RealOrangeOne, line 330: Issue: Can you add some tests for `IS_EXTERNAL_ENV` here.

cms/private_media/tests/test_images.py
```diff
@@ -0,0 +1,369 @@
...
+        renditions = image.create_renditions(Filter("fill-10x10"), Filter("fill-20x20"))
+        self.assertEqual(len(renditions), 2)
+
+        # For public images with up-to-date file permissions, rendition.url should return the file url
+        for rendition in renditions.values():
+            self.assertEqual(rendition.url, rendition.file.url)
+            # And cache keys should include 'public'
+            self.assertIn("-public-", rendition.get_cache_key())
+
+        # However, if the file permissions are outdated, rendition.url should return a serve URL
```
RealOrangeOne, line 158: Nit: a lot of the tests in this PR are quite large, and test multiple things in a single case. It might be a good idea to separate them for clarity

cms/private_media/storages.py
```diff
@@ -0,0 +1,54 @@
...
+        obj = self.bucket.Object(file.name)
+        try:
+            obj_acl = obj.Acl()
+        except ClientError as e:
+            logger.warning("Failed to retrieve ACL for %s: %s", file.name, repr(e))
```
RealOrangeOne, line 33: Issue: These should use `logger.exception` so that errors can be surfaced.
```suggestion
            logger.exception("Failed to retrieve ACL for %s", file.name)
```

cms/private_media/tests/test_storages.py
```diff
@@ -0,0 +1,60 @@
...
+            f"WARNING:cms.private_media.storages:Failed to set ACL for {self.document.file.name}: ",
+            logs.output[0],
+        )
+
+    def test_make_private_client_error_on_acl_get(self):
+        object_mock = mock.MagicMock()
+        object_mock.Acl.side_effect = ClientError({"Error": {"Code": "AccessDenied"}}, "GetObjectAcl")
+        # initialize the bucket to allow it to be patched
+        self.storage.bucket  # noqa: B018 # pylint: disable=pointless-statement
+        with mock.patch.object(self.storage._bucket, "Object", return_value=object_mock):  # pylint: disable=protected-access  # noqa: SIM117
```
RealOrangeOne, line 39: Issue: The point of using `mock_aws` is that this isn't necessary

cms/private_media/views.py
```diff
@@ -0,0 +1,34 @@
...
+
+if TYPE_CHECKING:
+    from django.http import HttpResponse
+
+    from cms.private_media.models import AbstractPrivateRendition, PrivateImageMixin
+
+
+class PrivateImageServeView(ServeView):
+    def serve(self, rendition: "AbstractPrivateRendition") -> "HttpResponse":
+        image: PrivateImageMixin = rendition.image
```
RealOrangeOne, line 16: Nit: This might as well be typed as the image model itself

pyproject.toml
```diff
@@ -52,7 +52,7 @@ Werkzeug = "~3.0.6"
 setuptools = "^75.2.0"
 honcho = "^2.0.0"
 detect-secrets = "^1.5.0"
-moto = {extras = ["iam"], version = "^5.0.21"}
+moto = {version = "^5.0.22", extras = ["s3,iam"]}
```
RealOrangeOne, line 55: Question: Should `extras` not be a list, rather than a comma-separated string?

cms/private_media/wagtail_hooks.py
```diff
@@ -0,0 +1,22 @@
...
+
+@hooks.register("before_serve_document")
+def protect_private_documents(document: "PrivateDocumentMixin", request: "HttpRequest") -> None:
+    """Block access to private documents if the user has insufficient permissions."""
+    if document.is_private and (
+        not request.user.is_authenticated
+        or not permission_policy.user_has_any_permission_for_instance(
+            request.user, ["choose", "add", "change"], document
+        )
+    ):
```
RealOrangeOne, line 21: Thought: This logic is duplicated in a few places. Perhaps it should be a method on the mixin? Something like `user_can_access` or alike?

docs/custom-features/media_privacy.md
```diff
@@ -0,0 +1,36 @@
...
+
+The custom `PrivacySettingS3Storage` class logs failed file-setting attempts using the 'WARNING' log level, so should be available to review in most environments should issues occur.
+
+## Compatibility with upstream caching
+
+Because serve view responses and direct file urls can be cached, it's important that when the privacy of a media item changes, the cache
+is invalidated for any affected URLs. In environments where Wagtail's front-end-cache app is configured, all relevant URLs should be collected into a batch and sent to the relevant backend to initiate purging.
+
+NOTE: Image serve view responses are not cached, so don't need to be invalidated.
```
RealOrangeOne, line 36: Issue: This seems to be incorrect: https://github.com/wagtail/wagtail/blob/3e75c018be642eae97eeba22051041280f285825/wagtail/images/views/serve.py#L38

cms/private_media/views.py
zerolab: I am pretty sure pylint then complains that `is_public` and `file_permissions_are_outdated` are not valid attributes

cms/private_media/bulk_operations.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    results: dict[FieldFile, bool] = {}
+
+    def set_file_permission_and_report(file: "FieldFile") -> None:
+        storage = file.storage
+        handler: Callable[[FieldFile], bool] | None
+        if intended_privacy is Privacy.PRIVATE:
+            handler = getattr(storage, "make_private", None)
+        elif intended_privacy is Privacy.PUBLIC:
+            handler = getattr(storage, "make_public", None)
```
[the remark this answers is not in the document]
zerolab: https://docs.python.org/3/howto/enum.html#comparisons states identity comparison (`intended_privacy is Privacy.PRIVATE`) is fine

cms/private_media/views.py
RealOrangeOne: If that's the case, we should probably disable `pylint`'s attribute checking, and instead let `mypy` do it (which is probably better at it anyway).

cms/private_media/managers.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class PrivateMediaModelManager(models.Manager):
+    """A custom model `Manager` to be used by concrete subclasses of
+    `PrivateMediaCollectionMember`. It includes several methods for
+    applying privacy-related changes to multiple objects at the same
+    time.
+    """
+
+    def bulk_set_privacy(self, objects: Iterable["PrivateMediaMixin"], intended_privacy: Privacy) -> int:
```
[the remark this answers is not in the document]
ababic: I consider them to be more in-line with `bulk_update()`, which will operate on any kind of iterable of objects. Therefore, I think the manager is the best place for them.

cms/private_media/managers.py
```diff
@@ -0,0 +1,124 @@
...
+        now = timezone.now()
+        for obj in objects:
+            files = files_by_object[obj]
+            if all(results.get(file) for file in files):
+                obj.file_permissions_last_set = now
+
+        if save_changes:
+            self.bulk_update(objects, ["file_permissions_last_set"])  # type: ignore[arg-type]
+
+        return objects
```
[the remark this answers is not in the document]
ababic: Because it doesn't always save changes to the objects. When being used by the other manager methods (which already call `bulk_update()` to save changes to the DB), saving is not necessary. Where used in other processes, saving is necessary, and the number of updated items holds no value.

cms/private_media/models/images.py
```diff
@@ -0,0 +1,115 @@
...
+        image: PrivateImageMixin = self.image  # pylint: disable=no-member
+        if image.is_public and not image.file_permissions_are_outdated():
+            file_url: str = self.file.url
+            return file_url
+        return self.serve_url
+
+    @cached_property
+    def serve_url(self) -> str:
+        """Return a permission-checking 'ServeView' URL for this rendition."""
+        from wagtail.images.views.serve import generate_image_url  # pylint: disable=import-outside-toplevel
```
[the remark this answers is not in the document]
ababic: I'm pretty sure that results in a circular import - hence why it's inline.

cms/private_media/tests/test_storages.py
ababic: I'm no expert on `mock_aws`, but I couldn't figure out how to easily make it throw errors at different stages of the process like I need it to.

ababic: The main reassurance we are getting from `mock_aws` is actually to test that it can get as far as the ACL setting part (that we're using the right things from boto).

cms/private_media/views.py
ababic: Is is a big deal? This solves the issue whilst remaining concise

pyproject.toml
ababic: I dunno - this was set by poetry add and seems to work

cms/private_media/wagtail_hooks.py
ababic: I don't thing that's worth the added complexity to be honest. I think we'll eventually be overriding the permission policy, and can look at it at that point. For now, we've agreed that this PR should just use stock permissions stuff.

cms/private_media/models/images.py
```diff
@@ -0,0 +1,115 @@
...
+        return created_renditions
+
+    def get_privacy_controlled_serve_urls(self, sites: Iterable["Site"]) -> Iterator[str]:
+        """Return an iterator of fully-fledged serve URLs for this image, covering the domains for all
+        provided sites.
+
+        NOTE: Since image serve URLS in this project are exempt from caching, and this result is only
+        used for cache purging, it doesn't need to return anything.
+        """
+        return iter([])
```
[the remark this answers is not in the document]
ababic: In earlier feedback from Dan, we agreed that the overhead of purging URLs for every rendition outweighed the benefits of caching in the first place

pyproject.toml
ababic: Fixed in [e1ecdb3](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/e1ecdb3ebf071725c4e1e98515bbdf27d20fadca)

docs/custom-features/media_privacy.md
ababic: I believe that is cancelled-out by the fact that the url is included in `private_urls` in `urls.py`, which applies `never_cache` to each item

cms/private_media/models/mixins.py
```diff
@@ -0,0 +1,121 @@
...
+    managed field values are written to the database by the overridden
+    `save()` method.
+
+    Where multiple objects are updated at once (e.g. via a signal handler or
+    management command running on the server), changes to managed field values
+    are written to the database by the bulk update methods provided by
+    `PrivateMediaModelManager`.
+    """
+
+    _privacy = models.CharField(max_length=10, choices=Privacy.choices, default=Privacy.PRIVATE)
```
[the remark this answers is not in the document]
ababic: Because it isn't meant to really be set or modified directly. The property getter and setter do additional things.

cms/private_media/signal_handlers.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def unpublish_media_on_unpublish(instance: "Model", **kwargs: Any) -> None:
+    """Signal handler to be connected to the 'page_unpublished' and 'unpublished'
+    signals for all publishable models. It is responsible for identifying any
+    privacy-controlled media used solely by the object, and ensuring that it is
+    made private again.
+    """
+    page_ct = ContentType.objects.get_for_model(Page)
+    draft_page_ids = {str(pk) for pk in Page.objects.filter(live=False).values_list("id", flat=True)}
```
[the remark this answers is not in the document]
ababic: From what I can tell, there's no way other to query the reference index in a way that would a eliminate the references from drafts. The benefit of doing it this way is that the result can at least be shared by all of the queries for each model. Ultimately, it's a set of short strings, and there will likely be far more live pages than drafts. If it becomes a bottleneck in future, it can be looked at then - but we'd have to be in the millions of pages for it to cause any real problems.

cms/private_media/managers.py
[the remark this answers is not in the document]
RealOrangeOne: `bulk_update` is on the queryset, not the manager. It also only takes instances of the model it works on, rather than any which match a mixin.

cms/private_media/signal_handlers.py
```diff
@@ -0,0 +1,89 @@
...
+                "Did you override the manager class and forget to subclass PrivateMediaModelManager?",
+            )
+
+
+def register_signal_handlers() -> None:
+    """Register signal handlers for models using the private media system."""
+    page_published.connect(publish_media_on_publish, dispatch_uid="publish_media")
+    page_unpublished.connect(unpublish_media_on_unpublish, dispatch_uid="unpublish_media")
+    published.connect(publish_media_on_publish, dispatch_uid="publish_media")
+    unpublished.connect(unpublish_media_on_unpublish, dispatch_uid="unpublish_media")
```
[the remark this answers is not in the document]
ababic: Good spot! I originally thought the two signals were separate, but it turns out [PublishRevisionAction._after_publish](https://github.com/wagtail/wagtail/blob/main/wagtail/actions/publish_revision.py#L90C9-L90C18) emits `published` and [PublishPageRevision._after_publish](https://github.com/wagtail/wagtail/blob/main/wagtail/actions/publish_page_revision.py#L59) calls `super()` after emitting `page_published`, so there's no need for both.

cms/private_media/signal_handlers.py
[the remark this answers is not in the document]
RealOrangeOne: Returning that many rows into memory can still be costly, time and resource wise. I don't think it would be too difficult to inline the query so it's handled as a `JOIN` instead, removing the need to load all the pages into memory.

Alternatively, could this be handled in a different way? Fetch the referenced pages (a small list), then do an additional query to find which of those page ids are draft (`Page.objects.filter(live=False, id__in=...)`). That way there's no `JOIN`, but also no need to load a potentially large number of ids into memory and over the network.

cms/private_media/tests/test_storages.py
RealOrangeOne: Hmm, yeah I'm not sure how to mock getting errors back from the API, other than emulating a file not existing. Emulating IAM issues is probably far more complex.

docs/custom-features/media_privacy.md
RealOrangeOne: Might be worth adding a test to make sure this is the case.

cms/private_media/managers.py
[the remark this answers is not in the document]
ababic: Maybe, but defining on the manager allows subclasses of `PrivateMediaMixin` to easily inherit this behaviour without the need to define custom querysets.

cms/private_media/models/images.py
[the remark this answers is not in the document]
ababic: @MebinAbraham What would you prefer to do here? One to solve on this PR, or something we detattch for later?

cms/private_media/signal_handlers.py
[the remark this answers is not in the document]
ababic: > Alternatively, could this be handled in a different way? Fetch the referenced pages (a small list), then do an additional query to find which of those page ids are draft (Page.objects.filter(live=False, id__in=...)). That way there's no JOIN, but also no need to load a potentially large number of ids into memory and over the network.

It would result in an additional query for each model, but I suppose that's a trade-off worth making. Will revise as suggested.

cms/private_media/storages.py
```diff
@@ -0,0 +1,54 @@
...
+            obj_acl.put(ACL=acl_name)
+        except ClientError as e:
+            logger.warning("Failed to set ACL for %s: %s", file.name, repr(e))
+            return False
+
+        logger.info("ACL set successfully for %s", file.name)
+        return True
+
+
+class DummyPrivacySettingFileSystemStorage(FileSystemStorage):
```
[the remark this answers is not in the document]
ababic: I'll experiment locally, but I figured the lack of persistence might be an issue for tests that use `setUpTestData` instead of the usual `setup`

cms/private_media/tests/test_documents.py
```diff
@@ -0,0 +1,320 @@
...
+from .utils import PURGED_URLS
+
+
+class TestModelConfiguration(SimpleTestCase):
+    def test_document_model_using_private_document_mixin(self):
+        """Verify that the document model correctly inherits from PrivateDocumentMixin.
+        And that the manager is a subclass of PrivateDocumentManager.
+        """
+        document_model = get_document_model()
+        self.assertTrue(issubclass(document_model, PrivateDocumentMixin))
```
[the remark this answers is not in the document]
ababic: Resolved in [52267f2](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/52267f27ca61c2f9de0ff8ef4cdb8e45cdbdf8b9)

cms/private_media/tests/test_images.py
ababic: If they were all `SimpleTestCase` instances, I would agree with you. But, the overhead of doing that for a `TestCase` isn't worth it in my opinion. These tests are well commented and walk through each scenario thoroughly. Things could become more difficult to reason about if broken down further.

cms/private_media/storages.py
ababic: Resolved in [10fdba7](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/10fdba70c715d173719e8eca174c315fe4cef73e)

cms/private_media/tests/test_images.py
ababic: Added in [f67ac65](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/f67ac65c035666c2a343dfbb5222dfc577d296ad)

cms/private_media/tests/test_documents.py
```diff
@@ -0,0 +1,320 @@
...
+            self.assertFalse(image.file_permissions_are_outdated())
+
+        # Another attempt should result in no updates
+        with self.assertNumQueries(1):
+            # Query summary:
+            # 1. to fetch the documents
+            self.assertEqual(self.model.objects.bulk_make_private(self.model.objects.all()), 0)
+
+
+class TestDocumentServeView(TestCase):
```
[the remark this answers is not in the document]
ababic: Added in [f67ac65](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/f67ac65c035666c2a343dfbb5222dfc577d296ad)

cms/private_media/tests/test_documents.py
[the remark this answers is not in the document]
ababic: Oh, apparently not, so reverted in [f0c5d04](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/f0c5d04e0b8466533ff694982a9616a2bafc79ba)

cms/private_media/tests/test_documents.py
```diff
@@ -301,21 +301,25 @@ def test_private_document(self):
         response = self.client.get(self.private_document.url)
         self.assertEqual(response.status_code, 200)
 
-    def test_public_document(self):
+    @override_settings(IS_EXTERNAL_ENV=True)
+    def test_serve_private_document_external_env(self):
+        """Test the serve view behaviour for private documents in an external environment."""
+        self.test_serve_private_document()
```
RealOrangeOne, line 307: This feels like a really strange way of handling it. It might be better to use `subTest` to run the test twice.

ababic: "really strange" is a bit strong, no? To me, this makes the intention more obvious.

cms/private_media/signal_handlers.py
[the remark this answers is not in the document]
ababic: Revised in [a2748a6](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/a2748a6721f1fd8f418be13ba972a798af82e6bd)

cms/private_media/storages.py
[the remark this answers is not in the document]
ababic: Fixed in [439d5a1](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/439d5a1b04bbe433a3426e1c77db37da5cdc8c1b) and

nehakerung (changes requested):
LGTM - The doc was simple to understand and follow. I also tested as how to review suggested.

Everything was as said but for _Testing media privacy after a page has been published_, 
`media/images/{image filename}.fill-446x390.format-webp-.original.png` showed as `media/images/{image filename}.original.png` instead. I presume this is because I didn't set any fill/format for my images so I don't believe this to be important.

Just requesting an update on hyperlink :D

docs/custom-features/media_privacy.md
```diff
@@ -0,0 +1,97 @@
...
+
+1. Images and documents are 'private' by default, and will only be served to authenticated users with some level of permission on the media item.
+2. They only become 'public' when a page referencing them is published.
+3. When a live page unpublished, any images or documents referenced soley by that page (i.e. not used on any other live pages) are made private again.
+
+When a media item is 'private', file requests are handled by views that check the permissions of the requesting user, and only serve the file if the user has some level of permission on the media item (to support previews and reviews from other editors).
+
+When a media item is 'public', permission-checking in serve views is skipped, and the user is either redirected to the media storage URL (in the case of images), or served the file directly (in the case of documents).
+
+These cases (and more) are covered by integration tests in [cms/private_media/tests/test_signal_handlers.py](https://github.com/ONSdigital/dis-wagtail/tree/main/cms/private_media/tests/test_signal_handlers.py).
```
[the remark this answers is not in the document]
ababic: Not yet, but it will once merged :)

nehakerung (approved):

cms/private_media/views.py
```diff
@@ -0,0 +1,34 @@
...
+        if image.is_public and not image.file_permissions_are_outdated():
+            return redirect(rendition.file.url)
+
+        # Block access to private images if the user has insufficient permissions
+        user = self.request.user
+        if image.is_private and (
+            not user.is_authenticated
+            or not permission_policy.user_has_any_permission_for_instance(user, ["choose", "add", "change"], image)
+        ):
+            raise PermissionDenied
```
MebinAbraham, line 29: We need to ensure that the file can still be served if the image is private but has public pages referencing it. For example, if a page is published but the ACLs are not properly set due to network issues etc, the application should still serve the image until the next scheduled task attempts to fix and retry the ACL settings. We cannot have a page that features broken assets. The same principle applies to documents.

cms/private_media/wagtail_hooks.py
MebinAbraham, line 21: As above, we need to be able to serve even if it is private just incase ACL set failed.

cms/private_media/storages.py
```diff
@@ -0,0 +1,54 @@
...
+from storages.backends.s3 import S3Storage
+
+if TYPE_CHECKING:
+    from django.db.models.fields.files import FieldFile
+
+
+logger = logging.getLogger(__name__)
+
+
+class PrivacySettingS3Storage(S3Storage):
```
MebinAbraham, line 15: Nit: I don't find `PrivacySetting` very clear; what about something like `PrivacyAwareS3Storage` or `AccessControlledS3Storage`

MebinAbraham (commented):
Nice work.

I'm happy for it to be separate, but I suggest we disable the Media library from Admin. It will add an unnecessary layer of complexity.

cms/private_media/management/commands/retry_file_permission_set_attempts.py
```diff
@@ -0,0 +1,62 @@
...
+                    if obj.privacy is Privacy.PRIVATE:
+                        make_private.append(obj)
+                    elif obj.privacy is Privacy.PUBLIC:
+                        make_public.append(obj)
+
+                self.update_file_permissions(model, make_private, Privacy.PRIVATE)
+                self.update_file_permissions(model, make_public, Privacy.PUBLIC)
+
+    def update_file_permissions(
+        self, model_class: type["PrivateMediaMixin"], items: list["PrivateMediaMixin"], privacy: Privacy
```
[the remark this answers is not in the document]
ababic: I definitely agree where it comes to reusable code, but in the case of management commands (which are self-contained, and only really reusable if you're defining a base class for other commands to inherit), I think being explicit adds value.

cms/private_media/signal_handlers.py
```diff
@@ -0,0 +1,110 @@
...
+            .annotate(int_id=Cast("to_object_id", output_field=IntegerField()))
+            .values_list("int_id", flat=True)
+            .distinct()
+        )
+
+        queryset = (
+            model_class.objects.all()
+            .filter(pk__in=[int(pk) for pk in referenced_pks], _privacy=Privacy.PUBLIC)
+            .exclude(pk__in=live_page_referenced_media_pks)
+        )
```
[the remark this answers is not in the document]
ababic: This has now been refactored so that the potential list of pages only includes those that reference the same media referenced by the current object. If, like you say, the that case is rare, it should be very small indeed.

Its always better for important code like this to assume as little as possible about what other things might be doing, so even if we did manage to somehow restrict media use to a single object (I currently have no idea how that could be acheived), I'd still very much recommend keeping this.

Where the queryset result isn't explicitly evaluated to a list/set beforehand, Django usually keeps everything lazy and has the database evaluate everything in one go, so this is really quite performant (there should be some assertNumQueries items in the tests to confirm this). So, it isn't even necessarily bad performance wise.

cms/private_media/views.py
ababic: This is covered by signal handlers. If a page features any media that is private, it is made public when those changes are published, and whilst the page remains published, media referenced by it will always remain public. 

Also, the `_privacy` field should always be the canonical indicator of privacy for a media item. If we're having to make additional queries at the time of serving to figure out if that is really the case, then we've failed terribly.

cms/private_media/wagtail_hooks.py
ababic: Documents are different to images in thar they are ALWAYS served by the view, regardless of privacy values. So, in essence, the ACL is irrelevant here, as the storage backend uses its credentials to stream the file

cms/private_media/storages.py
ababic: Ah, naming things!! I like the 2nd suggestion. Let's go with that

cms/private_media/signal_handlers.py
```diff
@@ -0,0 +1,110 @@
...
+    for model_class in get_private_media_models():
+        model_ct = ContentType.objects.get_for_model(model_class)
+        referenced_pks = (
+            ReferenceIndex.get_references_for_object(instance)
+            .filter(to_content_type=model_ct)
+            .annotate(int_object_id=Cast("to_object_id", output_field=IntegerField()))
+            .values_list("int_object_id", flat=True)
+            .distinct()
+        )
+        queryset = model_class.objects.filter(pk__in=referenced_pks, _privacy=Privacy.PRIVATE)
```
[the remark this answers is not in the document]
ababic: The issue there is that checking the result at this point will evaluate the query (a separate database round trip), currently, all of the subqueries remain lazy until the final queryset is evaluated, and postgres handles it in one round-trip).

cms/private_media/bulk_operations.py
```diff
@@ -0,0 +1,57 @@
...
+
+        if handler is None:
```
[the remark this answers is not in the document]
ababic: Funnily enough, that's how I had it originally, but I changed my mind at some point. Should be very easy to restor.

cms/private_media/views.py
ababic: >  if a page is published but the ACLs are not properly set due to network issues etc, the application should still serve the image until the next scheduled task attempts to fix and retry the ACL settings.

This particular case is handled by the last line of the method.

cms/private_media/managers.py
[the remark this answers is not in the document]
ababic: But it is for operating applying changes to an arbitrary group of objects, which is what managers are for, and not a specific quereyset (which is why it's not a custom queryset class). I'm going to leave it as it is.

cms/private_media/management/commands/retry_file_permission_set_attempts.py
```diff
@@ -0,0 +1,62 @@
...
+                make_private = []
+                make_public = []
+                for obj in permissions_outdated:
+                    if obj.privacy is Privacy.PRIVATE:
+                        make_private.append(obj)
+                    elif obj.privacy is Privacy.PUBLIC:
+                        make_public.append(obj)
+
+                self.update_file_permissions(model, make_private, Privacy.PRIVATE)
+                self.update_file_permissions(model, make_public, Privacy.PUBLIC)
```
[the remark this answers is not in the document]
ababic: Funnily enough, that's how I had it originally, but it was before I broke handling off to `update_file_permissions`, and pylint had a problem with the number of return statements. It should be okay to do that again now that `update_file_permissions` exists

cms/private_media/storages.py
ababic: Resolved in [60fdb15](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/60fdb1593874fa6f8711694c777e52aec9557d24)

cms/private_media/managers.py
```diff
@@ -0,0 +1,124 @@
...
+        related files to reflect `intended_privacy`. Returns a list of the provided objects,
+        with their `file_permissions_last_set` datetime updated if all related files
+        were successfully updated.
+        """
+        files_by_object: dict[PrivateMediaMixin, list[FieldFile]] = defaultdict(list)
+        all_files = []
+        objects = list(objects)
+
+        for obj in objects:
+            for file in obj.get_privacy_controlled_files():
```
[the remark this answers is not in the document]
ababic: It could yes. It's up to the `PrivateMediaMixin` to define how `get_privacy_controlled_files()` works for that model. For example, you might have two or three individual file fields that need protecting (a data CSV and a thumbnail image for charts)

cms/private_media/managers.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def bulk_make_private(self, objects: Iterable["PrivateMediaMixin"]) -> int:
+        """Make an iterable of objects of this type 'private' as efficiently as
+        possible. Returns the number of objects that were actually updated
+        in the process.
+        """
+        return self.bulk_set_privacy(objects, Privacy.PRIVATE)
+
+    def bulk_set_file_permissions(
+        self, objects: Iterable["PrivateMediaMixin"], intended_privacy: Privacy, *, save_changes: bool = False
```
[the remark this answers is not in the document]
ababic: Resolved in [dee1193](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/dee11931d65f9207e1cc9d3b17b0e7a842db7df3)

cms/private_media/models/mixins.py
```diff
@@ -0,0 +1,121 @@
...
+            **kwargs: Additional keyword arguments passed to parent save method
+        """
+        # Save model field changes at this point
+        super().save(*args, **kwargs)
+
+        # Trigger file permission updates after-the-fact
+        if set_file_permissions and self.file_permissions_are_outdated():
+            results = bulk_set_file_permissions(self.get_privacy_controlled_files(), self.privacy)
+            # Only update 'file_permissions_last_set' if all updates were successfull
+            if set(results.values()) == {True}:
```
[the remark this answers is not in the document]
ababic: Resolved in [50f131d](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/50f131d6d0b5a3c8ff9d19961e543bcc2c917b2b)

cms/private_media/bulk_operations.py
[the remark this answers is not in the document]
ababic: Resolved in [16a05fb](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/16a05fb0c4f38f0b846f84dbf8de1284e6da094e)

cms/private_media/bulk_operations.py
```diff
@@ -0,0 +1,57 @@
...
+                storage.__class__.__name__,
+                intended_privacy,
+                file.name,
+            )
+            results[file] = True
+        else:
+            results[file] = handler(file)
+
+    with concurrent.futures.ThreadPoolExecutor(
+        max_workers=int(settings.PRIVATE_MEDIA_PERMISSION_SETTING_MAX_WORKERS)
```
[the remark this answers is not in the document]
ababic: I'd like to stick with the `PRIVATE_MEDIA_` prefix, as it makes it easier to mentally map settings to apps in the project. But, I do find the current name a little confusing, so have changed to `PRIVATE_MEDIA_BULK_UPDATE_MAX_WORKERS` in [ce09c50](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/ce09c50e132ff6f6c7be39ca3af93326ee60db9d).

Whilst it's still a bit lengthy, I think clarity is more important when it comes to settings.

cms/private_media/managers.py
```diff
@@ -0,0 +1,124 @@
...
+    ) -> list["PrivateMediaMixin"]:
+        """For an itrerable of objects of this type, set the file permissions for all
```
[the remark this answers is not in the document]
ababic: Resolved

cms/private_media/management/commands/retry_file_permission_set_attempts.py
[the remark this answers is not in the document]
ababic: Updated in [5198943](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/51989435c30917f3233920e1d77f2b5b72826519)

cms/private_media/management/commands/retry_file_permission_set_attempts.py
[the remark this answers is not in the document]
ababic: Resolved in [7f0de81](https://github.com/ONSdigital/dis-wagtail/pull/46/commits/7f0de81f3fe70d3fbaf9115c7f2a212bf337b56e)

cms/private_media/management/commands/retry_file_permission_set_attempts.py
```diff
@@ -0,0 +1,62 @@
...
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            dest="dry_run",
+            default=False,
+            help="Dry run -- don't change anything.",
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        self.dry_run = options["dry_run"]  # pylint: disable=attribute-defined-outside-init
```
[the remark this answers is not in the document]
ababic: This has been resolved

cms/private_media/models/mixins.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @property
+    def privacy(self) -> Privacy:
+        """Return the privacy status of the object."""
+        if isinstance(self._privacy, Privacy):
+            return self._privacy
+        for item in Privacy:
+            if item.value == self._privacy:
+                return item
+        raise ValueError(f"Invalid privacy value: {self._privacy}")
```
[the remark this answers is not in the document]
ababic: If can be both, depending what you're doing. `Privacy.PRIVATE` and `Privacy.PUBLIC` are both valid init/create kwarg values, as they both evaluate to a string on save.

cms/private_media/signal_handlers.py
```diff
@@ -0,0 +1,110 @@
...
+    from django.db.models import Model
+
+
+def publish_media_on_publish(instance: "Model", **kwargs: Any) -> None:
+    """Signal handler to be connected to the 'page_published' and 'published'
+    signals for all publishable models. It is responsible for identifying any
+    privacy-controlled media used by the object, and ensuring that it is also
+    made public.
+    """
+    for model_class in get_private_media_models():
```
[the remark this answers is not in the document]
ababic: Potentially, but parallelism is a fair bit more complicated when database connections are involved.

cms/private_media/models/images.py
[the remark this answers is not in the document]
ababic: Okay, I've restored caching for image serves, and cache purging for image rendition URLs

cms/private_media/signal_handlers.py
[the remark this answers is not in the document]
ababic: I suggest avoid this change based on the above reasoning. @MebinAbraham are you happy with that?

cms/private_media/signal_handlers.py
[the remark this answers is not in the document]
ababic: @MebinAbraham Happy to mark this as resolved based on the above explaination?

cms/private_media/views.py
ababic: @RealOrangeOne Could you mark this one as resolved please?

cms/private_media/models/images.py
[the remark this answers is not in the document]
ababic: @RealOrangeOne Happy to mark this one as resolved now?

cms/private_media/tests/test_images.py
ababic: @RealOrangeOne Happy to mark this one as resolved?

docs/custom-features/media_privacy.md
ababic: Now we're caching image urls again, this point is redundant. @RealOrangeOne are you happy to mark this as resolved?

cms/private_media/tests/test_documents.py
ababic: @RealOrangeOne Could you mark this as resolved please?

cms/private_media/models/mixins.py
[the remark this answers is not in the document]
MebinAbraham: Right, but I'm not following why we need the loop and exception handling?. How is it possible for the value not to be from Privacy? The code paths should prevent any other value being saved?.

cms/private_media/signal_handlers.py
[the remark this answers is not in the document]
MebinAbraham: That's true, let's revisit if it becomes a problem.

cms/private_media/views.py
MebinAbraham: > > if a page is published but the ACLs are not properly set due to network issues etc, the application should still serve the image until the next scheduled task attempts to fix and retry the ACL settings.
> 
> This particular case is handled by the last line of the method. The image is 'public' according to the system, so won't be permission-checked. It'll just be served, using the server's AWS creds to stream the file contents.

Is it not possible for it to still be private?. For example, `bulk_set_file_permissions` is called before updating the object privacy status; if `bulk_set_file_permissions` raises an exception other than `ClientError`, the object will not be updated, hence still marked as private. It currently handles it raising a ClientError only but looks like there could be other exceptions raised?

ababic: As discussed, I'll look more closely at the Exception types that ACL getting/setting operations can raise in Boto.

cms/private_media/wagtail_hooks.py
MebinAbraham: As discussed, one to revisit from performance testing as to why we wouldn't want to direct to S3 for documents also. Would result in fewer requests.

MebinAbraham (approved):
Great work @ababic 🎉

cms/private_media/views.py
ababic: @MebinAbraham Looking at the boto3 codebase, and I think we can safely assume that anything we encounter here will either be a `ClientError` (from deeper down in botocore) or a `Boto3Error` (Which all [exceptions in boto3](https://github.com/boto/boto3/blob/develop/boto3/exceptions.py) inherit from). So, I have updated the code to reflect that.

Happy to mark this one as resolved?

cms/private_media/storages.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+
+    def make_private(self, file: "FieldFile") -> bool:
+        """Pretend to make the provided file private."""
+        logger.info("Skipping private file permission setting for '%s'.", file.name)
+        return True
+
+    def make_public(self, file: "FieldFile") -> bool:
+        """Pretend to make the provided file public."""
+        logger.info("Skipping public file permission setting for '%s'.", file.name)
```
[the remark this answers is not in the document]
ababic: @MebinAbraham This was addressed in 84b6dc5b0b9855abd3c2b7c99bb16a24f6f5e445. Are you happy to mark as resolved?

MebinAbraham (approved):

zerolab (approved):

sanjeevz3009 (commented):
Reviewed!

PR #47, opened 2024-12-03
Disable pylint's `django-manager-missing` globally

zerolab:
Tentative global disable for `django-manager-missing`.
Technically this is pylint-django, not pylint, but still.

IMHO this is adding loads of noise, but folk may deem it needed hence "tentative"

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -438,6 +438,7 @@ disable=raw-checker-failed,
         too-few-public-methods,
         missing-class-docstring,
         missing-function-docstring,
+        django-manager-missing,
         fixme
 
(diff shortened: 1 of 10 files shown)


MebinAbraham (approved):

zerolab:
Gah, this is a mypy, not pylint issue. reworking

zerolab:
Will have to close as mypy currently doesn't support disabling error codes provided by plugins. See https://github.com/python/mypy issue 12987

sanjeevz3009 (commented):
Reviewed!

PR #33, opened 2024-11-20
Initial support for bundling

zerolab:
### What is the context of this PR?

This PR adds basic bundling functionality, ported from alpha. https://jira.ons.gov.uk/browse/CMS-205/
It also reverts #22

To-Do:

- [x] finish test coverage

### How to review

See walkthroughs.

Screencast: https://www.loom.com/share/23a483f107e948e6994305ac365f80b1

### Follow-up Actions

- Enhance functionality ([CMS-232](https://jira.ons.gov.uk/browse/CMS-232))
   - page-level workflows
   - teams-based access for non-publishing staff
   - reduce potential impact of Slack notifications call
   - add `assertNumQueries` for the `publish_scheduled` management command
- Add docs (e.g. how to make a page bundleable)

PR #50, opened 2024-12-04
Bundles: limit release calendar pages to provisional/confirmed

zerolab:
This builds on #33 and adds a separate chooser for the Bundle's release calendar page chooser that limits the options to only RC pages that are not cancelled or already published.

https://www.loom.com/share/2f26efc2772142af83b3f3beff7b44e6

also ensures chosen pages are deduplicated


cms/release_calendar/viewsets.py
```diff
@@ -0,0 +1,51 @@
...
+            self.title_column,  # type: ignore[attr-defined]
+            Column("release_date"),
+            Column("release_status", label=_("Status"), accessor="get_status_display"),
+            DateColumn(
+                "updated",
+                label=_("Last Updated"),
+                width="12%",
+                accessor="latest_revision_created_at",
+            ),
+            PageStatusColumn("status", label=_("Page status"), width="12%"),
```
MebinAbraham, line 28: Non-blocker: I found seeing both Status and Page status in the chooser confusing. Is page status important enough to show, it doesn't impact the selection?. Though, that might just be me 🤷

MebinAbraham (commented):
Food for thought: This prevents the chooser from showing Cancelled or Published but doesn't restrict those server-side. You could also have selected one that was Confirmed and subsequently cancelled, but no errors would occur. I'm happy for the resolution/consideration of that to be separate.

cms/release_calendar/viewsets.py
zerolab: I didn't have it originally, but then I though it may be useful, added it. Will drop

zerolab: cc4da27

MebinAbraham (approved):

sanjeevz3009 (approved):
Reviewed and locally everything is as expected! LGTM =)

PR #55, opened 2024-12-09
Styling for the release page plus some refactoring

helenb:

Styling of the release calendar page

- Tidied up commenting on scss files
- Refactored some heading styles to make them common to all pages
- Moved `release` styles to the common-header
- Fixed spacing with the accredited logo
- Added conditionals around page content where necessary
- Create a re-usable spacing class for streamfield and other content so we have consistent spacing between page blocks
- Adjusted spacing below paragraphs for all rich text blocks after discussion with Gareth
- Styling of accredited logo and census logo both in the header and in the main content
- Ensure the census logo is visible in high contrast mode
- Add a reusable component for contact details

### How to review

In your local build, check out this branch, and view a release page.

Check it in all it's various different states at both desktop and mobile:
- published
- confirmed
- provisional
- cancelled

Check the logos in the following combinations at both desktop and mobile:

- Page is accredited but does not have census data
- Page has census data but is not accredited
- Page is accredited and has census data

### Follow-up Actions

None


cms/jinja2/templates/components/census/census-information.html
```diff
@@ -1,4 +1,4 @@
 <div class="ons-u-mb-xl census-2021">
-    {% include "templates/components/census/census-logo.html" %}
+    <img src="https://cdn.ons.gov.uk/assets/images/census-logo/logo-census-2021-purple-landscape.svg" alt="Census 2021 logo" width="171" class="ons-u-mb-s">
```
helenb, line 2: Removed the include here as we need a different class in this case.

zerolab: question: could we pass the class to the include?

helenb: I tried that, but nunjucks doesn't have a `with` option for includes. I think we'd have to make it a macro, which feels like overkill.

zerolab:
```jinja2
{% with census_logo_classes="ons-u-mb-s" %}
    {% include "templates/components/census/census-logo.html" %}
{% endwith %}
```

should do it.

helenb: ooh hello.

cms/jinja2/templates/components/streamfield/stream_block--related-links.html
```diff
@@ -1,4 +1,4 @@
-<ol class="ons-list ons-list--bare ons-u-bt ons-u-mb-xl">
```
helenb, line 1: This is different to how the related links are presented on the analysis page. They are using slightly different streamfield blocks. I'm unsure if there was a deliberate reason for this so have left them for now, but they may need rationalising.

cms/jinja2/templates/components/contact_details/contact_details.html
```diff
@@ -0,0 +1,21 @@
+{% from "components/text-indent/_macro.njk" import onsTextIndent %}
+
+{% if page.contact_details %}
```
zerolab, line 3: suggestion: make this more agnostic. That is, have it expect a `contact_details` variable (which is a `ContactDetails` instance

helenb: Done.

nehakerung (commented):
LGTM -tested locally, looks great

zerolab (approved):
Tested locally and LGTM

nehakerung (approved):
LGTM

PR #34, opened 2024-11-20
CMS-205: Update design system version

helenb:
### What is the context of this PR?

Update the design system to the latest version, and associated changes to utility classes and variable names
https://jira.ons.gov.uk/browse/CMS-209

### How to review

- Check out the branch
- When in the docker shell, run `make load-design-system-templates`
- Review various kinds of pages on your local build, but especially:
    - documents on an analysis page
    - the release calendar page

See the instructions for the major design system update here: https://github.com/ONSdigital/design-system/blob/main/migration_guides/70.x.x-to-72.0.0-migration-guide.md

I have split the changes into commits to make it easier to review, but it is best not to test each commit separately as some functionality will break between commits.

PR #56, opened 2024-12-10
UAT fixes

zerolab:
https://jira.ons.gov.uk/browse/CMS-265
https://jira.ons.gov.uk/browse/CMS-225

and bumps dependencies to chiefly get us Django 5.1.4 (to silence Dependabot https://github.com/ONSdigital/dis-wagtail/security/dependabot/5)


cms/jinja2/templates/base.html
```diff
@@ -350,7 +350,7 @@
     onsCookiesBanner({
       'lang': "cy" if LANGUAGE_CODE == "cy" else "en",
       'serviceName': COOKIE_BANNER_SERVICE_NAME,
-      'settingsLinkURL': MANAGE_COOKIE_SETTINGS_URL,
+      'settingsLinkUrl': MANAGE_COOKIE_SETTINGS_URL,
```
zerolab, line 353: missed in https://github.com/ONSdigital/dis-wagtail/pull/34/files#diff-bd7285587bee7451cbc78ed59b209035ad01327b8f0dcbfe38341cea13801682L120

kacperpONS (approved):
LGTM!

cms/core/models/snippets.py
```diff
@@ -11,7 +12,7 @@ class ContactDetails(index.Indexed, models.Model):
     Note that this is registered as a snippet in core.wagtail_hooks to allow customising the icon.
     """
 
-    name = models.CharField(max_length=255)
+    name = models.CharField(max_length=255, unique=True)
```
MebinAbraham, line 15: We don't have the requirements for this. I can see a case where they may have the same name but different email addresses. Given that both the name and email are mandatory, I suggest we use a unique constraint based on these two values. Thoughts?

zerolab: still trying to get confirmation. will make name+email unique together

kacperpONS: That makes sense to, but then it would be good to add the email in another column in the table on the page so they can be easily distinguished.

zerolab: 72f51bd now adds:

Screenshot 2024-12-10 at 11 43 09
Screenshot 2024-12-10 at 11 43 16

cms/core/models/snippets.py
```diff
@@ -26,11 +27,24 @@ class ContactDetails(index.Indexed, models.Model):
...
     class Meta:
-        verbose_name_plural = "contact details"
+        verbose_name = _("contact details")
+        verbose_name_plural = _("contact details")
+        constraints: ClassVar[list[models.BaseConstraint]] = [
+            models.UniqueConstraint(
+                name="%(app_label)s_%(class)s_name_unique",
+                fields=["name", "email"],
+            ),
+        ]
```
MebinAbraham, line 42: We should have a test to validate this.

cms/core/models/snippets.py
kacperpONS: Haven't run it myself yet, but on the screenshot the name column header seems to be inconsistent between the two pages - probably should be "Name" for both.

cms/core/models/snippets.py
zerolab: 92b319b

cms/core/models/snippets.py
zerolab: Good spot @kacperpONS, fixed in 8d9f466

cms/core/models/snippets.py
MebinAbraham, line 42: I'm not sure what the answer should be, but these are currently case-sensitive. It feels more appropriate if they were case-insensitive checks; otherwise, you can have tw that look almost the same as long as they differ in case.

MebinAbraham: Nice one, small issue, the error message is not useful to the end user. It currently displays:

image

MebinAbraham (approved):

kacperpONS (approved):

sanjeevz3009 (commented):
Reviewed!

PR #66, opened 2025-01-07
Bundles: exclude Release Calendar pages with a date in the past

zerolab:
Builds on #50 which excluded cancelled/published RC pages.

To test:
- create a release calendar page (provisional, release_date to now)
- go to a bundle and open the release calendar chooser
- it should not be there

Also, this bumps the dependencies too. Chiefly Wagtail to [6.3.2](https://docs.wagtail.org/en/stable/releases/6.3.2.html) and Pillow to 11.1.0 (which should speed up PNG saving in some cases).
Supersedes #61 too


AdamHawtin (approved):

nehakerung (commented):
LGTM - Tested it and works as promised

nehakerung (approved):
LGTM

PR #68, opened 2025-01-08
Ensure the default DB is **always** used for writes

RealOrangeOne:

@zerolab raised an issue to me that trying to delete snippets would hang. On further inspection, the `read_replica` connection was being used for the `DELETE` query, which isn't correct. The "hang" is likely due to a lock between the 2 database connections.

The issue was in our database router, where we weren't enforcing that the default connection was used for writes. Without the enforcement, Django often defaults to the same connection used to obtain the instance, which is likely the read instance.

The fix itself is fairly small, and the rest of the PR is about test hardening.

### Follow-up Actions

Handling queries like this is quite brittle, so some follow-up to potentially change the way this is handled and move it to the infrastructure layer.


cms/core/tests/test_db_router.py
```diff
@@ -55,5 +68,27 @@ def test_uses_correct_db_in_transaction(self):
...
+        with self.assertNumQueriesConnection(replica=1):
             list(HomePage.objects.search("Home"))
+
+    def test_deletes_from_default_when_read_from_replica(self):
+        user = User.objects.create_user("user", "email@example.com", "password")
+
+        self.assertEqual(user._state.db, DEFAULT_DB_ALIAS)  # pylint: disable=protected-access
+
+        # HACK: The replica connection can't see the user right now for some reason
+        user._state.db = READ_REPLICA_DB_ALIAS  # pylint: disable=protected-access
```
zerolab, line 80: note: I wish we didn't need to resort to hacks, but it does the job for the purpose of the test.

zerolab (approved):
LGTM. 
and it makes sense since Django will default to using the connection it used for read.
Works nicely locally too

AdamHawtin (approved):
LGTM

PR #72, opened 2025-01-10
Fix to release content on the release page

helenb:

The release content block on the release page was not correctly filtering rich text, and had some spacing issues

### How to review

Create a new release page or edit an existing one, and ensure it is published. Add a few links for the release content.
Ensure that they display the description correctly and that there is a 16px gap between each link.

Compare to the release content on beta: https://ons-beta.torchbox.dev/releases/helen-test/

### Follow-up Actions

None


cms/jinja2/templates/components/streamfield/release_content_section.html
```diff
@@ -3,14 +3,16 @@ <h2>{{ value.title }}</h2>
...
+            {% if item.link %}
+                <li class="ons-list__item">
+                    <div class="ons-u-mb-s">
+                        <a href="{{ item.link.url }}">{{ item.link.text }}</a>
+                        {% if item.link.description %}
+                            {{ item.link.description|richtext }}
+                        {% endif %}
+                    </div>
+                </li>
+            {% endif %}
```
zerolab, line 15: These changes LGTM as they are.

I was cross-referencing this with the `onsList` component (well, DS macro) and it looks like it adds specific classes on the `a` tag:

- https://github.com/ONSdigital/design-system/blob/main/src/components/list/_macro.njk#L88 for internal links
- https://github.com/ONSdigital/design-system/blob/main/src/components/external-link/_macro.njk#L4 for external links

which I think are worth following. What do you think?

zerolab: on a second thought - let's do this as a follow up task (can you add it as a task to the statistical article or methodology DS integration ticket)?

zerolab (approved):

nehakerung (approved):
LGTM!

cms/jinja2/templates/components/streamfield/release_content_section.html
helenb: I'll resolve this conversation as there is nothing further for @zerolab to check - I have added a comment on the statistical article page ticket.

PR #74, opened 2025-01-10
Fix for Methodology page 

kacperpONS:

A small fix to make the equations and embeds render correctly.


zerolab (commented):
This works, even if inneficient.
Let's get it in so it unblock further testing. Will submit a separate PR that should eliminate the need for these custom properties on page models

zerolab (approved):

sanjeevz3009 (approved):
Pulled it down, tested it locally. Tested every single block possible. All works/ LGTM =)

zerolab:
Follow up #75

PR #75, opened 2025-01-10
Make the equations and ons embed block check better

zerolab:

This is a follow up/replacement for #74

### How to review

Add an equation / ONS embed block to an information page, methodology page, statistical article page. Check they render as expected

### Follow-up Actions

N/A


kacperpONS:
I've just merged my branch, so after you merge main into this one, the two functions in the Methodology page model can be removed too 👍

zerolab:
@ababic what do you think of this approach? Considered a mixin too, but it is prone to errors

AdamHawtin (approved):
👍 LGTM equation rendered as expected.

cms/core/models/base.py
```diff
@@ -66,9 +66,13 @@ def has_equations(self) -> bool:
...
         """
-        if streamfield := getattr(self, self.content_field_name):
+        if streamvalue := getattr(self, self.content_field_name):
+            # run the check on the StreamBlock itself, if it supports it
+            if hasattr(streamvalue.stream_block, "has_equations"):
+                return bool(streamvalue.stream_block.has_equations(streamvalue))
+
             try:
-                return streamfield.first_block_by_name(block_name="equation") is not None
+                return streamvalue.first_block_by_name(block_name="equation") is not None
```
ababic, line 75: @zerolab I feel like we might be better to remove this alternative option and just let the StreamBlock implementation be the canonical API. What do you think?

cms/core/models/base.py
```diff
@@ -79,9 +83,13 @@ def has_ons_embed(self) -> bool:
...
         """
-        if streamfield := getattr(self, self.content_field_name):
+        if streamvalue := getattr(self, self.content_field_name):
+            # run the check on the StreamBlock itself, if it supports it
+            if hasattr(streamvalue.stream_block, "has_ons_embed"):
+                return bool(streamvalue.stream_block.has_ons_embed(streamvalue))
+
             try:
-                return streamfield.first_block_by_name(block_name="ons_embed") is not None
+                return streamvalue.first_block_by_name(block_name="ons_embed") is not None
```
ababic, line 92: Same here. I think we should embrace the new API and return `False` if not implemented.

ababic (commented):
@zerolab Really like the new API. Just had a couple of recommendations to remove old checking mechanism completely, as it makes it a bit difficult to know what you need to do when adding a new field

cms/core/models/base.py
kacperpONS: Just a note - if we decided to do that and remove those methods from `BasePage`, then the `content_scripts.html` template would have to be modified too.
We wouldn't be able to simply do  `{% if page.has_equations %}` anymore and would have to filter out the relevant `StreamBlock` and call its own `has_....` function.

zerolab: Good shout. 68232d5

cms/core/models/base.py
zerolab: 68232d5

cms/core/models/base.py
zerolab: @kacperpONS we're not removing the methods, just the fallback check. See 68232d5

kacperpONS (approved):

PR #76, opened 2025-01-14
Update dev and test configs to streamline developer experience

AdamHawtin:
Update the development and test defaults, so that running the app and tests locally works out of the box, without setting additional environment variables. Developing and testing in the docker container should also still work with no addtional steps, as the compose file already overrides the environment variables to its needs.

### How to review

- Check out this branch locally
- Remove or rename any previous `.env` files or environment configuration you previously had for native local development
- Follow the README steps for "Running locally with supporting services in Docker"
- Try to start the app locally, either from your shell or IDE of choice
- Check the app runs successfully without you having to configure additional environment variables (other than setting the `DJANGO_SETTINGS_MODULE`, if you are running by a method other than the `make runserver` command, as the README says).
- Run the unit/component tests locally. Note there may be an unrelated issue with running the tests in the parallel locally with the `make test` command, if this fails with an error try running `poetry run coverage run ./manage.py test --shuffle --settings=cms.settings.test` instead to run them in series.
- Stop the locally running app
- Try running the app and tests in the docker container, following the README instructions, to check the config changes haven't broken that method.

### Follow-up Actions
None


nehakerung (commented):
LGTM - followed instructions which worked 🫡

README.md
```diff
@@ -145,20 +145,8 @@ You can then access the admin at `http://0.0.0.0:8000/admin/` or `http://localho
...
-some variables (such as `DATABASE_URL` and `REDIS_URL`) in the `.env` file.
-
-> [!NOTE]
-> When running the application locally in a virtual environment via Poetry the `.env` file will not be picked up automatically.
-> For this to work you'll need to install the [poetry-plugin-dotenv](https://github.com/pivoshenko/poetry-plugin-dotenv).
-> However, if you installed Poetry with `brew` rather than `pip` that currently isn't going to
-> work (see the [issue](https://github.com/pivoshenko/poetry-plugin-dotenv/issues/327)) and you'll need to install an older and
-> seemingly no longer maintained [poetry-dotenv-plugin](https://github.com/mpeteuil/poetry-dotenv-plugin).
+The correct development configuration should be used by default when running from the shell, for IDE's you may need to add
+`DJANGO_SETTINGS_MODULE=cms.settings.dev` to the runtime configuration (for PyCharm see [here](https://www.jetbrains.com/help/pycharm/run-debug-configuration.html#createExplicitly)).
```
zerolab, line 149: nitpick:
```suggestion
The correct development configuration should be used by default when running from the shell. For IDEs you may need to add
`DJANGO_SETTINGS_MODULE=cms.settings.dev` to their runtime configuration (e.g [PyCharm configuration documentation](https://www.jetbrains.com/help/pycharm/run-debug-configuration.html#createExplicitly)).
```

README.md
```diff
@@ -193,6 +181,21 @@ In order to run it:
     make runserver
     ```
 
+#### Environment Configuration
+
+By default, `make` targets will use the `cms.settings.dev` configuration unless their commands explicitly use a different setting, or `DJANGO_SETTINGS_MODULE`
```
zerolab, line 186:
```suggestion
By default, `make` targets will use the `cms.settings.dev` settings unless their commands explicitly use a different setting (via the `--settings` parameter or `DJANGO_SETTINGS_MODULE` environment variable).
```

Makefile
```diff
@@ -2,6 +2,10 @@ DESIGN_SYSTEM_VERSION=`cat .design-system-version`
 
 .DEFAULT_GOAL := all
 
+.EXPORT_ALL_VARIABLES:
+# Default to development config if DJANGO_SETTINGS_MODULE is not set
+DJANGO_SETTINGS_MODULE ?= cms.settings.dev
```
zerolab, line 7: TIL!

zerolab (commented):
A couple of very minor suggestions, otherwise it works well.

Locally have to use `poetry run ./manage.py test ...` as `make test` does fail with `AttributeError: 'DatabaseWrapper' object has no attribute '_test_serialized_contents'` and while tempting, we have many other tasks to get on with

zerolab (approved):
This is marked improvement. 👍🏼 from me

sanjeevz3009 (approved):
LGTM

README.md
```diff
@@ -193,6 +181,21 @@ In order to run it:
...
+To override settings in the environment, you can use a `.env` file. Note, however, that settings this file may also be picked up in the docker container, so
+you may need to remove or rename the file, or comment out specific variables if you switch to running the app in the container.
+
+> [!NOTE]
+> When running the application locally in a virtual environment via Poetry the `.env` file will not be picked up automatically.
+> For this to work you'll need to install the [poetry-plugin-dotenv](https://github.com/pivoshenko/poetry-plugin-dotenv).
+> However, if you installed Poetry with `brew` rather than `pip` that currently isn't going to
+> work (see the [issue](https://github.com/pivoshenko/poetry-plugin-dotenv/issues/327)) and you'll need to install an older and
+> seemingly no longer maintained [poetry-dotenv-plugin](https://github.com/mpeteuil/poetry-dotenv-plugin).
+
```
kacperpONS, line 198: It should have been done this way from the start 😄 
I remember I started with this approach, but for a reason I don't recall now I shifted to doing it via the `.env` file.

Everything looks good, but we might want to get rid of the bit I've highlighted since we're no longer using the `.env` file and it won't be relevant for most people unless someone decides to use their own `.env` (but again it'd now be easier to modify the settings files..)
Not a huge issue tho as it looks that others are fine with it!

kacperpONS (approved):
It should have been done this way from the start 😄 
I remember I started with this approach, but for reason I don't recall now I shifted to doing it via the `.env` file.

Everything looks good, but we might want to get rid of the bit I've highlighted since we're no longer using the `.env` file. Not a huge issue tho as it looks that others are fine with it!

README.md
zerolab: I think the note is useful. Let's leave it in and revisit should someone new try this and stumble.

PR #79, opened 2025-01-16
CMS-299: Add check to ensure site settings work in external env

RealOrangeOne:

When defining a Wagtail setting, it needs to inherit from our custom base classes. These do 2 things:

- Ensure reads are routed to the read replica rather than the write (`.get_or_create` will always use the write).
- Use default values if an instance doesn't exist in the external env

However, these base classes are easy to miss. This PR adds a check to ensure they're being used for all settings.

### How to review

The fact the checks pass show the current state of the codebase is correct. Since this is based on model definitions, it's hard to unittest this specifically.

### Follow-up Actions

I've intentionally added an implementation for `GenericSetting`, however it's not currently used. That means in future, it should be easy for someone to start using it. However, there are no tests for `GenericSetting`, which is unfortunate.


zerolab (commented):
Tested locally by changing imports in the `core.models.settings`:


```
django.core.management.base.SystemCheckError: SystemCheckError: System check identified some issues:

ERRORS:
core.SocialMediaSettings: Site setting does not extend project base.
	HINT: Ensure site setting extends <class 'cms.core.models.base.BaseSiteSetting'>.
core.SystemMessagesSettings: Site setting does not extend project base.
	HINT: Ensure site setting extends <class 'cms.core.models.base.BaseSiteSetting'>.
```

Any chance for a quick test?

sanjeevz3009 (approved):
LGTM

AdamHawtin (approved):
LGTM

PR #83, opened 2025-01-21
Fix: Only ignore the root/automatically-compiled 'static' dirs

ababic:

The .gitignore file is currently a little bit too overbearing, and doesn't account for when apps that have their own staticfiles. This caused pain most recently on https://github.com/ONSdigital/dis-wagtail/pull/80, where it worked fine locally, but not on anybody Helen's machine. But, as the project grows, it's likely to trip developers up unnecessarily on other threads of work too.

### How to review

Add a `static` folder to an existing app and plonk a random file in it. `git status` should reflect the addition of that file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -185,7 +185,7 @@ copier_scripts/
 /cms/jinja2/layout
 /cms/static_compiled
 
-/node_modules/
+node_modules/
 /static/
 
 # mkdocs
--- a/.gitignore
+++ b/.gitignore
@@ -185,8 +185,8 @@ copier_scripts/
 /cms/jinja2/layout
 /cms/static_compiled
 
-node_modules/
-static/
+/node_modules/
+/static/
 
 # mkdocs
 /site


.gitignore
```diff
@@ -185,8 +185,8 @@ copier_scripts/
 /cms/jinja2/layout
 /cms/static_compiled
 
-node_modules/
-static/
+/node_modules/
```
zerolab, line 188: Any reason not to leave the `node_modules` greedy? That is, would we ever want to commit `node_modules` from anywhere in the tree?

helenb: Agree we would never want to commit `node_modules`.

.gitignore
```diff
@@ -185,8 +185,8 @@ copier_scripts/
...
+/static/
```
helenb, line 189: Ideally we would keep all static files in `static_src` - we can have a separate folder for `admin` files that relate to wagtail admin if needed.

.gitignore
ababic: This is less of an immediate problem, admittedly, but related. Ultimately, I think it's reasonable for an app to have a `node_modules` directory where it needs generate complex JS for Wagtail or something (something that might not be bundled up with the global FE media).

helenb: In that scenario we can add an extra endpoint in webpack that targets a different js file, e.g. `admin.js`.

.gitignore
ababic: Just to clarify: `static_src` is good for FE stuff that needs to be 'packaged up' for end users. And that should rightly never be in `.gitignore`. It's also valid for apps to have static files that are only needed for working inside Wagtail, which shouldn't be bundled in the same way. This change is to allow for that second scenario.

.gitignore
zerolab: @ababic while the scenario you outline is reasonable, I'd rather we moved this further up (ie. the project `package.json`, `static_src` etc). Or even better, publish it to PyPI if allowed

ababic: Okay, happy to revert this one.

ababic: Reverted in b1f0308ff6b14277531ec440e2c86d2e241d106e

zerolab (approved):
LGTM

.gitignore
helenb: Yep, but this can still be done via files that live in the `static_src` directory -example on the bucks project (with apologies to ONS devs who won't be able to follow this link): https://git.torchbox.com/buckinghamshire-council/bc/-/tree/release/bc/static_src/javascript?ref_type=heads

ababic: @helenb Nothing about this PR is really suggesting what the end-goal approach should be (which would be one to iron out in review). Merely that, it's simply the easiest way to solve a problem a lot of the time (e.g. when there hasn't been any FE involvement yet), and ignoring those files by default doesn't make sense.

helenb: Ok sure happy to approve.

helenb (approved):

PR #87, opened 2025-01-27
Bring in some of the styling from the proof of concept

helenb:

Bring in some styling from the [Highcharts poc](https://confluence.ons.gov.uk/display/NWP/Highcharts+proof+of+concept) to the datavis poc in wagtail.

### How to review

To discuss with @ababic 

### Follow-up Actions


cms/datavis/models/charts.py
```diff
@@ -96,7 +85,8 @@ def media(self) -> Media:
 
     def get_context(self, request: Optional["HttpRequest"] = None, **kwargs: Any) -> dict[str, Any]:
         config = self.get_component_config(self.primary_data_source.headers, self.primary_data_source.rows)
-        return super().get_context(request, config=config, **kwargs)
+        theme = self.theme
```
helenb, line 88: Should already be accessible on the chart object passed through to the template

PR #89, opened 2025-01-27
Fix: Document not accessible after a page featuring it is published

ababic:

- Removes redundant "wagtail-storages" from installed apps
- Adds a custom class-based view to replace `wagtail.documents.views.serve.serve` (that cannot be affected by hooks or other configuration)
- Rewrites the custom image serve view to no longer extend `wagtail.images.views.serve.ServeView`, allowing it to share a similar pattern to the document serve view and to provide better control over cache headers (and avoid behaviour changes based on hooks or other configuration)
- Applies appropriate cache headers to responses from both views to suit different scenarios
- Moves the `wagtaildocs_serve` serve path definition out of `private_urls`, so that headers set by the view are respected
- Overrides the document's `url` property to mirror the logic used for renditions

### How to review

The test cases in `mol.private_media.tests.test_serve_views` should cover everything that needs to be covered. Because we're no longer leaning on any built-in views for serving, the test cases have been extended to include handling of invalid ID, signature and filename requests.


cms/private_media/tests/test_serve_views.py
```diff
@@ -0,0 +1,150 @@
...
+            for is_external_env in [True, False]:
+                with self.subTest(is_external_env=is_external_env) and override_settings(
+                    IS_EXTERNAL_ENV=is_external_env
+                ):
+                    response = self.client.get(rendition.serve_url)
+                    self.assertEqual(response.status_code, 200)
+
+    def test_serve_public_image(self):
+        """Test the serve view behaviour for public image renditions."""
+        # For public image renditions, the serve view should redirect to the file path orURL.
```
RealOrangeOne, line 49:
```suggestion
        # For public image renditions, the serve view should redirect to the file path or URL.
```

cms/private_media/tests/test_serve_views.py
```diff
@@ -0,0 +1,150 @@
...
+        cls.private_image_renditions = cls.private_image.create_renditions(Filter("fill-10x10"), Filter("fill-20x20"))
+        cls.public_image_renditions = cls.public_image.create_renditions(Filter("fill-10x10"), Filter("fill-20x20"))
+        cls.superuser = get_user_model().objects.create(username="superuser", is_superuser=True)
+
+    def test_serve_private_image(self):
+        """Test the serve view behaviour for private image renditions."""
+        # If not authenticated, permission checks should fail and a Forbidden response returned
+        for rendition in self.private_image_renditions.values():
+            for is_external_env in [True, False]:
+                with self.subTest(is_external_env=is_external_env) and override_settings(
```
RealOrangeOne, line 31: Issue: Include `rendition` in the `subTest`

cms/private_media/views.py
```diff
@@ -1,34 +1,252 @@
...
 
 if TYPE_CHECKING:
-    from django.http import HttpResponse
+    from django.http import HttpRequest, HttpResponseBase, HttpResponseRedirect
+    from wagtail.documents.models import AbstractDocument
+    from wagtail.images.models import AbstractImage, AbstractRendition
 
-    from cms.private_media.models import AbstractPrivateRendition, PrivateImageMixin
 
+class ImageServeView(View):
```
RealOrangeOne, line 25: There's a lot of new code here copied from Wagtail, which is going to be a burden to maintain as packages upgrade. 

What are the differences between this and `ServeView`? It looks like the old view could handle changing how to serve public / private images by changing functionality in `serve`?

ababic: As outlined in the description, we want to use different cache headers etc in different scenarios, but `ServeView` adds it's 'public' ones for everything, which isn't really suitable for this project.

cms/private_media/tests/test_serve_views.py
ababic: Resolved in [f8b0cdc](https://github.com/ONSdigital/dis-wagtail/pull/89/commits/f8b0cdcd25c61e96a99376bffaf5c88d201ada46)

cms/private_media/tests/test_serve_views.py
ababic: Resolved in [1dcdcb3](https://github.com/ONSdigital/dis-wagtail/pull/89/commits/1dcdcb311e437d3b5043cd407bf131536a25e423)

cms/private_media/views.py
ababic: We're also doing the permission check / rejection here before attempting to find/create the rendition, which I think is slightly nicer.

ababic: Oh, and the Wagtail version also raises a `PermissionDenied` exception when the signature is invalid - which is not the best choice, and could definitely make things difficult to reason about should we face any issues in an environment. Raising `SuspiciousOperation` instead helps to resolve that.

zerolab (commented):
Tested locally with the dev S3 credentials.
This works. I do share Jake's note on the custom serve views, but ultimately that was to be expected given the very bespoke requirements.

@ababic based on the view changes here - what core view changes would you like to see that would enable us to make fewer modifications

ababic:
> @ababic based on the view changes here - what core view changes would you like to see that would enable us to make fewer modifications

- Just breaking `get()` up a little bit into separate methods would help massively. It'd give you a lot more override points.
- Move the cache-header setting from `get()` on to a `get_file_response()` method or similar. 
- Raise SuspiciousOperation instead of PermissionDenied when a signature is invalid

AdamHawtin (approved):
I'm not set up to test this manually myself, but it looks good and the test cases look comprehensive and well written 👍

MebinAbraham (approved):
I share the same concerns as Jake, but if we can make changes to core to reduce the duplication that would be great. Otherwise, LGTM, works as expected. 👍

zerolab (approved):

sanjeevz3009 (commented):
Looks good to me!

sanjeevz3009 (approved):

PR #85, opened 2025-01-22
Main Menu Template Tags

sanjeevz3009:
### What is the context of this PR?

This is a follow-up PR to the https://github.com/ONSdigital/dis-wagtail/pull/41.

This PR introduces template tags for the main menu, by removing front-end logic. By separating concerns, this ensures templates are focused on presentation while business logic remains in Python. This improves readability and testability, simplifying the template structure. Additionally, we will have a more maintainable code by reducing complexity and allowing for performance optimisations when the page is rendered.

### How to review

Test the main menu per norm as instructed in https://github.com/ONSdigital/dis-wagtail/pull/41 description. Everything should work normally when adding a main menu -> highlights -> columns - > sections -> topics.

> :warning: This PR will be merged with main once the https://github.com/ONSdigital/dis-wagtail/pull/41 gets merged to main.

### Follow-up Actions

List any follow-up actions (if applicable), like needed documentation updates or additional testing.

PR #91, opened 2025-01-29
Try out the statistical article hero wtih the hero component from the design system

helenb:

This is an initial test to try out the new grey hero on the statistical article page - relating to https://jira.ons.gov.uk/browse/CMS-307

It is very much draft and not intended to be merged yet - it shows the two hero blocks one above the other on the statistical article page, in order to highlight some of the styling and implementation issues with switching, which are going to be discussed with the design system team.

Pushing this as I would like some initial feedback on the change to move the breadcrumb logic to the back-end, which is necessary in order to use the `breadcrumbs` option in the hero.

Update: breadcrumb filter is now moved to be a global function, and I have followed @sanjeevz3009's approach in https://github.com/ONSdigital/dis-wagtail/pull/85 and created a `navigation_tags` file.

### How to review

You can check this branch out to test locally if desired, but it is not intended to be tested - more here to get feedback on the breadcrumbs approach.

### Follow-up Actions

- Discuss design differences with design team
- Make summary fields consistent as rich text across page types.

### DIFF
--- a/cms/core/templatetags/util_tags.py
+++ b/cms/core/templatetags/util_tags.py
@@ -3,7 +3,6 @@ from typing import TYPE_CHECKING, Optional
 import jinja2
 from django import template
 from django.template.loader import render_to_string
-from django.utils.translation import gettext_lazy as _
 from django_jinja import library
 
 from cms.core.models import SocialMediaSettings
@@ -38,17 +37,3 @@ def social_image(page: "Page", site: "Site") -> Optional["CustomImage"]:
 def include_django(context: jinja2.runtime.Context, template_name: str) -> "SafeString":
     """Allow importing a pre-rendered Django template into jinja2."""
     return render_to_string(template_name, context=dict(context), request=context.get("request", None))
-
-
-# Breadcrumbs
-@register.filter(name="breadcrumbs")
-def breadcrumbs(page: "Page") -> list[dict[str, str]]:
-    """Returns the breadcrumbs as a list of dictionaries for the given page."""
-    breadcrumbs = []
-    for ancestor_page in page.get_ancestors().specific().defer_streamfields():
-        if not ancestor_page.is_root():
-            if ancestor_page.depth <= 2:
-                breadcrumbs.append({"url": "/", "text": _("Home")})
-            elif not getattr(ancestor_page, "exclude_from_breadcrumbs", False):
-                breadcrumbs.append({"url": ancestor_page.get_url(), "text": ancestor_page.title})
-    return breadcrumbs
--- a/cms/core/templatetags/util_tags.py
+++ b/cms/core/templatetags/util_tags.py
@@ -3,6 +3,7 @@ from typing import TYPE_CHECKING, Optional
 import jinja2
 from django import template
 from django.template.loader import render_to_string
+from django.utils.translation import gettext_lazy as _
 from django_jinja import library
 
 from cms.core.models import SocialMediaSettings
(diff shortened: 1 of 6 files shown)


cms/core/templatetags/util_tags.py
```diff
@@ -37,3 +38,17 @@ def social_image(page: "Page", site: "Site") -> Optional["CustomImage"]:
 def include_django(context: jinja2.runtime.Context, template_name: str) -> "SafeString":
     """Allow importing a pre-rendered Django template into jinja2."""
     return render_to_string(template_name, context=dict(context), request=context.get("request", None))
+
+
+# Breadcrumbs
+@register.filter(name="breadcrumbs")
+def breadcrumbs(page: "Page") -> list[dict[str, str]]:
```
zerolab, line 45: suggestion

```suggestion
@jinja2.pass_context
def breadcrumbs(context: jinja2.runtime.Context, page: "Page") -> list[dict[str, str]]:
```

a la https://github.com/ONSdigital/dis-wagtail/pull/85/files#diff-bfd42dbff73bbdea18630c438176ac08db2c376e9a8c3205a430e5f15caa3264R75-R76

then you can do `.get_url(request=context.get("request"))`.


`@register.filter(name="breadcrumbs")` is for the Django template library, which we only use for the Wagtail admin

zerolab: Could you also update the breadcrumbs template to remove https://github.com/ONSdigital/dis-wagtail/blob/main/cms/jinja2/templates/components/navigation/breadcrumbs.html#L4-L14 and change https://github.com/ONSdigital/dis-wagtail/blob/main/cms/jinja2/templates/components/navigation/breadcrumbs.html#L19 to `page|breadcrumbs` ?

cms/jinja2/templates/pages/statistical_article_page.html
```diff
@@ -1,9 +1,103 @@
...
+        {% set census_html = "" %}
+    {% endif %}
+    {# Currently missing links as the onsDescriptionList does not allow markup #}
+    {% if page.is_latest %}
+        {% set releases_text = _("View previous releases") %}
+    {% else %}
+        {% set releases_text = _("View latest release") %}
+    {% endif %}
+    {% set next_release_date_text = page.next_release_date|date("DATE_FORMAT") or _("To be announced") %}
+    {% with breadcrumbs=page|breadcrumbs %}
```
zerolab, line 39: I believe you can pass `page|breadcrumbs` directly to the macro.

cms/core/templatetags/util_tags.py
helenb: I think it would make sense to use the `navigation` app for the breadcrumbs like the ones in @sanjeevz3009's merge request so will update.

zerolab: > it seems to be being used for the social_image and social_text filters currently which are used on the front-end not in the admin.

that was part of kit and it makes sense to clean them

helenb: Happy to leave them if that makes sense - did you mean 'keep them' not 'clean them'?

helenb: I will leave the breadcrumbs template for now and see what becomes of this pull request - it may be that we can get rid of it entirely once we're using the new heroes more consistently.

cms/jinja2/templates/pages/statistical_article_page.html
helenb: Sorted

zerolab (changes requested):
A couple of suggestions which should hopefully improve the template readability

zerolab:
Superseded by #129

PR #96, opened 2025-02-05
Topic page enhancements

zerolab:

This PR adds:
- the ability to feature an article series and always display the latest article in the series)
- the ability to highlight articles and/or methodologies (CMS-294)
  - the chosen items take priority and will display first in the chosen order. the rest up to 3 is backfilled from other relevant statistical articles (or methodologies, for its section)
- an explore more StreamField block with internal and external links
- wires everything in the template

To-Do:
- [ ] limit the featured series chooser to relevant series (created under the given topic)
- [ ] limit the highlights chooser to relevant pages (created under the topic)
- [ ] tests

### How to review

Check the branch out locally. Run migrations. Create or edit a topic page and sequantially fill in the various fields (features, highlighted articles, highlighted methodologies, explore more). Also create several series / articles in the series, and methodologies, under the given topic page

### Follow-up Actions

* enhance with related articles/methodologies by taxonomy, when the feature is merged.


AdamHawtin (commented):
Functionally all seems good, I think the styling/spacing might need a little touching up, the related articles section looks cramped under the featured article:

Screenshot 2025-02-10 at 13 27 36

Happy for there to be a separate PR pass on the styling if that's the preferred approach.

cms/topics/models.py
```diff
@@ -1,24 +1,229 @@
...
+            "featured_series",
+            heading=_("Featured"),
+            widget=FeaturedSeriesPageChooserWidget(linked_fields={"topic_page_id": "#id_topic_page_id"}),
+        ),
+        InlinePanel(
+            "related_articles",
+            heading=_("Highlighted articles"),
+            help_text=_(
+                f"Choose up to {MAX_ITEMS_PER_SECTION} articles to highlight. "
+                "The 'Related articles' section will be topped up automatically."
```
nehakerung, line 99: Query: When should we use this instead of the regular help text? Or is it only appropriate in this specific case?
I assumed it was more of an info box as used by bundles in the admin page. 🤔 
Screenshot 2025-02-10 at 11 14 39

nehakerung (commented):
LGTM - looks good, just have a small question.

zerolab:
@AdamHawtin can you run `npm run build`, that should sort out the featured block. (I missed that from the instructions, now updated in the ticket description)

cms/topics/models.py
zerolab: @nehakerung this is the help block for `InlinePanel` which is a collection of fields, rather than just one field

AdamHawtin (approved):
Looks good to me, the functionality works as expected.

I have some questions about the behaviour with highlighting/featuring draft pages. The way it's handled makes sense technically, with editors able to select to feature/highlight draft pages, but those draft pages then not appearing on the published topic page until they are published, but this isn't communicated in the editor, and I wonder if it will be understood by users? Perhaps draft pages need to be highlighted in the editor with a warning that they will not show until they are published.

sanjeevz3009:
> Looks good to me, the functionality works as expected.
> 
> I have some questions about the behaviour with highlighting/featuring draft pages. The way it's handled makes sense technically, with editors able to select to feature/highlight draft pages, but those draft pages then not appearing on the published topic page until they are published, but this isn't communicated in the editor. I wonder if it will be understood by users? Perhaps draft pages need to be highlighted in the editor with a warning that they will not show until they are published.

@AdamHawtin Echoing what Adam said. This is the same behaviour on the Generic Index page.
@kacperpONS

zerolab:
@AdamHawtin @sanjeevz3009 I hear you both. We have a backlog item that looks specifically at previews in a holistic manner (CMS-356) and one that looks at release preview in particular (CMS-350)

kacperpONS:
LGTM, two quick questions:
- in your walkthrough it seems that the article titles are displayed without the article series title prefix, but when running this branch it shows both. Just wanted to clarify whether that has changed since you recorded the screencast? 
- why is the `label` field defined as a method? wouldn't it be sufficient if it was defined as a [class variable](https://stackoverflow.com/questions/68645/class-static-variables-and-methods#:~:text=Variables%20declared%20inside%20the%20class%20definition%2C%20but%20not%20inside%20a%20method%20are%20class%20or%20static%20variables%3A)?

zerolab:
@kacperpONS 
1. I noticed the same while doing the screencast, so followed it up in c45deb1
2. Old habits. class var works just as well, so updated in 586f1c2. Cheers!

cms/topics/viewsets.py
```diff
@@ -0,0 +1,132 @@
...
+
+from cms.articles.models import ArticleSeriesPage, StatisticalArticlePage
+from cms.methodology.models import MethodologyPage
+
+if TYPE_CHECKING:
+    from wagtail.models import Page
+    from wagtail.query import PageQuerySet
+
+
+class FeaturedSeriesPageMixin:
```
sanjeevz3009, line 18: Nit: Shouldn't this be in the `mixins.py` file. We also have another mixin below called `HighlightedChildPageMixin`.

We have other `mixins` files in the codebase, so thinking it might make sense to extract out the mixins into one mixin file.

Nothing pressing or major!

sanjeevz3009 (commented):
Everything looks good to me! Works as expected! One minor comment =)

sanjeevz3009 (approved):

cms/topics/viewsets.py
zerolab: this is more of contextul decision. You'd move them if they are more general purpose (i.e. to be used in other apps/modules) or if there is a bigger number of them

Leaving as is (but improved their name a tad in 708da0f)

sanjeevz3009: Thanks, Dan, for the class name improvements!

sanjeevz3009 (approved):

cms/topics/models.py
```diff
@@ -1,23 +1,225 @@
...
+        article: StatisticalArticlePage | None = StatisticalArticlePage.objects.none()
+        if self.featured_series:
+            article = (
+                StatisticalArticlePage.objects.child_of(self.featured_series)
+                .live()
+                .public()
+                .order_by("-release_date")
+                .first()
+            )
+        return article
```
RealOrangeOne, line 138: Nit: The type of `article` isn't quite right. The easiest fix is to just remove it.

```suggestion
        if self.featured_series:
            return (
                StatisticalArticlePage.objects.child_of(self.featured_series)
                .live()
                .public()
                .order_by("-release_date")
                .first()
            )
        return None
```

cms/topics/models.py
```diff
@@ -1,23 +1,225 @@
...
+
+    @cached_property
+    def processed_methodologies(self) -> list[MethodologyPage]:
+        """Returns the latest methodologies relevant for this topic.
+        TODO: extend when Taxonomy is in.
+        """
+        # check if any methodologies were highlighted. if so, fetch in the order they were added.
+        highlighted_page_pks = tuple(
+            page_id for page_id in self.related_methodologies.values_list("page_id", flat=True)
+        )
```
RealOrangeOne, line 190:
```suggestion
        highlighted_page_pks = list(
            self.related_methodologies.values_list("page_id", flat=True)
        )
```

cms/topics/tests/test_viewsets.py
```diff
@@ -0,0 +1,217 @@
...
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, "wagtailadmin/generic/chooser/results.html")
+
+        self.assertListEqual(
+            response.context["items"].object_list,
+            [self.second_methodology],
+        )
+
+
+class ViewSetConfigurationTest(TestCase):
```
RealOrangeOne, line 189: Question: Why are these in a separate test case rather than being included in the ones above? It feels like they'd be better grouped with the tests for how the view works itself.

cms/topics/models.py
zerolab: something, something appease mypy. tweaked closer to what you suggested 069f8cd

cms/topics/models.py
zerolab: 👍🏼 069f8cd

cms/topics/tests/test_viewsets.py
zerolab: initially I had a view that this would be more extensive, but in the end it was not needed, so it kind of stayed. merged in the test cases above (069f8cd), thanks

RealOrangeOne (approved):

sanjeevz3009 (approved):
LGTM!

PR #98, opened 2025-02-11
Allow rendering a bundle without a "created by" user

RealOrangeOne:

If the user who created a bundle is deleted, `created_by` is `None`, and the admin site can't be rendered.

### How to review

Delete the user who created a bundle.

### Follow-up Actions

This will need cherry-picking into `feature/datavis`, too.


sanjeevz3009 (approved):

zerolab:
@RealOrangeOne approved so the DataViz instance can be unblocked. Would be good to have a test for it (happy to cover it)

PR #101, opened 2025-02-12
Fix broken dev Redis default config

AdamHawtin:
#76 Introduced an issue with the developer config where I incorrectly assumed `REDIS_URL` set in `dev.py` would properly populate the Redis configuration.

### How to review

Check out the this branch
Run `make compose-dev-up`
Rename or remove any local `.env` or `local.py` config files
Try starting the wagtail app from the command line or IDE
The app should successfully start, not hit an error around database cache tables.


kacperpONS (approved):
Works for me! thanks for the quick fix

sanjeevz3009 (commented):
Reviewed! Everything works as expected! Thanks @AdamHawtin !

PR #102, opened 2025-02-12
Methodology page UI tests

kacperpONS:

This PR adds functional tests for the methodology page.

### How to review

Run the below command to run only the test for the methodology page
```
poetry run behave functional_tests/ -i methodology_page.feature
```
or 
```
make functional-tests
``` 
to run the whole suite to make sure that there's no conflicts/ unintended side effects.

### Follow-up Actions


functional_tests/steps/article_page.py
```diff
@@ -0,0 +1,10 @@
+from behave import given
+from behave.runner import Context
+
+from cms.articles.tests.factories import ArticleSeriesFactory, StatisticalArticlePageFactory
+
+
+@given("a statistical article page has been published under the existing theme page")
+def create_article_page(context: Context):
+    context.article_series = ArticleSeriesFactory(parent=context.topic_page)
+    context.statistical_article = StatisticalArticlePageFactory(parent=context.article_series)
```
zerolab, line 10: @kacperpONS worth noting I have a similar one in #96 - https://github.com/ONSdigital/dis-wagtail/pull/96/files#diff-326f13bd4cb7217b4cf1b63f44d0e10ec184fb2310ddd8efb83af4ab61198e95R1-R15

kacperpONS: I'll change the wording wording to avoid conflicts later on. 

One thing to note, you don't need to pass the title to the article page factory as it gets automatically populated by Faker. It will then be referenced with `context.article.title` so having the hard-coded value there doesn't add much value.

kacperpONS: And a question - it feels like the two sequences of steps:
1. Create theme page -> create topic page
2. Create article series -> create an article

will be used *almost always* together. 

How about we use the specific steps only when it's necessary and for scenarios where they're used together define steps that will perform the whole sequence to avoid duplication?

zerolab: > One thing to note, you don't need to pass the title to the article page factory as it gets automatically populated by Faker. It will then be referenced with context.article.title so having the hard-coded value there doesn't add much value.

I am well aware of that. In some cases I do want these to be deterministic

zerolab: > How about we use the specific steps only when it's necessary and for scenarios where they're used together define steps that will perform the whole sequence to avoid duplication?

100% on this with you.

functional_tests/features/topic_page.feature
```diff
@@ -2,20 +2,19 @@ Feature: CMS users can draft, edit, and publish topic pages
...
+        And the user visits the topic page
         Then the topic page with the example content is displayed
         And the user can see the topic page featured article
 
     Scenario: The feature series on a topic page displays the latest article
-        Given a topic page exists
+        Given a topic page exists under a theme page
         And the user has created a statistical article in a series
         And the user has featured the series
-        When the user creates a new statistical article in the series
```
zerolab, line 19: @kacperpONS this condition is valid. The scenario is:
"If there's a featured series that has an article and the user publishes another article in the series, then that last article is features on the topic page"

kacperpONS: Yup, my bad.
It seems the test step "the user has created a statistical article in a series" doesn't actually create the first article so the test doesn't fully test this functionality. 
Am I misunderstanding something or is that the case? I'm happy to add the extra line.

zerolab: hmm,  it should

```python
@given("the user has created a statistical article in a series")
@given("a statistical article page has been published under the topic page")
def create_article_in_a_series(context: Context):
    context.article_series = ArticleSeriesPageFactory(parent=context.topic_page)
    context.article = StatisticalArticlePageFactory(title="January 2025", parent=context.article_series)
```

kacperpONS: Have a look at: [45d784b](https://github.com/ONSdigital/dis-wagtail/pull/102/commits/45d784bd05183344e6e2a5433cff89b90796ca4a)
It's also going to need the date to be set explicitly so that it's after the "old" article.

functional_tests/steps/methodology_page.py
```diff
@@ -0,0 +1,120 @@
...
+        context.page.locator("#panel-child-content-child-title-errors .error-message").get_by_text(
+            "This field is required"
+        )
+    ).to_be_visible()
+
+    expect(
+        context.page.locator("#panel-child-content-child-summary-errors .error-message").get_by_text(
+            "This field is required"
+        )
+    )
```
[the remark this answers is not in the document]
kacperpONS: As a follow up to this - the tests seems to be failing after I've added the assertion. 
It's passing when run locally on its own however, I'm going to investigate this.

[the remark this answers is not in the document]
kacperpONS: Fixed, see [6fa3b21](https://github.com/ONSdigital/dis-wagtail/pull/102/commits/6fa3b217019fd744fd47cfc512f958b677adfc54) and [0c8bdbc](https://github.com/ONSdigital/dis-wagtail/pull/102/commits/0c8bdbc74dcd6abeabc9001ea5c17a3ecbc3c7d8).

nehakerung (approved):
LGTM - the tests are working and look good 👍

functional_tests/steps/methodology_page.py
```diff
@@ -0,0 +1,132 @@
...
+
+    expect(context.page.locator(".help-block.help-critical").get_by_text("This field is required")).to_be_visible()
+
+
+@then("the preview is visible with the populated data")
+def preview_is_visible(context: Context):
+    context.page.get_by_role("button", name="Toggle preview").click()
+
+    # add a small delay to allow the preview to render
+    context.page.wait_for_timeout(500)
```
AdamHawtin, line 116: Can we use an `expect` to wait for an element in the preview, rather than a hardcoded time?

functional_tests/features/methodology_page.feature
```diff
@@ -0,0 +1,65 @@
...
+        And the user clicks "Publish page"
+        And the user clicks "View Live" on the publish confirmation banner
+        Then the Contact Details are visible on the page
+
+    Scenario: The mandatory fields raise validation errors when left empty on the Methodology page.
+        Given a topic page exists under a theme page
+        And a CMS user logs into the admin site
+        When the user creates a methodology page as a child of the existing topic page
+        And the user clicks the "Save Draft" button and waits for the page to reload
+        Then the mandatory fields raise a validation error
```
AdamHawtin, line 39: Perhaps worth making this step wording specific to methodology pages, e.g.
```cucumber
        Then the methodology page mandatory fields raise validation errors
```
As we'd probably have to do this down the line for future pages with different mandatory fields.

functional_tests/steps/methodology_page.py
kacperpONS: Turns out I've prematurely added the delay in this case as the test is passing fine without it - probably the Playwright built-in timeouts are kicking in and waiting for when the preview is visible.

functional_tests/features/methodology_page.feature
kacperpONS: Done in [e9cb0a5](https://github.com/ONSdigital/dis-wagtail/pull/102/commits/e9cb0a594d0d3f78ee306940ee230caf746cc565)

functional_tests/steps/methodology_page.py
```diff
@@ -0,0 +1,129 @@
+from behave import given, then, when  # pylint: disable=no-name-in-module
+from behave.runner import Context
+from playwright.sync_api import expect
+
+
+@given("the user creates a methodology page as a child of the existing topic page")
+@when("the user creates a methodology page as a child of the existing topic page")
```
AdamHawtin, line 7:
```suggestion
@step("the user creates a methodology page as a child of the existing topic page")
```
Lets you match the step wording on any prefix

AdamHawtin: Resolving so I'm not blocking while away

kacperpONS: [9755549](https://github.com/ONSdigital/dis-wagtail/pull/102/commits/9755549f29603593566656eadbce80d9a3e04ae9)

sanjeevz3009 (approved):
LGTM!

No comments within the scope of this work. @kacperpONS and I had discussions on making some things done in this PR reusable and generic as part of the [POM](https://jira.ons.gov.uk/browse/CMS-317) ticket we have.

zerolab (approved):
Let's get this in.

PR #103, opened 2025-02-14
Improvements to data labels

helenb:

- Previously our custom code to move data labels inside the bars of a bar chart, if the bar chart was wide enough, was only for the situation where the bar was 50px or wider. This does a more accurate test to check the actual width of the data label, and compare it to the width of the bar.
- Also declares functions more consistently with fat arrows

### How to review

- Create a bar chart visualisation with varied bar widths.
- Tick 'show value labels' in the styling options
- Observe if the data labels are inside or outside the bars

### Follow-up Actions

### DIFF
--- a/cms/static_src/javascript/components/highcharts-base-chart.js
+++ b/cms/static_src/javascript/components/highcharts-base-chart.js
@@ -26,12 +26,13 @@ class HighchartsBaseChart {
       );
     }
 
-    // Hide data labels for clustered bar charts with more than 2 series
-    if (
-      this.chartType === 'bar' &&
-      this.useStackedLayout === false &&
-      this.apiConfig.series.length > 2
-    ) {
+    // Hide data labels for clustered bar charts with more than 2 series, and also for stacked bar charts
+    const hideDataLabels =
+      (this.chartType === 'bar' &&
+        this.useStackedLayout === false &&
+        this.apiConfig.series.length > 2) ||
+      this.useStackedLayout === true;
+    if (hideDataLabels) {
       this.apiConfig.series.forEach((series) => {
         /* eslint-disable no-param-reassign */
         series.dataLabels = {
@@ -58,7 +59,9 @@ class HighchartsBaseChart {
         if (this.useStackedLayout === false) {
           this.updateBarChartHeight(event);
         }
-        this.postLoadDataLabels(event);
+        if (!hideDataLabels) {
+          this.postLoadDataLabels(event);
+        }
       }
     };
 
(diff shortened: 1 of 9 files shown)


cms/static_src/javascript/components/highcharts-base-chart.js
```diff
@@ -58,7 +59,9 @@ class HighchartsBaseChart {
         if (this.useStackedLayout === false) {
           this.updateBarChartHeight(event);
         }
-        this.postLoadDataLabels(event);
+        if (!hideDataLabels) {
+          this.postLoadDataLabels(event);
+        }
```
nimasmi, line 64: **Question**: Previously were we calling `postLoadDataLabels` for any bar, even if we'd set `series.dataLabels.enabled` to false?

cms/static_src/javascript/components/highcharts-base-chart.js
```diff
@@ -131,34 +134,26 @@ class HighchartsBaseChart {
...
-      if (point.shapeArgs.height > 50) {
-        point.update(options, false);
-      }
+    currentChart.series.forEach((series) => {
+      const points = series.data;
+      points.forEach((point) => {
+        // Get the actual width of the data label
+        const labelWidth = point.dataLabel && point.dataLabel.absoluteBox.width;
+        // Move the data labels inside the bar if the bar is wider than the label
+        if (point.shapeArgs.height > labelWidth) {
```
nimasmi, line 149: **Praise**: Gotcha. Much better. Good change.

nimasmi (approved):
Thanks Helen. The code looks good. I haven't time to user-test this, sorry.

cms/static_src/javascript/components/highcharts-base-chart.js
```diff
@@ -26,12 +26,13 @@ class HighchartsBaseChart {
...
-      this.chartType === 'bar' &&
-      this.useStackedLayout === false &&
-      this.apiConfig.series.length > 2
-    ) {
+    // Hide data labels for clustered bar charts with more than 2 series, and also for stacked bar charts
+    const hideDataLabels =
+      (this.chartType === 'bar' &&
+        this.useStackedLayout === false &&
+        this.apiConfig.series.length > 2) ||
+      this.useStackedLayout === true;
```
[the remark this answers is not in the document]
helenb: Could be a simple bar without being stacked or clustered, or could be a clustered bar chart with 2 series - in either of those cases `hideDataLabels` would be falst.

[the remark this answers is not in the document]
helenb: For stacked layouts we always want to hide data labels.

cms/static_src/javascript/components/highcharts-base-chart.js
helenb: Previously we hid them for clustered bar charts with more than 2 series - this updates to hide them for stacked bar charts too.

PR #105, opened 2025-02-18
Charts: pass config and annotations to the macro as json

helenb:

After a discussion with the design system team, they don't want to encode json themselves in the macro - they'd rather receive the config directly as json.

This updates both the annotations and the config to be encoded as json before they are passed to the template.
Instead of using a script tag, the macro passes a json encoded string as a data attribute.

### How to review

Test that creating charts in an information page all still works as expected.

### Follow-up Actions

Update the macro parameters documentation in the charts spec
Potentially remove the custom version of `json_script` from the repo, although TBC as this could still be a useful addition.

### DIFF
--- a/cms/datavis/models/charts.py
+++ b/cms/datavis/models/charts.py
@@ -1,8 +1,6 @@
-import json
 from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any, ClassVar, Literal, Optional
 
-from django.core.serializers.json import DjangoJSONEncoder
 from django.db import models
 from django.forms import Media
 from django.utils.functional import cached_property, classproperty
@@ -99,12 +97,8 @@ class Chart(Visualisation):
 
     def get_context(self, request: Optional["HttpRequest"] = None, **kwargs: Any) -> dict[str, Any]:
         config = self.get_component_config(self.primary_data_source.headers, self.primary_data_source.rows)
-        encoded_config = json.dumps(config, cls=DjangoJSONEncoder)
         annotations_values = self.get_annotations_values()
-        encoded_annotations_values = json.dumps(annotations_values, cls=DjangoJSONEncoder)
-        return super().get_context(
-            request, config=encoded_config, annotations_values=encoded_annotations_values, **kwargs
-        )
+        return super().get_context(request, config=config, annotations_values=annotations_values, **kwargs)
 
     general_panels: ClassVar[Sequence["Panel"]] = [
         FieldPanel("name"),
--- a/cms/datavis/models/charts.py
+++ b/cms/datavis/models/charts.py
@@ -1,6 +1,8 @@
+import json
 from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any, ClassVar, Literal, Optional
 
+from django.core.serializers.json import DjangoJSONEncoder
 from django.db import models
 from django.forms import Media
 from django.utils.functional import cached_property, classproperty
@@ -97,8 +99,12 @@ class Chart(Visualisation):
 
     def get_context(self, request: Optional["HttpRequest"] = None, **kwargs: Any) -> dict[str, Any]:
         config = self.get_component_config(self.primary_data_source.headers, self.primary_data_source.rows)
+        encoded_config = json.dumps(config, cls=DjangoJSONEncoder)
         annotations_values = self.get_annotations_values()
-        return super().get_context(request, config=config, annotations_values=annotations_values, **kwargs)
+        encoded_annotations_values = json.dumps(annotations_values, cls=DjangoJSONEncoder)
+        return super().get_context(
+            request, config=encoded_config, annotations_values=encoded_annotations_values, **kwargs
+        )
 
     general_panels: ClassVar[Sequence["Panel"]] = [
         FieldPanel("name"),
--- a/cms/jinja2/templates/datavis/_macro.njk
+++ b/cms/jinja2/templates/datavis/_macro.njk
@@ -34,13 +34,13 @@
 
         {# script tags to pass json data to the javascript component #}
         <script type="application/json" data-highcharts-config="{{ params.uuid }}">
-            {# Note that in the design system this may need to be a custom filter #}
-            {{ params.config|safe }}
+            {# Note that in the design system this will need to be a custom filter #}
+            {{ params.config|tojson }}
         </script>
         {% if params.annotations_values %}
             <script type="application/json" data-highcharts-annotations-values="{{ params.uuid }}">
-                {# Note that in the design system this may need to be a custom filter #}
-                {{ params.annotations_values|safe }}
+                {# Note that in the design system will may need to be a custom filter #}
+                {{ params.annotations_values|tojson }}
             </script>
         {% endif %}
     </div>
--- a/cms/jinja2/templates/datavis/_macro.njk
+++ b/cms/jinja2/templates/datavis/_macro.njk
@@ -1,23 +1,22 @@
 {% macro onsChart(params) %}
-    <div data-highcharts-base-chart data-highcharts-type="{{ params.chartType }}" data-highcharts-theme="{{ params.theme }}" data-highcharts-title="{{ params.title }}" {% if params.use_stacked_layout %}data-highcharts-use-stacked-layout{% endif %}>
-        <figure id="{{ params.domId }}">
+    <div
+        data-highcharts-base-chart
+        data-highcharts-type="{{ params.chartType }}"
+        data-highcharts-theme="{{ params.theme }}"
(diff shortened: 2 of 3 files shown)


cms/jinja2/templates/datavis/_macro.njk
```diff
@@ -1,24 +1,23 @@
 {% macro onsChart(params) %}
-    <div data-highcharts-base-chart data-highcharts-type="{{ params.chartType }}" data-highcharts-theme="{{ params.theme }}" data-highcharts-title="{{ params.title }}" {% if params.use_stacked_layout %}data-highcharts-use-stacked-layout{% endif %}>
-        <figure id="{{ params.domId }}">
+    <div
+        data-highcharts-base-chart
+        data-highcharts-type="{{ params.chartType }}"
+        data-highcharts-theme="{{ params.theme }}"
+        data-highcharts-config="{{ params.config }}"
+        data-highcharts-title="{{ params.title }}"
+        {% if params.annotations_values %}data-highcharts-annotations-values="{{ params.annotations_values }}"{% endif %}
```
RealOrangeOne, line 8: Issue: I'm afraid passing plain data like this isn't secure, as a literal `<script>` tag in the content will still be interpreted.

Is there a way we can keep using the `json_script` setup as is there currently for the large bits of data like that? It's a much better way of passing large blobs of JSON to the client.

helenb: Now using a `script` tag instead - but wondering if there is a better way than using the `|safe` filter to escape the json, as I suspect this is a django filter not available in the design system templates (which is what this file is representing).

cms/datavis/models/charts.py
```diff
@@ -97,8 +99,12 @@ def media(self) -> Media:
 
     def get_context(self, request: Optional["HttpRequest"] = None, **kwargs: Any) -> dict[str, Any]:
         config = self.get_component_config(self.primary_data_source.headers, self.primary_data_source.rows)
+        encoded_config = json.dumps(config, cls=DjangoJSONEncoder)
```
RealOrangeOne, line 102: I think that, because you (will be) using `json_script`, you don't need to encode this manually. As it'll likely end up with extra quotes

cms/jinja2/templates/datavis/_macro.njk
```diff
@@ -33,5 +31,15 @@
                 {% endif %}
             </ol>
         {% endif %}
+
+        <!-- script tags to pass json data to the javascript component -->
+        <script type="application/json" data-highcharts-config--{{ params.uuid }}>
```
RealOrangeOne, line 36: Suggestion: Rather than customizing the data-attribute itself, perhaps worth using a value? `data-highcharts="{{ params.uuid }}"`.

cms/jinja2/templates/datavis/_macro.njk
```diff
@@ -33,5 +31,15 @@
                 {% endif %}
             </ol>
         {% endif %}
+
+        <!-- script tags to pass json data to the javascript component -->
```
RealOrangeOne, line 35: Nit: Whilst I like the comment, I'd suggest it be a jinja2 comment rather than an HTML comment, to keep it out the final page

```suggestion
        {# script tags to pass json data to the javascript component #}
```

cms/jinja2/templates/datavis/_macro.njk
RealOrangeOne: Following some discussions, it's best to use the `|tojson` filter inside a `script` tag. That avoids the need to use `|safe`, and avoids the escaping issues around dumping content into a JS literal.

helenb: `tojson` results in an error when I switch, and the charts don't render. I wonder if this is because we are already converting to json in the backend, and then perhaps `tojson` is double encoding it?

This is the contents of the script tag with `tojson`
```
"{\"chart\": {\"type\": \"bar\"}, \"legend\": {\"enabled\": true}, \"xAxis\": {\"type\": \"linear\", \"title\": {\"enabled\": true, \"text\": \"Y axis label\"}, \"reversed\": false, \"categories\": [\"One\", \"Two\", \"Three\", \"Four\", \"Five\", \"Six\", \"Seven\", \"Eight\", \"Nine\", \"Ten\"]}, \"yAxis\": {\"reversed\": false, \"title\": {\"enabled\": true, \"text\": \"x axis label\"}, \"tickInterval\": 4.0}, \"series\": [{\"name\": \"Values\", \"data\": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10], \"animation\": false, \"dataLabels\": {\"enabled\": true}}]}"
```

This is the content of the script tag with `safe`:
```
{"chart": {"type": "bar"}, "legend": {"enabled": true}, "xAxis": {"type": "linear", "title": {"enabled": true, "text": "Y axis label"}, "reversed": false, "categories": ["One", "Two", "Three", "Four", "Five", "Six", "Seven", "Eight", "Nine", "Ten"]}, "yAxis": {"reversed": false, "title": {"enabled": true, "text": "x axis label"}, "tickInterval": 4.0}, "series": [{"name": "Values", "data": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10], "animation": false, "dataLabels": {"enabled": true}}]}
```

Or is there another step needed in the JavaScript with `tojson`? I am using `JSON.parse()`.

cms/datavis/models/charts.py
helenb: The design system team explicitly wanted to receive the data as json rather than as a python object. But let me know how you would see it working.

cms/jinja2/templates/datavis/_macro.njk
helenb: good idea - now done.

cms/jinja2/templates/datavis/_macro.njk
helenb: agreed! Updated.

cms/jinja2/templates/datavis/_macro.njk
helenb: For comparison this is using my copy of django's `to_json` which was being parsed fine by the js:

```"{\"chart\": {\"type\": \"bar\"}, \"legend\": {\"enabled\": true}, \"xAxis\": {\"type\": \"linear\", \"title\": {\"enabled\": true, \"text\": \"Y axis label\"}, \"reversed\": false, \"categories\": [\"One\", \"Two\", \"Three\", \"Four\", \"Five\", \"Six\", \"Seven\", \"Eight\", \"Nine\", \"Ten\"]}, \"yAxis\": {\"reversed\": false, \"title\": {\"enabled\": true, \"text\": \"x axis label\"}, \"tickInterval\": 4.0}, \"series\": [{\"name\": \"Values\", \"data\": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10], \"animation\": false, \"dataLabels\": {\"enabled\": true}}]}"```

Comment (helenb) [reply cms/jinja2/templates/datavis/_macro.njk]: The `to_json` from django and the `tojson` from jinja look identical making me think this is a bug in my JS, so I will investigate further...

File: cms/datavis/models/charts.py (hunk 2 above)
Comment (RealOrangeOne) [reply cms/datavis/models/charts.py:102]: Do they want the macro argument to be a string, or just what's put on the page? If it's the latter, it's still fine to remove this.

File: cms/jinja2/templates/datavis/_macro.njk (hunk 1 above)
Comment (RealOrangeOne) [reply cms/jinja2/templates/datavis/_macro.njk]: I suggest using the `tojson` from jinja, as that's the one which definitely sanitises things properly, and is what we'll replicate in the DS.

Yes, it's likely double encoding it, so you'll want to remove the encode from the backend

File: cms/datavis/models/charts.py (hunk 2 above)
Comment (helenb) [reply cms/datavis/models/charts.py]: Yep let's make it a string and then do the json encoding with the custom filter as discussed.

File: cms/static_src/javascript/components/highcharts-base-chart.js [hunk 5]
```diff
@@ -48,8 +51,9 @@ class HighchartsBaseChart {
     // Will only run once per page load
     this.setCommonChartOptions();
 
-    // Configure any annotations that have been specified
+    // Configure any annotations that have been specified (will be an empty array if no annotations are specified)
     if (this.annotationsValues) {
+      console.log('annotations found');
```
RealOrangeOne, line 56: Nit: Debug

RealOrangeOne (approved):

cms/static_src/javascript/components/highcharts-base-chart.js
helenb: oops! will remove and then merge.

PR #106, opened 2025-02-20
[UR tweaks] Show the rich text toolbar and the minimap by default

kacperpONS:

Following User Research we've identified the need to change the default behaviour of the Wagtail editor - we want the rich text toolbar to be pinned at the top of the input field and the editor minimap to be shown.

Functional tests have been added to check for the default behaviour and storing user preferences.

Walkthrough: https://www.loom.com/share/52fb408fa2d24b3290a2977b3ff919f2

### How to review
1. Run Wagtail and open it in an incognito window (alternatively clear local storage in the dev browser console).
2. Go to page editor and check that: 
    - the rich text toolbar is pinned at the top of the input field.
    - the page editor minimap is shown on the right
3. Then unpin the toolbar and hide the editor then refresh the page to check that the preference is saved.

Also, run the functional tests.

### Link to Jira
https://jira.ons.gov.uk/browse/CMS-353


cms/core/wagtail_hooks.py
```diff
@@ -20,4 +22,13 @@ def register_icons(icons: list[str]) -> list[str]:
     ]
 
 
+@hooks.register("insert_editor_js")
+def editor_js():
+    js_files = [
```
zerolab, line 27: note: let's merge this two into a `editor-tweaks.js` (or similarly named) file. Fewer things to load or look out for

Then this hook implementation becomes:

```python
@hooks.register("insert_editor_js")
def editor_js():
    return format_html('<script src="{}"></script>', static("js/editor-tweaks.js"))
```

functional_tests/features/information_page.feature
```diff
@@ -10,3 +10,14 @@ Feature: A general use information page
...
+        Given a CMS user logs into the admin site
+        When the user navigates to the pages menu
+        And the user clicks add child page and chooses information page type
+        Then the rich text toolbar is displayed
+
+    Scenario: Minimap is shown by default
+        Given a CMS user logs into the admin site
+        When the user navigates to the pages menu
+        And the user clicks add child page and chooses information page type
+        Then the minimap is displayed
```
MebinAbraham, line 23: These will test the default, but you will need to add a test to ensure the final acceptance criteria passes which is that the user's preference is used if set.
> And if I unpin the toolbar or hide the minimal panel, then that choice is remembered in my browser

cms/jinja2/assets/js/wagtail-editor-customisations.js
```diff
@@ -0,0 +1,15 @@
...
+  
+
+  function setInitialRichtextToolbarStickyByDefault(){
+    if (localStorage.getItem("wagtail:draftail-toolbar") === null){
+      localStorage.setItem('wagtail:draftail-toolbar', 'sticky');
+    }
+  };
+  
+setInitialMinimapExpandedByDefault();
+setInitialRichtextToolbarStickyByDefault();
```
MebinAbraham, line 15: Looks like this file is not being formatted due to this exclusion in https://github.com/ONSdigital/dis-wagtail/blob/main/.prettierignore#L12

I suggest replacing `jinja2` with `.min.js` and `.min.css` to prevent the formatting of minified js/css. Alternatively, exclude the `third-party` folder. I don't see a reason to include all of the `jinja2` folders. HTML is already excluded.

MebinAbraham (commented):
Can you add a screencast for this please.

cms/jinja2/assets/js/wagtail-editor-customisations.js
kacperpONS: [de3d237](https://github.com/ONSdigital/dis-wagtail/pull/106/commits/de3d23708f50f28261cab4beba9b2bc776f33ef4)

functional_tests/features/information_page.feature
kacperpONS: That was added in [33206a5](https://github.com/ONSdigital/dis-wagtail/pull/106/commits/33206a5773b7559562f0c9976aa5ed17d3c52011) and slightly reworded in [c73178c](https://github.com/ONSdigital/dis-wagtail/pull/106/commits/c73178c9ec364fba8afb6129e474f7c1131c7f89)

MebinAbraham (approved):

sanjeevz3009 (commented):
Reviewed!

PR #108, opened 2025-02-28
Add an ONS table block

zerolab:

This PR adds:
- a new table block using TinyMCE limited to the table plugin
- an ONS Table Block with source/footer notes

[Screen recording](https://www.loom.com/share/d69e45684dab40bf949ff8f3fb7a70c9)

To-Do: 
- [x] integration test that adds a table and checks the output

### How to review

Check the branch out, add/edit a statistical artcile/ methodology page.

### Follow-up Actions

- [x] Package and open source TinyTableBlock - https://github.com/torchbox/wagtail-tinytableblock/


cms/jinja2/templates/components/streamfield/table_block.html
```diff
@@ -1,2 +1,13 @@
...
+
+    {{ onsTable(options) }}
+
+    {%- if source or footnotes %}
+        <footer>
+            {% if source %}<p>{{ _("Source") }}: {{ source }}</p>{% endif %}
+            {% if footnotes %}{{ footnotes|richtext }}{% endif %}
+        </footer>
+    {% endif %}
+</figure>
```
MebinAbraham, line 13: Currently, publishing is possible without any table data. We should either prevent saving the page when there is no table data or only display the supporting element when table data exists like `cms/tiny_table_block/templates/tiny_table_block/table_block.html` does.

cms/tiny_table_block/static/tiny_table_block/js/tiny-table-block.js
```diff
@@ -0,0 +1,74 @@
+/* global tinymce */
```
MebinAbraham, line 1: Hopefully this vendor files doesn't live here long if we can open source and package it separately. Did you consider npm installing?

cms/tiny_table_block/utils.py
```diff
@@ -0,0 +1,70 @@
...
+            headers.append([get_cell_data(cell) for cell in header_row.find_all("th")])
+
+        if tbody_rows := table.find("tbody"):
+            table_rows = tbody_rows.find_all("tr")
+        else:
+            # If there's a thead but no tbody, get all rows not in thead
+            table_rows = [row for row in table.find_all("tr") if row.parent != thead]
+    else:
+        table_rows = table.find_all("tr")
+        # TODO: extract headers from tr if all cells are th?
```
MebinAbraham, line 57: Can you clarify?. Is this needed given `get_cell_data` below is only looping over `td`/`th`?

MebinAbraham (commented):
Done an initial pass, this looks really good. Open sourcing TinyTableBlock 🎉 👍 

- The `Row header` option does not work as expected when using the UI buttons. It works as expected when copying/pasting, but it does not "understand" that we have row headers and the icon is not highlighted. The column header is working as expected.
- We should add functional tests to validate the table behaviours. I feel like there are a few gaps that we don't test. We should test the different functionality and conditional cases such as each element is rendered if input, such as title, caption etc.
- There is an alignment issue with the table/page. It is indented too much. See attached image: 
image

cms/tiny_table_block/static/tiny_table_block/js/tiny-table-block.js
zerolab: I have, but I am using a subset (tinymce + 2 or 3 plugins). Also, bundling it means downstream doesn't need to worry about FE tooling

cms/tiny_table_block/utils.py
zerolab: header rows are wrapped in `<thead>` by the design system - https://github.com/ONSdigital/design-system/blob/main/src/components/table/_macro.njk#L19

it is possible to paste the following table

```
<table>
<tr><th>header cell</th><th>header cell</th></tr>
<tr><td>cell</td><td>cell</td></tr>
</table>
```

which is valid.

I want to get this in front of folk sooner to get feedback, and sort this out while working on the other couple of table-related tickets (links and cell alignment)

cms/jinja2/templates/components/streamfield/table_block.html
zerolab: Funny enough, I have been thinking about the same thing this morning. Will sort

zerolab:
> There is an alignment issue with the table/page. It is indented too much. See attached image

That is because I wrapped the whole output in a `figure` element. Will change to a `div`

cms/core/blocks/markup.py
```diff
@@ -105,3 +108,31 @@ def render(self, value: dict, context: dict | None = None) -> Union[str, "SafeSt
...
+            "title": value.get("title"),
+            "options": {
+                "caption": value.get("caption"),
+                "ths": data.get("headers", []),
+                "trs": [{"tds": row} for row in data.get("rows", [])],
+            },
+            "source": value.get("source"),
+            "footnotes": value.get("footnotes"),
+            **context,
+        }
```
kacperpONS, line 134: While I appreciate the idea here, we tend to have the values passed to the macro directly in the template files. It has the advantage that you can see what are the used options at a glance when looking at the template files. 
Since `options` don't really have a justification until it gets to the macro - let's return a flat dict here and construct the dict in the template files.

cms/jinja2/templates/components/streamfield/table_block.html
```diff
@@ -1,2 +1,13 @@
...
+    {% if title %}<h3 class="heading-4">{{ title }}</h3>{% endif %}
+
+    {{ onsTable(options) }}
+
+    {%- if source or footnotes %}
+        <footer>
+            {% if source %}<p>{{ _("Source") }}: {{ source }}</p>{% endif %}
+            {% if footnotes %}{{ footnotes|richtext }}{% endif %}
+        </footer>
+    {% endif %}
```
kacperpONS, line 12: Couldn't the `title` and `footnotes` be used in the custom macro for the table instead? 

I'm assuming the long term plan is to include this new version of the table into the DS. This way it'd reduce the need for modification of this file in the future.

zerolab: Fair point, but that is a Design System question. 
I wanted to make minimum modifications as I don't know the direction in which DS will take it. If it gets incorporated, then will happily use it

cms/jinja2/templates/components/streamfield/table_block.html
zerolab: 5139d29
went with hide if empty table

cms/jinja2/templates/components/streamfield/table_block.html
```diff
@@ -1,2 +1,13 @@
...
+{% from "component_overrides/table/_macro.njk" import onsTable %}
+<figure id="table-{{ block_id }}">
+    {% if title %}<h3 class="heading-4">{{ title }}</h3>{% endif %}
+
+    {{ onsTable(options) }}
+
+    {%- if source or footnotes %}
+        <footer>
+            {% if source %}<p>{{ _("Source") }}: {{ source }}</p>{% endif %}
+            {% if footnotes %}{{ footnotes|richtext }}{% endif %}
```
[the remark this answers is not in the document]
zerolab: 5139d29

cms/core/blocks/markup.py
sanjeevz3009: I second Mebin, but could be biased from the main menu work. But I do side with avoiding logic in templates as much as possible. I like Dan's approach!

zerolab:
@MebinAbraham I've re-tagged you. Will shortly switch to using the PyPI package for TinyTableBlock, but I think the changes are worth checking here

cms/core/tests/test_blocks.py
```diff
@@ -245,7 +246,7 @@ def test_relatedlinksblock__toc(self):
             block.to_table_of_contents_items(block.to_python([])), [{"url": "#related-links", "text": "Related links"}]
         )
 
-    def test_basictableblock_get_context(self):
+    def test_basictableblock__get_context(self):
```
sanjeevz3009, line 249: Nit: `test_basictableblock__get_context` to `test_basic_table_block__get_context`

sanjeevz3009, line 249: Nit: `test_basictableblock__get_context` to `test_basic_table_block__get_context` or  `test_basic_tableblock__get_context`
```suggestion
    def test_basic_table_block__get_context(self):
```

sanjeevz3009 (commented):
Minor comments. LGTM!

cms/core/tests/test_blocks.py
zerolab: if we do this, then we need to update all the other test names which follow the `test_BlockClassToLowerCase__extra` pattern :)

sanjeevz3009: In that case, all good by me!

MebinAbraham:
> * The `Row header` option does not work as expected when using the UI buttons. It works as expected when copying/pasting, but it does not "understand" that we have row headers and the icon is not highlighted. The column header is working as expected.

This still seems to be outstanding.

Also, I feel like there are still gaps in the test. We should have confidence in the feature of TinyMCE provide as merged cells, row headers, column headers etc. For headers, we have tests that validate the value is displayed but not whether it is syntactically correct. At the moment, we don't validate the rendered version for things like merged cells, row headers, column headers etc.

RealOrangeOne (approved):

sanjeevz3009 (approved):
LGTM! Tested everything locally.

cms/jinja2/component_overrides/table/_macro.njk
```diff
@@ -0,0 +1,81 @@
...
+    {% if 'scrollable' in variants %}
+    <div class="ons-table-scrollable ons-table-scrollable--on">
+        <div class="ons-table-scrollable__content" tabindex="0" role="region" aria-label="{{ params.caption }}. {{ params.ariaLabel | default("Scrollable table") }}">
+    {% endif %}
+            <table {% if params.id %}id="{{ params.id }}"{% endif %} class="ons-table{% if params.tableClasses %} {{ params.tableClasses }}{% endif %}{% if variants %}{% if variants is not string %}{% for variant in variants %} ons-table--{{ variant }}{% endfor %}{% else %} ons-table--{{ variants }}{% endif %}{% endif %}" {% if params.sortBy and 'sortable' in variants %}data-aria-sort="{{ params.sortBy }}" data-aria-asc="{{ params.ariaAsc }}" data-aria-desc="{{ params.ariaDesc }}"{% endif %}>
+                {% if params.caption %}
+                <caption class="ons-table__caption{{ " ons-u-vh" if params.hideCaption }}">{{ params.caption }}</caption>
+                {% endif %}
+                <thead class="ons-table__head">
+                    {% set headers = params.headers if params.headers else parms.ths|list %}
```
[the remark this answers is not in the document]
zerolab: fc6bc43

zerolab:
@MebinAbraham 2f43015 should take care of the header rows highlight in the editor and output, 35ac138 adds tests for rendering with a few variations, a6a2b4f improves the macro

MebinAbraham:
> @MebinAbraham [2f43015](https://github.com/ONSdigital/dis-wagtail/commit/2f43015ea02c68e463edd626709f063c77a362cb) should take care of the header rows highlight in the editor and output, [35ac138](https://github.com/ONSdigital/dis-wagtail/commit/35ac13887d6c4a2e13fb3a170d9e2e43f1c5307d) adds tests for rendering with a few variations, [a6a2b4f](https://github.com/ONSdigital/dis-wagtail/commit/a6a2b4fa3b11656d294a9770cccf3feacd2ab389) improves the macro

@zerolab The row header still does not seem to work when using the UI. If I click the row header option, the row disappears. Headers work fine when copy pasting table data. 
image

MebinAbraham (commented):
~Can we add a test for links support. I can't seem to get links to work. They don't render in anchor tags even though the admin UI looks like it has got a link.
Looks like the DS component override has not been updated to support it?.~

Edit: Fixed after installed latest deps.

functional_tests/steps/statistical_article_page.py
```diff
@@ -0,0 +1,77 @@
...
+@then("the published statistical article page is displayed with the populated data")
+def the_statistical_article_page_is_displayed_with_the_populated_data(context: Context):
+    expect(context.page.get_by_text("Statistical article", exact=True)).to_be_visible()
+    expect(context.page.get_by_role("heading", name="The article page")).to_be_visible()
+    expect(context.page.get_by_text("Page summary")).to_be_visible()
+    expect(context.page.get_by_text("11 January 2025", exact=True)).to_be_visible()
+    expect(context.page.get_by_role("heading", name="Cite this analysis")).to_be_visible()
+
+    expect(context.page.get_by_role("heading", name="Heading")).to_be_visible()
+    expect(context.page.get_by_role("heading", name="Content")).to_be_visible()
```
MebinAbraham, line 70: We should update to check footnotes accordian renders correctly.

MebinAbraham, line 70: We should update this to include the footnotes accordion to ensure that it renders/behaves correctly.

functional_tests/steps/statistical_article_page.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@then("the published statistical article page has the added table")
+def _published_statistical_article_page_has_the_added_table(context: Context):
+    expect(context.page.get_by_role("table")).to_be_visible()
+    expect(context.page.get_by_text("cell1")).to_be_visible()
+    expect(context.page.get_by_text("cell2")).to_be_visible()
```
MebinAbraham, line 77: We should update to check footnotes accordion renders/behaves correctly. It would be better if we changed to check the data instead of visibility in these cases.

MebinAbraham (approved):
Happy with as-is, though, I do think we should expand the UI tests even as a future exercise to cover more of TinyMCE's functions and assure ourselves of our implementation.

functional_tests/steps/statistical_article_page.py
zerolab: 5181add

MebinAbraham (approved):

functional_tests/steps/statistical_article_page.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+@then("the user can expand the footnotes")
+def expand_footnotes(context: Context):
+    page = context.page
+
+    # page.wait_for_timeout(750)  # added to allow JS to be ready
```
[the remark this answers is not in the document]
zerolab: 5596769

PR #110, opened 2025-03-04
Glossary terms snippet

kacperpONS:
### What is the context of this PR?

We want to add the Glossary term snippet and allow it to be used on Methodology and Statistical Article pages.

### How to review

Make sure code looks ok, run tests.

### Links
https://jira.ons.gov.uk/browse/CMS-81

PR #113, opened 2025-03-11
Add a User chooser

zerolab:

Noticed while reviewing #110. Adds a Wagtail chooser for Users.

Screencast: https://www.loom.com/share/4da321f60e92471f876c560eb21d798f

### How to review

Check out on top of the branch from #110, go to add/edit a glossary term

### Follow-up Actions

Try to get this in core

### DIFF
--- a/cms/users/tests/test_viewsets.py
+++ b/cms/users/tests/test_viewsets.py
@@ -1,0 +1,58 @@
+from django.test import TestCase
+from django.urls import reverse
+from wagtail.test.utils import WagtailTestUtils
+
+from cms.users.tests.factories import UserFactory
+from cms.users.viewsets import user_chooser_viewset
+
+
+class TestUserChooserViewSet(WagtailTestUtils, TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.chooser_url = user_chooser_viewset.widget_class().get_chooser_modal_url()
+        cls.chooser_results_url = reverse(user_chooser_viewset.get_url_name("choose_results"))
+
+        cls.user_without_name = UserFactory(username="noname", first_name="", last_name="", is_active=False)
+        cls.user_with_first_name = UserFactory(username="first_name_user", first_name="Data", last_name="")
+        cls.user_with_last_name = UserFactory(username="last_name_user", first_name="", last_name="La Forge ")
+        cls.user_with_full_name = UserFactory(
+            username="full_name_user", first_name="Jean-Luc", last_name="Picard", is_superuser=True
+        )
+
+    def setUp(self):
+        self.login(user=self.user_with_full_name)
+
+    def test_chooser_viewset(self):
+        response = self.client.get(self.chooser_url)
+
+        self.assertContains(response, self.user_without_name.username, 2)  # Name and username
+        self.assertContains(response, self.user_with_first_name.username)
+        self.assertContains(response, self.user_with_last_name.username)
+        self.assertContains(response, self.user_with_full_name.username)
+
+        self.assertContains(response, "Data")
+        self.assertContains(response, "La Forge")
+        self.assertContains(response, "Jean-Luc Picard")
+
+        self.assertContains(response, "Active", 3)
+        self.assertContains(response, "Inactive", 1)
+
+    def test_chooser_search(self):
+        """Tests that the chooser search results work as expected."""
+        response = self.client.get(f"{self.chooser_results_url}?q={self.user_with_first_name.first_name}")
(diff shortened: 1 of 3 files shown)


cms/users/viewsets.py
```diff
@@ -0,0 +1,88 @@
...
+                accessor=lambda u: "Active" if u.is_active else "Inactive",
+                primary=lambda u: u.is_active,
+                label="Status",
+                width="10%",
+            ),
+            DateColumn(
+                "last_login",
+                label="Last login",
+                width="15%",
+            ),
```
MebinAbraham, line 68: I suggest not displaying this to other users.

cms/users/viewsets.py
```diff
@@ -0,0 +1,88 @@
...
+from .models import User
+
+if TYPE_CHECKING:
+    from django.db.models import QuerySet
+
+
+class UserFilterForm(SearchFilterMixin, BaseFilterForm):
+    @cached_property
+    def model_fields(self) -> set[str]:
+        return {f.name for f in User._meta.get_fields()}
```
MebinAbraham, line 22: Nit:
```suggestion
        return {field.name for field in User._meta.get_fields()}
```

MebinAbraham (commented):
Nice enhancement. Hopefully something like this can get into core.

Let's add a few tests to ensure the chooser is as expected, otherwise LGTM.

cms/users/viewsets.py
zerolab: 42d1703

MebinAbraham (approved):

cms/users/tests/test_viewsets.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+
+    def setUp(self):
+        self.login(user=self.user_with_full_name)
+
+    def test_chooser_viewset(self):
+        response = self.client.get(self.chooser_url)
+
+        self.assertContains(response, self.user_without_name.username, 2)  # Name and username
+        self.assertContains(response, self.user_with_first_name.username)
```
MaciekBaron, line 29: I think the comments should be on the line below, as `user_with_first_name` has both name and username, while `user_without_name` just has the username and no name. Unless I misunderstood the comment.
```suggestion
        self.assertContains(response, self.user_without_name.username, 2)
        self.assertContains(response, self.user_with_first_name.username)  # Name and username
```

cms/users/tests/test_viewsets.py
```diff
@@ -0,0 +1,58 @@
...
+
+class TestUserChooserViewSet(WagtailTestUtils, TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.chooser_url = user_chooser_viewset.widget_class().get_chooser_modal_url()
+        cls.chooser_results_url = reverse(user_chooser_viewset.get_url_name("choose_results"))
+
+        cls.user_without_name = UserFactory(username="noname", first_name="", last_name="", is_active=False)
+        cls.user_with_first_name = UserFactory(username="first_name_user", first_name="Data", last_name="")
+        cls.user_with_last_name = UserFactory(username="last_name_user", first_name="", last_name="La Forge ")
```
MaciekBaron, line 17: Nit-pick: Dangling space

```suggestion
        cls.user_with_last_name = UserFactory(username="last_name_user", first_name="", last_name="La Forge")
```

MaciekBaron (changes requested):

cms/users/tests/test_viewsets.py
zerolab: not quite. if the user doesn't have a fullname `get_user_display_name` defaults to the username - https://github.com/wagtail/wagtail/blob/bfcd0c7496279137c28ebd00a304fa2a514df47b/wagtail/admin/utils.py#L52, so this checks that the username is displayed twice

cms/users/tests/test_viewsets.py
zerolab: Good catch. And that is on purpose as `get_user_display_name` does `.strip()`. Granted it is a core method which should be tested, but I left it on purpose

cms/users/tests/test_viewsets.py
MaciekBaron: Makes sense, thanks!

MaciekBaron (approved):

MaciekBaron:
LGTM

sanjeevz3009 (approved):
LGTM! Works as expected!

kacperpONS (approved):

PR #115, opened 2025-03-13
Search Integration - Kafka

sanjeevz3009:

- This feature ensures that content updates in Wagtail (publishing and unpublishing) are reflected in the ONS Search Service.
- This uses Wagtail hooks to trigger Kafka events for content changes.
- We want to ensure content from both Wagtail and the existing CMS can be surfaced to the users.

### How to review

**Search spec links:** 
- https://github.com/ONSdigital/dp-search-data-extractor/blob/develop/specification.yml#L53
- https://github.com/ONSdigital/dis-search-upstream-stub/blob/main/docs/contract/resource_metadata.yml
- https://github.com/ONSdigital/dp-search-data-importer/blob/30fb507e90f2cf1974ec0ca43bb0466307e2f112/specification.yml#L186

0. Make sure the publishers.py/ producer code/ logic aligns with the search spec.
1. Pull down this branch/ checkout to this branch.
2. Spin up the Wagtail web app locally per usual.
3. Create -> publish pages (Information Page, Methodology Page, Release Calendar Page, Index Page, and Statistical Article Page) -> you should see the producer logs on your terminal -> unpublish these pages -> you should see the producer logs on your terminal.
4. Try to create and publish excluded pages (Article series, Home page, Release calendar index, Theme page, and Topic Page) -> you should not see any events get fired/ logged on the terminal-> try unpublish these pages -> you should not see any events fire/ logged on the terminal.
5. Spin up the Kafka broker container to test the compatibility with an actual Kafka broker service with the command `make compose-dev-pull` -> `make compose-dev-up` -> set the env vars to 
```
SEARCH_INDEX_PUBLISHER_BACKEND = "kafka"
KAFKA_SERVER = "localhost:9092"
KAFKA_CHANNEL_CREATED_OR_UPDATED = "search-content-updated"
KAFKA_CHANNEL_DELETED = "search-content-deleted"
KAFKA_API_VERSION = tuple(map(int, "3,5,1".split(",")))
```
-> Now repeat steps 3 to 4 and you should see the events being fired and sent to the Kafka broker/ logged on the terminal.

### Follow-up Actions

- [ ] Configure the env vars in Heroku for `SEARCH_INDEX_PUBLISHER_BACKEND`, `KAFKA_SERVER`, `KAFKA_CHANNEL_CREATED_OR_UPDATED`,  `KAFKA_CHANNEL_DELETED` and `KAFKA_API_VERSION`.
- [ ] Determine if we can enqueue the Kafka messages for a bit of async work (for context: this relates to `result = future.get(timeout=10)`)


cms/search/migrations/__init__.py
zerolab, line 1: the `migrations` dir is unnecessary

cms/settings/dev.py
```diff
@@ -68,6 +68,8 @@
 # Django Defender
 ENABLE_DJANGO_DEFENDER = False
 
+# PUBLISHER_BACKEND = os.getenv("PUBLISHER_BACKEND", "log")  # Options: log, kafka, none
```
zerolab, line 71: suggestion: move to `local.py.example`

cms/settings/base.py
```diff
@@ -850,3 +851,8 @@
 SLACK_NOTIFICATIONS_WEBHOOK_URL = env.get("SLACK_NOTIFICATIONS_WEBHOOK_URL")
 
 ONS_API_BASE_URL = env.get("ONS_API_BASE_URL")
+
+PUBLISHER_BACKEND = os.getenv("PUBLISHER_BACKEND", "kafka")
+KAFKA_SERVER = os.getenv("KAFKA_SERVER", "localhost:9092")
+KAFKA_TOPIC_CREATED_OR_UPDATED = os.getenv("KAFKA_TOPIC_CREATED_OR_UPDATED", "search-content-updated")
+KAFKA_TOPIC_DELETED = os.getenv("KAFKA_TOPIC_DELETED", "search-content-deleted")
```
zerolab, line 858: Let's avoid falling back on localhost paths in the base settings.
The rule of thumb for integrations like this:
1. are the key settings present in the env? In this case this would be `PUBLISHER_BACKEND` and `KAFKA_SERVER`. 
2. if the backend env var is "kafka" and we have the server env var, set up all the settings. If the backend env var is set to something else, use that

The bottom line is `localhost:9092` is not a good fallback for a prod env

cms/search/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class SearchConfig(AppConfig):
+    default_auto_field = "django.db.models.BigAutoField"
```
zerolab, line 5:
```suggestion
    default_auto_field = "django.db.models.AutoField"
```

at most. Given you don't have any special config, I'd say `apps.py` is not needed here

cms/search/publishers.py
```diff
@@ -0,0 +1,202 @@
...
+
+        mapping = {
+            "ReleaseCalendarPage": "release",
+            "StatisticalArticlePage": "bulletin",
+            "InformationPage": "static_page",
+            "IndexPage": "static_landing_page",
+            "MethodologyPage": "static_methodology",
+        }
+
+        return mapping.get(page_class_name)
```
zerolab, line 142:
```suggestion
        return mapping.get(page.__class__.__name__)
```

without `page_class_name = page.__class__.__name__`

cms/search/test_wagtail_hooks.py
```diff
@@ -0,0 +1,51 @@
...
+        self.__class__.__name__ = class_name
+
+
+class WagtailHooksTests(TestCase):
+    @patch("cms.search.wagtail_hooks.publisher")
+    def test_page_published_excluded_page_type(self, mock_publisher):
+        """Pages in EXCLUDED_PAGE_TYPES should not trigger any publish calls."""
+        for excluded_type in EXCLUDED_PAGE_TYPES:
+            page = MockPage(excluded_type)
+            page_published(MockRequest(), page)
```
zerolab, line 28: please use [`get_dummy_request()`](https://github.com/wagtail/wagtail/blob/bfcd0c7496279137c28ebd00a304fa2a514df47b/wagtail/coreutils.py#L413)

cms/search/publishers.py
```diff
@@ -0,0 +1,202 @@
...
+import logging
+from abc import ABC, abstractmethod
+
+from django.conf import settings
+from kafka import KafkaProducer
+
+logger = logging.getLogger(__name__)
+
+
+class BasePublisher(ABC):
```
zerolab, line 11: this needs a link to the search spec at the very least.
I suggest adding a docs page about the integration with the relevant detail

zerolab (commented):
Did a quick skim and left a few notes/questions.
Will review when fully ready

cms/settings/base.py
sanjeevz3009: Yes, I was going to clean up all of this, as this is not the correct way at all. Currently, it's only serving me to test some things.

cms/search/apps.py
sanjeevz3009: Removed!

cms/search/publishers.py
sanjeevz3009: Done!

cms/search/test_wagtail_hooks.py
sanjeevz3009: Done + using the factories we have now.

cms/search/publishers.py
sanjeevz3009: Will get to this at the end =)

sanjeevz3009: Added relevant links for now!

cms/search/publishers.py
```diff
@@ -0,0 +1,202 @@
...
+    dp-search-data-importer spec link:
+    https://github.com/ONSdigital/dp-search-data-importer/blob/30fb507e90f2cf1974ec0ca43bb0466307e2f112/specification.yml#L186
+    contract: https://github.com/ONSdigital/dis-search-upstream-stub/blob/main/docs/contract/resource_metadata.yml
+    """
+
+    def publish_created_or_updated(self, page: "Page") -> None:
+        """Build the message for the created/updated event.
+        Delegate sending to the subclass's _publish_to_service().
+        """
+        topic = self.get_topic_created_or_updated()
```
zerolab, line 33: nitpick: while this follows the terminology used in the dp-search schema, I think we should adapt it to our project. Speficially, we have `Topic` from taxonomy, and Topic page. But here "topic" is the operation https://github.com/ONSdigital/dp-search-data-extractor/blob/develop/specification.yml#L13-L50

Let's make this relevant, and avoid confusion. So `operation` and `get_created_or_updated_operation`. Or worst case scenarion `channel` (https://github.com/ONSdigital/dp-search-data-extractor/blob/develop/specification.yml#L13-L28)

cms/search/publishers.py
```diff
@@ -0,0 +1,202 @@
...
+        message = self._construct_message_for_create_update(page)
+        logger.info(
+            "BasePublisher: About to publish created/updated message=%s to topic=%s",
```
zerolab, line 36:
```
BasePublisher: About to publish created/updated message={...}
LogPublisher: topic=log-created-or-updated message={...}
```

Doesn't flow. IMHO, the log messages should be consistent and specify the actual publisher. ie. if I am using `LogPublisher` then all log messages should state that.

Additionally, KafkaPublisher._publish_to_service does:

```
KafkaPublisher: Publishing to topic=%s, message=%s
KafkaPublisher: Publish result for topic %s: %s
```

so we end up with 

```
BasePublisher: About to publish created/updated message={...}
KafkaPublisher: Publishing to topic=%s, message=%s
KafkaPublisher: Publish result for topic %s: %s
```

cms/search/publishers.py
```diff
@@ -0,0 +1,202 @@
...
+           - ReleasePayload: extends StandardPayload with release-specific fields
+        """
+        # For pages that are 'release' content, we add release fields below.
+        content_type = self._map_page_type_to_content_type(page)
+
+        # Common fields for StandardPayload (also part of ReleasePayload)
+        # The schema requires at minimum: uri, title, content_type
+        message = {
+            "uri": page.url_path,
+            "content_type": page.content_type_id,
```
zerolab, line 87:
```suggestion
            "content_type": content_type,
```

if I am reading https://github.com/ONSdigital/dis-search-upstream-stub/blob/main/docs/contract/resource_metadata.yml#L18-L46 correctly, no?

`page.content_type_id` is an integer

cms/search/publishers.py
```diff
@@ -0,0 +1,202 @@
...
+            "release_date": (
+                page.release_date.isoformat().replace("+00:00", "Z")
+                if content_type == "release" and getattr(page, "release_date", None)
+                else None
+            ),
+            "summary": page.summary,
+            "title": page.title,
+            "topics": list(page.topics.values_list("topic_id", flat=True)) if hasattr(page, "topics") else [],
```
zerolab, line 95: This needs more work for `StatisticalArticlePage` because the topics are on the series.

suggestion: add a cached `topics` property on `StatisticalArticlePage`. Even better, add a `topic_ids` property to the taxonomy mixins that would return the list of topic IDs. And this then becomes `"topics": getattr(page, "topic_ids", []),`

cms/search/publishers.py
```diff
@@ -0,0 +1,202 @@
...
+        # The schema requires at minimum: uri, title, content_type
+        message = {
+            "uri": page.url_path,
+            "content_type": page.content_type_id,
+            "release_date": (
+                page.release_date.isoformat().replace("+00:00", "Z")
+                if content_type == "release" and getattr(page, "release_date", None)
+                else None
+            ),
+            "summary": page.summary,
```
zerolab, line 93: this needs to be wrapped in [`expand_db_html()`](https://github.com/wagtail/wagtail/blob/8f640a8cdb0dfcf14e611d33cbff388d5db7cf9d/wagtail/templatetags/wagtailcore_tags.py#L124-L127) so that any internal links are properly converted. Unless the search API needs plain text, in which case `get_text_for_indexing(force_str(page.summary))` ([ref](https://github.com/wagtail/wagtail/blob/8f640a8cdb0dfcf14e611d33cbff388d5db7cf9d/wagtail/rich_text/__init__.py#L65))

cms/search/publishers.py
```diff
@@ -0,0 +1,202 @@
...
+        """Build a dict that matches the 'content-deleted' schema:
+        - required: uri
+        - optional: trace_id.
+        """
+        message = {
+            "uri": page.url_path,
+        }
+
+        if hasattr(page, "trace_id"):
+            message["trace_id"] = page.trace_id
```
zerolab, line 135: We don't have `trace_id`

cms/search/publishers.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def __init__(self) -> None:
+        # Read Kafka configs settings
+        self.kafka_server = settings.KAFKA_SERVER
+        self._topic_created_or_updated: Optional[str] = settings.KAFKA_TOPIC_CREATED_OR_UPDATED
+        self._topic_deleted: Optional[str] = settings.KAFKA_TOPIC_DELETED
+
+        self.producer = KafkaProducer(
+            bootstrap_servers=[self.kafka_server],
+            api_version=(3, 8, 0),
```
zerolab, line 166: question: I take it this needs to match the installed Kafka version. We should have a least a note about it. And if that is case, we need a note in the infra IAC

cc @MebinAbraham

cms/search/publishers.py
```diff
@@ -0,0 +1,202 @@
...
+            value_serializer=lambda v: json.dumps(v).encode("utf-8"),
+        )
+
+    def get_topic_created_or_updated(self) -> str:
+        if self._topic_created_or_updated is None:
```
zerolab, line 171: This check and the check below should either happen in `__init__`, or even better as a system check https://docs.djangoproject.com/en/5.1/topics/checks/

cms/search/publishers.py
```diff
@@ -0,0 +1,202 @@
...
+        if self._topic_deleted is None:
+            raise ValueError("Topic must not be None here.")
+        return self._topic_deleted
+
+    def _publish_to_service(self, topic: str, message: dict) -> "RecordMetadata":
+        """Send the message to Kafka."""
+        logger.info("KafkaPublisher: Publishing to topic=%s, message=%s", topic, message)
+        future = self.producer.send(topic, message)
+        # Optionally block for the result, capturing metadata or error
+        result = future.get(timeout=10)  # Wait up to 10s for send to complete
```
zerolab, line 185: question: What happens if there's an unpredictable error?

thought: the 10s timeout will impact bundle publishing if there are lots of pages. It may have to be a follow up PR, but I wonder if django-tasks could come in play here: https://docs.wagtail.org/en/stable/releases/6.4.html#support-for-background-tasks-using-django-tasks

Unless we do fire and forget. 

cc @RealOrangeOne

cms/search/publishers.py
```diff
@@ -0,0 +1,202 @@
...
+        logger.info("KafkaPublisher: Publish result for topic %s: %s", topic, result)
+        # breakpoint()
```
zerolab, line 187: Please clean-up debug or unused code.

cms/search/tests/test_kafka_integration.py
```diff
@@ -0,0 +1,84 @@
...
+        logging.getLogger(__name__).info("KafkaIntegrationTests: setUpClass complete.")
+
+    @classmethod
+    def tearDownClass(cls):
+        # Close the consumers
+        cls.consumer_created.close()
+        cls.consumer_deleted.close()
+        super().tearDownClass()
+
+    def _poll_for_message(self, consumer, expected_uri, timeout_secs=5):
```
zerolab, line 52:
```suggestion
    def _poll_for_message(self, consumer, expected_uri, timeout_seconds=5):
```

let's be more verbose here

cms/search/tests/test_kafka_integration.py
```diff
@@ -0,0 +1,84 @@
...
+        self.assertIsNotNone(publish_result)  # We get some metadata from Kafka
+
+        msg_found = self._poll_for_message(self.consumer_created, page.url_path)
+        self.assertTrue(msg_found, "No matching message found in 'search-content-updated' topic.")
+
+    def test_publish_deleted_integration(self):
+        """Publish a "deleted" message to Kafka and consume it from 'search-content-deleted'."""
+        page = InformationPageFactory()
+
+        _ = self.publisher.publish_deleted(page)
```
zerolab, line 81:
```suggestion
        self.publisher.publish_deleted(page)
```

cms/search/tests/test_wagtail_hooks.py
```diff
@@ -0,0 +1,96 @@
...
+
+        cls.included_factories = [
+            InformationPageFactory,
+            MethodologyPageFactory,
+            ReleaseCalendarPageFactory,
+            StatisticalArticlePageFactory,
+        ]
+
+        cls.index_page = IndexPageFactory(slug="custom-slug-1")
+        cls.included_factories.append(lambda: cls.index_page)
```
zerolab, line 54: considering you're initialising the pages anyway, why not store them as instances?

cms/search/wagtail_hooks.py
```diff
@@ -0,0 +1,41 @@
+from typing import TYPE_CHECKING
+
+from wagtail import hooks
+from wagtail.models import Page
```
zerolab, line 4: nitpick: move to `if TYPE_CHECKING`

cms/search/wagtail_hooks.py
```diff
@@ -0,0 +1,41 @@
...
+
+@hooks.register("after_unpublish_page")
+def page_unpublished(request: "HttpRequest", page: Page) -> None:
+    """This hook is called after a page is unpublished from the Wagtail admin."""
+    if _should_exclude_page(page):
+        return
+    publisher.publish_deleted(page)
+
+
+EXCLUDED_PAGE_TYPES = (
```
zerolab, line 35: nitpick: this should be in settings

cms/search/__init__.py
```diff
@@ -0,0 +1,13 @@
+from typing import Union
+
+from django.conf import settings
+
+from .publishers import KafkaPublisher, LogPublisher
+
+
+def get_publisher() -> Union[KafkaPublisher, LogPublisher]:
```
zerolab, line 8: also, it may be a good idea to cache the publisher. [reference](https://adamj.eu/tech/2023/03/02/django-profile-and-improve-import-time/#do-other-kinds-of-work-lazily)

cms/search/wagtail_hooks.py
```diff
@@ -0,0 +1,41 @@
...
+    """Return True if the page should be excluded from publishing."""
+    return page.__class__.__name__ in EXCLUDED_PAGE_TYPES
+
+
+@hooks.register("after_publish_page")
+def page_published(request: "HttpRequest", page: Page) -> None:
+    """This hook is called after a page is published from the Wagtail admin."""
+    if _should_exclude_page(page):
+        return
+    publisher.publish_created_or_updated(page)
```
zerolab, line 24:
```suggestion
    if not _should_exclude_page(page):
        publisher.publish_created_or_updated(page)
```

and I would go as far as doing `if not page.__class__.__name__ in EXCLUDED_PAGE_TYPES:` ...

docker-compose-dev.yml
```diff
@@ -23,5 +23,24 @@ services:
     ports:
       - 6379:6379
 
+  broker:
```
zerolab, line 26: question: should this not be part of the main `docker-compose.yml` in that it is useful whether you run everything in Docker, or partially

zerolab (changes requested):
Lots of good stuff in here. Left a number of questions and suggestions

cms/search/publishers.py
```diff
@@ -0,0 +1,202 @@
...
+        # For pages that are 'release' content, we add release fields below.
+        content_type = self._map_page_type_to_content_type(page)
+
+        # Common fields for StandardPayload (also part of ReleasePayload)
+        # The schema requires at minimum: uri, title, content_type
+        message = {
+            "uri": page.url_path,
+            "content_type": page.content_type_id,
+            "release_date": (
+                page.release_date.isoformat().replace("+00:00", "Z")
```
MebinAbraham, line 89: We shouldn't need to change to Zulu; `+00:00` is a valid ISO 8601 format. Was this done for some specific reason?

cms/search/publishers.py
MebinAbraham: `get_text_for_indexing` 👍

cms/search/publishers.py
```diff
@@ -0,0 +1,202 @@
...
+        """Maps the class name of a given page object to a corresponding content type string."""
+        mapping = {
+            "ReleaseCalendarPage": "release",
+            "StatisticalArticlePage": "bulletin",
+            "InformationPage": "static_page",
+            "IndexPage": "static_landing_page",
+            "MethodologyPage": "static_methodology",
+        }
+
+        return mapping.get(page.__class__.__name__)
```
MebinAbraham, line 149: Nit: instead of having to keep this in sync with page name changes/new types, I suggest we add a property to each page model called something like `search_index_content_type`

cms/search/wagtail_hooks.py
```diff
@@ -0,0 +1,41 @@
...
+from . import get_publisher
+
+if TYPE_CHECKING:
+    from django.http import HttpRequest
+
+publisher = get_publisher()
+
+
+def _should_exclude_page(page: Page) -> bool:
+    """Return True if the page should be excluded from publishing."""
```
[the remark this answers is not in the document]
MebinAbraham: Nit: This could be a property on the base page model.

cms/search/wagtail_hooks.py
```diff
@@ -0,0 +1,41 @@
...
+    return page.__class__.__name__ in EXCLUDED_PAGE_TYPES
+
+
+@hooks.register("after_publish_page")
+def page_published(request: "HttpRequest", page: Page) -> None:
+    """This hook is called after a page is published from the Wagtail admin."""
```
MebinAbraham, line 21: Let's clarify this doc string. Might be Admin UI or programmatic. Not sure we need the doc string. Could link to doc source.

cms/settings/base.py
```diff
@@ -851,4 +852,9 @@
 
 ONS_API_BASE_URL = env.get("ONS_API_BASE_URL")
 
+PUBLISHER_BACKEND = os.getenv("PUBLISHER_BACKEND")
```
MebinAbraham, line 855: Let's make the purpose a bit clearer, maybe something like `SEARCH_INDEX_PUBLISHER_BACKEND` or just `SEARCH_INDEX_PUBLISHER`.

MebinAbraham (changes requested):

cms/search/__init__.py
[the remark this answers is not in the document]
sanjeevz3009: Done!

cms/settings/base.py
sanjeevz3009: Okay, made the change! Clearer now!

cms/search/publishers.py
sanjeevz3009: Good spot, removed!

cms/search/wagtail_hooks.py
sanjeevz3009: Okay, understood. Done!

cms/search/wagtail_hooks.py
sanjeevz3009: Great idea, moved to settings!

cms/search/wagtail_hooks.py
sanjeevz3009: Damn it, imported it for that purpose, ended up not moving it! Done!

cms/search/tests/test_kafka_integration.py
sanjeevz3009: Roger, understood! Done!

cms/search/tests/test_kafka_integration.py
sanjeevz3009: Done!

cms/search/wagtail_hooks.py
sanjeevz3009: Happy with that! Done!

cms/search/__init__.py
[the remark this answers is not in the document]
zerolab: Did you forget to push?

cms/search/publishers.py
zerolab: Let's go with channel

cms/search/__init__.py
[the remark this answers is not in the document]
sanjeevz3009: Yes, pushed everything now!

cms/search/publishers.py
sanjeevz3009: Done!

cms/search/publishers.py
sanjeevz3009: Done!

cms/search/__init__.py
sanjeevz3009: Done!

cms/search/publishers.py
```diff
@@ -0,0 +1,202 @@
...
+        message = self._construct_message_for_delete(page)
+        logger.info(
+            "BasePublisher: About to publish deleted message=%s to topic=%s",
+            message,
+            topic,
+        )
+        return self._publish_to_service(topic, message)
+
+    @abstractmethod
+    def _publish_to_service(self, topic: str, message: dict) -> None:
```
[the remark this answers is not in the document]
sanjeevz3009: Done!

cms/search/publishers.py
sanjeevz3009: Done!

cms/search/publishers.py
```diff
@@ -0,0 +1,176 @@
...
+
+        resource_metadata.yml:
+           - StandardPayload: requires {uri, title, content_type}, optional fields
+           - ReleasePayload: extends StandardPayload with release-specific fields
+        """
+        # Common fields for StandardPayload (also part of ReleasePayload)
+        # The schema requires at minimum: uri, title, content_type
+        message = {
+            "uri": page.url_path,
+            "content_type": page.search_index_content_type,
```
zerolab, line 88: suggestion: either use `getattr()` with a generic defualt, or add an attribute on `BasePage` with a generic default. Or add a Django system check that if someone adds a new page type to the list to send to the search index, it has a `search_index_content_type` attribute.

cms/articles/models.py
```diff
@@ -232,3 +232,8 @@ def is_latest(self) -> bool:
             .first()
         )
         return bool(self.pk == latest_id)  # to placate mypy
+
+    @property
+    def search_index_content_type(self) -> str:
+        """Returns the content type string for the search index."""
+        return "bulletin"
```
zerolab, line 239: question: why not an actual property on the class?

cms/settings/base.py
```diff
@@ -852,6 +870,19 @@
 
 ONS_API_BASE_URL = env.get("ONS_API_BASE_URL")
 
+SEARCH_INDEX_PUBLISHER_BACKEND = os.getenv("SEARCH_INDEX_PUBLISHER_BACKEND")
+KAFKA_SERVER = os.getenv("KAFKA_SERVER")
+KAFKA_CHANNEL_CREATED_OR_UPDATED = os.getenv("KAFKA_CHANNEL_CREATED_OR_UPDATED")
+KAFKA_CHANNEL_DELETED = os.getenv("KAFKA_CHANNEL_DELETED")
+
+EXCLUDED_PAGE_TYPES = (
```
zerolab, line 878:
```suggestion
SEARCH_INDEX_EXCLUDED_PAGE_TYPES = (
```

cms/settings/base.py
```diff
@@ -852,6 +870,19 @@
 
 ONS_API_BASE_URL = env.get("ONS_API_BASE_URL")
 
+SEARCH_INDEX_PUBLISHER_BACKEND = os.getenv("SEARCH_INDEX_PUBLISHER_BACKEND")
```
zerolab, line 873:
```suggestion
#
# Configuration for the External Search service 
#
SEARCH_INDEX_PUBLISHER_BACKEND = os.getenv("SEARCH_INDEX_PUBLISHER_BACKEND")
```

cms/search/tests/test_wagtail_hooks.py
```diff
@@ -0,0 +1,96 @@
...
+# Methodology Page: Topics: TESTED
+# Release Calendar Page: No topics:
+# Index Page: No topics
+# Statistical Article: No topics
+
+
+class WagtailHooksTests(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.mock_request = wagtail.coreutils.get_dummy_request()
```
zerolab, line 36:
```suggestion
        cls.mock_request = get_dummy_request()
```

and the import `from wagtail.coreutils import get_dummy_request`

cms/search/publishers.py
```diff
@@ -0,0 +1,176 @@
...
+    """Publishes messages to Kafka for 'search-content-updated' (created or updated)
+    and 'search-content-deleted' (deleted) events, aligning with the StandardPayload
+    / ReleasePayload / content-deleted schema definitions.
+    """
+
+    def __init__(self) -> None:
+        # Read Kafka configs settings
+        self.kafka_server = settings.KAFKA_SERVER
+        self._channel_created_or_updated: Optional[str] = settings.KAFKA_CHANNEL_CREATED_OR_UPDATED
+        self._channel_deleted: Optional[str] = settings.KAFKA_CHANNEL_DELETED
```
zerolab, line 137: needs Django system check as per https://github.com/ONSdigital/dis-wagtail/pull/115#discussion_r1999871934

cms/search/wagtail_hooks.py
```diff
@@ -0,0 +1,33 @@
...
+    from wagtail.models import Page
+
+publisher = get_publisher()
+
+
+@hooks.register("after_publish_page")
+def page_published(request: "HttpRequest", page: "Page") -> None:
+    """This hook is called after a page is published, either via the Wagtail admin UI or programmatically.
+
+    Wagtail hook documentation: https://docs.wagtail.io/en/stable/reference/hooks.html#after-publish-page
```
zerolab, line 20: The Wagtail hooks are called only for actions via the UI.

Please use the corresponding signals:
- https://docs.wagtail.org/en/stable/reference/signals.html#page-published and 
- https://docs.wagtail.org/en/stable/reference/signals.html#page-unpublished
- and https://docs.djangoproject.com/en/5.1/ref/signals/#post-delete for page deletion

zerolab (changes requested):
Getting there. Adde a few additional thoughts

cms/articles/models.py
sanjeevz3009: Good question, now thinking about it we can just have a static, class-level value (rather than a dynamic, instance-level property that can be overridden or extended as we don't need that from my understanding).

I have updated this on the other models too now.

cms/search/publishers.py
sanjeevz3009: Wanted the time in UTC without explicitly showing the offset, but no other reason, we should be good with `+00:00`.

`Z` Removed!

cms/settings/base.py
sanjeevz3009: Done

cms/settings/base.py
sanjeevz3009: Done

cms/search/publishers.py
sanjeevz3009: Done!

cms/search/publishers.py
sanjeevz3009: Done!

cms/search/checks.py
```diff
@@ -0,0 +1,44 @@
+from django.conf import settings
+from django.core.checks import Error, register
+
+
+@register()
+def check_kafka_settings(app_configs, **kwargs) -> list[Error]:
+    """Check that required Kafka settings are present."""
+    errors = []
+
+    kafka_settings = [
```
zerolab, line 10: these should only kick in if the provider is "kafka"

cms/search/publishers.py
sanjeevz3009: Done

cms/search/checks.py
sanjeevz3009: Done!

cms/search/publishers.py
sanjeevz3009: Went with the Django system check!

cms/search/publishers.py
sanjeevz3009: Done! Set the fallback to `3.5.1` also.

docker-compose-dev.yml
sanjeevz3009: Good shout, added to the main `docker-compose.yml` file also.

cms/search/tests/test_wagtail_hooks.py
sanjeevz3009: Like this suggestion. Done!

cms/search/publishers.py
sanjeevz3009: Done! Added the `topic_ids` property to the `GenericTaxonomyMixin` for now.

cms/search/wagtail_hooks.py
sanjeevz3009: Removed the use of Wagtail hooks, using Signals now.

cms/search/checks.py
```diff
@@ -0,0 +1,88 @@
...
+    for setting, hint_value, missing_id, empty_id in kafka_settings:
+        if not hasattr(settings, setting):
+            errors.append(
+                Error(
+                    f"Missing required setting {setting}",
+                    hint=f"Add {setting} to your Django settings.",
+                    id=missing_id,
+                )
+            )
+        elif not getattr(settings, setting):
```
MaciekBaron, line 44: I think relying on the setting being falsy might cause issues later down the line, if for some reason we add a setting which could have a value of `False`, `0` or `""`.

Perhaps we could explicitly compare with `None` here? I assume the setting is missing if `os.getenv()` returns `None`.

MaciekBaron (changes requested):

cms/search/checks.py
sanjeevz3009: Yes, that's correct.

```
# Empty KAFKA_CHANNEL_DELETED
>>> os.getenv("KAFKA_CHANNEL_DELETED") is None
True
>>> 
```

I am happy to go with an explicit comparison in this case =)

Updated the line with `elif getattr(settings, setting) is None:`

cms/search/signals.py
```diff
@@ -0,0 +1,38 @@
...
+@receiver(page_unpublished)
+def on_page_unpublished(sender: "Page", instance: "Page", **kwargs: dict) -> None:  # pylint: disable=unused-argument
+    """Called whenever a Wagtail Page is unpublished (UI or code).
+    instance is the unpublished Page object.
+    """
+    if instance.__class__.__name__ not in SEARCH_INDEX_EXCLUDED_PAGE_TYPES:
+        publisher.publish_deleted(instance)
+
+
+@receiver(post_delete)
```
MaciekBaron, line 31: Add sender here
```suggestion
@receiver(post_delete, sender=Page)
```

cms/search/signals.py
```diff
@@ -0,0 +1,46 @@
+from django.apps import apps
```
zerolab, line 1: This should be `signal_handlers.py` for consistency

cms/search/publishers.py
zerolab: Adding to next steps

cms/search/tests/test_publishers.py
```diff
@@ -0,0 +1,384 @@
...
+    def setUpTestData(cls):
+        cls.included_factories = [
+            InformationPageFactory,
+            MethodologyPageFactory,
+            ReleaseCalendarPageFactory,
+            StatisticalArticlePageFactory,
+        ]
+
+        cls.index_page = IndexPageFactory(slug="custom-slug-1")
+        cls.included_factories.append(lambda: cls.index_page)
```
zerolab, line 54: @sanjeevz3009 please do the same as for https://github.com/ONSdigital/dis-wagtail/pull/115/files/9761f03988ca2eb248cc898fc58cc71307887d72#diff-c7dfda5eebe1d604ad2b46e2ce74f2daa029434e1736d470dd73541a35f9fc78

cms/search/signals.py
```diff
@@ -0,0 +1,46 @@
...
+@receiver(page_unpublished)
+def on_page_unpublished(sender: "Page", instance: "Page", **kwargs: dict) -> None:  # pylint: disable=unused-argument
+    """Called whenever a Wagtail Page is unpublished (UI or code).
+    instance is the unpublished Page object.
+    """
+    if instance.__class__.__name__ not in SEARCH_INDEX_EXCLUDED_PAGE_TYPES:
+        publisher.publish_deleted(instance)
+
+
+def on_page_deleted(sender: "Page", instance: "Page", **kwargs: dict) -> None:  # pylint: disable=unused-argument
```
zerolab, line 32:
```suggestion
@receiver(post_delete, sender=Page)
def on_page_deleted(sender: "Page", instance: "Page", **kwargs: dict) -> None:  # pylint: disable=unused-argument
```

then you don't need `connect_page_delete_signal`

zerolab: https://docs.djangoproject.com/en/5.1/topics/signals/#connecting-to-signals-sent-by-specific-senders

same for the other two receivers

zerolab (changes requested):
Almost there. A bit of tidy up for the signals and it should be good to go

cms/search/signals.py
zerolab:
```
@receiver(post_delete, sender=Page)
def on_page_deleted(sender: "Page", instance: "Page", **kwargs: dict) -> None:  # pylint: disable=unused-argument
    """Catches all subclass deletions of Wagtail's Page model.
    Only fires if the page is not in SEARCH_INDEX_EXCLUDED_PAGE_TYPES.
    """
    # Only proceed if `sender` is a subclass of Wagtail Page
    if instance.specific_class.__name__ not in SEARCH_INDEX_EXCLUDED_PAGE_TYPES:
        publisher.publish_deleted(instance)
```

does the job

sanjeevz3009: https://onswebsite.slack.com/archives/C07TZEU09FS/p1743002471280039

The core issue is that using a global `post_delete receiver` (i.e. without sender=...) has two side‐effects:

It intercepts deletions for `all models`, including `User`, which, breaks the `database‐router test` by sending the `M2M` deletes to the wrong `DB`.

When I try to narrow it with @receiver(post_delete, sender=Page), Django will only fire the signal when the exact sender is` Page`, not when `sender` is a `subclass` of `Page` like `HomePage`. Thus `page` deletion test fails.

Django will only call `on_page_deleted` if the literal sender in `post_delete.send(sender=..., ...)` is `Page`. Subclasses like` HomePage`, `ArticlePage`, etc. will not trigger it.

**Problem**: I want to fire `page deleted` logic for all `Wagtail Page subclasses` but not for every `model` in the project. Using a global `post_delete` with an issubclass(...) check changes Django’s internal DB routing for `User ` `(breaking the M2M‐delete test)`, while using `sender=Page` does not catch `HomePage` and other `subclasses`.

**Solution I found**: Dynamically connect `post_delete` to each concrete subclass of Page (and to Page itself). This way we skip subclasses nor inadvertently register for non‐Page models like User.

https://github.com/ONSdigital/dis-wagtail/pull/115/commits/d58de837170623f5774018d089580c5529a384fc

zerolab: Try the snippet I shared please

cms/search/tests/test_publishers.py
sanjeevz3009: Done!

cms/search/signals.py
sanjeevz3009: Done!

cms/search/signals.py
sanjeevz3009: I am so code blind.

I kept trying, only now when I copy pasted your snippet I realised I was doing `if instance.__class__.__name__ not in SEARCH_INDEX_EXCLUDED_PAGE_TYPES:` not `if instance.specific_class.__name__ not in SEARCH_INDEX_EXCLUDED_PAGE_TYPES:`.

Mistake was myself using `instance.__class__.__name__` and not using `instance.specific_class.__name__`.

Recievers and tests updated =)

This works for `post_delete`. Adding `sender=Page` to the other published and unpublished receivers doesn't let any events fire.

Thanks @zerolab

cms/search/checks.py
```diff
@@ -0,0 +1,90 @@
...
+    for setting, hint_value, missing_id, empty_id in kafka_settings:
+        if not hasattr(settings, setting):
+            errors.append(
+                Error(
+                    f"Missing required setting {setting}",
+                    hint=f"Add {setting} to your Django settings.",
+                    id=missing_id,
+                )
+            )
+        else:
```
MebinAbraham, line 44: Nit: Since you're performing both if and else checks without early returns, you could simplify the logic by flipping the structure - check for the positive case (hasattr) first to improve readability. For example:
```py
if hasattr(settings, setting):
    ...
else:
    ...
```

Additionally, consider whether this level of validation is necessary. You're currently checking if the value is either None or an empty string, but in practice, that only covers a narrow range of invalid values. For instance, setting the value to 1 would pass this check, even though it might not be a valid configuration.

If we're not enforcing strict type or format validation for each setting, it might be more pragmatic to simplify the condition to:
```py
if not getattr(settings, setting, None):
```
This would catch both missing and falsy values in a single, more concise check.

cms/search/checks.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # Get all Wagtail Page subclasses
+    page_models = get_page_models()
+
+    for model in page_models:
+        # Skip anything that appears in the exclusion list by class name
+        if model.__name__ in excluded_types:
+            continue
+
+        if not hasattr(model, "search_index_content_type"):
```
MebinAbraham, line 76: Nit: Use getattr and check for truthiness?

cms/search/publishers.py
```diff
@@ -0,0 +1,163 @@
...
+        """Constructs and returns a dictionary with release-specific fields."""
+        release_fields = {
+            "finalised": page.status in ["CONFIRMED", "PROVISIONAL"],
+            "cancelled": page.status == "CANCELLED",
+            "published": page.status == "PUBLISHED",
+            "date_changes": [],
+        }
+
+        # If page.changes_to_release_date exists, add the first item's reason & previous date
+        if hasattr(page, "changes_to_release_date") and page.changes_to_release_date:
```
MebinAbraham, line 104: Nit:
```suggestion
        if getattr(page, "changes_to_release_date", None):
```

cms/search/publishers.py
```diff
@@ -0,0 +1,163 @@
...
+            first_change = page.changes_to_release_date[0].value
+            release_fields["date_changes"].append(
+                {
+                    "change_notice": first_change.get("reason_for_change"),
+                    "previous_date": str(first_change.get("previous_date").isoformat())
+                    if first_change.get("previous_date")
+                    else None,
+                }
+            )
```
MebinAbraham, line 113: This is only adding the first change. There might be multiple changes to the release date. The previous date is mandatory so you don't need the `if` statement.
Need to update tests.
```suggestion
            release_fields["date_changes"] = [
                {
                    "change_notice": change.value.get("reason_for_change"),
                    "previous_date": change.value.get("previous_date").isoformat(),
                }
                for change in page.changes_to_release_date
            ]
```

cms/search/publishers.py
```diff
@@ -0,0 +1,163 @@
...
+        return release_fields
+
+
+class KafkaPublisher(BasePublisher):
+    """Publishes messages to Kafka for 'search-content-updated' (created or updated)
+    and 'search-content-deleted' (deleted) events, aligning with the StandardPayload
+    / ReleasePayload / content-deleted schema definitions.
+    """
```
MebinAbraham, line 121: Add a link to the spec?

MebinAbraham (commented):
I couldn't find any tests that explicitly verify the payload being sent for different page types. Have I missed them? If not, we should add unit tests to assert that the generated payload matches expectations for each page type.

Specifically, we should test:
- That the exclusion list is respected.
- That pages with differing payload structures (e.g. Release vs Bulletin) produce the correct payload format.

.github/workflows/ci.yml
```diff
@@ -97,6 +97,24 @@ jobs:
           - 5432:5432
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
 
+      broker:
+        image: apache/kafka:3.8.0
```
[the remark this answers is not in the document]
sanjeevz3009: Forgive my ignorance, but I can't seem to find a Kafka `3.5.1` image on here: https://hub.docker.com/r/apache/kafka/tags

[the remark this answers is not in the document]
zerolab: It would have to be the bitnami image: https://hub.docker.com/r/bitnami/kafka/tags?name=3.5.1

cms/search/tests/test_publishers.py
```diff
@@ -0,0 +1,380 @@
...
+
+    def setUp(self):
+        self.publisher = DummyPublisher(
+            channel_created_or_updated="dummy-channel-created",
+            channel_deleted="dummy-channel-deleted",
+        )
+
+    def test_publish_created_or_updated_calls_publish(self):
+        """Verify `publish_created_or_updated` calls `_publish` with the correct channel & message."""
+        with patch.object(self.publisher, "_publish", return_value=None) as mock_method:
```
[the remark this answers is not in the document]
sanjeevz3009: Referencing `self.publisher` inside the decorator causes `self` is not defined error.

I tried but this didn't work as it is as `self.publisher` isn’t defined until after the class is instantiated `(in setUp())`. By the time Python sees `@patch.object(...)` at class-definition time, `self.publisher` does not exist yet.

There are some other work arounds to this to make this a decorator such as patching the class instead of the instance or patching by module path.

I went with patching the class.

.github/workflows/ci.yml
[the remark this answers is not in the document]
sanjeevz3009: Done!

cms/search/checks.py
sanjeevz3009: Went with the latter!

cms/search/checks.py
sanjeevz3009: Done!

cms/search/publishers.py
sanjeevz3009: Done!

cms/search/publishers.py
sanjeevz3009: Done!

cms/search/publishers.py
sanjeevz3009: Done!

cms/search/tests/test_publishers.py
```diff
@@ -0,0 +1,380 @@
...
+        self.assertIn("date_changes", message)
+
+        self.assertEqual(len(message["date_changes"]), 1)
+
+        date_change = message["date_changes"][0]
+        self.assertEqual(date_change["change_notice"], page.changes_to_release_date[0].value["reason_for_change"])
+        self.assertEqual(
+            date_change["previous_date"],
+            page.changes_to_release_date[0].value["previous_date"].isoformat(),
+        )
```
MebinAbraham, line 188: Let's update this to test multiple, not just the first one.

sanjeevz3009: Done + added tests for testing topics (topic tagging) also.

MaciekBaron (approved):

zerolab (approved):

cms/search/tests/test_publishers.py
```diff
@@ -0,0 +1,425 @@
...
+    def setUp(self):
+        self.publisher = DummyPublisher(
+            channel_created_or_updated="dummy-channel-created",
+            channel_deleted="dummy-channel-deleted",
+        )
+
+        self.client.force_login(self.superuser)
+
+        self.root_page = Page.objects.get(id=1)
+        # self.root_topic = Topic.objects.root_topic()
```
kacperpONS, line 86:
```suggestion
```

kacperpONS (approved):
Good stuff, works for me!

cms/search/tests/test_publishers.py
zerolab: 🦅 👀 !

cms/search/checks.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    errors: list[Error] = []
+
+    # Fetch the excluded list (defaulting to empty tuple if not set)
+    excluded_types = getattr(settings, "SEARCH_INDEX_EXCLUDED_PAGE_TYPES", ())
+
+    # Get all Wagtail Page subclasses
+    page_models = get_page_models()
+
+    for model in page_models:
```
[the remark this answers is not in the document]
sanjeevz3009: Done!

cms/search/tests/test_publishers.py
sanjeevz3009: Great spot! I was going to remove that line + other needless lines of code like that in taxonomy too when I wrote the tests there, and ended up leaving it in last time.

Removed now!

kacperpONS (approved):

cms/search/publishers.py
```diff
@@ -0,0 +1,166 @@
...
+        # Common fields for StandardPayload (also part of ReleasePayload)
+        # The schema requires at minimum: uri, title, content_type
+        message = {
+            "uri": page.url_path,
+            "content_type": page.search_index_content_type,
+            "release_date": (
+                page.release_date.isoformat()
+                if page.search_index_content_type == "release" and getattr(page, "release_date", None)
+                else None
+            ),
```
MebinAbraham, line 83: 1. If the `release_date` is `None`, then you need to add the property `provisional_date` and populate that with the value from `release_date_text`. Currently, this is only done for `release` types therefore it should be done in `_construct_release_specific_fields`.
2. What implied this field is only for type release?. I can't see the spec suggesting this (we can resolve in a follow-up PR to get things moving, but I think we need to populate release date for articles etc - have asked for confirmation (lets follow up))

MebinAbraham (commented):
Few things to address

cms/articles/models.py
```diff
@@ -90,6 +90,7 @@ class StatisticalArticlePage(BundledPageMixin, RoutablePageMixin, BasePage):  #
 
     parent_page_types: ClassVar[list[str]] = ["ArticleSeriesPage"]
     subpage_types: ClassVar[list[str]] = []
+    search_index_content_type: ClassVar[str] = "bulletin"
```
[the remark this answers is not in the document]
zerolab: `statistical_article` is not in the https://github.com/ONSdigital/dp-search-data-extractor/blob/develop/specification.yml#L121-L145 list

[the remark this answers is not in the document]
MebinAbraham: Maybe I was being too eager as they have started making some changes, plus by the time we have an "integrated env" it would be supported. Happy to keep as is and come back given it's a two line change.

cms/search/publishers.py
```diff
@@ -0,0 +1,166 @@
...
+            "summary": get_text_for_indexing(force_str(page.summary)),
+            "title": page.title,
+            "topics": getattr(page, "topic_ids", []),
```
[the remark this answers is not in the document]
zerolab: Good catch! I could swear I tested this specifically when I re-reviewed since I flagged it in a previous comment 🙈

cms/search/publishers.py
zerolab: Another 🦅 👁️! 💪🏼 
Alas, I intepreted `release_date` as being relevant only to release calendar, but the spec is clear it is in the main corpus https://github.com/ONSdigital/dis-search-upstream-stub/blob/main/docs/contract/resource_metadata.yml#L58

MebinAbraham: I forgot we had the `release_date` field on the article, so that's where this should populate from for that page. Worth resolving here:
- Articles/Calendar Entry - `release_date` field
- Methodology - `publication_date` field

cms/articles/models.py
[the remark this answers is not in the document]
sanjeevz3009: Sounds good! Leaving it out. Got confused on this one!

cms/search/publishers.py
```diff
@@ -0,0 +1,166 @@
...
+    https://github.com/ONSdigital/dp-search-data-importer/blob/30fb507e90f2cf1974ec0ca43bb0466307e2f112/specification.yml#L186
+    contract: https://github.com/ONSdigital/dis-search-upstream-stub/blob/main/docs/contract/resource_metadata.yml
+    """
+
+    def __init__(self) -> None:
+        # Read Kafka configs settings
+        self.kafka_server = settings.KAFKA_SERVER
+        self.kafka_api_version = settings.KAFKA_API_VERSION
+        self._channel_created_or_updated: Optional[str] = settings.KAFKA_CHANNEL_CREATED_OR_UPDATED
+        self._channel_deleted: Optional[str] = settings.KAFKA_CHANNEL_DELETED
```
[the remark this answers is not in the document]
sanjeevz3009: @kacperpONS caught this (making the abstract methods into abstract properties) actually early on and mentioned it to me in the office, was going to get around to it _one day_. We were surprised how no one caught it =0

`values are from settings, there's no need to store them as object attributes`, love this.

Additionally, I took out the instance attributes from the constructor `self.kafka_server = settings.KAFKA_SERVER`, `self.kafka_api_version = settings.KAFKA_API_VERSION` as they are not used else where class.

Done now!

cms/search/publishers.py
```diff
@@ -0,0 +1,166 @@
...
+            "title": page.title,
+            "topics": getattr(page, "topic_ids", []),
+        }
+
+        # If it's a Release, we add the extra fields from ReleasePayload
+        if page.search_index_content_type == "release":
+            message.update(self._construct_release_specific_fields(page))
+        return message
+
+    def _construct_release_specific_fields(self, page: "Page") -> dict:
```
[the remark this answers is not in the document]
sanjeevz3009: Correct! I think @kacperpONS also mentioned this in person, and I thought of it too, but never got around to it

cms/search/publishers.py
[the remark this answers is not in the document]
sanjeevz3009: I thought I got this covered, hmmm, because Dan or someone raised this, but I can't remember. It was on my list of things to implement, and I ticked it off. Sad =/

I realised that when I added a property to return topics in the `GenericTaxonomyMixin`, which returns `topic_ids` for `ArticleSeriesPage`, it won't return the `topic_ids` for the `StatisticalArticlePage`. Missed this! Didn't think of it to be honest!

It's done now!

cms/search/publishers.py
sanjeevz3009: Yes, I went `spec` blind at this point, working on this haha. Correlated `release_date` with `release_calendar`, made an assumption, similar to @zerolab , even though `spec` says otherwise.

`Methodology - publication_date field (could create an alias release_date property to simplify logic)` @zerolab thankfully has already done this 2 months ago, so this works seamlessly!

I made it so we look for the `release_date` for all the pages we care about, if they don't have a `release_date` then we set `release_date` to `None`.

If the page type is `release` then it either has to have a `release_date` or `provisional_date`.

Done!

cms/search/publishers.py
```diff
@@ -97,12 +93,17 @@ def _construct_message_for_create_update(self, page: "Page") -> dict:
...
         release_fields = {
+            "provisional_date": None,
             "finalised": page.status in ["CONFIRMED", "PROVISIONAL"],
             "cancelled": page.status == "CANCELLED",
             "published": page.status == "PUBLISHED",
             "date_changes": [],
         }
 
+        # If we do NOT have release_date but we do have release_date_text, treat that as provisional_date
+        if not page.release_date and page.release_date_text:
```
MebinAbraham, line 104: Non-blocker, these are mutually exclusive already, there is validation to prevent this already so, you can just check if you have `release_date_text`.

cms/search/publishers.py
```diff
@@ -97,12 +93,17 @@ def _construct_message_for_create_update(self, page: "Page") -> dict:
     def _construct_release_specific_fields(page: "Page") -> dict:
         """Constructs and returns a dictionary with release-specific fields."""
         release_fields = {
+            "provisional_date": None,
```
MebinAbraham, line 96: This is a string in the api; however, let's omit this. Only add if needed as done below.

cms/search/publishers.py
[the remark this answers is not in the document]
MebinAbraham: Great, however, I didn't see a test to verify the behaviour, let's add one :+1:

cms/search/publishers.py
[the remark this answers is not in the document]
kacperpONS: GMTA 😌

AdamHawtin (approved):
LGTM 👍

cms/search/signal_handlers.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+@receiver(post_delete, sender=Page)
+def on_page_deleted(sender: "Page", instance: "Page", **kwargs: dict) -> None:  # pylint: disable=unused-argument
+    """Catches all subclass deletions of Wagtail's Page model.
+    Only fires if the page is not in SEARCH_INDEX_EXCLUDED_PAGE_TYPES.
+    """
+    # Only proceed if `sender` is a subclass of Wagtail Page
+    if instance.specific_class.__name__ not in SEARCH_INDEX_EXCLUDED_PAGE_TYPES:
+        publisher.publish_deleted(instance)
```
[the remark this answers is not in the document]
sanjeevz3009: Done!

cms/search/publishers.py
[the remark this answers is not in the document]
sanjeevz3009: Done!

cms/search/publishers.py
sanjeevz3009: Done!

cms/search/publishers.py
sanjeevz3009: Done!

cms/articles/models.py
```diff
@@ -269,3 +270,9 @@ def previous_version(self, request: "HttpRequest", version: int) -> "TemplateRes
         )
 
         return response
+
+    @property
+    def topic_ids(self) -> list[str]:
+        """Returns a list of topic IDs associated with the parent article series page."""
+        parent = self.get_parent().specific
+        return list(parent.topics.values_list("topic_id", flat=True)) if hasattr(parent, "topics") else []
```
zerolab, line 278:
```suggestion
        return list(self.get_parent().specific_deferred.topics.values_list("topic_id", flat=True))
```

1. `specific_deferred` will help not load the whole thing
2. we know the parent has a `topic` attribute

sanjeevz3009: Performance improvements and a lazy approach here are awesome.

Done.

MebinAbraham (approved):
👍

PR #118, opened 2025-03-14
Kubernetes Deployment

RealOrangeOne:

Code changes needed to build the project and deploy it into Kubernetes.

- Add Makefile commands for building and publishing contianer (in a separate file to remove need for everyone to have the AWS CLI)
- Fix location of elasticache credentials provider so it can be imported in settings
- Explicitly pass a region for DB auth
- Maintain DB connections for much longer, since opening them with IAM auth is expensive. This is configurable, for example to allow the public instance to close its write connection after a lack of use.
- Add liveness probe endpoint (for use with Kubernetes). This intentionally doesn't conform to [standard](https://github.com/ONSdigital/dp-standards/blob/main/HEALTH_CHECK_SPECIFICATION.md) - that will come later.

### How to review

Check the diff, and cross-reference against the latest commits which contain them. I'm hoping most of the changes are fairly self-explanatory though.

### Follow-up Actions


.docker/Makefile
```diff
@@ -0,0 +1,77 @@
+# The following is a temporary build script until a Concourse CI pipeline exists
+# It is just designed for building a production ready image using --target=web
+# To use these commands you will need AWS access and your profiles configured correctly
```
zerolab, line 3: q: (to confirm) is the idea this will be run manually?

also, suggestion: AWS cli is a dependency too, so perhaps add a note about it?

RealOrangeOne: This is very temporary, so I don't want to sink too much time into perfecting this.

For now, yes it will be run manually. The AWS CLI is only needed for deployment, which should happen outside the container anyway.

MaciekBaron (approved):
LGTM!

AdamHawtin (approved):
LGTM, I wasn't able to test the new make commands yet, but they look fine as a temporary manual solution.

PR #120, opened 2025-03-18
Add basic preview teams model

zerolab:

Adds the basic preview Team model, and its viewset/chooser. Relates to CMS-384 (62a1fe1) and CMS-289 (e9d5462)

### How to review

Check out, run migrations.
Set `ALLOW_TEAM_MANAGEMENT = True` in `local.py` - you should be able to manage teams.
To add users, you can do so via the Django shell.

### Follow-up Actions

Test will come in a separate PR


cms/teams/models.py
```diff
@@ -0,0 +1,37 @@
+from functools import cached_property
+from typing import ClassVar
+
+from django.db import models
+from wagtail.admin.utils import get_user_display_name
+from wagtail.search import index
+
+
+class Team(index.Indexed, models.Model):
+    identifier = models.CharField(max_length=255, unique=True)
```
MebinAbraham, line 10: When management is allowed, this error as it is `unique` true and create view doesn't allow creation of this identifier. So we need to add that this the create view or disable uniqueness for now.

cms/teams/viewsets.py
```diff
@@ -0,0 +1,97 @@
...
+        "created_at",
+        "updated_at",
+        "is_active",
+        "total_members",
+        "users",
+    ]
+
+    @property
+    def form_fields(self) -> list[str]:
+        # note: when ALLOW_TEAM_MANAGEMENT is removed, conver this to an actual property
```
MaciekBaron, line 56: Nit-pick: typo
```suggestion
        # note: when ALLOW_TEAM_MANAGEMENT is removed, convert this to an actual property
```

MaciekBaron (changes requested):

cms/teams/viewsets.py
zerolab: 23cef93 + `cached_property` for `get_teams_display`

MaciekBaron (approved):

MebinAbraham (approved):

sanjeevz3009 (approved):
LGTM! Tested everything locally, all seems to be working well!

cms/teams/models.py
```diff
@@ -0,0 +1,38 @@
...
+    name = models.CharField(max_length=255)
+    precedence = models.PositiveIntegerField(null=True)
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+    is_active = models.BooleanField(default=True)
```
kacperpONS, line 15: Question: what's the role of the `is_active` field? I can't find a reference to it in the attached cards. 
Also, wouldn't we want to make it changeable in the UI?

kacperpONS (commented):
LGTM, left a question

cms/teams/models.py
zerolab: Teams can be marked as inactive in Florence. @MebinAbraham can expand on this more

kacperpONS (approved):

cms/teams/models.py
MebinAbraham: Discussed and resolved with Kacper. If the user is `Enabled` in AWS Cognito, they are considered active.

sanjeevz3009 (approved):

PR #121, opened 2025-03-18
Update wagtail-tinytableblock to show the copy/paste buttons

zerolab:

This came up in a conversation with the UR team - roughly half of the POs use keyboard shortcuts, while the other half rely on editor buttons.

### How to review

`poetry install`, add a table block and see the new copy/paste buttons in the toolbar and the context menu. Note that Firefox prevents access to the clipboard, so they will see a notice to that effect

### Follow-up Actions

N/A

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2591,13 +2591,13 @@ files = [
 
 [[package]]
 name = "wagtail-tinytableblock"
-version = "0.2.1"
+version = "0.2.2"
 description = "A Wagtail StreamField block powered by TinyMCE and its table plugin"
 optional = false
 python-versions = ">=3.12"
 files = [
-    {file = "wagtail_tinytableblock-0.2.1-py3-none-any.whl", hash = "sha256:8cae541ac0aaeb33af17b37ab14d68cbe6aee432b24081bb313608839c9ffd59"},
-    {file = "wagtail_tinytableblock-0.2.1.tar.gz", hash = "sha256:bca9225e2e15ae6259a1f14ff63fcbea81880c1fad503137bf1fc9c77dc6516e"},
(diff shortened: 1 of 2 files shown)


MaciekBaron (approved):

sanjeevz3009:
Hey @zerolab I am testing this on chrome but I get the notice `Your browser doesn't support direct access to the clipboard. Please use the ⌘+X/C/V keyboard shortcuts instead.`? Maybe I am missing something?

zerolab:
Hi @sanjeevz3009 @kacperpONS 
These actions rely on the browser [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) which will be available depending on the browser

You can give that a try on https://www.tiny.cloud/docs/tinymce/latest/basic-example/
Also in Google Docs (https://docs.new) -> Edit menu -> Paste

Will have a look whether I can change these to simulate the user pressing "Ctrl/Cmd+V"

zerolab:
OK, so after further digging, I am afraid this is very much browser dependent - https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API

I would argue that it is better to have the paste button as it will at least give users a hint on what to do

zerolab:
One possible improvement I can think of is to disable the TinyMCE contextual menu so that the browser native one is present. This way users can right-click and use the native browser paste directive

zerolab:
now with browser contextual menu

Screenshot 2025-03-19 at 12 10 11

MaciekBaron (approved):

nehakerung (approved):
Right clicking, copy and paste works for both safari and chrome + it is the browser contextual menu that shows.
And as covered before clipboard doesn't work on all browsers (which was true for chrome but works on safari for me).
😎 so LGTM

sanjeevz3009 (approved):
Happy with the browser contextual menu option. Looks good!

PR #122, opened 2025-03-19
Add correction and notice block

MaciekBaron:

This relates to CMS-195. This adapts the correction and notice block from the alpha.

Currently the HTML markup has been imported as it was written in the alpha.

This change also adds the ability add additional tabs by specifying them in `additional_panel_tabs`, which is a list of tuples: the first element is a list of panels, and the second is the string used for the name of the tab.

### How to review

* Pull and checkout the branch
* Run migrations
* Create or edit an existing statistical article
* Go to the "Corrections and updates" tab
* In the Updates section, click the plus button to add a correction
* Publish your page
* View the page. You should see a correction box with a link to an old version.

### Follow-up Actions

Tests need to be added.


nehakerung (approved):
✅ LGTM - functionally tested it and works as intended.

cms/settings/base.py
```diff
@@ -345,8 +345,10 @@
 STATICFILES_DIRS = [
     # "static_compiled" is a folder used by the front-end tooling
     # to output compiled static assets.
+    # "static" is a folder used for any static assets that are not compiled.
     (PROJECT_DIR / "jinja2" / "assets"),
     (PROJECT_DIR / "static_compiled"),
+    (PROJECT_DIR / "core" / "static"),
```
RealOrangeOne, line 351: This shouldn't be necessary, as Django will use the app dirs finder to find the static files.

functional_tests/steps/statistical_article_page.py
```diff
@@ -89,3 +118,62 @@ def expand_footnotes(context: Context):
...
+    expect(context.page.get_by_text("Notice text")).to_be_hidden()
+
+
+@then("the user can edit the correction")
+def user_cannot_edit_the_correction(context: Context):
+    page = context.page
+    page.wait_for_timeout(500)  # added to allow JS to be ready
+    page.locator("#tab-label-corrections_and_notices").click()
+    expect(page.locator("#corrections-0-value-when")).to_be_editable()
+    page.wait_for_timeout(50)  # added to prevent flakiness
```
RealOrangeOne, line 169: Question: Why is a wait needed _after_ an assertion?

cms/core/tests/test_blocks.py
```diff
@@ -532,3 +536,43 @@ def test_render_block__ds_component_markup(self):
...
+        rendered = block.render(self.notice_data)
+
+        self.assertIn("1 January 2025", rendered)
+        self.assertIn("Notice text", rendered)
+
+
+class CorrectionBlockTestCase(TestCase):
+    def setUp(self):
+        self.series = ArticleSeriesPageFactory()
+        self.statistical_article = StatisticalArticlePageFactory(parent=self.series)
```
RealOrangeOne, line 560: Suggestion: These should be in `setUpTestData` so they're only generated once

cms/core/static/js/correction-block.js
RealOrangeOne, line 1: Suggestion: These should be in a subdirectory rather than in the root of `/static/js`.

Issue: These files should be excluded from the external environment using https://github.com/ONSdigital/dis-wagtail/blob/main/cms/core/whitenoise.py

cms/core/blocks/panels.py
```diff
@@ -37,3 +42,54 @@ class Meta:
...
+    when = blocks.DateTimeBlock()
+    text = blocks.RichTextBlock(features=settings.RICH_TEXT_BASIC)
+
+    class Meta:
+        abstract = True
+
+
+class CorrectionBlock(CorrectionOrNoticeBlock):
+    previous_version = PreviousVersionBlock(required=False)
+    frozen = blocks.BooleanBlock(required=False)
```
RealOrangeOne, line 61: Question: Should this have an explicit default of `false`?

functional_tests/features/statistical_article_page.feature
```diff
@@ -19,3 +19,58 @@ Feature: Statistical Article Page components
         Then the published statistical article page is displayed with the populated data
         And the published statistical article page has the added table
         And the user can expand the footnotes
+
+    Scenario: A CMS user can add a correction to a Statistical Article Page
```
AdamHawtin, line 23: How easy would it be to test some of these scenarios in unit tests, instead of in the functional tests?

Given that the functional tests are quite slow to run and are proving a bit flaky/tricky to develop/maintain, I think it would be beneficial to test as much of the logic as we can on the unit side and limit the functional tests to aspects which are better (or only) testable from the browser.

MaciekBaron: Normally unit tests should test a "unit" of code. In this case we would need to setup a page and test the output of a page with a given streamfield, which would turn more into an integration test.

Moreover, a lot of these tests evaluate the code on the admin side, i.e. is it possible to add a correction using the admin panel with JS loaded etc..

We can however later on add unit tests which cover the rendering output (which I purposefully omitted as Jane will be working on the frontend side of things).

functional_tests/steps/statistical_article_page.py
MaciekBaron: I know this looks strange, but there is some weird race condition here where the test *following* this test would fail occasionally if there was no wait here. I don't want to do cargo cult programming here but this seems to have fixed it.

cms/core/blocks/panels.py
MaciekBaron: Added default value.

cms/core/tests/test_blocks.py
MaciekBaron: Moved.

cms/settings/base.py
MaciekBaron: Thanks, removed.

cms/core/static/js/correction-block.js
```diff
@@ -0,0 +1,38 @@
...
+    const block = super.render(
+      placeholder,
+      prefix,
+      initialState,
+      initialError,
+    );
+
+    const parent = `[data-streamfield-child]:has([id^="${prefix}"])`;
+
+    if (initialState.frozen) {
```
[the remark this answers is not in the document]
MaciekBaron: Added additional validation for this.

cms/articles/signals.py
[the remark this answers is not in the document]
MaciekBaron: Renamed.

functional_tests/features/statistical_article_page.feature
AdamHawtin: Poor choice of words on my part, our Python are really a mix of unit and integration tests.

Fair points, I'm just getting concerned about the reliability of our functional tests, which should be addressed separately.

functional_tests/steps/statistical_article_page.py
AdamHawtin: Did you find that only occurs in this scenario? I wonder if we'd benefit from putting a very short wait like this in the after scenario hook, if this might potentially happen in other scenarios. Something to look at in a separate ticket.

AdamHawtin (approved):
LGTM 👍 Working as expected, acknowledging the styling is being addressed in follow-up.

RealOrangeOne (approved):

functional_tests/steps/statistical_article_page.py
MaciekBaron: Thanks Adam, I've added some extra context to the comment now.

AdamHawtin (approved):

sanjeevz3009 (commented):
Reviewed! Looks good!

PR #127, opened 2025-03-21
Use HTML text inputs in numeric mode

nimasmi:

This converts several numeric fields in the charts editor to use Django TextInput widgets, which render as `<input type="text">`, and then specify validation.

### How to review

1. Check out the branch.
2. Create an InformationPage.
3. Add a chart.
4. Add an annotation block (these inputs are used elsewhere, but this is the only place to see an integer input).
5. Observe the inputs, minus the unhelpful up/down increment buttons.
6. Try to enter bad data (a float in the x-position integer input, or some non-numeric text in the y-position float input).
7. View the inputs on a mobile device, and see an appropriate keyboard in use for integer fields (x-position), though not for float inputs (y-position).

### Follow-up Actions

Sort out the merge target. This is targeting `feature/streamfield-charts` for now, to keep the diff small.

### DIFF
--- a/cms/datavis/blocks/utils.py
+++ b/cms/datavis/blocks/utils.py
@@ -1,0 +1,24 @@
+from django.forms import TextInput
+from wagtail import blocks
+
+
+class TextInputCharBlock(blocks.IntegerBlock):
+    """A text input widget that only allows numeric input.
+
+    See https://design-system.service.gov.uk/components/text-input/#numbers
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.field.widget = TextInput(attrs={"inputmode": "numeric", "pattern": "[0-9]*"})
+
+
+class TextInputFloatBlock(blocks.FloatBlock):
+    """A text input widget that only allows decimal input.
+
+    See https://design-system.service.gov.uk/components/text-input/#numbers
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.field.widget = TextInput(attrs={"inputmode": "decimal", "pattern": "[0-9]*\\.?[0-9]*"})
--- a/cms/datavis/blocks/utils.py
+++ b/cms/datavis/blocks/utils.py
@@ -21,4 +21,5 @@ class TextInputFloatBlock(blocks.FloatBlock):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        self.field.widget = TextInput(attrs={"inputmode": "decimal", "pattern": "[0-9]*\\.?[0-9]*"})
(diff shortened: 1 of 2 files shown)


cms/datavis/blocks/utils.py
```diff
@@ -0,0 +1,25 @@
...
+class TextInputFloatBlock(blocks.FloatBlock):
+    """A text input widget that only allows decimal input.
+
+    See https://design-system.service.gov.uk/components/text-input/#numbers
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        # NB inputmode is not "decimal" as per the guidance linked in the docstring above
+        self.field.widget = TextInput(attrs={"inputmode": "text", "pattern": "[0-9]*\\.?[0-9]*"})
```
MaciekBaron, line 25: This might be nit-picky but the pattern allows to input just `.` (also numbers without leading or trailing zeroes).

How about something like this?

```suggestion
        self.field.widget = TextInput(attrs={"inputmode": "text", "pattern": "\\d*(\\.\\d+)?"})
```

(We could also use a raw string to avoid the double slashes: `r"\d*(\.\d+)?"`)

MaciekBaron (changes requested):

cms/datavis/blocks/utils.py
nimasmi: Thanks for calling me out on this. The integer input pattern was verbatim from the GDS blog post, but I think the float one was suggested by autocomplete in my code editor, and I admit I did not check it.

I spent some time this morning thinking about edge cases, and doing TDD, and went down a rabbit hole of whether it was right to permit inputs like `0123` (integer with leading zeroes), `-0` (negative zero), `.5` (shorthand decimal), or `3.`, i.e. things which are legitimate but not _well-formed_.

Then I saw that the GDS docs examples at https://design-system.service.gov.uk/components/text-input/#numbers and https://design-system.service.gov.uk/components/text-input/#asking-for-decimal-numbers don't have patterns at all, and remembered that GDS advises to disable browser validation with `novalidate`. Wagtail already does this, so I think we can remove these regex patterns, and my tests can burn.

For posterity, I settled on:

```python
INTEGER_PATTERN = r"^-?([1-9][0-9]*|0)$"
FLOAT_PATTERN = r"^-?([1-9][0-9]*|0)(\.[0-9]+)?$"
```

nimasmi: Novalidate docs ref: https://design-system.service.gov.uk/patterns/validation/#turn-off-html5-validation

MaciekBaron (approved):

nimasmi:
(I merged to the original target branch anyway, as I want to test that as one first before merging all.)

sanjeevz3009 (commented):
Reviewed!

PR #128, opened 2025-03-21
RBAC  - Set up Wagtail user groups

kacperpONS:

This PR sets up user groups and assigns permissions to them as defined in the [RBAC matrix](https://confluence.ons.gov.uk/display/NWP/RBAC). 
Also see the [dev copy of the RBAC matrix](https://confluence.ons.gov.uk/display/NWP/Dev+copy+of+RBAC+matrix) to see technical details in the **Dev Notes** column.

#### Unit tests:
Run the new unit tests:
```
poetry run coverage run ./manage.py test --parallel --settings=cms.settings.test cms.core.tests.test_permissions --keepdb
```

#### Group permissions
1. Run the migration and start the app.
1. Log in as a superuser and in Wagtail admin go to `Settings` -> `Groups` and verify the permissions match the specification in the RBAC matrix

#### Permission functionality
1. Create a `test_user`:
```
make createtestuser
```
2. Add the `test_user` to a group by either
- run `make make-admin`/ `make-officer`/ `make-viewer`
-  log in as a superuser and assign the `test_user` to a group
3. In an Incognito window log in as the `test_user` (password is the same as login) and confirm that the permissions work as expected

### Follow-up Actions

WIP: I'll create cards to track the permissions that are specified in the RBAC matrix but are blocked on features that aren't yet implemented.


cms/core/migrations/0006_update_user_groups.py
```diff
@@ -0,0 +1,192 @@
+from django.conf import settings
+from django.db import migrations
+from wagtail.models import PAGE_PERMISSION_CODENAMES, GroupCollectionPermission, GroupPagePermission
```
zerolab, line 3: please move this to the op functions as 

`GroupPagePermission = apps.get_model("wagtailcore.GroupPagePermission")` and `GroupCollectionPermission = apps.get_model("wagtailcore.GroupCollectionPermission")`

cms/core/migrations/0006_update_user_groups.py
```diff
@@ -0,0 +1,192 @@
...
+
+WAGTAIL_PERMISSION_TYPES = ["add", "change", "delete"]
+
+WAGTAIL_PAGE_PERMISSION_TYPES = [codename.replace("_page", "") for codename in PAGE_PERMISSION_CODENAMES]
```
zerolab, line 7: let's go for the layman's approach for this 

`WAGTAIL_PAGE_PERMISSION_TYPES = ["add", "change", "bulk_delete", "lock", "publish", "unlock"]`

cms/core/migrations/0006_update_user_groups.py
```diff
@@ -0,0 +1,192 @@
...
+    for model, app in snippet_classes_dict.items():
+        for permission in WAGTAIL_PERMISSION_TYPES:
+            assign_permission_to_group(
+                apps,
+                settings.PUBLISHING_ADMINS_GROUP_NAME,
+                permission_codename=f"{permission}_{model.lower()}",
+                app=app,
+                model=model,
+            )
+        if model in ("MainMenu", "FooterMenu"):
```
zerolab, line 99: suggestion: make `snippet_classes_dict` into a list of tuples
```python
snippet_classes = [
    # app, model, add publish permission
    ("core", "GlossaryTerm", False),
    # ...
]

for app,  model, add_publish_permisison in snippet_classes:
    ...
```

cms/core/migrations/0006_update_user_groups.py
```diff
@@ -0,0 +1,192 @@
...
+    )
+
+
+def create_image_permissions(apps):
+    """Allow Publishing Admins to add and edit images in the image library.
+    Note that the 'choose_image' permission isn't required
+    as it's assigned to all users which have the "Access the Wagtail admin" permission.
+    See: https://github.com/wagtail/wagtail/blob/125a749a9ab785757dd898e2f88bfb8fd3f65e11/wagtail/images/migrations/0023_add_choose_permissions.py#L23.
+    """
+    # TODO:  these shouldn't be imported directly and should be using historical models instead
```
zerolab, line 63: 💯

cms/core/tests/test_permissions.py
```diff
@@ -0,0 +1,110 @@
...
+    @classmethod
+    def create_user(cls, first_name, last_name, group_name):
+        # Create user
+        cls.user = User(first_name=first_name, last_name=last_name)
+        cls.user.save()
+
+        # Assign user to group
+        group = Group.objects.get(name=group_name)
+        group.user_set.add(cls.user)
+        group.save()
```
zerolab, line 23: You should not need `save` for adding m2m relations. That is only needed if you're creating a new instance (in this case a new Group instance)

https://docs.djangoproject.com/en/5.1/topics/db/examples/many_to_many/

cms/core/tests/test_permissions.py
```diff
@@ -0,0 +1,110 @@
...
+    def all_permissions_check_helper(self, app: str, model: str):
+        """A helper method to check if a user has all permissions for a model."""
+        for permission_type in WAGTAIL_PERMISSION_TYPES:
+            self.assertTrue(self.user.has_perm(f"{app}.{permission_type}_{model}"))
+
+
+class PublishingAdminPermissionsTestCase(PermissionsTestCase):
+    @classmethod
+    def setUpTestData(cls):
+        super().create_user(
```
zerolab, line 34:
```suggestion
        cls.create_user(
```

should do

cms/core/migrations/0006_update_user_groups.py
```diff
@@ -0,0 +1,192 @@
...
+
+def create_topic_page_featured_series_permission(apps):
+    """Create a custom permission that will be registered using a Wagtail hook,
+    given to the Publishing Admins group,
+    and used on the TopicPage's featured article series FieldPanel.
+    """
+    assign_permission_to_group(
+        apps,
+        group_name=settings.PUBLISHING_ADMINS_GROUP_NAME,
+        permission_codename="add_featured_article_series_on_topic_page",
```
zerolab, line 159: question: where does this requirement come from? The RBAC matrix mentions "headline figures", not "highlighted/featured series"

cms/core/migrations/0006_update_user_groups.py
```diff
@@ -0,0 +1,192 @@
...
+                model=bundle_class,
+            )
+
+    # View-only access for the Viewers group
+    assign_permission_to_group(apps, settings.VIEWERS_GROUP_NAME, "view_bundle", bundles_app, bundle_class)
+
+
+def create_page_permissions(apps):
+    """Allow Publishing Admins and Publishing Officers all permissions on the HomePage and its children."""
+    # TODO: these shouldn't be imported directly and should use historical models instead
```
zerolab, line 133: 💯

zerolab (commented):
Did a first pass and left a number of suggestions and questions

cms/core/migrations/0006_update_user_groups.py
kacperpONS: From [CMS-62 Topic page](https://jira.ons.gov.uk/browse/CMS-62):

> Automation logic:
> Headline figures - if no headline figures are added editorially, but there is a featured bulletin series, attempt to pull the headline figures from the latest bulletin in the featured series

So apart from the manual headline figure selection as proposed in [CMS-349 Headline figures spike](https://jira.ons.gov.uk/browse/CMS-349), if we want to only allow the Publishing Admins to set the headline figures, we'll need to restrict who can add the Featured series

cms/core/migrations/0006_update_user_groups.py
```diff
@@ -0,0 +1,192 @@
...
+    from django.contrib.auth.models import Group
+
+    from cms.home.models import HomePage
+
+    # HomePage = apps.get_model("home.HomePage")
+    # Group = apps.get_model("auth.Group")
+
+    home_page = HomePage.objects.first()
```
zerolab, line 141: I suggest giving access to the root page `Page.objects.get(depth=1)`. This way they will inherit the same permission for the Welsh copy of the tree

cms/core/migrations/0006_update_user_groups.py
kacperpONS: Had a chat with Dan and removed in [3dbd513](https://github.com/ONSdigital/dis-wagtail/pull/128/commits/3dbd513619abdce3feb87767ae19487560ad0bf0)

cms/core/migrations/0006_update_user_groups.py
```diff
@@ -0,0 +1,175 @@
...
+
+def create_user_groups(apps):
+    """Remove the existing Wagtail user groups and create new ones,
+    and give them  the "Access the Wagtail admin" permission.
+    """
+    Group = apps.get_model("auth.Group")
+
+    # Remove existing Wagtail groups
+    for wagtail_group in Group.objects.all():
+        wagtail_group.delete()
```
zerolab, line 36:
```suggestion
    Group.objects.all().delete()
```

cms/core/migrations/0006_update_user_groups.py
```diff
@@ -0,0 +1,175 @@
...
+    content_type = ContentType.objects.get_for_model(model_class)
+    permission, _was_created = Permission.objects.get_or_create(
+        content_type=content_type,
+        codename=permission_codename,
+        defaults={"name": f"Can {' '.join(permission_codename.split('_'))}"},
+    )
+    group = Group.objects.get(name=group_name)
+
+    group.permissions.add(permission)
+    group.save()
```
zerolab, line 25:
```suggestion
```

should be enough

cms/core/tests/test_permissions.py
```diff
@@ -0,0 +1,102 @@
...
+
+WAGTAIL_PAGE_PERMISSION_TYPES = ["add", "change", "bulk_delete", "lock", "publish", "unlock"]
+
+
+class PermissionsTestCase(TestCase):
+    @classmethod
+    def create_user(cls, first_name, last_name, group_name):
+        # Create user
+        cls.user = User(first_name=first_name, last_name=last_name)
+        cls.user.save()
```
zerolab, line 17: q: why not use `UserFactory`?
```python
cls.user = UserFactory(first_name=first_name, last_name=last_name)
```

zerolab (changes requested):
A couple of notes to start with. Will test is locally and do a full review pass

cms/core/tests/test_permissions.py
kacperpONS: Omission from when I was first writing the tests, thanks for spotting this

cms/core/migrations/0006_update_user_groups.py
```diff
@@ -0,0 +1,182 @@
...
+            permission=Permission.objects.get(content_type__app_label="wagtailimages", codename="choose_image"),
+        )
+
+
+def create_snippet_permissions(apps):
+    """Allow Publishing Admins to create, edit and delete snippets."""
+    snippet_classes = [
+        # app, model, add publish permission
+        ("core", "GlossaryTerm", False),
+        ("core", "ContactDetails", False),
```
zerolab, line 88: this is missing:

 Object | Role | Permissions
--------|------|--------------
 Glossary terms | PO | add
 Glossary terms | PA | add, change
 Team | PO | view
 Team | PA | view

cms/core/migrations/0006_update_user_groups.py
```diff
@@ -0,0 +1,182 @@
...
+
+def create_reporting_permissions(apps):
+    """Allow Publishing Admins to view the log entry reports."""
+    # The "Reports" section and everything under it
+    assign_permission_to_group(
+        apps, settings.PUBLISHING_ADMINS_GROUP_NAME, "view_logentry", "wagtailcore", "PageLogEntry"
+    )
+
+
+def create_image_permissions(apps):
```
zerolab, line 54: Needs the same for Document

cms/core/migrations/0006_update_user_groups.py
```diff
@@ -0,0 +1,182 @@
...
+            assign_permission_to_group(
+                apps,
+                group_name=group_name,
+                permission_codename=f"{permission}_{bundle_class.lower()}",
+                app=bundles_app,
+                model=bundle_class,
+            )
+
+    # View-only access for the Viewers group
+    assign_permission_to_group(apps, settings.VIEWERS_GROUP_NAME, "view_bundle", bundles_app, bundle_class)
```
zerolab, line 130: can you add it to the other two roles please?

kacperpONS:
In the above commit I've added some convenient Makefile commands for:
- creating a `test_user`
- removing them from all groups they belong to 
- adding them to the Publishing Admins, Publishing Officers and Viewers groups

These are not meant to be merged and only used to speed up manual testing of this PR.

cms/core/migrations/0006_update_user_groups.py
kacperpONS: Regarding the first two: 
- The matrix doesn't mention the PO being able to add a Term (as in create one), only "Add to Content" and this is handled by the Page permissions
- PA has both `add` and change `change` as they're in the `WAGTAIL_PERMISSION_TYPES` list 

With the latter two, me and Mebin agreed to leave that as the next steps, but since the Teams PR has gone in in the meantime I'm happy to add that here.

cms/core/migrations/0006_update_user_groups.py
kacperpONS: Sure, but I'm not entirely clear why we'd need that as the `add` and `change` permissions imply that they can view it (and looking at the `Bundles` page in the UI the PA and PO still have the `Inspect` option for each Bundle).
Is this just for the sake of being explicit?

cms/core/migrations/0006_update_user_groups.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def update_user_groups(apps, schema_editor):
+    """Create user groups for the CMS."""
+    create_user_groups(apps)
+    create_page_permissions(apps)
+    create_image_permissions(apps)
+    create_snippet_permissions(apps)
+    create_bundle_permissions(apps)
+    create_reporting_permissions(apps)
```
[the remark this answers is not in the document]
kacperpONS: Nice one, I'll add these to the matrix

cms/core/migrations/0006_update_user_groups.py
zerolab: >  Is this just for the sake of being explicit?

Yes

cms/core/migrations/0006_update_user_groups.py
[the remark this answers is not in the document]
sanjeevz3009: Piggybacking of @zerolab. During the conversations I had with @MebinAbraham, there were couple of options on the table such as:

- Populating the nav programmatically.
- Allowing PO and or PA to go ahead and populate the nav using the UI and then disabling access to configure the nav later on.
- Or we devs using the UI to populate the nav and then disabling it once the nav have been populated and configured.

FYI @kacperpONS

cms/core/tests/test_permissions.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def test_publishing_officer_can_manage_pages(self):
+        """Check that the Publishing Officer can create and edit the root page and its children."""
+        for permission_type in ("add", "change"):
+            self.page_permission_check_helper(permission_type)
+
+    def test_publishing_officer_can_choose_images(self):
+        """Check that the Publishing Officer can choose images on pages."""
+        self.image_collection_permission_check_helper("choose")
+
```
sanjeevz3009, line 123: You're currently testing to check image collection permissions, but not checking for document collection permissions.

- `Publishing Admin` gets to add, change, delete + choose on both images and documents.
- `Publishing Officer` gets to choose both images and documents.

Suggestion here is to:
- Tests for `choose_document` for both `Publishing Admin` and `Publishing Officer`.
- Tests for `add_document`, `change_document`, `delete_document` for `Publishing Admin`.

kacperpONS: Yup, tests on the way, just finishing the UI ones now 🫡

cms/core/tests/test_permissions.py
```diff
@@ -0,0 +1,134 @@
...
+
```
sanjeevz3009, line 155: I think it would be good to add some **negative tests** as well. This way we can check that certain user groups don't have specific permissions.

Such as:

- `Publishing Officer` does not have `bulk_delete`, `lock`, `unlock` and `publish` page.
- Viewer only has `view_bundle` and no other such as `add`, `change`, and `delete` on bundles.

Writing some negative tests here will be really good + there may be other cases I haven't mentioned that should be tested too. So good to think about some other negative test cases as well.

cms/core/tests/test_permissions.py
```diff
@@ -0,0 +1,165 @@
...
+    def test_publishing_officer_can_choose_images(self):
+        """Check that the Publishing Officer can choose images on pages."""
+        self.collection_permission_check_helper("choose", "image")
+
+    def test_publishing_officer_can_choose_documents(self):
+        """Check that the Publishing Officer can choose documents on pages."""
+        self.collection_permission_check_helper("choose", "document")
+
+
+class ViewerPermissionsTestCase(PermissionsTestCase):
```
sanjeevz3009, line 156: Testing for Snippet permissions would be nice too.

Currently, I think you're assigning snippet permissions `add`, `change`, `delete` and `publish` to `Publishing Admins`?

So it would be good to confirm that `Publishing Officers` and `Viewers` lack snippet permissions.

Some negative tests around this will be awesome.

sanjeevz3009 (commented):
Good work everywhere =) Added some comments around improving our test coverage here.

cms/core/tests/test_permissions.py
kacperpONS: I get where you're coming from, but this kinda goes agains how permissions work in Django, where we start of with a "blank" group and then following the principle of least privilege only assign the permissions that the role explicitly needs.
Question is how meticulous we want to be here because if we go with this you could argue that you need the negative tests for _all_ the permissions that the PAs and POs have but the Viewers don't. 
Happy to add this, let's wait for other's input.

sanjeevz3009: I understand your point, happy to wait and see what others think. @zerolab @MebinAbraham @AdamHawtin. Pinged some devs here.

I came across this same dilemma during testing the main menu, as how many different scenarios would I want to test, at the end I realised I can have unit tests checking 24 different combinations but at the end settled on 20 tests/ combinations as being pragmatic, and having the balance that if one or some of the 20 tests fails, then that's enough to know there is something going wrong.

Reflecting on that, I could have settled on 10 - 15 tests, too, for the main menu unit tests.

On the other hand, for taxonomy work, I went all in as it was crucial that I test as much as possible.

I think there is a balance that needs to be found for this particular situation, too =)

So I agree with you, do we need this? If so, could we settle with `x` amount of tests and that shall suffice even though we can have `y` amount.

zerolab: I think in this case we do want to be more comprehensive. if we change permissions via code in the future, that would be flagged, so updating the tests would be an explicit action.

Saying that, I suggest we do this as a follow-up PR immediately after this one as the scope keeps balooning

sanjeevz3009: Thanks @zerolab 

On that note:

@kacperpONS Can we please write a small JIRA card (Task) that would do for this and put it in this thread for the follow-up? Thanks!

MebinAbraham: Happy to follow up as Dan suggested to help move things forward. When it comes to permissions, though, we want to be thorough. From a security standpoint, negative testing is essential. It's far better for someone to be temporarily blocked from doing something they are authorised to do than for someone to be able to do something they aren’t authorised to do.

zerolab: 💯

cms/core/migrations/0006_update_user_groups.py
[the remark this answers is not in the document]
MebinAbraham: In the interest of moving things on, let's keep nav perms as is. We will revisit at a later date. Likely, we will hide from UI (even if they technically can manage via URLs) / add a warning panel that these are not synced across systems.

Makefile
```diff
@@ -161,6 +161,30 @@ migrate: ## Apply the database migrations
...
+
+.PHONY: make-officer
+make-officer: clear-groups
+	poetry run python ./manage.py shell -c "from cms.users.models import User; user=User.objects.filter(username='test_user').first(); from django.conf import settings; from django.contrib.auth.models import Group; Group.objects.get(name=settings.PUBLISHING_OFFICERS_GROUP_NAME).user_set.add(user);"
+
+.PHONY: make-viewer
+make-viewer: clear-groups
+	poetry run python ./manage.py shell -c "from cms.users.models import User; user=User.objects.filter(username='test_user').first(); from django.conf import settings; from django.contrib.auth.models import Group; Group.objects.get(name=settings.VIEWERS_GROUP_NAME).user_set.add(user);"
+
+#####
```
MebinAbraham, line 186: Reminder to revert before merge.

cms/core/tests/test_permissions.py
kacperpONS: https://github.com/ONSdigital/dis-wagtail/pull/128#discussion_r2026761530

cms/core/tests/test_permissions.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def test_publishing_admin_can_manage_contact_details(self):
+        self.all_permissions_check_helper("core", "contactdetails")
+
+    def test_publishing_admin_can_manage_main_menu(self):
+        self.all_permissions_check_helper("navigation", "mainmenu")
+
+    def test_publishing_admin_can_manage_footer_menu(self):
+        self.all_permissions_check_helper("navigation", "footermenu")
+
```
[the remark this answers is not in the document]
kacperpONS: [f49d5fc](https://github.com/ONSdigital/dis-wagtail/pull/128/commits/f49d5fcbfd4fddc8a046dd7713a774513dbdc50b)

MebinAbraham (approved):
👍

cms/core/tests/test_permissions.py
sanjeevz3009: @kacperpONS Has this been addressed yet?

kacperpONS: Yup, first after this comment: [f49d5fc](https://github.com/ONSdigital/dis-wagtail/pull/128/commits/f49d5fcbfd4fddc8a046dd7713a774513dbdc50b)

zerolab (approved):

sanjeevz3009 (approved):
LGTM!

PR #134, opened 2025-03-27
Sort corrections and check date

MaciekBaron:

This improves #122 by sorting the corrections when saving, so that users do not get confused when adding new corrections. This is an outcome of a conversation I had with Ade who expressed his concerns about this.

### How to review

1. Pull branch
2. Create a statistical article page
3. Publish the page
4. Add a correction
5. Publish the page
6. Add a new correction below the existing one
7. Save a draft
8. Notice that the new correction is now at the top

### Follow-up Actions

Review with the QA team to ensure this approach is more user-friendly.

### DIFF
--- a/cms/core/forms.py
+++ b/cms/core/forms.py
@@ -1,12 +1,13 @@
 from django.forms import ValidationError
+from wagtail.blocks.stream_block import StreamValue
 from wagtail.models import PageLogEntry
 
 from cms.taxonomy.forms import DeduplicateTopicsAdminForm
 
 
 class PageWithCorrectionsAdminForm(DeduplicateTopicsAdminForm):
-    def clean_corrections(self) -> list:  # noqa: C901
-        corrections: list = self.cleaned_data["corrections"]
+    def clean_corrections(self) -> StreamValue:  # noqa: C901
+        corrections: StreamValue = self.cleaned_data["corrections"]
 
         if self.instance.pk is None:
             if corrections:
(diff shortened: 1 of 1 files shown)


cms/jinja2/templates/pages/statistical_article_page.html
```diff
@@ -83,13 +83,17 @@ <h1 class="ons-u-fs-3xl common-header__heading">
...
+                            {% include_block correction %}
+                        {% endfor %}
+                    </div>
                 {% endif %}
                 {% if page.notices %}
                     <h3>{{ _("Notices") }}</h3>
-                    {{ page.notices }}
+                    <div id="notices">
+                        {{ page.notices }}
+                    </div>
```
zerolab, line 96: random thought: considering the article body has sections with editorial titles which the get slugified and used as IDs for the table of contents, there is a chance that one of the sections may be "Notices" so we'd have clashing IDs. So, a) could this be more general b) is it needed at all? Or are you pre-empting the DS component?

sanjeevz3009 (approved):
Tested everything locally, looks good!

cms/jinja2/templates/pages/statistical_article_page.html
MaciekBaron: To be honest I only did this to simplify the locator in the test:

https://github.com/ONSdigital/dis-wagtail/blob/7925f572d9e1c4080b26bb312a998cf794356b38/functional_tests/steps/statistical_article_page.py#L178-L180

...but I see your point. I could either prefix the ID with something like `"cn_"` or rework the locator so that the id here can be removed. What do you think?

zerolab (approved):
Other than the id thought, it LGTM.
Even that is non-blocking per se

cms/jinja2/templates/pages/statistical_article_page.html
zerolab: let's leave as is for now

PR #135, opened 2025-03-28
Audit logging improvements

RealOrangeOne:

This makes a few changes to the logging configuration:

- Add log events for login, logout and failed login
- Change log format to JSON, for easier ingesting and parsing
- Ensure `gunicorn` logs as JSON
- Add `django-xff` to allow correctly identifying a client's IP address (mostly unconfigured for now, but can be configured through environment variables as needed).

See also https://jira.ons.gov.uk/browse/CMS-425 and https://jira.ons.gov.uk/browse/CMS-426

### How to review

Python's `assertLogs` hooks in before formatters, so it's not possible to easily test the formats.

### Follow-up Actions

The log formats are close, but don't necessarily follow the [logging standards](https://github.com/ONSdigital/dp-standards/blob/main/LOGGING_STANDARDS.md). That should be reviewed later.


cms/users/signal_handlers.py
```diff
@@ -0,0 +1,53 @@
...
+            "email": user.email,
+            "event": "user_logged_in",
+            "ip_address": get_client_ip(request),
+            "user_agent": request.headers.get("User-Agent"),
+        },
+    )
+
+
+@receiver(user_logged_out)
+def audit_user_logged_out(sender: Any, request: HttpRequest, user: User, **kwargs: Any) -> None:  # pylint: disable=unused-argument
```
MaciekBaron, line 30: `user` can be `None` here if the user was not authenticated before logging out (in other words, this signal is still sent if a user attempted a logout without first logging in). This will throw an exception in the code below.

You can verify this by writing a test which attempts a logout without a prior login. Perhaps we could also add a test to cover this?

https://docs.djangoproject.com/en/5.1/ref/contrib/auth/#django.contrib.auth.signals.user_logged_out

MaciekBaron (changes requested):

cms/users/signal_handlers.py
RealOrangeOne: Great catch, I missed that. I'll fix the types and add a test!

MaciekBaron (approved):
LGTM now, thanks!

cms/users/signal_handlers.py
```diff
@@ -0,0 +1,57 @@
...
+
+@receiver(user_logged_in)
+def audit_user_logged_in(sender: Any, request: HttpRequest, user: User, **kwargs: Any) -> None:  # pylint: disable=unused-argument
+    logger.info(
+        "User logged in",
+        extra={
+            "user_id": user.id,
+            "email": user.email,
+            "event": "user_logged_in",
+            "ip_address": get_client_ip(request),
```
MebinAbraham, line 23: As per the [dissemination guidance](https://github.com/ONSdigital/dp-standards/blob/main/LOGGING_STANDARDS.md#sensitive-information), they do not want apps to log ip addresses. Where has this requirement come from?

MebinAbraham (changes requested):

cms/core/utils.py
```diff
@@ -43,3 +43,11 @@ def get_formatted_pages_list(
             }
         data.append(datum)
     return data
+
+
+def get_client_ip(request: "HttpRequest") -> str | None:
```
sanjeevz3009, line 48: I don't think we should be logging `IP addresses`: https://github.com/ONSdigital/dp-standards/blob/main/LOGGING_STANDARDS.md#sensitive-information.

This functionality may need to be removed, and then this test would not be needed.

@MebinAbraham

sanjeevz3009 (commented):
Did an initial round of review. A minor comment written.

cms/core/utils.py
sanjeevz3009: Realised @MebinAbraham  pointed this out too https://github.com/ONSdigital/dis-wagtail/pull/135#discussion_r2022481984

cms/core/logs.py
```diff
@@ -0,0 +1,51 @@
...
+            "method": record_args["m"],
+            "scheme": record_args["{wsgi.url_scheme}e"],
+            "host": record_args["{host}i"],
+            "path": record_args["U"],
+            "query": record_args["q"],
+            "status_code": record_args["s"],
+            "ended_at": response_time,
+            "duration": record_args["D"] * 1000,
+            "response_content_length": record_args["B"],
+        }
```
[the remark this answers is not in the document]
RealOrangeOne: The keys of the dict are some level of documentation. There's a link to the gunicorn docs above, but I'll move it closer so it's obvious what it's for

cms/users/signal_handlers.py
RealOrangeOne: I was told explicitly to log IP address - specifically for auth events. I'll raise it again and double check.

RealOrangeOne: It's a requirement, especially useful for auditing events. I've opened a PR to the standards so it's more widely visible. https://github.com/ONSdigital/dp-standards/pull/17. 

I think proper conclusion of that PR blocks this one.

cms/core/utils.py
sanjeevz3009: Resolving based on: https://github.com/ONSdigital/dis-wagtail/pull/135#discussion_r2024503727

zerolab (approved):

sanjeevz3009 (approved):
Happy with this. Approving on the basis: https://github.com/ONSdigital/dis-wagtail/pull/135#discussion_r2024503727

cms/users/signal_handlers.py
MebinAbraham: We’re unlikely to see progress on this any time soon, as any changes here would require a clear and tangible justification due to the significant governance work involved, particularly around DPIA and GDPR compliance.

In the interest of moving this on, I suggest we revert the IP logging addition for now so the merge can proceed. If it's later agreed that IP logging is acceptable, we can revisit and implement it with the necessary approvals.

RealOrangeOne: Sounds sensible to me

RealOrangeOne: Removed in https://github.com/ONSdigital/dis-wagtail/pull/135/commits/1141a66e614f7c83f71a1401e156a1cde2f146e0. I'll open it as a separate PR once this is merged

kacperpONS (approved):

sanjeevz3009 (approved):

MebinAbraham (approved):

sanjeevz3009 (approved):

sanjeevz3009:
@RealOrangeOne Squash and merge this away as the UI tests are being so flaky, I must have re-run them 2 billion times by now, and it's not passing.

PR #138, opened 2025-04-01
Add Custom Date Formatting Filter For Jinja Templates

sanjeevz3009:

This PR is to update the time format so it is consistent with the ONS style guide. Currently, Wagtail displays a 12-hour format `12.30a.m.`. This PR overrides the `TimeFormat` class's `a` method, which is inherited by the `DateFormat` class, so we can remove the `period` between and at the end of `a.m.`, but keep the 12-hour format. E.g. `12.30am`.

Loom: WILL BE ADDED HERE.

### How to review

- Spin up the local dev web application per usual.
- Test out the following pages:
    - `Statistical Article Page` to see if the override is working as expected.
    - `Release Calendar Page` with various statuses.
- Run the UI and integration tests.

> [!NOTE]
> @sanjeevz3009 Will take over this PR, address any comments, add UI tests and clean up.

### Follow-up Actions


cms/core/custom_dateformat.py
```diff
@@ -0,0 +1,22 @@
...
+class NewDateFormat(DateFormat):
+    def a(self) -> str:
+        """Am or pm."""
+        # To fix error: Item "date" of "date | time" has no attribute "hour"
+        if isinstance(self.data, datetime | time):
+            return "pm" if self.data.hour > PM_TIME else "am"
+        return ""
+
+
+def new_date_format(value: datetime, format_string: str) -> str:
```
zerolab, line 20:
```suggestion
def ons_date_format(value: datetime, format_string: str) -> str:
```

cms/core/custom_dateformat.py
```diff
@@ -0,0 +1,22 @@
...
+from datetime import datetime, time
+
+from django.utils.dateformat import DateFormat
+
+# Custom date format class that overrides the default date format to use am/pm instead of a.m./p.m.
+
+PM_TIME = 11
+
+
+class NewDateFormat(DateFormat):
```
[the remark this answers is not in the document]
nehakerung: Updated

cms/core/custom_dateformat.py
nehakerung: Updated

cms/core/jinja2tags.py
```diff
@@ -33,5 +33,7 @@ def __init__(self, environment: "Environment"):
                 "social_text": social_text,
                 "social_image": social_image,
                 "setAttributes": set_attributes_filter,
+                # define a filter to use custom_date_format her
+                "custom_date": custom_date_format,
```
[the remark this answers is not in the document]
nehakerung: I used `ons_date_format_filter` instead per comment below

cms/core/templatetags/util_tags.py
```diff
@@ -48,3 +50,12 @@ def set_attributes_filter(attributes: dict, new_attributes: dict) -> dict:
     """
     attributes.update(new_attributes)
     return attributes
+
+
+@register.filter(name="custom_date")
+def custom_date_format(value: datetime, format_string: str) -> str:
```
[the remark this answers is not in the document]
nehakerung: I used `ons_date_format_filter` instead the suggested does not overlap with ons_date_format from [cms/core/custom_dateformat.py](https://github.com/ONSdigital/dis-wagtail/pull/138/files/1a35c1abdb8e8955eec5f07d5a921314f191952f#diff-2bb67f1aa413ae2c9b739a84f2c402f141abbb8c81178894ba62929053e7960b)

cms/core/templatetags/util_tags.py
```diff
@@ -48,3 +50,12 @@ def set_attributes_filter(attributes: dict, new_attributes: dict) -> dict:
...
+    if not value:
+        return ""
+    if format_string == "DATETIME_FORMAT":
+        format_string = "j F Y g:ia"
```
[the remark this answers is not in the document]
nehakerung: Of course 🤦‍♀️  - removed [9ec6031](https://github.com/ONSdigital/dis-wagtail/pull/138/commits/9ec6031bada5b83d46668a4aff10b3daad8120d4)

cms/jinja2/templates/components/streamfield/correction_block.html
```diff
@@ -1,7 +1,7 @@
 <div class="ons-corrections-notices__item">
     <div class="ons-corrections-notices__content-meta">
         <h3 class="ons-u-fs-r--b ons-corrections-notices__content-heading">{{ _("Correction") }}</h3>
-        <time class="ons-corrections-notices__content-date">{{ value.when|date("DATETIME_FORMAT") }}</time>
+        <time class="ons-corrections-notices__content-date">{{ value.when|custom_date("DATETIME_FORMAT") }}</time>
```
zerolab, line 4: `custom_date` should now be `ons_date` everywhere ;)

nehakerung: 😭 noticed - am fixing it!

nehakerung: 🤦‍♀️ made the changes now

functional_tests/features/release_page.feature
```diff
@@ -14,3 +16,17 @@ Feature: CMS users can draft, edit, and publish release pages
...
+        And the user clicks "Publish"
+        And the user clicks "View Live" on the publish confirmation banner
+        Then the datetime is displayed with "am"
+    
+    Scenario: A CMS user inputs PM datetime on a release calendar page and "pm" is displayed
+        When the user creates and populates a release calendar page
+        And the user inputs an PM datetime
+        And the user clicks "Publish"
+        And the user clicks "View Live" on the publish confirmation banner
+        Then the datetime is displayed with "pm"
```
MaciekBaron, line 32: Nit-pick: Missing blank line at end of file

cms/core/custom_date_format.py
```diff
@@ -0,0 +1,33 @@
...
+    PM_CUTOFF_HOUR = 11
+
+    def a(self) -> str:
+        if isinstance(self.data, datetime):
+            return "pm" if self.data.hour > self.PM_CUTOFF_HOUR else "am"
+        return ""
+
+
+@cache
+def _resolve_format(fmt: str) -> str:
```
zerolab, line 25: get_format is already caching things:
https://github.com/django/django/blob/main/django/utils/formats.py#L16 and https://github.com/django/django/blob/main/django/utils/formats.py#L116C16-L116C29

I would be curiouse to understand why this needs the additional caching. For things like this, before/after performance profiling data would be great to have

sanjeevz3009: Ooo, `get_format is already caching things` didn't end up spotting this.

I tried to do some performance profiling, and this was the result.

```
>>> import timeit
>>> from functools import cache
>>> from django.utils.formats import get_format
>>> 
>>> @cache
... def cached(fmt):
...     return get_format(fmt)
... 
>>> def bench(stmt):
...     return timeit.timeit(stmt, number=200_000, globals=globals())
... 
>>> print("get_format direct:", bench("get_format('DATETIME_FORMAT')"))
print("two-layer cache  :", bench("cached('DATETIME_FORMAT')"))
get_format direct: 1.12417606300005
>>> print("two-layer cache  :", bench("cached('DATETIME_FORMAT')"))
two-layer cache  : 0.017881286999909207
>>> 
```

Results:
```
get_format direct: 1.12417606300005
two-layer cache  : 0.017881286999909207
```
56x faster?

Unless I did this wrong.

sanjeevz3009: Again, I don't think this will be some crazy improvement, and all of these are small differences. So happy to remove the caching here =)

sanjeevz3009: Had a nice dig deep dive session with @MebinAbraham around this, and we realised the `56x` performance improvement was because I wasn't taking `locale` into account, and this would break if we had a drastically different locale.

There are some considerations and other things we can do here, but for now, we decided to remove it =)

MaciekBaron (approved):
LGTM!

zerolab (approved):

functional_tests/steps/release_page.py
```diff
@@ -114,3 +114,30 @@ def check_selected_datasets_are_displayed(context: Context):
...
+
+
+@when("the user inputs a PM datetime")
+def add_pm_datetime(context: Context):
+    context.page.get_by_label("Release date", exact=True).fill("2025-3-1 17:00")
+
+
+@then('the datetime is displayed with "pm"')
+def display_pm_datetime(context: Context):
+    expect(context.page.get_by_text("March 2025 5:00pm")).to_be_visible()
```
kacperpONS, line 143: These don't need to be separate functions, you can parametrize it on based on `MeridiemIndicator` to reduce duplication

sanjeevz3009: I actually tried this, and then it ended up being an over abstraction, so I left it as is due to readability.

But I will push up what I had and get your feedback.

sanjeevz3009: @kacperpONS Done!

AdamHawtin (approved):

PR #142, opened 2025-04-03
Update CMS configuration options for line charts

nimasmi:

This change makes the editor UI for line charts more closely match the eventual configuration options.
It sets out a pattern for defining features per chart type, with blocks that subclass the base chart block.

Some options, which are related to configurable axes min/max and suffixes, will be dealt with as part of work on the axes, rather than this specific chart type.

### How to review

Add a Line Chart block, and enter some data.
Test that the fields in the ticket scope can be edited, and are reflected in the rendered chart.

Disregard that the rendered chart itself does not yet correspond to the spec. The updated version of the chart is in the Design System repo.

<details><summary>Screenshot</summary>
<p>

image

</p>
</details>

### DIFF
--- a/cms/datavis/blocks/base.py
+++ b/cms/datavis/blocks/base.py
@@ -20,6 +20,7 @@ class BaseVisualisationBlock(blocks.StructBlock):
     title = blocks.CharBlock()
     subtitle = blocks.CharBlock()
     table = SimpleTableBlock(label="Data table")
+    source = blocks.CharBlock()
 
     highcharts_chart_type: ClassVar[str]
 
--- a/cms/datavis/blocks/base.py
+++ b/cms/datavis/blocks/base.py
@@ -143,7 +143,9 @@ class BaseVisualisationBlock(blocks.StructBlock):
             "categories": [r[0] for r in rows],
         }
 
-        # Only add axis title if supported and provided
+        # Only add x-axis title if supported and provided, as the Highcharts
+        # x-axis title default value is undefined. See
+        # https://api.highcharts.com/highcharts/xAxis.title.text
         if (title := attrs.get("title")) and getattr(self, "supports_x_axis_title", False):
             config["title"] = {
                 "text": title,
@@ -166,8 +168,11 @@ class BaseVisualisationBlock(blocks.StructBlock):
             # "reversed": self.y_reversed,
         }
 
-        # Only add axis title if supported and provided
-        if (title := attrs.get("title")) and getattr(self, "supports_y_axis_title", False):
+        # Only add y-axis title if supported
+        if getattr(self, "supports_y_axis_title", False):
+            # Highcharts y-axis title default value is "Values". Set to undefined to
+            # disable. See https://api.highcharts.com/highcharts/yAxis.title.text
+            title = attrs["title"] or None
             config["title"] = {
                 "text": title,
             }
--- a/cms/datavis/blocks/base.py
+++ b/cms/datavis/blocks/base.py
@@ -1,13 +1,14 @@
 from collections.abc import Sequence
 from typing import Any, ClassVar, Optional
 
-from django.forms.widgets import Media
+from django.forms.widgets import Media, RadioSelect
 from django.utils.functional import cached_property
 from wagtail import blocks
 from wagtail.blocks.struct_block import StructValue
 
 from cms.datavis.blocks.table import SimpleTableBlock
 from cms.datavis.blocks.utils import TextInputFloatBlock, TextInputIntegerBlock
+from cms.datavis.constants import HighchartsTheme
 
 
 class PointAnnotationBlock(blocks.StructBlock):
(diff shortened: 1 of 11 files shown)


cms/datavis/blocks/base.py
```diff
@@ -87,64 +123,69 @@ def get_component_config(self, value: "StructValue") -> dict[str, Any]:
 
         return {
             "chart": {
-                "type": self.highcharts_chart_type,
+                "type": self.highcharts_chart_type,  # TODO: remove this once we upgrade to the latest version of the DS
```
nimasmi, line 125: **Comment**: See https://github.com/ONSdigital/design-system/pull/3506/files#diff-2f969b9706709ca622d1a927bd21417d5a2f6fcc817e223dc44d902cb692940aR20, for the corresponding change in the Design System repo. The chart type is part of the constructor, not a part of the config.

cms/datavis/blocks/base.py
```diff
@@ -87,64 +123,69 @@ def get_component_config(self, value: "StructValue") -> dict[str, Any]:
...
             "categories": [r[0] for r in rows],
         }
-        if (tick_interval := x_value.get("tick_interval")) is not None:
+
+        # Only add x-axis title if supported and provided, as the Highcharts
+        # x-axis title default value is undefined. See
+        # https://api.highcharts.com/highcharts/xAxis.title.text
+        if (title := attrs.get("title")) and getattr(self, "supports_x_axis_title", False):
+            config["title"] = {
+                "text": title,
```
nimasmi, line 150: **Comment**: This removes the default x-axis value, which was previously taken from the first header in the dataset.

cms/datavis/blocks/base.py
```diff
@@ -87,64 +123,69 @@ def get_component_config(self, value: "StructValue") -> dict[str, Any]:
...
+
+        # Only add y-axis title if supported
+        if getattr(self, "supports_y_axis_title", False):
+            # Highcharts y-axis title default value is "Values". Set to undefined to
+            # disable. See https://api.highcharts.com/highcharts/yAxis.title.text
+            title = attrs["title"] or None
             config["title"] = {
-                "enabled": True,
-                "text": label,
+                "text": title,
```
nimasmi, line 176: **Comment**: Due to a Highcharts quirk, we must explicitly remove the title if we don't want one, by setting it to `None`.

cms/datavis/blocks/charts.py
```diff
@@ -4,27 +4,44 @@
...
     supports_markers = True
-    x_axis_type = "linear"
+    extra_item_attributes: ClassVar = {
+        "connectNulls": True,
+    }
 
-    # Remove the use_stacked_layout field from the block
+    # Remove unsupported features
     use_stacked_layout = None
+    show_data_labels = None
```
nimasmi, line 19: **Comment**: This is the best suggestion for the time being for how to have optional fields in subclasses of BaseVisualisationBlock. We could _add_ them in the subclasses which need them, but that would affect the field ordering, since StructBlock doesn't have a panels attribute. A possible improved approach might be [How to build custom StreamField blocks &mdash; Wagtail Documentation 6.4.1 documentatio # Custom editing interfaces for StructBlock](https://docs.wagtail.org/en/stable/advanced_topics/customization/streamfield_blocks.html#custom-editing-interfaces-for-structblock).

cms/datavis/blocks/base.py
```diff
@@ -166,23 +207,33 @@ def get_annotations_config(self, value: "StructValue") -> list[dict[str, Any]]:
...
+            if self.supports_data_labels:
+                item["dataLabels"] = {
+                    "enabled": value.get("show_data_labels"),
                 }
+
+            if getattr(self, "supports_markers", False):
+                item["marker"] = {"enabled": value.get("show_markers")}
+
+            for key, val in getattr(self, "extra_item_attributes", {}).items():
+                item[key] = val
```
nimasmi, line 234: **Comment**: This is a way of allowing a block type to define hard-coded series options. E.g. for line charts: `"connectNulls": True`.

cms/datavis/blocks/base.py
helenb: Worth adding that as a comment in code.

cms/datavis/blocks/base.py
helenb: Again a comment in code might be useful here.

cms/datavis/tests/test_table_block.py
```diff
@@ -0,0 +1,61 @@
...
+from cms.datavis.tests.factories import TableDataFactory
+
+
+class SimpleTableBlockTestCase(TestCase):
+    table_data: TableDataType
+
+    def setUp(self):
+        self.block = SimpleTableBlock()
+        self.block_data = TableDataFactory(
+            table_data=[
```
MaciekBaron, line 13: **Question**: Isn't the `table_data` you are providing here the same as the default data used in the factory?

https://github.com/ONSdigital/dis-wagtail/blob/3911fd4661f7827fb3266cc9f39fd96b3dd51d0a/cms/datavis/tests/factories.py#L19-L23

cms/datavis/blocks/base.py
nimasmi: Thanks, yes, this is already commented on line 171, and 173–174.

```python
        # Only add y-axis title if supported
        if getattr(self, "supports_y_axis_title", False):
            # Highcharts y-axis title default value is "Values". Set to undefined to
            # disable. See https://api.highcharts.com/highcharts/yAxis.title.text
            title = attrs["title"] or None
            config["title"] = {
                "text": title,
            }
```

cms/datavis/blocks/base.py
nimasmi: Thanks for the suggestion. Even just making this comment yesterday I realised it was badly named, so I have fixed that too, in 51281aa.

cms/datavis/tests/test_table_block.py
nimasmi: Yes. I think originally this came before I had added the factory, and then I refactored this. It's probably important to set it differently and explicitly only in those test methods which look at the table value, so I've updated it (as well as refactoring to rely less on properties of `self`) in c580f62.

MaciekBaron (approved):

PR #145, opened 2025-04-04
Set KafkaPublisher Retries to 5 To Fix NotLeaderForPartitionError

sanjeevz3009:

This is a quick follow-up PR to the https://github.com/ONSdigital/dis-wagtail/pull/115 work. This PR address an issue where when the kafka tries to produce and send events for the first time, the topics `search-content-updated` and `search-content-deleted` doesn't exists in the Kafka  `broker` container.

This PR adds a retry of `5`. If the first attempt hits “NotLeaderForPartition,” it will retry. By attempt #2 or #3, the topic will have been created and will have a leader.

### How to review

- Spin up a fresh Kafka broker container locally by removing the existing broker containers.
- Run the search tests, and everything should be passing on the first go.

### Follow-up Actions

### DIFF
--- a/cms/search/publishers.py
+++ b/cms/search/publishers.py
@@ -132,6 +132,7 @@ class KafkaPublisher(BasePublisher):
             bootstrap_servers=[settings.KAFKA_SERVER],
             api_version=settings.KAFKA_API_VERSION,
             value_serializer=lambda v: json.dumps(v).encode("utf-8"),
+            retries=5,
         )
 
     @property
--- a/cms/search/tests/test_publishers.py
+++ b/cms/search/tests/test_publishers.py
@@ -367,6 +367,7 @@ class KafkaPublisherTests(TestCase):
             bootstrap_servers=["localhost:9092"],
             api_version=(3, 5, 1),
             value_serializer=ANY,  # or a lambda
+            retries=5,
         )
 
         self.assertEqual(publisher.created_or_updated_channel, "search-content-updated")  # pylint: disable=W0212


MaciekBaron:
Could you just add context to the PR description so its recorded for posterity? Thanks!

sanjeevz3009:
> Could you just add context to the PR description so its recorded for posterity? Thanks!

Yes, was adding it haha! Rushed to get this PR in, so devs don't have to re-run the unit tests =)

cms/search/publishers.py
```diff
@@ -132,6 +132,7 @@ def __init__(self) -> None:
             bootstrap_servers=[settings.KAFKA_SERVER],
             api_version=settings.KAFKA_API_VERSION,
             value_serializer=lambda v: json.dumps(v).encode("utf-8"),
+            retries=5,
```
MebinAbraham, line 135: Let's making this a setting with a default.

zerolab:
On the CI issue, https://stackoverflow.com/questions/68128973/kafka-broker-with-inconsistent-data-notleaderforpartitionerror suggests https://kafka.apache.org/documentation/#topicconfigs_unclean.leader.election.enable

In addition to this, we need to revisit https://github.com/ONSdigital/dis-wagtail/blob/main/cms/search/publishers.py#L149-L150

That is, properly handle exceptions. @MebinAbraham you mentioned we need to surface whether we had any issues with sending data to Kafka, so this would be good place to do it. We do want to consider how we store the info.

sanjeevz3009:
> On the CI issue, https://stackoverflow.com/questions/68128973/kafka-broker-with-inconsistent-data-notleaderforpartitionerror suggests https://kafka.apache.org/documentation/#topicconfigs_unclean.leader.election.enable
> 
> In addition to this, we need to revisit https://github.com/ONSdigital/dis-wagtail/blob/main/cms/search/publishers.py#L149-L150
> 
> That is, properly handle exceptions. @MebinAbraham you mentioned we need to surface whether we had any issues with sending data to Kafka, so this would be good place to do it. We do want to consider how we store the info.

I did look at these links posted already, but after digging, instead of configuring anything in the Kafka container, I went ahead with the `retries`, assuming we may face similar issues in the future when Kafka events fire for the first time, as when I publish a page or delete a page the Kafka events received by the consumer aren't successful but the subsequent attempts are.

Regarding exceptions, yes, that was going to be on a follow-up card, which I was going to create.

MebinAbraham:
> On the CI issue, [stackoverflow.com/questions/68128973/kafka-broker-with-inconsistent-data-notleaderforpartitionerror](https://stackoverflow.com/questions/68128973/kafka-broker-with-inconsistent-data-notleaderforpartitionerror) suggests [kafka.apache.org/documentation#topicconfigs_unclean.leader.election.enable](https://kafka.apache.org/documentation/#topicconfigs_unclean.leader.election.enable)
> 
> In addition to this, we need to revisit [`main`/cms/search/publishers.py#L149-L150](https://github.com/ONSdigital/dis-wagtail/blob/main/cms/search/publishers.py#L149-L150)
> 
> That is, properly handle exceptions. @MebinAbraham you mentioned we need to surface whether we had any issues with sending data to Kafka, so this would be good place to do it. We do want to consider how we store the info.

I suggest we come back to the sync/async as it needs some analysis about how we store and display the error on which interfaces and in what format. We also need to think about how the issue would be surfaced for non-bundled pages.

Let's fix the issues that are blocking CI for now.

zerolab:
Adding the retries is the good first step (and the scope of this PR) 👍🏼 

To be clear: I was not suggesting we solve everything here, more of a reminder for all of us as it was the right context

MebinAbraham (approved):

sanjeevz3009:
> Adding the retries is the good first step (and the scope of this PR) 👍🏼
> 
> To be clear: I was not suggesting we solve everything here, more of a reminder for all of us as it was the right context. And subject to 2 follow up tickets for sure

Good stuff! After discussions with @MebinAbraham. This is good to go in as a temporary CI fix, I'll create the cards to capture everything discussed in this PR and what we need to do moving forward =)

cms/search/publishers.py
sanjeevz3009: We are leaving this as is =)

AdamHawtin (approved):

kacperpONS (approved):

MaciekBaron (approved):

PR #150, opened 2025-04-10
Bump tinytableblock and Django versions

zerolab:

Bumps versions for:
- wagtail-tinytableblock - improves paste handling. See thread in Slack
- Django - it doesn't affect us, but will keep Dependabot happy ([ref](https://docs.djangoproject.com/en/5.1/releases/5.1.8/))

And relabels the table block "Caption" field to "sub-heading" (this was highlighted at today's UR findings session)

### How to review

`poetry install`, create a new table and paste content from Excel or a table copied from a web page (including precending/follow-up text)

### Follow-up Actions

N/A


cms/core/blocks/markup.py
```diff
@@ -114,6 +114,11 @@ class ONSTableBlock(TinyTableBlock):
     source = blocks.CharBlock(label="Source", required=False)
     footnotes = blocks.RichTextBlock(label="Footnotes", features=settings.RICH_TEXT_BASIC, required=False)
 
+    def __init__(self, *, local_blocks=None, search_index=True, **kwargs):
```
kacperpONS, line 117: To resolve the linter failure
```suggestion
    from wagtail.blocks.base import Block
    def __init__(self, *, local_blocks: list[Block] | None = None, search_index: bool = True, **kwargs: Any) -> None:
```

cms/core/blocks/markup.py
```diff
@@ -114,6 +114,11 @@ class ONSTableBlock(TinyTableBlock):
...
+        super().__init__(local_blocks=local_blocks, search_index=search_index, **kwargs)
+        # relabeled to match editorial stuff terminology
```
kacperpONS, line 119: For the comment maybe something like: 
```suggestion
        # relabeled to match the publishing team's terminology
```
<br>
Also - the functional tests will need to updated to reflect this.

cms/core/blocks/markup.py
zerolab: argh, good shout! 06077ec

kacperpONS (approved):

nehakerung (approved):

sanjeevz3009 (approved):
Works as expected, let's get this in ;)

PR #151, opened 2025-04-11
Update design system to latest version (72.4.0 - 72.6.0)

janehughes:
### What is the context of this PR?

This updates the design system version to the latest 72.6.0 to bring in a few new components and enhancements which we need to update the website templates.

### How to review

- Pull branch locally
- Run `make load-design-system-templates`
- See that the latest design system changes have been added to your local build (refer to [release notes](https://github.com/ONSdigital/design-system/releases) for updates)

PR #153, opened 2025-04-14
Feature/ccb 84 ds upgrade

nimasmi:

The chart builder was originally written using placeholder chart templates and JS functions, pending separate work in the https://github.com/ONSdigital/design-system/ repository. The initial version of line charts was merged in version 72.5.0, so the chart builder can switch and start to use the Design System directly to render charts.

Changes:

- 

#### Draft state notes

This draft PR temporarily upgrades to Design System version 72.5.0.

Due to the incompatibility fixed here in 907ad3b, and in the Design System in https://github.com/ONSdigital/design-system/pull/3550, this cannot yet be merged.

### How to review

### Follow-up Actions


nimasmi:
> FWIW, `main` is on 72.6.0 as of #151

Thanks, yes, I understand. The DS chart rendering will not be compatible with the CMS chart builder until https://github.com/ONSdigital/design-system/pull/3550 is merged, so this will stay in draft for a bit, and hopefully this will eventually upgrade to DS > 72.6.0.

cms/core/jinja2.py
```diff
@@ -0,0 +1,15 @@
...
+from typing import Any
+
+from jinja2 import ChainableUndefined
+
+
+def undefined_json_handler(obj: Any) -> Any:
+    """Custom JSON handler that can handle Nunjucks's ChainableUndefined objects."""
+    if isinstance(obj, ChainableUndefined):
+        return None
+    raise TypeError(f"Object of type {type(obj).__name__} is not JSON serializable")
```
nimasmi, line 12: **Comment**: Python docs ref: https://docs.python.org/3/library/json.html#:~:text=Specializing%20JSON%20object%20encoding

cms/core/jinja2.py
```diff
@@ -0,0 +1,15 @@
+import json
```
nimasmi, line 1: **Comment**: I wasn't completely sure about the best name or location for this module. Suggestions are welcome.

cms/settings/base.py
```diff
@@ -12,6 +12,8 @@
 from django.utils.translation import gettext_lazy as _
 from django_jinja.builtins import DEFAULT_EXTENSIONS
 
+from cms.core.jinja2 import custom_json_dumps
```
nimasmi, line 15: **Comment**: This is the only function import into settings from the local codebase. Are we happy with that approach? Django initialises fine with it.

cms/core/tests/test_tags.py
```diff
@@ -0,0 +1,31 @@
...
+from cms.core.templatetags.util_tags import extend
+
+
+class AppendTagTest(TestCase):
+    def test_append(self):
+        """Test the original list is modified in place."""
+        series = [{"name": "Series 1"}, {"name": "Series 2"}]
+        series_item = {"name": "Series 3"}
+        extend(series, series_item)
+        self.assertEqual([{"name": "Series 1"}, {"name": "Series 2"}, {"name": "Series 3"}], series)
```
zerolab, line 12: nitpick: `self.assertListEqual`

cms/settings/base.py
zerolab: #118 adds another inport from the codebase (ref: https://github.com/ONSdigital/dis-wagtail/pull/118/files#diff-bc1615013e08a4b8745001b986f376d9d23645bc1a278152d5dac09d88cb5b7cR14), so yes

cms/settings/base.py
```diff
@@ -169,6 +171,9 @@
                 "wagtail.contrib.settings.jinja2tags.settings",
                 "cms.core.jinja2tags.CoreExtension",
             ],
+            "policies": {
```
zerolab, line 174: suggestion: link to the Jinja docs for context
```suggestion
            "policies": {
                # https://jinja.palletsprojects.com/en/stable/api/#policies
```

zerolab (commented):
Did a quick pass and left a few notes.

cms/core/tests/test_tags.py
zerolab: To be honest, it was something we've been doing (although it may have come from a different project and I am muddling things up). Thank you for linking to the relevant Python docs.

cms/core/jinja2.py
zerolab: That is a great link, add to the function docblock?

cms/datavis/blocks/base.py
```diff
@@ -146,12 +144,11 @@ def get_x_axis_config(
...
         # https://api.highcharts.com/highcharts/xAxis.title.text
         if (title := attrs.get("title")) and getattr(self, "supports_x_axis_title", False):
-            config["title"] = {
-                "text": title,
-            }
+            config["title"] = title
 
         if (tick_interval := attrs.get("tick_interval")) is not None:
-            config["tickInterval"] = tick_interval
+            config["tickIntervalMobile"] = tick_interval
```
helenb, line 150: These changes don't seem relevant to the PR

nimasmi: DS version 72.8.0 introduces configuralble tick intervals as separate mobile and desktop parameters in https://github.com/ONSdigital/design-system/pull/3551. This PR is about compatibility with a new version of the Design System. If not here, then when and how should we update these parameters in the chart builder?

helenb: Oh I see - I thought they had crept in from somewhere else. Ignore me in that case.

nimasmi: Understood, thanks.

PR #156, opened 2025-04-15
Remove the sitemap

zerolab:

Currently there is no sitemap on the live site. We can't go day one with a partial one.

### How to review

Checkout and check `/sitemap.xml` is 404

### Follow-up Actions

N/A


kacperpONS (approved):
Works as expected 👍

PR #154, opened 2025-04-14
Update statistical article & topic page hero banner

janehughes:
### What is the context of this PR?

Describe what you have changed and why, link to other PRs or Issues as appropriate.

### How to review

Describe the steps required to test the changes (include screenshots if appropriate).

### Follow-up Actions

List any follow-up actions (if applicable), like needed documentation updates or additional testing.

PR #158, opened 2025-04-15
Make release date mandatory on release pages

kacperpONS:

This PR makes the release date field mandatory on all release pages no matter their `status`

Previously, release pages with status _provisional_ didn't require the `release_date` to be set. This has led to a bug where if we try to link a bundle to a release page that has the status _provisional_ and doesn't have the release date set, we would get a programatic error and the page would crash. 

After conversation with the publishing team in order to keep consistent with the current system, I've also made the following changes:
- allow `release_date_text` to override `release_date` on the _provisional_ release pages
- added JS to hide the `release_date_text` panel in the editor when the status is not _provisional_

I also had to update the search indexing so that if `release_date_text` is present then it is sent as `provisional_date` and `release_date` isn't sent to mimic the overwrite mechanism used for provisional pages.

### How to review

- Check that the release date field is mandatory on release page no matter the `status`.
- Check that the `release_date_text` panel is hidden in the editor when the `status` other than _provisional_
- Check that on _provisional_ pages, you can override `release_datet` with `release_date_text` 
- Also try linking a bundle schedule to a provisional release page to confirm that the issue has been resolved.
- Run unit tests
```
poetry run coverage run ./manage.py test --settings=cms.settings.test cms.release_calendar.tests --keepdb
```

### Follow-up Actions
- I've spoken with Joe Nation and a change of the design of Release pages is likely to follow - it will make use of the corrections and notices UI component for the  `changes_to_release_date` and `notice` fields. This ticket should also move these two fields to the "Corrections and notices" tab in the page editor. 
- Add support for Welsh month names for release date text field [CMS-449](https://jira.ons.gov.uk/browse/CMS-449)


cms/jinja2/assets/js/hide-date-text-on-non-provisional-releases.js
```diff
@@ -0,0 +1,16 @@
+function toggleDateTextVisibility() {
+  var statusField = document.getElementById('id_status');
```
MaciekBaron, line 2: Should be a `const` - `var`s should not be used anymore unless for some very specific reasons when you need a function scoped variable. Here and on line 14.
```suggestion
  const statusField = document.getElementById('id_status');
```

cms/jinja2/assets/js/hide-date-text-on-non-provisional-releases.js
```diff
@@ -0,0 +1,16 @@
...
+  const panelWrapper = dateTextInput.closest('.w-panel__wrapper');
+
+  panelWrapper.style.display = isProvisional ? 'block' : 'none';
+}
+
+document.addEventListener('DOMContentLoaded', function () {
+  toggleDateTextVisibility();
+
+  var statusField = document.getElementById('id_status');
+  statusField.addEventListener('change', toggleDateTextVisibility);
```
MaciekBaron, line 15: This code gets run in every editor context, so `statusField` can be `null`. You should only add the event listener if `statusField` exists, otherwise you will get errors in the console.

cms/jinja2/assets/js/hide-date-text-on-non-provisional-releases.js
MaciekBaron, line 2: Additionally what you could consider is using `event.target` which will be the element which triggered the change event, so you wouldn't need to re-retrieve it using `getElementById`.

https://developer.mozilla.org/en-US/docs/Web/API/Event/target

functional_tests/features/release_page.feature
```diff
@@ -14,3 +14,12 @@ Feature: CMS users can draft, edit, and publish release pages
...
         And the user can see the breadcrumbs
+
+    Scenario: Hiding of release date text field for non-provisional releases
+        Given a superuser logs into the admin site
+        When the user navigates to the release calendar page
+        And clicks "add child page" to create a new draft release page
+        And the page status is set to "Provisional" and the release date text field is visible
+        And the user sets the page status to "Confirmed"
+        Then the date text field is not visible
+
```
MaciekBaron, line 25: Nit-pick: Missing blank line at the end of the file

MaciekBaron (changes requested):

cms/jinja2/assets/js/hide-date-text-on-non-provisional-releases.js
kacperpONS: Thanks for refreshing my rusty JS knowledge 😄

cms/jinja2/assets/js/hide-date-text-on-non-provisional-releases.js
kacperpONS: I was also wondering if there's maybe a way to access the page details here? If so we could only include the Release pages for this.  Page type is known at this point but is it accessible in JS? 

Or maybe there's a way to use the `insert_editor_js` hook only for given page types? 

Either one would exclude this code sooner, but I'm happy to go with this if there's no mechanism to do it.

cms/jinja2/templates/pages/release_calendar/release_calendar_page--cancelled.html
```diff
@@ -3,14 +3,12 @@
 {% from "components/table-of-contents/_macro.njk" import onsTableOfContents %}
 
 {% block release_meta %}
-    {% if page.release_date or page.release_text %}
-        <div class="ons-container common-header__releases common-header__releases--single-row">
-            <div>
-                <span class="ons-u-fs-r--b">{{ _("Release date") }}:</span>
-                <span class="ons-u-nowrap">{{ _("Cancelled") }}</span>
-            </div>
+    <div class="ons-container common-header__releases common-header__releases--single-row">
```
zerolab, line 6: note: #154 updates the heroes, so all templates changes will need updating once that is merged

cms/jinja2/assets/js/hide-date-text-on-non-provisional-releases.js
MaciekBaron: Afaik conditional inclusion isn't possible using hooks, but I don't think it's a problem to include this in all editor instances, provided that we do the check mentioned.

cms/jinja2/assets/js/hide-date-text-on-non-provisional-releases.js
kacperpONS: I don't think that would work - note that `toggleDateTextVisibility` is called on DOM load even when there's no change to `status` to trigger it. This is to handle the case when the user opens editor on an already existing page that is non-provisional.

MaciekBaron: That's a good point @kacperpONS – I'll resolve.

cms/jinja2/templates/pages/release_calendar/release_calendar_page--cancelled.html
kacperpONS: [595751](https://github.com/ONSdigital/dis-wagtail/pull/158/commits/65957513e647fc52a1bdc2ccd26d1d15fefb6892)

cms/jinja2/assets/js/hide-date-text-on-non-provisional-releases.js
kacperpONS: [d5632aa](https://github.com/ONSdigital/dis-wagtail/pull/158/commits/d5632aa9f8ce5f1e73229f8d81c88af5c4a2424c)

cms/jinja2/assets/js/hide-date-text-on-non-provisional-releases.js
kacperpONS: [d5632aa](https://github.com/ONSdigital/dis-wagtail/pull/158/commits/d5632aa9f8ce5f1e73229f8d81c88af5c4a2424c)

cms/release_calendar/forms.py
```diff
@@ -37,13 +37,13 @@ def clean(self) -> dict:
...
             self.validate_bundle_not_pending_publication(status)
 
-        # Input field is hidden with custom JS for non-provisional releases, set to None to avoid unexpected behavior
         if self.instance.status != ReleaseStatus.PROVISIONAL:
-            self.instance.release_date_text = None
+            # Input field is hidden with custom JS for non-provisional releases,
+            # set to None to avoid unexpected behavior
+            cleaned_data["release_date"] = None
 
         if (
```
zerolab, line 45: this whole if block needs indenting so it is at the same level as line 43

cms/jinja2/templates/pages/release_calendar/release_calendar_page--provisional.html
```diff
@@ -37,9 +20,12 @@
                 "descriptionListLabel": "",
                 "termCol": "6",
                 "descriptionCol": "6",
-                "itemsList": [
-                    release_date_html
-                ]
+                "itemsList": [{
+                    "term": release_date_label,
+                    "descriptions": [{
+                        "description": page.release_date_text|page.release_date|ons_date("DATETIME_FORMAT")
```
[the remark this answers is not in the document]
kacperpONS: It's all work in progress 😅

cms/jinja2/templates/pages/release_calendar/release_calendar_page.html
```diff
@@ -1,47 +1,48 @@
...
-        {%- set release_date_html = "" -%}
-    {% endif %}
-
-    {% if page.next_release_date or page.next_release_text %}
-        {%- set next_release_date_term = _("Next release:") -%}
-        {# fmt:off #}
-        {%- set next_release_html =
-            {
-                "term": next_release_date_term,
+    {# fmt:off #}
```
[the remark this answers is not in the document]
kacperpONS: The problem with this suggestion is that there's now no way of conditionally including the items - say neither `next_release_date_text`  nor `next_release_date` exits  - "Next release date:" will still be shown on the page. 

And dict list + for loop seemed like a neater solution but I'm happy to revert.

cms/release_calendar/tests/test_models.py
```diff
@@ -314,6 +314,27 @@ def test_rendered__changes_to_release_date(self):
 
                 self.assertEqual("The reason" in str(response.content), is_shown)
 
+    def test_release_date_text_overrides_release_datetime(self):
+        """Check that release date text overrides release datetime if both are present."""
+        self.page.status = ReleaseStatus.PROVISIONAL
+        self.page.release_date = timezone.now()
+        self.page.release_date_text = "January 2024"
+        self.page.save_revision().publish()
```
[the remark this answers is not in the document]
kacperpONS: Would the next release date need an equivalent property then? In the theme of keeping the logic out of the templates. Also it would lead to insistency - release date would have the logic in the model, and next_release in the template.


> so this is only technically true in that if we have both, the template will give release_date_text precedence.

I'm not sure I understand this part, this was exactly the idea for the test. To be more precise it tests this line in the provisional template: 
```
    "description": page.release_date_text|page.release_date|ons_date("DATETIME_FORMAT")
```

cms/search/utils.py
```diff
@@ -35,8 +35,11 @@ def build_release_specific_fields(page: "Page") -> dict:
         "published": page.status == ReleaseStatus.PUBLISHED,
         "date_changes": [],
     }
-    # If page.release_date_text is present, treat it as provisional_date
+
+    # If page.release_date_text is present, treat it as a provisional_date
+    # and do not add release_date to the payload.
```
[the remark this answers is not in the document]
kacperpONS: A confirmed release would have the `release_date_text` set to `None` in the `clean` method so this condition doesn't apply for it.

cms/jinja2/templates/pages/release_calendar/release_calendar_page.html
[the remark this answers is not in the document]
zerolab: You are right!

cms/jinja2/templates/pages/release_calendar/release_calendar_page.html
```diff
@@ -1,47 +1,48 @@
...
+        "value": page.release_date,
+        "text": page.release_date_text
+        },
+        {
+        "type": "next_release_date",
+        "label": next_release_date_label,
+        "value": page.next_release_date,
+        "text": page.next_release_date_text
+        }
+    ] %}
```
zerolab, line 26: How about

```suggestion
    {%- set release_date_details = [
        {
            "label": release_date_label,
            "value": page.release_date_text or page.release_date|ons_date("DATETIME_FORMAT"),
        },
        {
            "label": next_release_date_label,
            "value": page.next_release_date_text or page.next_release_date|ons_date("DATETIME_FORMAT"),
        }
    ] -%}
```

since type is not used, and we do a valur or text anyway.

Then:

```jinja2
    {%- set release_dates_list = [] -%}
    {%- for release_date in release_date_details -%}
        {# fmt:off #}
        {% if release_date.value %}
            {# fmt:off #}
            {%- do release_dates_list.append({
                "term": release_date.label,
                "descriptions": [{"description": release_date.value}]
            }) -%}
            {# fmt:on #}
        {% endif %}
    {%- endfor -%}
```

cms/release_calendar/tests/test_models.py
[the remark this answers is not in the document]
zerolab: What I was trying to convey was

Given `test_release_date_text_overrides_release_datetime` is quite generic, we should have a test that looks at the other statuses and ensures for the confirmed/published only `page.release_date` is used.


> so this is only technically true in that if we have both, the template will give release_date_text precedence.

If we have both values in the db (for whatever reason) then if the status is confirmed or published precedence will be given to `release_date_text`, but we want `release_date`

you can change line 319 to `self.page.status = ReleaseStatus.CONFIRMED` to check

cms/jinja2/templates/pages/release_calendar/release_calendar_page--confirmed.html
```diff
@@ -6,6 +6,41 @@
...
+    {# fmt:off #}
+    {% set release_date_details = [
+        {
+        "type": "release_date",
+        "label": _("Release date:"),
+        "value": page.release_date,
+        "text": page.release_date_text
+        },
+        {
+        "type": "next_release_date",
```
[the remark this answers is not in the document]
kacperpONS: I took the liberty of putting leaving them there: [902abf5](https://github.com/ONSdigital/dis-wagtail/pull/158/commits/902abf5e7ab80eb6a8bdbef68389a07aba335292)

cms/jinja2/templates/pages/release_calendar/release_calendar_page.html
kacperpONS: Went with this in: [902abf5](https://github.com/ONSdigital/dis-wagtail/pull/158/commits/902abf5e7ab80eb6a8bdbef68389a07aba335292)

In this commit I also included the next release dates in the templates for provisional, confirmed and published releases. Let's keep the convo about this part in this thread: https://github.com/ONSdigital/dis-wagtail/pull/158#discussion_r2063907359

cms/release_calendar/tests/test_models.py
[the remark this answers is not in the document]
kacperpONS: Made it a property in [902abf5](https://github.com/ONSdigital/dis-wagtail/pull/158/commits/902abf5e7ab80eb6a8bdbef68389a07aba335292). I'll be working on improving the tests now.

cms/jinja2/templates/pages/release_calendar/release_calendar_page--confirmed.html
```diff
@@ -3,8 +3,35 @@
 
 {% block header_area %}
     {%- set hero_topic = _("Release") -%}
-    {%- set release_date_label = _("Release date:") -%}
     {%- set breadcrumbs_label = _("Breadcrumbs") -%}
+    {%- set release_date_label = _("Release date:") -%}
+    {%- set next_release_date_label = _("Next release date:") -%}
```
zerolab, line 8: Back to the original comment: the current live site doesn't show a next release date, nor were the requirements asking for this.

What prompted this change?

cms/release_calendar/forms.py
```diff
@@ -66,13 +65,6 @@ def clean(self) -> dict:
             raise ValidationError({"next_release_date": "The next release date must be after the release date."})
 
         release_date_text = cleaned_data.get("release_date_text")
-        if cleaned_data.get("release_date") and release_date_text:
-            error = "Please enter the release date or the release date text, not both."
-            raise ValidationError({"release_date": error, "release_date_text": error})
-
-        if cleaned_data.get("next_release_date") and cleaned_data.get("next_release_text"):
-            error = "Please enter the next release date or the next release text, not both."
-            raise ValidationError({"next_release_date": error, "next_release_text": error})
```
zerolab, line 75: The next release date or next relese date text logic is still relevant, so this check should stay. Unless there was a change in requirements that I am not aware of?

cms/jinja2/templates/pages/release_calendar/release_calendar_page--confirmed.html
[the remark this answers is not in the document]
zerolab: >  I think it makes sense to add the option to have the next release date for provisional, confirmed and published

This needs to be confirmed with DP, please

cms/jinja2/templates/pages/release_calendar/release_calendar_page--confirmed.html
zerolab: closing in favour of the discussion in https://github.com/ONSdigital/dis-wagtail/pull/158#discussion_r2063907359

cms/release_calendar/forms.py
kacperpONS: I went with the same mechanism as for the release date, where the date text overwrites the datetime for consistency. Arguably this case is different because `next_release_date` is nullable, but it might be counter intuitive to see it enforced for next date and not for date.

cms/jinja2/templates/pages/release_calendar/release_calendar_page--confirmed.html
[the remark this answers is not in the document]
kacperpONS: I've messaged Paula

cms/search/utils.py
```diff
@@ -35,8 +35,11 @@ def build_release_specific_fields(page: "Page") -> dict:
...
     if page.release_date_text:
+        data["release_date"] = None
```
sanjeevz3009, line 42: Superfluous. Why are we setting `release_date` to `None` here? This is unnecessary as we already set it to None here: https://github.com/ONSdigital/dis-wagtail/blob/5b12c7eba1ccad1251caa68b6a850706cd75aed2/cms/search/utils.py#L22

sanjeevz3009 (commented):
Did an initial round of reviews, so far looks good. Will do another round of review once the other comments are resolved and business decisions have been made.

cms/search/utils.py
sanjeevz3009: Okay, wasn't aware, or maybe I seen some convo around this somewhere on this PR/ or somewhere else and got confused. Thanks for info, this is weird but if this is how is done, then that's fine.

kacperpONS:
@sanjeevz3009 
Very good spot, I've just made the field nullable and fixed the validation. 

Regarding the styling - I think it is more logical the way it is now - you have one line for release date and another one for **next** release date. If we went ahead with this suggestion we would lose that and have it mixed which might be confusing - why have `release date` and `next release date text` on the same line? Moreover to a less observant user it might appear that it's the `next_release_date_text` that's disappearing, not `release_date_text`.
In either case I don't think we should worry about this in this ticket - if that comes as an issue in UR we can revisit.

cms/release_calendar/forms.py
```diff
@@ -37,11 +37,10 @@ def clean(self) -> dict:
...
 
-        if status in [ReleaseStatus.CONFIRMED, ReleaseStatus.PUBLISHED]:
-            if not cleaned_data.get("release_date"):
-                raise ValidationError(
-                    {"release_date": "The release date field is required when the release is confirmed"}
-                )
+        if self.instance.status != ReleaseStatus.PROVISIONAL:
+            # Input field is hidden with custom JS for non-provisional releases,
+            # set to None to avoid unexpected behavior
+            cleaned_data["release_date_text"] = None
```
zerolab, line 43:
```suggestion
            cleaned_data["release_date_text"] = ""
```

kacperpONS: shall I revert making it nullable?

zerolab: Yes please. `CharField` should not be nullable per se.

kacperpONS: [12e3f92](https://github.com/ONSdigital/dis-wagtail/pull/158/commits/12e3f9241dd780cbed6f58a67c077b226a07ee51)

sanjeevz3009:
> @sanjeevz3009 Very good spot, I've just made the field nullable and fixed the validation.
> 
> Regarding the styling - I think it is more logical the way it is now - you have one line for release date and another one for **next** release date. If we went ahead with this suggestion we would lose that and have it mixed which might be confusing - why have `release date` and `next release date text` on the same line? Moreover to a less observant user it might appear that it's the `next_release_date_text` that's disappearing, not `release_date_text`. In either case I don't think we should worry about this in this ticket - if that comes as an issue in UR we can revisit.

Thanks, will re-review once it's all fixed.

Happy with the latter when it comes to the editor design, not too concerned with that.

cms/jinja2/templates/pages/release_calendar/release_calendar_page--confirmed.html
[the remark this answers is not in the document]
kacperpONS: Removed from the templates in [3a31c3c](https://github.com/ONSdigital/dis-wagtail/pull/158/commits/3a31c3c846dc1578d7733b476f1c3b304f8cf370). We can resolve this comment now and pick it up as a separate task once DP gets back.

cms/release_calendar/tests/test_forms.py
```diff
@@ -190,34 +180,6 @@ def test_form_clean__validates_release_date_when_confirmed__happy_path(self):
...
-        data["release_date_text"] = "November 2024"
-        data = nested_form_data(data)
-        form = self.form_class(instance=self.page, data=data)
-
-        self.assertFalse(form.is_valid())
-        message = ["Please enter the release date or the release date text, not both."]
-        self.assertFormError(form, "release_date", message)
-        self.assertFormError(form, "release_date_text", message)
-
-    def test_form_clean__validates_either_next_release_date_or_text(self):
```
zerolab, line 207: this test needs to stay

cms/release_calendar/models.py
```diff
@@ -174,6 +177,20 @@ def get_context(self, request: "HttpRequest", *args: Any, **kwargs: Any) -> dict
...
+            return self.release_date_text
+        return ons_date_format(self.release_date, "DATETIME_FORMAT")
+
+    @property
+    def next_release_date_value(self) -> str | None:
+        if self.next_release_date:
+            return ons_date_format(self.next_release_date, "DATETIME_FORMAT")
+        if self.next_release_date_text:
+            return self.next_release_date_text
+        return None
```
zerolab, line 192: The help text for `next_release_date_text` says "Override next release date". So this should be flipped (it was correct before 080b253

cms/release_calendar/tests/test_models.py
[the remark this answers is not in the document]
kacperpONS: Tests added in [080b253](https://github.com/ONSdigital/dis-wagtail/pull/158/commits/080b253f7b5d428023aa2dec99ef4cba974be20b). 
Note that for the `next_release_date_value` property I simply copied the logic from the template, where datetime takes precedence over date text if both are present (would have to be added programatically as we prohibit both at the same time during form validation).

cms/release_calendar/models.py
zerolab: saying that, we do have validation in the form that prevents both to be added. so let's leave as is.

cms/release_calendar/tests/test_models.py
```diff
@@ -314,6 +314,27 @@ def test_rendered__changes_to_release_date(self):
...
+    def test_next_release_date_text_overrides_next_release_datetime(self):
+        """Check that next release date text overrides next release datetime if both are present."""
+        self.page.status = ReleaseStatus.PUBLISHED
+        self.page.release_date = timezone.now()
+        self.page.next_release_date = timezone.now()
+        self.page.next_release_date_text = "January 2024"
+        self.page.save_revision().publish()
+
+        response = self.client.get(self.page.url)
+        self.assertIn("January 2024", str(response.content))
```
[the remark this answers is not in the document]
kacperpONS: [d59b34d](https://github.com/ONSdigital/dis-wagtail/pull/158/commits/d59b34d69cc9ea32907262ad54c9cd28982dc1cb)

functional_tests/steps/release_page.py
```diff
@@ -116,6 +116,17 @@ def check_selected_datasets_are_displayed(context: Context):
         expect(context.page.get_by_text(dataset["description"])).to_be_visible()
 
 
+@when('the page status is set to "Provisional" and the release date text field is visible')
```
zerolab, line 119: nitpick: perhaps make this a `@step`?

My understanding of BDD is:
- `given` adds context. This is where you can either preset things (create user, page, etc) and do some actions (e.g. log in)
- `when` sets the scene. That is does actions
- `then` is the bit that checks the results.


I wonder if perhaps this scenario needs to be two:
1. when status is provisional, release date text is visiblwe
2. when changing to any other, it gets hidden

zerolab: Other than this, LGTM!

kacperpONS: I was trying to be overly concise here - I agree it makes sense to have it as two scenarios. 
I'm not too sure about making it a `@step` as 1 - it's only used once (with `Then`), and 2 - it only holds assertions so according to what you said here it's most likely going to stay a `Then`.

kacperpONS: [2c0d58a](https://github.com/ONSdigital/dis-wagtail/pull/158/commits/2c0d58acd74d8c6d51a23b1639ad34ac4046e402)

cms/search/utils.py
zerolab: Resolving based on the comment above ^

zerolab (approved):

cms/release_calendar/forms.py
```diff
@@ -37,26 +37,26 @@ def clean(self) -> dict:
...
+        if status != ReleaseStatus.PROVISIONAL:
+            # Input field is hidden with custom JS for non-provisional releases,
+            # set to None to avoid unexpected behavior
+            cleaned_data["release_date_text"] = ""
+
+        if (
+            status in [ReleaseStatus.CONFIRMED, ReleaseStatus.PUBLISHED]
+            and self.instance.release_date
+            and self.instance.release_date != cleaned_data.get("release_date")
+            and len(self.instance.changes_to_release_date) == len(cleaned_data.get("changes_to_release_date", []))
```
AdamHawtin, line 49: I don't think this validation around `changes_to_release_date` quite works at the moment, since you are allowed to add arbitrary items to `changes_to_release_date` without actually changing the date, you can bypass this particular validation by _removing_ an existing item rather than adding an new one.

Perhaps we should only be allowing additions to `changes_to_release_date` when the release date actually has been changed, and we should only allow one at a time?

cms/release_calendar/models.py
```diff
@@ -53,15 +54,17 @@ class ReleaseCalendarPage(BasePage):  # type: ignore[django-manager-missing]
...
 
-    release_date = models.DateTimeField(
-        blank=True, null=True, help_text="Required once the release has been confirmed."
-    )
+    release_date = models.DateTimeField(blank=False, null=False)
     release_date_text = models.CharField(
-        max_length=50, blank=True, help_text="Format: 'Month YYYY', or 'Month YYYY to Month YYYY'."
+        max_length=50,
+        blank=True,
+        help_text="Override release date for provisional entries. Format: 'Month YYYY', or 'Month YYYY to Month YYYY'.",
```
nehakerung, line 61:
```suggestion
        help_text="Override release date for provisional entries. Formats: 'Month YYYY', or 'Month YYYY to Month YYYY'.",
```
Nit

cms/release_calendar/forms.py
kacperpONS: I've talked about this with Adam on Slack and we agreed that this is out of scope for this PR. 

I suggested that we could use the same mechanism as for **Corrections & notices** on Statistical Article pages where the block becomes `frozen` on page publication and editing is then disabled on the client side. 

There's also a broader idea of aligning the **Changes to release date** and **Notices** to the **Corrections and notices** blocks so they have a similar workflow making it more intuitive for the users. 
This would include aligning the validation (you can for instance add `changes_to_release_date` block even before the page is published, which isn't possible for the C&N), disabling editing of `frozen` blocks and moving the blocks to a separate tab.
This seems like a bigger task, so it might be better to keep the scope of CMS-355 limited and revisit this on a separate card. 

Also - Adam highlighted the fact that we don't validate month name in `release_date_text` on Welsh release pages - there's a TODO in the code and I've created a card: [CMS-449](https://jira.ons.gov.uk/browse/CMS-449) https://github.com/ONSdigital/dis-wagtail/blob/b29adc4924c66997f40d87ae046b8f2a7900ff97/cms/release_calendar/forms.py#L77

cms/release_calendar/models.py
nehakerung: Pointed out to keep consistency with `the next_release_date_text` help text - I would suggest changing its help text below to `Format`.

nehakerung (approved):
LGTM - functionally tested!

cms/release_calendar/models.py
kacperpONS: Good spot! fixed in [2a70680](https://github.com/ONSdigital/dis-wagtail/pull/158/commits/2a706807fd6940deb42db8a465d763761db15062) 😄

kacperpONS:
[Loom video for this PR](https://www.loom.com/share/91e8ed505f47472a95de19cf6993264b)

Now just waiting for the functional tests to pass 🤞

sanjeevz3009 (commented):
Reviewed!

PR #159, opened 2025-04-15
Search Integration Resources API Endpoint

sanjeevz3009:

- The search service requires a mechanism to fetch all indexable content from Wagtail.
- This API endpoint will provide structured pagination to manage resource retrieval efficiently.
- An API endpoint in Wagtail that returns all indexable resources, so we can fetch and re-index them efficiently.

### How to review

- Run `django-admin collecstatic --noinput` in the enviroment that runs the Django server
- Set `IS_EXTERNAL_ENV` to `True`.
- Create some pages such as `Generic Index Page`, `Statistical Article Page`, `Methodology Page`, `Information Page`, and `Release Calendar Page`.
- Pages: `Article Series`, `Home Page`, `Release Calendar Index`, `Theme` and `Topic Pages` should be excluded.
- Navigate to `http://127.0.0.1:8000/v1/resources/` to view the `Browsable API` page.
- Run the search tests: `./manage.py test cms.search --settings=cms.settings.test --keepdb` and make sure everything passes.

**Useful Specs:**
https://github.com/ONSdigital/dis-search-upstream-stub/blob/main/specification.yml
https://github.com/ONSdigital/dis-search-upstream-stub/blob/main/docs/contract/resource_metadata.yml

**Dissemination's API Standards**: https://github.com/ONSdigital/dp-standards/blob/main/API_STANDARDS.md

### Follow-up Actions


cms/search/serializers.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def get_summary(self, obj):
+        if hasattr(obj, "summary"):
+            return get_text_for_indexing(obj.summary)
+        return ""
+
+    def get_topics(self, obj):
+        if hasattr(obj, "topic_ids"):
+            return obj.topic_ids
+        return []
```
zerolab, line 32:
```suggestion
        return getattr(obj, "topic_ids", []):
```

cms/settings/base.py
```diff
@@ -918,3 +919,5 @@
 # FIXME: remove before going live
 ENFORCE_EXCLUSIVE_TAXONOMY = env.get("ENFORCE_EXCLUSIVE_TAXONOMY", "true").lower() == "true"
 ALLOW_TEAM_MANAGEMENT = env.get("ALLOW_TEAM_MANAGEMENT", "false").lower() == "true"
+
+IS_EXTERNAL_ENV = os.getenv("IS_EXTERNAL_ENV", "True").lower() == "true"
```
zerolab, line 923: this is already defined on line 37

cms/search/views.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def get_queryset(self):
+        """Returns a queryset of 'published' pages that are indexable,
+        excluding pages we do not want to index.
+        """
+        # all published pages
+        qs = Page.objects.live().specific()
+
+        # exclude pages we do not want to index
+        qs = [page for page in qs if page.specific_class.__name__ not in SEARCH_INDEX_EXCLUDED_PAGE_TYPES]
```
zerolab, line 48: something like

```python
qs = Page.objects.live().public().type(*SEARCH_INDEX_EXCLUDED_PAGE_TYPES).specific().defer_streamfields()
```

will scale much much better.

Ref: https://docs.wagtail.org/en/stable/reference/pages/queryset_reference.html#wagtail.query.PageQuerySet.type

cms/search/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path
+
+from .views import ResourceListView
+
+urlpatterns = [
+    path("v1/resources/", ResourceListView.as_view(), name="resources-list"),
```
zerolab, line 6: note: to cross-reference with the updated URL structure needs

cms/search/serializers.py
```diff
@@ -0,0 +1,80 @@
...
+                }
+                for c in obj.changes_to_release_date
+            ]
+        return []
+
+    def get_provisional_date(self, obj):
+        """If the page has 'release_date_text', return it as 'provisional_date'."""
+        if hasattr(obj, "release_date_text") and obj.release_date_text:
+            return obj.release_date_text
+        return None
```
MaciekBaron, line 80:
```suggestion
return getattr(obj, "release_date_text", None) or None
```

The `or None` at the end guarantees every falsy value to return `None`, if this is what you intended. Otherwise you can omit that.

MaciekBaron (changes requested):
Type annotations are missing, but I'm guessing this is still in progress.

cms/settings/base.py
sanjeevz3009: Removed! Residue from testing!

cms/search/serializers.py
sanjeevz3009: Sweet! Removed!

sanjeevz3009:
> Type annotations are missing, but I'm guessing this is still in progress.

@MaciekBaron Type hints added =)

cms/urls.py
```diff
@@ -82,7 +82,10 @@
         debug_urlpatterns = [path("__debug__/", include(debug_toolbar.urls)), *debug_urlpatterns]
 
 # Public URLs that are meant to be cached.
-urlpatterns: list[Union["URLResolver", "URLPattern"]] = []
+urlpatterns: list[Union["URLResolver", "URLPattern"]] = [
+    path("", include("cms.search.urls")),
+]
```
MebinAbraham, line 87: Currently, it 404s if it is not an external environment. We should change the behaviour so the route is only registered if it is external, so the endpoint doesn't even get registered in internal.

cms/search/serializers.py
```diff
@@ -0,0 +1,89 @@
+from typing import TYPE_CHECKING
```
MebinAbraham, line 1: My concern is that this is duplicating a lot of what we already have in the search publisher. Have you looked at reusing/refactoring existing code so we are not constructing the message in multiple places? Ideally, we construct in 1 place, for search and the view. Right now, they are out of sync, and changes would have to be made in both places.

MebinAbraham (changes requested):
Just initial thoughts.

zerolab:
@kacperpONS `django-admin collecstatic --noinput`

kacperpONS:
> @kacperpONS `django-admin collecstatic --noinput`

That's not it unfortunately, this is run as part of `make dev-init`. 
The templates themselves are fine, the expected page comes up when I go to `/test404` and the prod (non-debug) 500 pages that do come up are using the `500.html` template without any issues.

cms/search/serializers.py
```diff
@@ -0,0 +1,21 @@
...
+
+from rest_framework import serializers
+
+from cms.search.utils import build_resource_dict
+
+if TYPE_CHECKING:
+    from wagtail.models import Page
+
+
+class ResourceSerializer(serializers.Serializer):
```
zerolab, line 11: question: why not use `serializers.BaseSerializer`? https://www.django-rest-framework.org/api-guide/serializers/#read-only-baseserializer-classes recommnds using it for read-only cases, and you don't need to override `create`/`update`

cms/search/tests/test_api_views.py
```diff
@@ -0,0 +1,132 @@
...
+            InformationPageFactory(slug=f"Extra-Page-{i}", parent=self.index_page)
+
+        response = self.call_view_as_external(f"/v1/resources/?limit={500 + 999}")
+        self.assertEqual(response.status_code, 200)
+
+        data = self.parse_json(response)
+        self.assertEqual(len(data["items"]), 500)
+        self.assertEqual(data["count"], 500)
+        self.assertEqual(data["limit"], 500)
+        self.assertGreater(data["total_count"], 500)
```
zerolab, line 132: please override the max limit setting rather than create so many pages.

but if you need to creat things in bulk, `InformationPageFactory.create_batch(number)` ([ref](https://factoryboy.readthedocs.io/en/stable/reference.html#factory.Factory.create_batch))

cms/search/tests/test_api_views.py
```diff
@@ -0,0 +1,132 @@
...
+    @classmethod
+    def setUpTestData(cls):
+        # Pages that are excluded from the search index
+        cls.excluded_pages = [
+            ArticleSeriesPageFactory(),
+            HomePage(),
+            ReleaseCalendarIndex(),
+            ThemePageFactory(),
+            TopicPageFactory(),
+        ]
```
zerolab, line 27: Can you add a draft instance of one of the pages that would normally be included. We need a test that confirms that only _live_ pages are shown

cms/search/utils.py
```diff
@@ -0,0 +1,67 @@
...
+        "summary": get_text_for_indexing(force_str(page.summary)),
+        "title": page.title,
+        "topics": getattr(page, "topic_ids", []),
+    }
+
+
+def build_release_specific_fields(page: "Page") -> dict:
+    """Builds the extra fields that only apply to release content_type."""
+    data = {
+        "finalised": page.status in ["CONFIRMED", "PROVISIONAL"],
```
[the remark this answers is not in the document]
sanjeevz3009: Done!

cms/settings/base.py
```diff
@@ -921,3 +922,6 @@
 # FIXME: remove before going live
 ENFORCE_EXCLUSIVE_TAXONOMY = env.get("ENFORCE_EXCLUSIVE_TAXONOMY", "true").lower() == "true"
 ALLOW_TEAM_MANAGEMENT = env.get("ALLOW_TEAM_MANAGEMENT", "false").lower() == "true"
+
+DEFAULT_LIMIT_PAGE_SIZE = int(os.getenv("DEFAULT_LIMIT", "20"))
+DEFAULT_MAXIMUM_LIMIT_PAGE_SIZE = int(os.getenv("MAX_LIMIT", "500"))
```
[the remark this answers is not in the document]
sanjeevz3009: Done! Yes, Mebin mentioned this too, I used my creativity to come up with something good here now 👀

cms/search/serializers.py
sanjeevz3009: Done! Correct, remember reading this, lost it.

cms/settings/base.py
```diff
@@ -921,3 +922,6 @@
 # FIXME: remove before going live
 ENFORCE_EXCLUSIVE_TAXONOMY = env.get("ENFORCE_EXCLUSIVE_TAXONOMY", "true").lower() == "true"
 ALLOW_TEAM_MANAGEMENT = env.get("ALLOW_TEAM_MANAGEMENT", "false").lower() == "true"
+
+DEFAULT_LIMIT_PAGE_SIZE = int(os.getenv("DEFAULT_LIMIT", "20"))
+DEFAULT_LIMIT_MAXIMUM_PAGE_SIZE = int(os.getenv("MAX_LIMIT", "500"))
```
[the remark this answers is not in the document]
sanjeevz3009: Done! https://github.com/ONSdigital/dis-wagtail/pull/159#discussion_r2055865532

cms/search/tests/test_api_views.py
```diff
@@ -0,0 +1,132 @@
...
+        cls.total_resources = len(cls.pages) + 1  # +1 for the parent
+
+    def test_default_pagination_returns_first_slice(self):
+        """With no limit/ offset specified we should get DEFAULT_LIMIT (or all if fewer)."""
+        response = self.call_view_as_external("/v1/resources/")
+        self.assertEqual(response.status_code, 200)
+
+        data = self.parse_json(response)
+
+        expected_items = min(self.total_resources, 20)  # Default page size is 20
```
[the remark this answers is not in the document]
sanjeevz3009: Done. Went with the latter!

cms/search/tests/test_api_views.py
sanjeevz3009: Done!

I had to patch it, due to CustomLimitOffsetPagination.max_limit is fixed when pagination.py is first imported, and the `@override_settings` decorator by itself doesn't seem to override it. I was going to either reload the `pagination` module after the decorator has set the new value in settings, or `Monkey-patch` the frozen constant/ attribute.

Went with patching!

cms/search/tests/test_api_views.py
sanjeevz3009: Done!

cms/search/tests/test_api_views.py
```diff
@@ -0,0 +1,142 @@
...
+        self.assertEqual(data["count"], 5)
+        self.assertEqual(data["limit"], 5)
+
+    @patch("cms.search.pagination.CustomLimitOffsetPagination.max_limit", 20)
+    def test_limit_exceeds_max_uses_max(self):
+        """When limit exceeds max_limit, results should be capped at max_limit.
+        We add enough extra pages to go past MAX_LIMIT of 20.
+        """
+        InformationPageFactory.create_batch(
+            20, parent=self.index_page, slug=factory.Sequence(lambda n: f"extra_page_{n + 1}")
```
zerolab, line 132: Considering we have 13 pages created in `setupTestData`, we can patch with 10, and use that without needing to batch create even more pages? Or am I missing something obvious?

sanjeevz3009: Nope, you're not missing anything obvious.

This is something I thought of when making the other changes requested by you. I was going to either increase the number of pages in the `setupTestData` originally, but patching it with `10` seems better.

Done!

cms/search/tests/test_api_views.py
```diff
@@ -0,0 +1,138 @@
...
+        We add enough extra pages to go past MAX_LIMIT of 20.
+        """
+        response = self.call_view_as_external(f"/v1/resources/?limit={30}")
+        self.assertEqual(response.status_code, 200)
+
+        data = self.parse_json(response)
+        self.assertEqual(len(data["items"]), 10)
+        self.assertEqual(data["count"], 10)
+        self.assertEqual(data["limit"], 10)
+        self.assertEqual(data["total_count"], 13)
```
zerolab, line 138: nitpick: 
```suggestion
        self.assertEqual(data["total_count"], self.total_resources)
```

other than that LGTM

cms/search/pagination.py
```diff
@@ -0,0 +1,26 @@
...
+    default_limit = settings.SEARCH_API_DEFAULT_PAGE_SIZE  # Default number of items per page
+    max_limit = settings.SEARCH_API_MAX_PAGE_SIZE  # Maximum number of items per page
+
+    def get_paginated_response(self, data: list) -> Response:
+        """Override DRF's default so we output the keys required by
+        dp-standards.
+        """
+        return Response(
+            {
+                "count": len(data),  # number in this slice
```
MaciekBaron, line 20: Huge nit-pick but I think this reads better
```suggestion
                "count": len(data),  # size of this slice
```

cms/search/tests/test_api_views.py
```diff
@@ -0,0 +1,138 @@
...
+            self.total_resources, settings.SEARCH_API_DEFAULT_PAGE_SIZE
+        )  # Default page size is 20 currently
+        self.assertEqual(len(data["items"]), expected_items)
+        self.assertEqual(data["count"], expected_items)
+        self.assertEqual(data["total_count"], self.total_resources)
+        self.assertEqual(data["offset"], 0)
+
+    def test_second_slice_returns_remaining_items(self):
+        """Requesting offset=<7> should return whatever is left after the first slice."""
+        response = self.call_view_as_external(f"/v1/resources/?offset={7}")
```
MaciekBaron, line 108: Do we need an f-string here? Here and on line 131. I see a mixture of approaches (e.g. line 118)

```suggestion
        response = self.call_view_as_external("/v1/resources/?offset=7")
```

MaciekBaron (approved):
Apart from two minor things, LGTM!

cms/search/tests/test_api_views.py
sanjeevz3009: Done!

cms/search/pagination.py
sanjeevz3009: Done!

cms/search/tests/test_api_views.py
sanjeevz3009: Done!

I have a `RESOURCE_ENDPOINT = "/v1/resources/"` constant and I now utilise that consistently in the whole file =)

MaciekBaron: Thanks, do we still need brackets for the integers though? Is that for readability?

sanjeevz3009: It was initally because it was _readable/ stuck out to me_ but doing some quick google search on the best practices, standards, and the pythonic way. `{}` in f-strings shouldn't be used for this and this is not idiomatic.

So went with the style of `response = self.call_view_as_external(f"{RESOURCE_ENDPOINT}?offset=7")` etc

kacperpONS (approved):
LGTM, don't forget to add the change dev settings file we talked about 😌

sanjeevz3009:
> LGTM, don't forget to add the change in the dev settings file we talked about 😌

@kacperpONS Done!

kacperpONS (approved):
All good, the last change fixes the issue I mentioned in https://github.com/ONSdigital/dis-wagtail/pull/159#issuecomment-2820922249 👍

zerolab (approved):

cms/search/utils.py
```diff
@@ -0,0 +1,69 @@
...
+if TYPE_CHECKING:
+    from wagtail.models import Page
+
+
+def build_standard_resource_dict(page: "Page") -> dict:
+    """Returns a dict with the standard resource fields.
+    This covers the non-release case (and also forms the base of the release case).
+    """
+    return {
+        "uri": page.url_path,
```
MebinAbraham, line 18: This currently contains the full path which includes the `home`path; we need to omit that. So instead of `"/home/releases/jan-2025/"` it should be `"/releases/jan-2025/"`. See: https://github.com/ONSdigital/dis-search-upstream-stub/blob/main/data/json_files/release_two_topics.json

MebinAbraham (approved):
Just the one comment to address, otherwise LGTM.

cms/search/utils.py
sanjeevz3009: Good spot! Done!

cms/search/utils.py
```diff
@@ -0,0 +1,70 @@
...
+if TYPE_CHECKING:
+    from wagtail.models import Page
+
+
+def build_standard_resource_dict(page: "Page") -> dict:
+    """Returns a dict with the standard resource fields.
+    This covers the non-release case (and also forms the base of the release case).
+    """
+    return {
+        "uri": page.url_path.removeprefix("/home"),
```
zerolab, line 18: I would not rely on the prefix being `/home/`

Instead:
```pycon
>>> s = "/home/releases/psf-december-2024-release/"
>>> s.strip("/").split("/", 1)
['home', 'releases/psf-december-2024-release']
>>> s.strip("/").split("/", 1)[-1]
'releases/psf-december-2024-release'
>>> "/home/".strip("/").split("/", 1)[-1]
'home'
```

so:

```suggestion
    path = page.url_path.strip("/").split("/", 1)[-1]

    return {
        "uri": f"/{path}" if not getattr(settings, "WAGTAIL_APPEND_SLASH", True) else f"/{path}/",
```

with a `from django.conf import settings` import

cms/search/utils.py
```diff
@@ -0,0 +1,70 @@
...
+        "content_type": page.search_index_content_type,
+        "release_date": (page.release_date.isoformat() if getattr(page, "release_date", None) else None),
+        "summary": get_text_for_indexing(force_str(page.summary)),
+        "title": page.title,
+        "topics": getattr(page, "topic_ids", []),
+        "language": force_str(page.locale.get_display_name()),
```
zerolab, line 24:
```suggestion
        "language": force_str(get_locales_display_names().get(page.locale_id)),
```

with `from wagtail.coreutils import get_locales_display_names` - this makes use of the cached locales display names, and reduces the need to additional queries/joins on the locale table.

cms/settings/local.py.example
```diff
@@ -2,9 +2,6 @@
 # uncomment to use memory cache rather than the Redis service (from docker compose)
 # CACHES["default"]["BACKEND"] = "django.core.cache.backends.locmem.LocMemCache"
 
-# Django Debug Toolbar
```
zerolab, line 5: I would still leave this as an option, should one want to disable the debug toolbar locally. thiks means `dev.py` needs to retain `SHOW_TOOLBAR = True # Override in `local.py`` and `"SHOW_TOOLBAR_CALLBACK": lambda x: SHOW_TOOLBAR`,

zerolab (changes requested):
Final tweaks and we should be good to go

cms/search/utils.py
sanjeevz3009: Done!

PR #160, opened 2025-04-15
Improve Broker Service Container Config To Enable Search Integration Tests To Run Locally and Within The Web Container.

sanjeevz3009:

It was missed during the [search integration: kafka events](https://github.com/ONSdigital/dis-wagtail/pull/115) work that the search integration tests only worked locally and not when run within the web container.

This PR addresses that issue.

### How to review

- Run the search tests `locally` -> `make compose-dev-pull` -> `make compose-dev-up` -> `./manage.py test cms.search --settings=cms.settings.test --keepdb`.

- Run within the `web container` -> `make compose-pull` -> `make compose-up` -> `./manage.py test cms.search --settings=cms.settings.test --keepdb`.

Everything should be passing, including the `search integration tests` that interact with the `broker container`.

### Follow-up Actions


cms/settings/local.py.example
```diff
@@ -4,5 +4,13 @@
...
- # PREVIOUS_RELEASES_PER_PAGE = 3
+# Override size of the pagination used on the previous_release page.
+PREVIOUS_RELEASES_PER_PAGE = 3
+
+# Search Integration Kafka Events Env Vars
+SEARCH_INDEX_PUBLISHER_BACKEND = "kafka" # if this env var is not set to "Kafka" then LogPublisher will be used
+KAFKA_SERVER = "localhost:19092"
+KAFKA_CHANNEL_CREATED_OR_UPDATED = "search-content-updated"
+KAFKA_CHANNEL_DELETED = "search-content-deleted"
+KAFKA_API_VERSION = tuple(map(int, "3,5,1".split(",")))
```
zerolab, line 15: This should be at most `KAFKA_API_VERSION = (3, 5, 1)` here. `base.py` already has 3.5.1 as a default, so best to leave it out.

cms/settings/local.py.example
```diff
@@ -4,5 +4,13 @@
 # Django Debug Toolbar
 SHOW_TOOLBAR = True
 
- # Override size of the pagination used on the previous_release page.
- # PREVIOUS_RELEASES_PER_PAGE = 3
+# Override size of the pagination used on the previous_release page.
+PREVIOUS_RELEASES_PER_PAGE = 3
+
+# Search Integration Kafka Events Env Vars
```
zerolab, line 10:
```suggestion
# Search Integration Kafka configuration
```

cms/settings/local.py.example
```diff
@@ -4,5 +4,13 @@
...
+SEARCH_INDEX_PUBLISHER_BACKEND = "kafka" # if this env var is not set to "Kafka" then LogPublisher will be used
```
zerolab, line 11:
```suggestion
SEARCH_INDEX_PUBLISHER_BACKEND = "kafka"  # comment this out to use LogPublisher
```

cms/search/tests/test_kafka_integration.py
```diff
@@ -8,9 +9,11 @@
 from cms.search.publishers import KafkaPublisher
 from cms.standard_pages.tests.factories import InformationPageFactory
 
+KAFKA_SERVER = os.environ.get("KAFKA_SERVER", "localhost:19092")
```
zerolab, line 12: question: why this change?

We have it `base.py` reading from the env. Let's leave it at that, and remove from `override_settings` below.

.github/workflows/ci.yml
```diff
@@ -122,7 +122,7 @@ jobs:
           # Enable auto-creation of topics
           KAFKA_CFG_AUTO_CREATE_TOPICS_ENABLE: 'true'
         ports:
-          - '9092:9092'
+          - '19092:19092'
```
zerolab, line 125: question: why can't it remain on 19092? The controller is still on 9093

zerolab (changes requested):
I feel this is needlessly complicated.
Given `docker-compose-dev.yml` is used for local dev, and `docker-compose.yml` is for container dev. Why not set the `KAFKA_SERVER` env var as `localhost:9092` in `docker-compose-dev.yml` and `broker:9092` in `docker-compose.yml` ?

sanjeevz3009:
> I feel this is needlessly complicated. Given `docker-compose-dev.yml` is used for local dev, and `docker-compose.yml` is for container dev. Why not set the `KAFKA_SERVER` env var as `localhost:9092` in `docker-compose-dev.yml` and `broker:9092` in `docker-compose.yml` ?

@zerolab Correct! I got carried away with my dual-listener setup. I just did a simple one-listener broker in both compose files.

> [!NOTE]
> The simplified setup won't work if we need both the local host and web container clients at the same time. So we'll be able to build the local dev containers (redis, postgres, broker) and run the tests locally, or build full-stack containers (redis, postgres, broker, web container), and the tests will work in the web container too, but we will not be able to run the tests locally if we spin up the web container.

cms/settings/local.py.example
sanjeevz3009: Removed!

cms/settings/local.py.example
sanjeevz3009: Done!

cms/settings/local.py.example
sanjeevz3009: Done!

.github/workflows/ci.yml
sanjeevz3009: Irrelevant now, as I have simplified the config.

cms/search/tests/test_kafka_integration.py
sanjeevz3009: Residue! Was testing something. Removed!

cms/search/tests/test_kafka_integration.py
```diff
@@ -25,15 +25,15 @@ def setUpClass(cls):
         # so we can verify messages that come in.
         cls.consumer_created = KafkaConsumer(
             "search-content-updated",
-            bootstrap_servers=["localhost:9092"],
+            bootstrap_servers=settings.KAFKA_SERVER,
```
zerolab, line 28: q: does this not need a list?

sanjeevz3009: Per their docs: No, it can be a string or a list of strings. It just needs at least one broker that will respond to the meta data API request. But to keep it consistent with rest of the search work in the codebase, I updated this now to a list.

zerolab:
@AWitcherONS can you please test this PR to confirm it fixes the test issues you were seeing?

cms/settings/local.py.example
```diff
@@ -4,5 +4,12 @@
...
 SHOW_TOOLBAR = True
 
- # Override size of the pagination used on the previous_release page.
- # PREVIOUS_RELEASES_PER_PAGE = 3
+# Override size of the pagination used on the previous_release page.
+PREVIOUS_RELEASES_PER_PAGE = 3
+
+# Search Integration Kafka configuration
+# SEARCH_INDEX_PUBLISHER_BACKEND = "kafka" # comment this out to use LogPublisher
+KAFKA_SERVER = "localhost:19092"
```
kacperpONS, line 12:
```suggestion
KAFKA_SERVER = "localhost:9092"
```

sanjeevz3009: Good spot! Updated!

sanjeevz3009:
> @AWitcherONS can you please test this PR to confirm it fixes the test issues you were seeing?

@zerolab @AWitcherONS This PR only fixes the issue of Kafka integration tests previously not being able to run within the web container.

As @MebinAbraham mentioned in stand-up, other issues related to do with the parallel test running (not supported at the moment on Mac OS, I think) and other unit/ integration tests failing locally will be investigated later on, as those are issues that have been ongoing for a while.

docker-compose.yml
```diff
@@ -50,6 +51,9 @@ services:
     extends:
       file: docker-compose-dev.yml
       service: broker
+    environment:
+      KAFKA_CFG_ADVERTISED_LISTENERS: 'PLAINTEXT://broker:9092'
+      KAFKA_CFG_CONTROLLER_QUORUM_VOTERS: '1@broker:9093'
```
kacperpONS, line 56: Since we're always running Kafka in a container, we could change these two settings in the `docker-compose-dev.yml` (which this file is extending) to use the `broker` container directly instead of `localhost`.
We could then remove these overwrites in this file and it would also eliminate the need to restart the container when switching from testing locally to inside the container.

sanjeevz3009: If I am understanding this right, this won't work.

`Since we're always running Kafka in a container` - I am not sure if there is a misunderstanding, the search integration Kafka work, all we are doing is producing and sending events as part of our web application to a Kafka consumer. We don't have a Kafka container per se. We have a broker container for integration testing purposes.

As if we go with the suggestion here:
- When the Kafka client tries to reach the broker in the web container -> it can resolve broker and docker DNS will point at the broker container, so Kafka integration tests will work in the web container.

But
- When we are running the web application on the host -> it won't be able to resolve broker, so the Kafka integration tests won't work locally.

That is why the base file `docker-compose-dev.yml` advertises localhost, and the full‑stack file overrides it to broker.

This suggestion would only work if we always want to run inside Docker.

kacperpONS: My bad, went ahead with that suggestion too quickly - it won't be able resolve the container address as the web app would be outside of the Docker network.

AdamHawtin (approved):
LGTM, works both locally and in the container without manual configuration 👍

MebinAbraham (approved):
Working as expected. 👍

MaciekBaron (approved):
LGTM!

docker-compose.yml
sanjeevz3009: Precisely. And apologies for the confusing terms being thrown around, `Kafka`, `broker`, etc. I have to be careful and not use them interchangeably. As either I am confusing someone or myself with the terms. I/ we should be using the terms more carefully as there is `Apache Kafka` platform itself and then it's architecture and components such as  `Producer`, `Topic`, `Partition`, `Broker ( Kafka server)`, `Consumer` etc.

PR #161, opened 2025-04-16
Log IP addresses

RealOrangeOne:

As an extension to #135, this PR adds IP address logging for both general access logs and auth events.

The IP address is correctly parsed from the [`X-Forwarded-For`](https://theorangeone.net/posts/x-forwarded-for/) header, correctly reading the correct value based on the environment.

The IP address is only extracted on the internal instance - the external instance will not log the user's IP (and will raise an error if someone tries to use the util.

### Follow-up Actions

Configuring the internal environment to read the correct value is out of scope of this PR, but the required hooks are exposed as environment variables.


sanjeevz3009 (approved):
LGTM!

MaciekBaron (approved):
Apart from that non-blockable suggestion LGTM!

RealOrangeOne:
Following some discussions, we concluded that the logging is only necessary in the internal environment - external shouldn't log IP addresses, and doesn't have a need to anyway.

MaciekBaron (approved):

sanjeevz3009 (approved):

PR #163, opened 2025-04-17
feat(auth): add AWS Cognito JWT authentication support

MebinAbraham:

This PR integrates Wagtail CMS with the existing AWS Cognito authentication. It allows users from the existing Cognito Pool to log into Wagtail CMS using Access and ID Tokens provisioned by [AWS Cognito](http://aws.amazon.com/cognito/features/).

**Key Changes**:
- Added middleware to handle JWT token validation and user authentication
- Added client-side JavaScript for session management and renewal using https://github.com/ONSdigital/dis-authorisation-client-js)
- Extended User model with a unique `user_id` field, given that a user's email may change. Includes a migration script to handle existing users who may not have a `user_id`.
- Makes the Cognito login flow the default mechanism for login. Wagtail login is feature-flagged using the `WAGTAIL_CORE_ADMIN_LOGIN_ENABLED` env var.

Useful materials in relation to this PR:
- [How to Integrate with existing Auth](https://confluence.ons.gov.uk/display/DIS/How+to+integrate+with+auth)
- [AWS Cognito Documentation](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools.html)
- JWT: https://jwt.io/

> [!NOTE]  
> When running the auth stub locally, you’ll need to access Wagtail through the proxy (stub url) provided by the stub. This is necessary due to how authentication works, specifically, browser security restrictions prevent cross-origin access to local storage, which the stub relies on. If you try to access Wagtail using the usual route (e.g. localhost:8000) while also logging in via the stub, you’ll be logged out automatically. 

#### Set up the authentication stub to mock Florence login flow.

1. Clone the stub: https://github.com/ONSdigital/dis-authentication-stub
2. Switch to branch `feature/fallback-to-wagtail`
3. Run the stub using `make debug`. You will need to have `go` installed.

#### Run Wagtail

1. Take a DB snapshot on main branch so you can test other branches even after applying migrations in this. `poetry run dslr snapshot latest-main`.
1. Check out this branch
2. Run migrations via `make migrate`
3. Install node dependencies by running `npm install`, then `npm run build:prod `
4. OPTIONAL: Set the `WAGTAILADMIN_HOME_PATH = "wagtail-admin/"` in local.py if you want to mimic Florence redirect path name. However, for local development,, this is not important, and functionally, the auth stub has been updated to not care about the wagtail admin path. Note if you set this, make sure to update this string in any following instructions accordingly.
5. Start Wagtail
6. Go to http://localhost:29500/florence/login?redirect=http://localhost:29500/admin/ and log in as one of the users. You can see which roles/groups users have access to by looking at the [users.json](https://github.com/ONSdigital/dis-authentication-stub/blob/feature/fallback-to-wagtail/static/json/users.json.)
7. Test various login and log-out flows and ensure everything is as expected.
8. Key things to test:
    - Attempt access with missing or expired tokens (user should be logged out)
    - Validate behaviour when only one of the tokens (access or ID) is present – user should be logged out
    - Users are assigned the right Django User Groups (Publishing Admins, Publishing Officers, Viewers) etc
    - Users are assigned the right Preview teams
    - User details in Wagtail are updated as per [users.json](https://github.com/ONSdigital/dis-authentication-stub/blob/feature/fallback-to-wagtail/static/json/users.json)
    - Users roles/teams are updated if the source changes (make changes to the [users.json](https://github.com/ONSdigital/dis-authentication-stub/blob/feature/fallback-to-wagtail/static/json/users.json.)
    - Session is refreshed near expiry (TODO: Fix bug in auth stub which returns incomplete JWT tokens) (Hold off testing this)
    - When Cognito login is enabled, ensure a Cognito user cannot log in via the normal login form.
    - Set `WAGTAIL_CORE_ADMIN_LOGIN_ENABLED` to `True` and ensure the default Wagtail login is disabled.
    - Ensure if the access/id_token is deleted once logged in, the user is logged out.
    - Ensure a Coginto user cannot be logged into Wagtail without the access/id token being present.
    - Check the console log to see that session management via JS is working as expected. (Read the https://github.com/ONSdigital/dis-authorisation-client-js README)
    - Ensure JS session management is not duplicated when content is iframed.
    - Verify AC's on the card.

Test as many flows as possible; there are many more not called out. For some flows, you may need to make changes to the auth stub to mimic unhappy paths.

### Follow-up Actions

- [ ] Update auth stub to handle correct JWT generation for refresh
- [ ] Add tests


cms/auth/middleware.py
```diff
@@ -0,0 +1,169 @@
...
+
+        # Check if the session is up-to-date.
+        access_jti = access_payload["jti"]
+        id_jti = id_payload["jti"]
+        if request.user.is_authenticated and request.session.get(JWT_SESSION_ID_KEY) == f"{access_jti}{id_jti}":
+            # The session is up-to-date.
+            return
+
+        # Update the session with token identifiers.
+        request.session[JWT_SESSION_ID_KEY] = f"{access_jti}{id_jti}"
```
MaciekBaron, line 70: Small suggestion, as we only use the two variables for concatenation:

```suggestion
        jwt_session_key = f"{access_payload['jti']}{id_payload['jti']}"
        if request.user.is_authenticated and request.session.get(JWT_SESSION_ID_KEY) == jwt_session_key:
            # The session is up-to-date.
            return

        # Update the session with token identifiers.
        request.session[JWT_SESSION_ID_KEY] = jwt_session_key
```

cms/settings/base.py
```diff
@@ -917,3 +919,42 @@
...
+# Groups
+PUBLISHING_ADMIN_GROUP_NAME = "Publishing Admins"
+PUBLISHING_OFFICER_GROUP_NAME = "Publishing Officers"
+VIEWERS_GROUP_NAME = "Viewers"
+ROLE_GROUP_IDS = {"role-admin", "role-publisher"}
+
+# Cookie Names
+ACCESS_TOKEN_COOKIE_NAME = "access_token"  # noqa: S105
+REFRESH_TOKEN_COOKIE_NAME = "refresh_token"  # noqa: S105
+ID_TOKEN_COOKIE_NAME = "id_token"  # noqa: S105
```
MaciekBaron, line 956: Maybe we can add a comment here saying that this name is used explicitly in the JS, and so if it gets changed, the JS needs to be updated? Unless we think this is so unlikely that we don't need to bother.

MaciekBaron (commented):
Just had a first quick read-through and nothing major stood out to me but will have another look soon.

.gitignore
```diff
@@ -186,7 +186,12 @@ copier_scripts/
 /cms/static_compiled
 
 node_modules/
-/static/
+# Ignore the top-level "static" directory
+/static
+
+# Don't ignore "static/" within subdirectories
+!/*/static/
```
MaciekBaron, line 193: I think this may be redundant? We already have folders names `static` which are not ignored, as the first rule is only applied to the root because of the leading slash.

cms/auth/middleware.py
```diff
@@ -0,0 +1,169 @@
...
+        user, created = User.objects.get_or_create(
+            user_id=user_id,
+            defaults={"email": email, "username": email},
+        )
+
+        # Update user details.
+        user.update_details(email=email, first_name=first_name, last_name=last_name, created=created)
+
+        # Assign groups if provided.
+        groups_ids = id_payload.get("cognito:groups") or []
```
MaciekBaron, line 166: ***Question***: Usually with JWT we want to separate authentication and authorisation (i.e. the token is used for identification, but not to define what the user is able to do within the system). If the groups have associated privileges, this enters authorisation territory. Consider a situation, where a user had a group assigned when the token was issued, but since then was removed from said group. This could lead to escalated privileges. I know this is an extreme edge case as the expiry time of the token is short and it's very difficult to exploit, but just wanted to check if this was considered or if it is a problem at all?

cms/auth/utils.py
```diff
@@ -0,0 +1,114 @@
...
+    extra_fields = ["username", "client_id"] if token_type == "access" else ["cognito:username", "email"]  # noqa: S105
+
+    try:
+        return _validate_jwt(token, extra_fields=extra_fields, token_type=token_type)
+    except ExpiredSignatureError:
+        logger.exception("Token has expired", extra={"token_type": token_type})
+    except InvalidTokenError:
+        logger.exception("Invalid token", extra={"token_type": token_type})
+    except Exception:  # pylint: disable=broad-except
+        logger.exception("Error decoding token", extra={"token_type": token_type})
```
MaciekBaron, line 57: To aid debugging, should we also consider two more exceptions here:

- **requests.exceptions.HTTPError** which can be raised by line 40
- **json.decoder.JSONDecodeError** which can be raised by line 41

This is just so that the logged information is a bit clearer.

cms/settings/base.py
```diff
@@ -804,6 +802,10 @@
 if "WAGTAILADMIN_BASE_URL" in env:
     WAGTAILADMIN_BASE_URL = env["WAGTAILADMIN_BASE_URL"]
 
+# https://docs.wagtail.org/en/latest/reference/settings.html#wagtailadmin-login-url
+if "WAGTAILADMIN_LOGIN_URL" in env:
+    WAGTAILADMIN_LOGIN_URL = env["WAGTAILADMIN_LOGIN_URL"]
```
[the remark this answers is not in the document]
MebinAbraham: Updated instructions. I had `DJANGO_SETTINGS_MODULE=cms.settings.dev` set hence I didn't see the error. Will add to the `checks.py` as this is not a required env var.

cms/auth/middleware.py
MebinAbraham: Unfortunately, this is a limitation of how AWS Cognito JWTs function and how the ONS has chosen to implement group-based privileges. This concern was previously raised, but the risk was accepted due to the short token expiry time. It's important to note that tokens are not actively revoked once issued, either. However, this too has been reviewed and approved by Information Assurance (IA) as an acceptable risk, again due to the short lifespan of the tokens.

That said, the authentication service does provide a way to force a logout for all users by revoking all tokens globally. This approach is used in the event of a security incident.

cms/auth/middleware.py
MebinAbraham: Addressed in [`19cd39d` (#163)](https://github.com/ONSdigital/dis-wagtail/pull/163/commits/19cd39de350fcc17519fe05f42b38c54af137d7a)

cms/settings/base.py
MebinAbraham: I've made it use the settings value, so it is not out of sync. Though in the JS lib, it does hardcode it, so we would need to change it in the app if the JS lib changes, but at least it is in one place for us to change now. [`19cd39d` (#163)](https://github.com/ONSdigital/dis-wagtail/pull/163/commits/19cd39de350fcc17519fe05f42b38c54af137d7a)

.gitignore
MebinAbraham: Revert, can't remember why I changed this. [`19cd39d` (#163)](https://github.com/ONSdigital/dis-wagtail/pull/163/commits/19cd39de350fcc17519fe05f42b38c54af137d7a)

webpack.config.js
```diff
@@ -12,6 +12,7 @@ const projectRoot = 'cms';
 const options = {
   entry: {
     main: `./${projectRoot}/static_src/javascript/main.js`,
+    auth: `./${projectRoot}/auth/static/js/auth.js`,
```
RealOrangeOne, line 15: Issue: I think these should stay in the `static_src` dir, so the scripts are kept together, and linters definitely run.

cms/auth/middleware.py
```diff
@@ -0,0 +1,168 @@
...
+        id_token = request.COOKIES.get(settings.ID_TOKEN_COOKIE_NAME)
+        if not access_token or not id_token:
+            self._handle_unauthenticated_user(request)
+            return
+
+        # Validate the tokens.
+        access_payload = validate_jwt(access_token, token_type="access")  # noqa: S106
+        id_payload = validate_jwt(id_token, token_type="id")  # noqa: S106
+        if not access_payload or not id_payload:
+            logger.info("Invalid or expired JWT tokens; logging out user.")
```
RealOrangeOne, line 50: Question: How does this interact with the rest of Django's auth system, such as the auth signals?

cms/auth/middleware.py
```diff
@@ -0,0 +1,168 @@
...
+if TYPE_CHECKING:
+    from django.http import HttpRequest
+
+logger = logging.getLogger(__name__)
+User = get_user_model()
+
+JWT_SESSION_ID_KEY = "jwt_session_id"
+
+
+class ONSAuthMiddleware(AuthenticationMiddleware):
```
RealOrangeOne, line 20: Question: Does this need to be a custom implementation, could something like this be used? https://docs.allauth.org/en/latest/socialaccount/providers/amazon_cognito.html

cms/auth/static/js/auth.js
RealOrangeOne, line 1: Thought: Given the importance of auth, should this be typescript?

cms/auth/wagtail_hooks.py
```diff
@@ -0,0 +1,30 @@
...
+        return
+
+    @hooks.register("insert_global_admin_js")
+    def global_admin_auth_js() -> SafeString:
+        """Register the auth.js script in the Wagtail admin to handle refresh logic."""
+        return format_html(
+            """
+            <script>
+                window.authConfig = {config};
+            </script>
```
RealOrangeOne, line 22: Issue: This should use the `json_script` template filter, rather than injecting the config in like this, to avoid injection and escaping issues.

cms/jinja2/templates/base_page.html
```diff
@@ -68,4 +68,12 @@ <h1 class="ons-u-fs-3xl common-header__heading">
 {% block scripts %}
     {{ super() }}
     {% include "templates/partials/content_scripts.html" %}
+
+    <!-- Load JS to refresh user session if required -->
+    {% if AWS_COGNITO_LOGIN_ENABLED and not IS_EXTERNAL_ENV and request.user.is_authenticated %}
+        <script>
+            window.authConfig = {{ AUTH_CONFIG|safe }};
```
RealOrangeOne, line 75: Issue: This needs to use the `json_script` template tag to avoid escaping issues.

cms/auth/static/js/auth.js
```diff
@@ -0,0 +1,131 @@
...
+  try {
+    const response = await fetchWithCsrf(logoutURL);
+    if (response?.ok) {
+      const redirectURL = response.redirected ? response.url : origin;
+      console.info('[WAGTAIL] Logout successful. Redirecting to:', redirectURL);
+      window.location.href = redirectURL;
+    } else {
+      console.error('[WAGTAIL] Logout failed:', response);
+      // Reload the page to handle any session inconsistencies.
+      window.location.reload();
```
MaciekBaron, line 74: I'm thinking that any developer will need to enable persistent logs to see the error in the console after reloading the window. If they don't have it enabled they might not notice it. However I do understand why you might want to reload the window.

.prettierignore
```diff
@@ -7,7 +7,6 @@ venv
 .mypy_cache
 .coverage
 media
-static
```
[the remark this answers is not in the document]
sanjeevz3009: Done

cms/auth/__init__.py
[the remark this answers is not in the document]
sanjeevz3009: No longer relevant!

cms/auth/static/js/auth.js
sanjeevz3009: Can write a follow-up card for this, if this is desired. I am not too keen on changing it to TS as part of this PR.

webpack.config.js
sanjeevz3009: Done!

cms/auth/wagtail_hooks.py
sanjeevz3009: Great spot! Done!

sanjeevz3009: Closes the XSS hole and passes CSP/ XSS audit.

I added deferred-loading auth.js speeds things up.

cms/jinja2/templates/base_page.html
sanjeevz3009: Done!

cms/auth/middleware.py
sanjeevz3009: From my understanding `django.contrib.auth.logout()` is called.

The auth signal `logout()` automatically emits `user_logged_out` before it flushes the session

Side-effects: `request.user` is reset to AnonymousUser, CSRF token is rotated, etc. The same behaviour you’d get if the user clicked a `Log out` button per se.

Successful tokens, but session not really `fresh`

If the jti combo stored in the session (JWT_SESSION_ID_KEY) doesn’t match the pair in the received tokens, `django.contrib.auth.login()` persists the user id & backend in the session.

`login()` emits `user_logged_in` (which in turn triggers Django’s built-in update_last_login receiver)
 
Because we only do this when the `jti` changes, the signal is sent once per Cognito session, not on every HTTP request.

If it's an already-up-to-date session, if the combined jti in the session does match the tokens, we exit early, no call to `login()`, therefore no additional signals.

For `user_login_failed`

That signal is sent by `authenticate()` when credentials are rejected. Because our path never calls `authenticate()`, invalid or expired JWTs won’t emit `user_login_failed`.

cms/auth/middleware.py
sanjeevz3009: Not sure if @MebinAbraham looked into this and saw this as an option, I am assuming he did. But looking into this:

I think in our architecture we really do need a custom middleware.

I think when using allauth, I don't think it can validate or refresh those cookies for us that we would be receiving from Florence/ users will be logging in from a another application/ UI, and we would still have to write essentially the same middleware we already have. Each request to our application. The browser sends the two JWT cookies; our middleware must check signature, exp, aud, client_id, jti pairing.

Where user groups / Wagtail teams are mapped	_authenticate_user(); customised for CMS logic.	Would need extra adapter overrides anyway in allauth.

Additionally, allauth would also have so many apps coming with it, templates and migrations without removing some of the middleware we already have.

Switching to allauth would also mandate a new login/ logout endpoints and template tweaks probably on the Florence login/ the application the user would be using to login before being authenticated into Wagtail.

We rely on has_usable_password() checks and a feature-flag fallback to the Wagtail admin login when Cognito is off. Those behaviours are outside allauth’s scope but are already covered in the custom code.

Our middleware fires user_logged_in once per new jti pair, preventing log spam. With allauth we would have to suppress redundant signals manually.

I think our custom middleware does that job with fewer moving parts and without introducing a heavyweight dependency.

@MaciekBaron @zerolab

cms/auth/static/js/auth.js
sanjeevz3009: Yes, I am inclined to keep this as it is.

cms/auth/__init__.py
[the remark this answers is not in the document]
RealOrangeOne: Why is it no longer relevant? There are still 2 separate apps.

cms/settings/base.py
```diff
@@ -946,3 +949,42 @@
...
+# if AWS_COGNITO_LOGIN_ENABLED:
+#     required_env_vars = [
+#         "IDENTITY_API_BASE_URL",
+#         "AUTH_TOKEN_REFRESH_URL",
+#         "AWS_COGNITO_APP_CLIENT_ID",
+#         "AWS_COGNITO_USER_POOL_ID",
+#     ]
+#     for var in required_env_vars:
+#         if not env.get(var):
+#             raise ImproperlyConfigured(f"Missing required environment variable: {var}")
```
zerolab, line 974: @sanjeevz3009 this is left to do

sanjeevz3009: I was going to do this as a follow-up PR straight after this and write some tests for it too. Thought it would be cleaner and easier to get this in first and do a follow-up straight after.

cms/auth/__init__.py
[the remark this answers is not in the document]
sanjeevz3009: Oops, sorry, I have been coding away in this PR, got this confused with something else.

I do see the reasons to merge `auth` with `users`, such as the project simplicity, it may be overengineering to split them, tight coupling may seem like a good idea, etc.

 I am more eager to keep the `auth` as a separate app per the current setup. This is due to the separation of concerns, mainly.
 
 - `users` should manage user models, permissions, and profile logic, and their respective tests can sit together.
- `auth` can encapsulate JWT handling, middleware, token validation, auth views, and login/ logout flows, their respective tests can sit together.

As with a lot of mature Django applications, they keep authentication and user management as separate concerns, such as `django-allauth`.

For now, I think keeping `auth` and `users` separate will help us maintain clear boundaries between authentication mechanisms (e.g., middleware, JWT validation) and user management (models, permissions, viewsets). This follows Django's modular design principles also.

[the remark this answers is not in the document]
zerolab: Let's keep them separate

sanjeevz3009:
> @sanjeevz3009 with the functional tests:
> 
> * let's limit the external auth to the auth feature
> * for all others set `AWS_COGNITO_LOGIN_ENABLED = False`

@zerolab Done!

cms/teams/viewsets.py
```diff
@@ -41,16 +41,9 @@ class TeamsViewSet(ModelViewSet):
     add_to_admin_menu = True
     menu_order = 200
     inspect_view_enabled = True
-    list_display: ClassVar[list[str]] = ["name", "total_members", "created_at", "updated_at"]
```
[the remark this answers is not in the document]
zerolab: Let's re-add them - they are useful as an overview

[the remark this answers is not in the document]
sanjeevz3009: Reverted!

sanjeevz3009:
Went through all the 2,763 lines of code changes, including the tests, etc, just to make sure of good test coverage, I am happy to say, all is good from my eyes  =)

cms/settings/base.py
```diff
@@ -828,6 +826,11 @@
 if "WAGTAILADMIN_BASE_URL" in env:
     WAGTAILADMIN_BASE_URL = env["WAGTAILADMIN_BASE_URL"]
 
+# https://docs.wagtail.org/en/latest/reference/settings.html#wagtailadmin-login-url
+# if "WAGTAILADMIN_LOGIN_URL" in env:
+#     WAGTAILADMIN_LOGIN_URL = env["WAGTAILADMIN_LOGIN_URL"]
+WAGTAILADMIN_LOGIN_URL = env.get("WAGTAILADMIN_LOGIN_URL", "/admin/login/")
```
418error, line 832: An observation around naming conventions and the logic maybe my misunderstanding
`WAGTAILADMIN_LOGIN_URL` appears to be a path rather than a URL
However in the dev.py it appears to be fully qualified URL
```python
WAGTAILADMIN_LOGIN_URL = env.get("WAGTAILADMIN_LOGIN_URL", f"{AUTH_STUB_URL}/florence/login"
```

sanjeevz3009: Good observation.

You're correct that the `WAGTAILADMIN_LOGIN_URL` env in `dev.py` is set to a fully qualified URL `WAGTAILADMIN_LOGIN_URL = env.get("WAGTAILADMIN_LOGIN_URL", f"{AUTH_STUB_URL}/florence/login"` (`AUTH_STUB_URL = "http://localhost:29500"`) and `base.py`/ the main settings we default to a path: `WAGTAILADMIN_LOGIN_URL = env.get("WAGTAILADMIN_LOGIN_URL", "/admin/login/")`.

The difference reflects the environments:

For local development and testing purposes, we use the external authentication stub (`AUTH_STUB_URL`, we spin up `dis-authentication-stub`) to simulate a separate a sign-in service (e.g. `Florence` or something). That service redirects back to Wagtail after login, and the full URL is needed for the correct redirection and local cookie flow.

In production, Wagtail's built in login path (`/admin/login/`) is used directly hence the path-only fallback.

You're also right that the name `WAGTAILADMIN_LOGIN_URL` might imply a full URL. However, this is the name expected by [Wagtail](https://docs.wagtail.org/en/stable/reference/settings.html#wagtailadmin-login-url), and it accepts either a path or full URL.

If needed, we can improve clarity by updating documentation and or in-code comments to highlight this behaviour explicitly.

RealOrangeOne: For context, the URL here is used as the value of the `Location` header when doing a redirect, so either an absolute path or URL is valid.

cms/static_src/javascript/auth.js
```diff
@@ -0,0 +1,153 @@
...
+const logoutURL = `${origin}/${wagtailAdminHomePath}logout/`;
+const extendSessionURL = `${origin}/${wagtailAdminHomePath}extend-session/`;
+
+function getCookieByName(name) {
+  console.debug('[WAGTAIL] Getting cookie by name:', name);
+  const cookies = document.cookie;
+  if (!cookies) {
+    return null;
+  }
+  const cookie = cookies
```
RealOrangeOne, line 41: Suggestion: Because of the sensitivity of these cookies, ideally we'd want to set them `httpOnly`, so this would stop working.

For CSRF, this can be obtained with `window.wagtailConfig.CSRF_TOKEN`

The ID token is _probably_ fine to keep accessible from JS, but ideally we could pull that from something in the request body instead. Given this is loaded on the page, there might be something in Wagtail already.

cms/static_src/javascript/auth.js
```diff
@@ -0,0 +1,153 @@
...
+    authConfig = JSON.parse(configEl.textContent);
+  } catch (e) {
+    console.error('[WAGTAIL] Failed to parse <script id="auth-config">:', e);
+  }
+}
+
+const {
+  wagtailAdminHomePath,
+  csrfCookieName,
+  csrfHeaderName,
```
RealOrangeOne, line 25: Suggestion: With the below, `csrfCookieName` can be removed. `csrfHeaderName` can also be obtained using `window.wagtailConfig.CSRF_HEADER_NAME `

cms/static_src/javascript/auth.js
```diff
@@ -0,0 +1,153 @@
...
+  },
+  onError: async (error) => {
+    console.error('[WAGTAIL] General error:', error);
+    await logout();
+  },
+};
+
+// Initialise SessionManagement only if the user is logged in via Cognito (indicated by the presence of an id_token cookie).
+// Don't load if it's in an iframe (e.g. the Wagtail admin interface preview).
+if (window.self === window.top && getCookieByName(idTokenCookieName)) {
```
RealOrangeOne, line 141: Question: Should this be started in an event handler, rather than as soon as the JS is executed? Perhaps `window.load`?

cms/teams/management/commands/sync_teams.py
```diff
@@ -0,0 +1,135 @@
...
+            self.stdout.write(f"Creating new team: {name} (ID: {group_id})")
+            if not dry_run:
+                Team.objects.create(
+                    identifier=group_id,
+                    name=name,
+                    precedence=precedence,
+                    created_at=created_at,
+                    updated_at=updated_at,
+                    is_active=True,
+                )
```
RealOrangeOne, line 135: Suggestion: I think this is just re-implementing `create_or_update`. It might be cleaner to reuse that method.

cms/users/models.py
```diff
@@ -1,13 +1,82 @@
...
 from django.utils.functional import cached_property
 
+from cms.teams.models import Team
+
 
 class User(AbstractUser):
     """Barebones custom user model."""
 
+    # Unique identifier for user lookup, as emails (used as usernames) may change.
+    user_id = models.UUIDField(unique=True, default=uuid.uuid4, editable=False)
```
RealOrangeOne, line 16: Question: Why does this need to be different to the user's primary key? Having 2 different values for a user's "id" is likely to get confusing, especially for auditing.

cms/teams/viewsets.py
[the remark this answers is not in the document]
zerolab: @MebinAbraham ah, that rings a bell.
I say we do the removal, and link to Florence in the follow up task since we still allow managing teams in Wagtail.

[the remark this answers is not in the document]
sanjeevz3009: Happy to create a follow-up for this @MebinAbraham

cms/core/management/commands/scheduler.py
```diff
@@ -43,6 +43,11 @@ def configure_scheduler(self) -> None:
         """Configures the scheduler and triggers."""
         # "second=0" run the task every minute, on the minute (ie when the seconds = 0)
         self.add_management_command("publish_bundles", CronTrigger(second=0))
+
         # Run every 5 minutes.
         # See https://apscheduler.readthedocs.io/en/3.x/modules/triggers/cron.html#expression-types
         self.add_management_command("publish_scheduled_without_bundles", CronTrigger(minute="*/5"))
+
+        # Sync teams
+        # :TODO: Enable in ONS AWS environment.
```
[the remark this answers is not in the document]
sanjeevz3009: I am pretty sure we can enable programmatically to only run when `AWS_COGNITO_LOGIN_ENABLED`, but I would like more context around this.

Will be really be running this every minute? Wouldn't it be resource-intensive? 

So pulling in @MebinAbraham

cms/settings/base.py
sanjeevz3009: https://jira.ons.gov.uk/browse/CMS-528

cms/static_src/javascript/auth.js
```diff
@@ -0,0 +1,155 @@
...
+
+    // These will be fetched from localStorage which is set by the auth service before redirect.
+    // If they do not exist, then the session is invalid and the user will be logged out.
+    // Only clear if neither value is set, i.e. brand-new tab with no prior expiry info:
+    if (
+      !localStorage.getItem('session_expiry_time') &&
+      !localStorage.getItem('refresh_expiry_time')
+    ) {
+      SessionManagement.setSessionExpiryTime(null, null);
+    }
```
MebinAbraham, line 153: These values are never set by the auth client lib directly. Instead, they're stored in `dis_auth_client_state` key as:
```js
{
  session_expiry_time: "2025-06-04T09:12:39.000Z",
  refresh_expiry_time: "2025-06-04T20:57:15Z"
}
```
As a result, the current if condition always evaluates to false.

However, you don’t need this check at all. If you only call `setSessionExpiryTime` when `dis_auth_client_state` doesn’t exist, then the event listeners (used for session refresh, etc.) are never set up. This is because `setSessionExpiryTime` is the function that initialises those listeners.

Even when we pass `null` to `setSessionExpiryTime`, it still fetches the actual refresh timers from local storage.

cms/users/models.py
sanjeevz3009: Thanks, Dan, updated!

cms/settings/test.py
```diff
@@ -74,3 +75,17 @@
...
+# strip out our custom ONSAuthMiddleware
+MIDDLEWARE = [mw for mw in MIDDLEWARE if mw != "cms.auth.middleware.ONSAuthMiddleware"]
+
+# ensure Django's AuthenticationMiddleware is present (just after SessionMiddleware)
+if "django.contrib.auth.middleware.AuthenticationMiddleware" not in MIDDLEWARE:
+    idx = MIDDLEWARE.index("django.contrib.sessions.middleware.SessionMiddleware")
+    MIDDLEWARE.insert(
+        idx + 1,
+        "django.contrib.auth.middleware.AuthenticationMiddleware",
+    )
```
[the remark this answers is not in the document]
sanjeevz3009: I added the `strip out our custom ONSAuthMiddleware` logic into the `functional_test` settings file as it's more appropriate there + added some comments around it.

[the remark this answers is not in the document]
sanjeevz3009: Added some nice comments in the `cms/settings/test.py` file, which explain a bit more in technical detail.

cms/core/management/commands/scheduler.py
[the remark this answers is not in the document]
RealOrangeOne: A separate variable makes sense to me :+1: At some point this scheduler will disappear, so let's not spend too much time on this.

cms/settings/base.py
```diff
@@ -846,6 +845,11 @@
 if "WAGTAILADMIN_BASE_URL" in env:
     WAGTAILADMIN_BASE_URL = env["WAGTAILADMIN_BASE_URL"]
 
+# https://docs.wagtail.org/en/latest/reference/settings.html#wagtailadmin-login-url
+# if "WAGTAILADMIN_LOGIN_URL" in env:
+#     WAGTAILADMIN_LOGIN_URL = env["WAGTAILADMIN_LOGIN_URL"]
```
RealOrangeOne, line 850:
```suggestion
```

cms/settings/base.py
```diff
@@ -971,3 +975,42 @@
...
+AUTH_TOKEN_REFRESH_URL = env.get("AUTH_TOKEN_REFRESH_URL")
+SESSION_COOKIE_AGE = env.get("SESSION_COOKIE_AGE", 60 * 15)  # 15 minutes to match Auth Service
+SESSION_EXPIRE_AT_BROWSER_CLOSE = True
+IDENTITY_API_BASE_URL = env.get("IDENTITY_API_BASE_URL")
+AWS_COGNITO_LOGIN_ENABLED = env.get("AWS_COGNITO_LOGIN_ENABLED", "false").lower() == "true"
+AWS_COGNITO_USER_POOL_ID = env.get("AWS_COGNITO_USER_POOL_ID")
+AWS_COGNITO_APP_CLIENT_ID = env.get("AWS_COGNITO_APP_CLIENT_ID")
+
+# TODO: Make system checks
+# if AWS_COGNITO_LOGIN_ENABLED:
```
RealOrangeOne, line 991: Nit: Given this isn't implemented yet, might as well remove the commented out code.

RealOrangeOne (commented):
Couple nits

cms/settings/test.py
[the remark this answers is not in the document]
sanjeevz3009: Had a chat with @MebinAbraham. We will be looking into/ debugging the issues mentioned above, re-assessing the test coverage, so we can have good confidence in Auth and the test coverage.
- Will be reviewing the test coverage.
- Looking into improving the Playwright tests to make it more UI E2E/ integration testing as currently we are mocking too much.
- Investigate the issues mentioned above.

cms/users/migrations/0005_user_remove_uuid_null.py
```diff
@@ -0,0 +1,20 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("users", "0004_user_populate_external_user_id"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="user",
+            name="external_user_id",
+            field=models.UUIDField(default=uuid.uuid4, editable=False, unique=True),
```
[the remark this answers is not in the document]
MebinAbraham: Originally did this to avoid needing if else checks in various places. At the moment, whether a user is external is determined by whether they have a usable password. But on second thought, we’ll go with your suggestion as it simplifies testing and makes the differentiation more explicit. Will make the change 👍

cms/settings/base.py
sanjeevz3009: Done!

cms/settings/base.py
sanjeevz3009: Done!

sanjeevz3009:
@zerolab @MaciekBaron @AdamHawtin @RealOrangeOne 
For everyone's information, when reviewing this PR, do note we have got the unit tests, integration and the playwright tests also. All of them should have good coverage.
- Integration tests have been refactored and cleaned up.
- Unit tests haven't been cleaned/ refactored yet; they have been improved, but no heavy refactoring has been done yet.
- The Playwright test will additionally be d-mocked in a follow-up ticket; this was not possible due to issues with writing integration tests at the start, hence the heavy mocking.

There will be two follow-up cards/ maybe a few minor ones too, from Mebin:
- Django systems checks + tests for system checks
- Playwright de-mocking + unit tests refactor

Apart from that, the test coverage/ confidence should be really good for now, and with the follow-up playwright de-mocking it should be even better, as currently the Playwright tests are mostly testing the auth.js file functionalities.

cms/users/migrations/0005_user_remove_uuid_null.py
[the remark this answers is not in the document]
MebinAbraham: Changed in [`38aca2c` (#163)](https://github.com/ONSdigital/dis-wagtail/pull/163/commits/38aca2cb335089d5c26ad281dbc8f78592dc5d57)

cms/auth/tests/test_auth_integration.py
```diff
@@ -0,0 +1,341 @@
...
+from django.conf import settings
+from django.contrib.auth import get_user_model
+from django.test import TestCase, override_settings
+from django.urls import reverse
+from wagtail import hooks
+
+from cms.auth import wagtail_hooks
+from cms.auth.tests.helpers import CognitoTokenMixin, build_jwt
+
+User = get_user_model()
```
zerolab, line 13: nitpick/non-blocking: we're importing from our users app directly in other places. `from cms.users.models import User`.

This pattern here is usually used in packages where you do not know what the user model is.

cms/auth/tests/helpers.py
```diff
@@ -0,0 +1,171 @@
...
+    WAGTAIL_CORE_ADMIN_LOGIN_ENABLED=True,
+    AUTH_TOKEN_REFRESH_URL="/auth/refresh/",
+    WAGTAILADMIN_HOME_PATH="/admin/",
+    CSRF_COOKIE_NAME="csrftoken",
+    SESSION_RENEWAL_OFFSET_SECONDS=300,
+)
+class CognitoTokenMixin(TestCase):
+    """Utilities reused by every auth-related TestCase."""
+
+    __test__ = False  # Prevents Django's test runner from treating this as a runnable test case
```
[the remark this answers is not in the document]
sanjeevz3009: Habit from the mixins in the helpers in the search tests, which is a bit different. Thanks, Dan, removed!

cms/settings/dev.py
```diff
@@ -107,6 +122,7 @@
         "0004_make_release_date_mandatory_and_rename_next_release_text",  # Ignoring NOT NULL and RENAMING constraints
         "0004_statisticalarticlepage_headline_figures_figure_ids",
         "0006_statisticalarticlepage_dataset_sorting_and_more",  # Ignoring NOT NULL constraint
+        "0005_user_remove_uuid_null",  # Ignoring NOT NULL constraint
```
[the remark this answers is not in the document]
sanjeevz3009: Removed!

cms/core/management/commands/scheduler.py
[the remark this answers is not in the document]
sanjeevz3009: Done!

cms/auth/tests/test_auth_integration.py
sanjeevz3009: Bad Sanjeev. I knew this as well. Updated in all the relevant places.

cms/auth/tests/test_views.py
```diff
@@ -0,0 +1,182 @@
...
+            "/admin/extend-session/",
+            HTTP_X_CSRFTOKEN=token,
+        )
+        # attach session
+        SessionMiddleware(lambda r: None).process_request(request)
+        request.session.save()
+
+        # attach CSRF cookie
+        request.COOKIES["csrftoken"] = token
+        CsrfViewMiddleware(lambda r: None).process_view(request, extend_session, (), {})
```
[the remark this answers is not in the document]
sanjeevz3009: This is redundant. In our case, since the view is decorated with @csrf_protect and we are already providing the CSRF token in the header, this manual middleware call is not needed here. Probably forgot to clean it up after doing something with it.

Removed

cms/core/management/commands/scheduler.py
```diff
@@ -43,6 +44,11 @@ def configure_scheduler(self) -> None:
...
         # "second=0" run the task every minute, on the minute (ie when the seconds = 0)
         self.add_management_command("publish_bundles", CronTrigger(second=0))
+
         # Run every 5 minutes.
         # See https://apscheduler.readthedocs.io/en/3.x/modules/triggers/cron.html#expression-types
         self.add_management_command("publish_scheduled_without_bundles", CronTrigger(minute="*/5"))
+
+        # Sync teams
+        if settings.AWS_COGNITO_TEAM_SYNC_ENABLED:
+            self.add_management_command("sync_teams", CronTrigger(minute="*/1"))
```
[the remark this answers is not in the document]
RealOrangeOne: Question: Once a minute is pretty often. Does it need to be that regularly or can it be say every 10? How regularly do the teams change?

cms/users/models.py
```diff
@@ -1,13 +1,90 @@
...
+        never possess a native password.
+        """
+        return self.external_user_id is not None and not self.has_usable_password()
+
+    def assign_groups_and_teams(self, /, groups_ids: Iterable[str]) -> None:
+        """Assign groups and teams to the user based on their Cognito groups."""
+        self._assign_groups(groups_ids)
+        self._assign_teams(groups_ids)
+
+        self.save()
```
RealOrangeOne, line 32: Nit: Setting groups and teams doesn't change anything on the model, so this change is unnecessary

cms/auth/tests/test_auth_integration.py
```diff
@@ -0,0 +1,341 @@
...
+    @override_settings(AWS_COGNITO_LOGIN_ENABLED=False)
+    def test_core_admin_login_available_when_cognito_disabled(self):
+        user = User.objects.create_superuser(username="test", email="test@example.com")
+        self.client.force_login(user)
+        self.assertLoggedIn()
+
+    def test_core_admin_login_available_when_cognito_enabled(self):
+        user = User.objects.create_superuser(username="test", email="test@example.com")
+        self.client.force_login(user)
+        self.assertLoggedIn()
```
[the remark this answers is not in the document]
sanjeevz3009: Done!

cms/auth/tests/helpers.py
```diff
@@ -0,0 +1,169 @@
...
+    WAGTAIL_CORE_ADMIN_LOGIN_ENABLED=True,
+    AUTH_TOKEN_REFRESH_URL="/auth/refresh/",
+    WAGTAILADMIN_HOME_PATH="/admin/",
+    CSRF_COOKIE_NAME="csrftoken",
+    SESSION_RENEWAL_OFFSET_SECONDS=300,
+)
+class CognitoTokenTestCase(TestCase):
+    """Utilities reused by every auth-related TestCase."""
+
+    JWT_SESSION_ID_KEY: str = "jwt_session_id"
```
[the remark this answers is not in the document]
sanjeevz3009: Done!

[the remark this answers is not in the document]
sanjeevz3009: Done!

cms/auth/tests/helpers.py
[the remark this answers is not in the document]
sanjeevz3009: @MaciekBaron Can you resolve on behalf

sanjeevz3009:
@MebinAbraham @AdamHawtin @jakeandandrews @zerolab The remaining comments/ nits will be addressed in a follow-up PR =) As we got couple of follow-up PRs for Auth lined up

AdamHawtin (approved):

MebinAbraham (commented):
✅ LGTM! Let’s raise follow up PRs for any tidy-up tasks, and get this integrated and tested. Great work 🎉👍

@RealOrangeOne just a quick reminder to set the following env vars for the internal sandbox:

<details>

<summary>Auth env vars</summary>

```
WAGTAIL_CORE_ADMIN_LOGIN_ENABLED=true
WAGTAILADMIN_LOGIN_URL=/florence/login
AUTH_TOKEN_REFRESH_URL=/tokens/self
IDENTITY_API_BASE_URL (not needed until service auth is wired up)
AWS_COGNITO_LOGIN_ENABLED=true
AWS_COGNITO_USER_POOL_ID
AWS_COGNITO_APP_CLIENT_ID
WAGTAILADMIN_HOME_PATH=wagtail-admin/
LOGOUT_REDIRECT_URL=/florence/logout
```

</details>

Let me know if you hit any issues.

cms/auth/tests/test_views.py
```diff
@@ -0,0 +1,181 @@
...
+    # Helper attach a working session and message storage
+    def _prep_request(self, req):
+        # give the request a session
+        SessionMiddleware(lambda r: None).process_request(req)
+        req.session.save()
+
+        # attach a real FallbackStorage
+        storage = FallbackStorage(req)
+        req._messages = storage  # pylint: disable=protected-access
+        return storage
```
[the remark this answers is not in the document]
sanjeevz3009: Will do on a follow-up PR.

cms/auth/utils.py
```diff
@@ -0,0 +1,122 @@
...
+
+def _parse_der_public_key(b64_der_key: str) -> RSAPublicKey:
+    """Parses a Base64 encoded DER public key and returns the public key object."""
+    try:
+        der = base64.b64decode(b64_der_key)
+        key = serialization.load_der_public_key(der, backend=default_backend())
+        return cast(RSAPublicKey, key)
+    except Exception:
+        logger.exception("Failed to parse DER public key")
+        raise
```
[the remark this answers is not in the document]
sanjeevz3009: Will do on a follow-up PR.

cms/static_src/javascript/auth.js
```diff
@@ -0,0 +1,149 @@
...
+  },
+  onError: async (error) => {
+    console.error('[WAGTAIL] General error:', error);
+    await logout();
+  },
+};
+
+// Initialise SessionManagement only if the user is logged in via Cognito (indicated by the presence of an id_token cookie).
+// Don't load if it's in an iframe (e.g. the Wagtail admin interface preview).
+document.addEventListener('DOMContentLoaded', () => {
```
MaciekBaron, line 141: Question: On line 18 we log an error if we couldn't retrieve the config. However we continue nonetheless. Should we add this event listener if and only if line 16 was successful, i.e. `authConfig` isn't still an empty object?

cms/users/models.py
MebinAbraham: This step is necessary to save the changes made by the `update_details` call before this function, which edits the model itself. If it makes the code easier to follow, we can move the `self.save()` method out of this function and place it in the middleware after `assign_groups_and_teams`.

Given it was removed and no test failed around user's details change, it implies we have a gap in the test. (Let's add in the follow up).

sanjeevz3009: Thanks @MebinAbraham. I agree with moving the `self.save()` method out of this function and place it after `assign_groups_and_teams ` in the middleware where it makes more sense.

I thought I had covered this test case, =/. Let me have a quick look if not will add to it to follow-up.

sanjeevz3009:
@MebinAbraham @zerolab @AdamHawtin @RealOrangeOne @MaciekBaron 

Addressing minor/ nits comments from the Auth PR: https://jira.ons.gov.uk/browse/CMS-559.

https://github.com/ONSdigital/dis-wagtail/pull/163#issuecomment-3013090182

AdamHawtin (approved):

cms/static_src/javascript/auth.js
MebinAbraham: Look like a sensible change.

cms/auth/middleware.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @staticmethod
+    def _handle_unauthenticated_user(request: "HttpRequest") -> None:
+        """Logs out the user if JWT tokens are missing and the session configuration is unsuitable."""
+        if not settings.WAGTAIL_CORE_ADMIN_LOGIN_ENABLED or (
+            request.user.is_authenticated and request.user.is_external_user
+        ):
+            logger.info(
+                "Terminating session due to missing JWT tokens or insufficient login configuration.",
+            )
```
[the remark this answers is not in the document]
MebinAbraham: I am not a fan of adding another if branch for the sake of logging; I suggest just logging out the external user ID; it may be none, which is fine.
```py
        if not settings.WAGTAIL_CORE_ADMIN_LOGIN_ENABLED or (
            request.user.is_authenticated and getattr(request.user, "is_external_user", False)
        ):
            logger.info(
                "Terminating session due to missing JWT tokens or insufficient login configuration.",
                extra={"external_user_id": getattr(request.user, "external_user_id", None)},
            )
            logout(request)
```

PS: you don't need getattr for `is_external_user` if user is authenticated as that will always exist, but non blocker.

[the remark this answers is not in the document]
sanjeevz3009: Will make the change, as it satisfies @MaciekBaron ask regardless.

cms/auth/tests/test_auth_integration.py
```diff
@@ -0,0 +1,340 @@
...
+class WagtailHookTests(CognitoTokenTestCase):
+    def setUp(self):
+        # clear any previously-registered hooks
+        hooks._hooks.pop("insert_global_admin_js", None)  # pylint: disable=protected-access
+
+    def test_wagtail_hook_injection(self):
+        # reload under AWS_COGNITO_LOGIN_ENABLED=True
+        importlib.reload(wagtail_hooks)
+        # stub out the module's static() so it doesn't hit a manifest
+        wagtail_hooks.static = lambda path: f"/static/{path}"
```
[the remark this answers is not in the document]
RealOrangeOne: Can we not. This could cause strange failures for others, and could hide other failures.

cms/auth/tests/test_auth_integration.py
```diff
@@ -0,0 +1,340 @@
...
+        # GET should return 405
+        response_2 = self.client.get(url)
+        self.assertEqual(response_2.status_code, 405)
+        self.assertJSONEqual(response_2.content, {"status": "error", "message": "Invalid request method."})
+
+
+class WagtailHookTests(CognitoTokenTestCase):
+    def setUp(self):
+        # clear any previously-registered hooks
+        hooks._hooks.pop("insert_global_admin_js", None)  # pylint: disable=protected-access
```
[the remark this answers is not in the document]
RealOrangeOne: Can we not. This could cause strange failures for others, and could hide other failures.

cms/auth/middleware.py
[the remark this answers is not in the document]
sanjeevz3009: Done!

cms/auth/tests/test_auth_integration.py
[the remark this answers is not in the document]
sanjeevz3009: After making the change for the PR comment above this, don't need this setUp class anymore. Removed.

cms/auth/tests/test_auth_integration.py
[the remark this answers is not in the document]
sanjeevz3009: Done!

cms/auth/tests/helpers.py
```diff
@@ -0,0 +1,169 @@
...
+    def setUpTestData(cls):  # pylint: disable=invalid-name
+        cls.user_uuid: str = str(uuid.uuid4())
+
+        # RSA keypair and JWKS stub
+        cls.keypair = generate_rsa_keypair()
+        public_b64 = base64.b64encode(cls.keypair.public_der).decode()
+        cls.jwks = {cls.keypair.kid: public_b64}
+
+        # Reload utils so module constants use overridden settings
+        importlib.reload(auth_utils)
```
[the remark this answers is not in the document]
sanjeevz3009: Done!

sanjeevz3009:
> Assuming everything to be done in future is ticketed and not forgotten about, this is fine.

@RealOrangeOne Got it captured https://github.com/ONSdigital/dis-wagtail/pull/163#issuecomment-3013090182. It will be all done straight after this =)

PR #165, opened 2025-04-22
Add statistical article related data page

AdamHawtin:

Statistical articles need to feature their related data, via a link to a separate page. This PR achieves that by adding the datasets to the statistical article page, with a separate page route to view them.

Loom demo: https://www.loom.com/share/14c0ae941cc34ea68b32570282dd35c0

### How to review
NOTE: Due to a conflicting migration, if you have previously run migrations on this branch, you may need to reset back to main and re-run the migrations.

- Check out this branch locally
- Spin up your local development environment
- Apply the migrations with `make migrations` (or `make dev-init` if starting from scratch)
- Open the wagtail editor and create or edit a statistical article page
- There should be a new `Related data` tab in the editor interface, navigate to this tab
- Try adding datasets in this tab, both from the API and manually linked
- Publish the page, observe a new "View data used in this article" link should appear on the article page once there are datasets. 
- View this related data page, check it is rendering sensibly (bear in mind there is still a design and UI tidy up pass due for this page, so it is currently best dev endeavours)
- Check the preview when editing the article, there should be a new preview mode available to view the related data page
- Remove all the datasets from the article, verify that the "View data used in this article" link no longer shows.
- Add a number of datasets to check the page correctly paginates. This can be made quicker/easier to test by setting the environment variable `RELATED_DATASETS_PER_PAGE` to a lower number, by default it is 20 per page.

### Follow-up Actions
A UI design pass of the related data page is already on the backlog.
There are known issues with links in the preview which will be followed up in another card.


cms/datasets/models.py
```diff
@@ -87,7 +87,7 @@ def formatted_edition(self) -> str:
 
     @property
     def url_path(self) -> str:
-        return f"/datasets/{self.namespace}/editions/{self.edition}/versions/{self.version}"
+        return f"datasets/{self.namespace}"
```
zerolab, line 90: q: is this because we're using the public API which doesn't have the new paths? If so, perhaps leave the other one with TODO comment?

cms/jinja2/templates/components/navigation/breadcrumbs.html
```diff
@@ -12,6 +12,9 @@
             {% endif %}
         {% endif %}
     {% endfor %}
+    {% if page.include_self_in_breadcrumbs %}
```
zerolab, line 15: This doesn't look used anywhere? It would be good to either clean up, or make use of it and add some dev comments for context

cms/jinja2/templates/pages/statistical_article_page--related_data.html
```diff
@@ -0,0 +1,52 @@
...
+
+{% from "components/document-list/_macro.njk" import onsDocumentList %}
+{% from "components/hero/_macro.njk" import onsHero %}
+{% from "components/pagination/_macro.njk" import onsPagination %}
+
+{% block header_area %}
+    {# fmt:off #}
+    {{
+        onsHero({
+            "topic": _("Related data"),
```
zerolab, line 11: could you extract the translatable strings outside of this. It looks like gettext doesn't pick on them for some reason. An example:

```jinja2
{%- set hero_topic = _("Related data") -%}
{%- set breadcrumbs_aria_label = _("Breadcrumbs") -%}
```

cms/jinja2/templates/pages/statistical_article_page--related_data.html
```diff
@@ -0,0 +1,52 @@
...
+
+{% block main %}
+
+    {# fmt:off #}
+    {{
+        onsDocumentList({
+        "documents": paginated_datasets
+        })
+    }}
+    {# fmt:on #}
```
zerolab, line 36: nitpick:

```suggestion
    {{ onsDocumentList({"documents": paginated_datasets}) }}
```

should be enough

or, if going the formatted way:

```suggestion
    {# fmt:off #}
    {{
        onsDocumentList({
            "documents": paginated_datasets
        })
    }}
    {# fmt:on #}
```

cms/jinja2/templates/pages/statistical_article_page--related_data.html
```diff
@@ -0,0 +1,52 @@
...
+
+    {% block pagination %}
+        {% if paginated_datasets.paginator.num_pages > 1 %}
+
+            {# fmt:off #}
+            {{
+                onsPagination({
+                "currentPageNumber": paginated_datasets.number,
+                "pages": ons_pagination_url_list
+                })
```
zerolab, line 46:
```suggestion
                onsPagination({
                    "currentPageNumber": paginated_datasets.number,
                    "pages": ons_pagination_url_list
                })
```

cms/jinja2/templates/pages/statistical_article_page.html
```diff
@@ -189,6 +189,20 @@ <h3 class="ons-corrections-notices__banner-title ons-u-mb-2xs">{{ corrections_an
...
             {% endwith %}
+
+            {% if page.datasets %}
+                {% with title=_("Explore Data"), aria_label=_("Data related to this page"), related_data_text=_("View data used in this article"), related_data_url=routablepageurl(page.specific, "related_data") %}
+                {# fmt:off #}
+                {{-
+                    onsTableOfContents({
+                        "title": title,
+                        "ariaLabel": aria_label,
+                        "itemsList": [{"url":related_data_url, "text":related_data_text}]
```
zerolab, line 200:
```suggestion
                        "itemsList": [{"url": related_data_url, "text": related_data_text}]
```

cms/jinja2/templates/pages/statistical_article_page.html
```diff
@@ -214,5 +228,7 @@ <h2 class="ons-u-fs-l">{{ _("Cite this analysis") }}</h2>
                 {% endwith %}
             {% endif %}
         </div>
+
+
```
zerolab, line 232: unnecessary newlines?

cms/datasets/models.py
AdamHawtin: This is changing because it's been agreed that for now the links should just direct to the dataset landing page, will addc a comment in code and on the ticket.

cms/jinja2/templates/pages/statistical_article_page.html
```diff
@@ -189,6 +189,20 @@ <h3 class="ons-corrections-notices__banner-title ons-u-mb-2xs">{{ corrections_an
                 }}
                 {# fmt:on #}
             {% endwith %}
+
+            {% if page.datasets %}
+                {% with title=_("Explore Data"), aria_label=_("Data related to this page"), related_data_text=_("View data used in this article"), related_data_url=routablepageurl(page.specific, "related_data") %}
```
zerolab, line 194:
```suggestion
                {% with title=_("Explore Data"), aria_label=_("Data related to this page"), related_data_text=_("View data used in this article"), related_data_url=routablepageurl(page, "related_data") %}
```

`page` here is already the specific instance

cms/articles/tests/test_pages.py
```diff
@@ -405,3 +409,104 @@ def test_correct_template_used_for_edit_page(self):
             "wagtailadmin/panels/headline_figures/figures_used_by_ancestor_data.html",
             "Template required by Headline Figures functionality was not used",
         )
+
+    def test_related_data_route(self):
```
zerolab, line 413: I missed this when testing the original dataset implementation, but it would be good to expand the `DatasetStoryBlock` tests to check the duplicate links validation.

Specifically: https://github.com/ONSdigital/dis-wagtail/blob/main/cms/datasets/blocks.py#L53-L55 should be
```python
url_path = (
                block.value.url_path if block.block_type == "dataset_lookup" else urlparse(block.value["url"])
)
```

note the `block.value["url"].path` -> `block.value["url"]` change. The `url` value is a string

cms/articles/tests/test_pages.py
```diff
@@ -405,3 +409,104 @@ def test_correct_template_used_for_edit_page(self):
...
+            stream_data=[("manual_link", dataset) for dataset in manual_datasets],
+        )
+        self.page.save_revision().publish()
+        response = self.client.get(self.page.url + "related-data/?page=2")
+        content = response.content.decode(encoding="utf-8")
+
+        self.assertIn('class="ons-pagination__item ons-pagination__item--previous"', content)
+        self.assertIn('class="ons-pagination__item ons-pagination__item--current"', content)
+        self.assertIn('class="ons-pagination__item ons-pagination__item--next"', content)
+        self.assertIn('class="ons-pagination__position">Page 2 of 4', content)
```
zerolab, line 512: we need a test that confirms that adding a new article in the series copies over the previous one's datasets and sorting setting

zerolab (changes requested):
This is looking pretty good.

Discovered a bug with the dataset uniqueness validation. And we want a test for the related datasets prepopulation from previous article in the series

cms/articles/tests/test_pages.py
AdamHawtin: There are already tests for this in https://github.com/ONSdigital/dis-wagtail/blob/b29adc4924c66997f40d87ae046b8f2a7900ff97/cms/datasets/tests/test_blocks.py#L19

I think it works correctly as is, the intent was to validate for duplicate URL paths, ignoring the domain, the `.path` is looked up after the call to `urlparse`, to get just the path from the string. 

I discussed this with Mebin at the time, I believe the issue with just comparing full URLs was that in testing environments we will likely be pointing the API lookup to different API environments, but users were likely to still test with URLs from our public domain. 

It probably needs a comment explaining this intent.

zerolab: Arguably https://example.com/something is different from https://ons.gov.uk/something

But let's revisit this should it come up

cms/jinja2/templates/pages/statistical_article_page.html
AdamHawtin: Addressed in [1a7a89b](https://github.com/ONSdigital/dis-wagtail/pull/165/commits/1a7a89be9bd8fb2dd583a492e5330ba3206d691c)

cms/articles/tests/test_pages.py
AdamHawtin: Added in [1a7a89b](https://github.com/ONSdigital/dis-wagtail/pull/165/commits/1a7a89be9bd8fb2dd583a492e5330ba3206d691c)

zerolab (approved):
Needs linting fixes, otherwise LGTM

cms/datasets/models.py
kacperpONS: A side effect of this is that now that the validation that checks for duplicated links between Manually added and Looked up datasets doesn't work as expected. 


https://github.com/ONSdigital/dis-wagtail/blob/40da75b9f34b2f9f4e4bc8a1fc5210cd5a326fac/cms/datasets/blocks.py#L51
^ In this dict the URLs that will act as keys will be different for the manually linked and looked up datasets - even though the Looked-up dataset will still show the version number in the admin UI. It could be worth specifying in the comment in the `clean` method that this part won't work until we revert that.

AdamHawtin: There's some confusion around this because we didn't get a definitive answer on where the links for datasets should go, for the moment we've agreed to just link to the dataset landing page rather than the specific edition or version. However, I don't think we want to force manual dataset links to conform to this, they might be useful as a way of linking to a specific edition or version if required, and I don't think we want to try and block them from linking to two different editions or versions of the same dataset in case that is valid behaviour.

zerolab (approved):

kacperpONS (approved):

sanjeevz3009 (commented):
Reviewed!

PR #169, opened 2025-04-25
Add bar/column chart

nimasmi:

Ticket [CCB-48](https://jira.ons.gov.uk/browse/CCB-48) and [CCB-49](https://jira.ons.gov.uk/browse/CCB-49).

This replaces the placeholder bar and column chart types with a single, switchable, bar/column chart.

To avoid confusion due to Highcharts's transposed y/x axis labels for horizontal bar charts, they are relabelled "Value axis" and "Category axis" for both variants.

### How to review

Add a Bar/Column Chart block, and enter some data.
Test that the fields in the ticket scope can be edited, and are reflected in the rendered chart.

NB for now there's an open question about whether a horizontal bar chart can have an editable horizontal/y/value axis.


cms/datavis/blocks/base.py
```diff
@@ -117,14 +122,18 @@ def get_context(self, value: "StructValue", parent_context: Optional[dict[str, A
         context["download"] = self.get_download_config(value)
         return context
 
+    def get_highcharts_chart_type(self, value: "StructValue") -> str:
+        """Chart type may be set by a field, or hardcoded in the subclass."""
+        if chart_type := value.get("select_chart_type"):
+            return cast(str, chart_type)
+        return self.highcharts_chart_type
```
nimasmi, line 129: **Comment**: This method is new because the bar/column chart can have different values for this attribute.

cms/datavis/blocks/base.py
```diff
@@ -136,7 +145,7 @@ def get_x_axis_config(
         rows: Sequence[list[str | int | float]],
     ) -> dict[str, Any]:
         config: dict[str, Any] = {
-            "type": "linear",
+            "type": "category",
```
nimasmi, line 148: **Comment**: This is unrelated cleanup, but all chart types so far should have used the `category` axis type, and this is now part of the Design System axis spec: https://github.com/ONSdigital/design-system/blob/main/src/components/chart/_macro-options.md#x_axis

cms/datavis/blocks/base.py
```diff
@@ -165,7 +174,7 @@ def get_y_axis_config(
         }
 
         # Only add y-axis title if supported
-        if getattr(self, "supports_y_axis_title", False):
+        if self.supports_y_axis_title:
```
nimasmi, line 177: **Comment**: This is now made possible by the default value. As mentioned, watch this space for a refactor removing these.

cms/datavis/blocks/base.py
```diff
@@ -219,9 +228,7 @@ def get_series_data(
                 "animation": False,
             }
             if self.supports_data_labels:
-                item["dataLabels"] = {
-                    "enabled": value.get("show_data_labels"),
-                }
+                item["dataLabels"] = value.get("show_data_labels")
```
nimasmi, line 231: **Comment**: This corresponds to a change in the DS spec, which only applies to horizontal bar charts. https://github.com/ONSdigital/design-system/blob/main/src/components/chart/_macro-options.md#series

cms/datavis/blocks/base.py
```diff
@@ -234,9 +241,6 @@ def get_series_data(
                 item["tooltip"] = {
                     "valueSuffix": tooltip_suffix,
                 }
-            if getattr(self, "supports_stacked_layout", False) and value["use_stacked_layout"]:
-                item["stacking"] = "normal"
```
nimasmi, line 238: **Comment**: The DS spec has moved this to a chart attribute, rather than per-series. https://github.com/ONSdigital/design-system/blob/main/src/components/chart/_macro-options.md

nimasmi (commented):
I have added commentary to help reviewers.

cms/datavis/blocks/charts.py
```diff
@@ -15,33 +19,71 @@ class LineChartBlock(BaseVisualisationBlock):
...
+            (
+                "title",
+                blocks.CharBlock(
+                    required=False,
+                    help_text="Only use axis titles if it is not clear from the title "
+                    "and subtitle what the axis represents.",
+                ),
+            ),
+            # ("min", TextInputFloatBlock(label="Minimum", required=False)),
+            # ("max", TextInputFloatBlock(label="Maximum", required=False)),
```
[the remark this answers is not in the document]
nimasmi: This really puzzled me. I think it's not there because it's not clear from the specs whether it should be. I tried it with and without, and I think it's harmless to leave it in for now. There is now a [feature mapping](https://confluence.ons.gov.uk/display/NWP/Feature+mapping) page to track things like this, so we can remove in future if necessary.

I have made these actual lines of code in 3b37fad.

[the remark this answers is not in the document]
nimasmi: Correction: this feature _used to be_ available when the DS passed through chart and axis configuration directly, but now it acts as more of a gatekeeper (rightly so), and this is not yet available as an approved axis property.

I have reverted my previous commit, and added an explanatory note in 491cb9d.

zerolab (approved):

PR #170, opened 2025-04-25
Add locale label to admin page tree view

MaciekBaron:

This PR addresses CMS-455. It adds a locale label to the breadcrumbs of the admin page tree view (a page list).

@AdamHawtin expressed concerns regarding editors being confused about which locale they are viewing. This should hopefully make it more obvious when viewing a list of pages.

### How to review

1. Check out branch
2. Log in to admin panel
3. Go to a page list
4. Confirm that a locale label is present next to the main breadcrumbs

### Follow-up Actions

N/A

### DIFF
--- a/cms/core/templates/wagtailadmin/pages/page_listing_header.html
+++ b/cms/core/templates/wagtailadmin/pages/page_listing_header.html
@@ -1,0 +1,11 @@
+{% extends 'wagtailadmin/pages/page_listing_header.html' %}
+{% load wagtailadmin_tags %}
+
+{% block breadcrumbs %}
+    {{ block.super }}
+    {% if parent_page.url %}
+        <div class="w-mr-4">
+            {% status parent_page.locale classname="w-status--label w-m-0" %}
+        </div>
+    {% endif %}
+{% endblock %}
--- a/cms/core/tests/test_views.py
+++ b/cms/core/tests/test_views.py
@@ -4,6 +4,9 @@ from http import HTTPMethod
 from django.conf import settings
 from django.test import Client, SimpleTestCase, TestCase
 from django.urls import reverse
+from wagtail.test.utils import WagtailTestUtils
+
+from cms.home.models import HomePage
 
 
 class CSRFTestCase(TestCase):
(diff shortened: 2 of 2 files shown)


cms/core/tests/test_views.py
```diff
@@ -54,3 +57,18 @@ def test_all_methods(self):
                 self.assertEqual(response.status_code, 204)
                 self.assertEqual(response.content, b"")
                 self.assertEqual(response.templates, [])
+
+
+class AdminPageTreeTestCase(WagtailTestUtils, TestCase):
```
MaciekBaron, line 62: Wasn't sure where to put this test. Feel free to suggest a better location.

AdamHawtin (approved):
Looks good, nicely addresses the issue I raised 👍

cms/core/templates/wagtailadmin/pages/page_listing_header.html
```diff
@@ -0,0 +1,11 @@
+{% extends 'wagtailadmin/pages/page_listing_header.html' %}
+{% load wagtailadmin_tags %}
+
+{% block breadcrumbs %}
+    {{ block.super }}
+    {% if parent_page.url %}
```
zerolab, line 6: question: why the URL check? note that this incurs additional queries as .url is a property linked to `get_url` ([ref](https://github.com/wagtail/wagtail/blob/main/wagtail/models/pages.py#L1416))

MaciekBaron: This was so that the label doesn't appear when you are at the root level, but maybe there is a more elegant solution?

zerolab: `if not page.is_root`

MaciekBaron: For some reason I thought I tried that already... Thanks! [c5ab360](https://github.com/ONSdigital/dis-wagtail/pull/170/commits/c5ab3607245ab1aa408799e8c151c6080a380695)

sanjeevz3009 (approved):
LGTM =) Works as expected!

zerolab (approved):

PR #171, opened 2025-04-28
Enforce taxonomies in topic and theme pages

MaciekBaron:

This PR relates to CMS-458 and was originally raised by @MebinAbraham:
> (...) currently the Welsh pages allow different taxonomy selection. I don't see a valid reason for that given we don't have localised topics. Therefore, we should probably keep it in sync by enforcing a one to one by treating English as the source.

**Loom video**:
https://www.loom.com/share/c07091596bbe4134ac23cb964c056193

### How to review

1. Checkout branch
2. Create a topic or theme page, or go to an existing one
3. Confirm that the English version allows you to set the taxonomy (topic) you want
4. Switch to the Welsh version of that page
5. Confirm that you are unable to change the topic, and instead are told that the page will use the English selection

### Follow-up Actions

Currently the taxonomy selection will be updated next time you update the translation page, i.e. when you edit the translated page you see the English selection, but the translated page doesn't actually have the updated value until you save.

We need to discuss if immediate data syncing of topic choices is necessary when updating the English topics and themes, and changing their taxonomy choices. If this is deemed necessary, and a decision is made when to perform the sync (e.g. at save or publish?), we should create a new ticket to sync this.


MaciekBaron:
Thanks @sanjeevz3009 for your thorough review! 

The last point definitely looks like a bug, which I'll address shortly.

I think we should discuss the first two points internally just so we understand all possible scenarios (e.g. what if a draft version of an English page changes the topic selection?)

nehakerung (approved):
Other than the points raised before - LGTM

MaciekBaron:
Hey @sanjeevz3009 so I think you found a bug in core Wagtail. It looks like when you create a draft version of any page in Welsh and not publish it, you cannot correctly switch to the English version.

Steps to replicate:

- Create any page under the Welsh locale
- Save as a draft
- Switch to the English locale
- Notice that it still displays the same page with the same ID
- Go to the English page tree
- Notice that the page exists there
- However, when you click it it goes to the Welsh version

Could you confirm this is what you also experience?

sanjeevz3009:
> Hey @sanjeevz3009 so I think you found a bug in core Wagtail. It looks like when you create a draft version of any page in Welsh and not publish it, you cannot correctly switch to the English version.
> 
> Steps to replicate:
> 
> * Create any page under the Welsh locale
> * Save as a draft
> * Switch to the English locale
> * Notice that it still displays the same page with the same ID
> * Go to the English page tree
> * Notice that the page exists there
> * However, when you click it it goes to the Welsh version
> 
> Could you confirm this is what you also experience?

@MaciekBaron I tested this out and yes, I was able to reproduce this.

zerolab:
@MaciekBaron @sanjeevz3009 https://www.loom.com/share/be84c3e431724499af995e1bc1ca6e83 - using Bakerydemo, I can confirm that looks like a bug in core

cms/taxonomy/panels.py
```diff
@@ -0,0 +1,31 @@
...
+        template_name = "wagtailadmin/panels/exclusive_taxonomy/exclusive_taxonomy_panel.html"
+
+        def get_context_data(self, parent_context: "Optional[RenderContext]" = None) -> "Optional[RenderContext]":
+            context = super().get_context_data(parent_context)
+
+            # If the instance is not saved yet, we don't translations to check against.
+            if not self.instance.pk:
+                return context
+
+            if self.instance.locale != self.instance.get_root().locale and (
```
zerolab, line 23:
```suggestion
            if self.instance.locale_id != self.instance.get_root().locale_id and (
```

cms/taxonomy/panels.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class ExclusiveTaxonomyFieldPanel(FieldPanel):
+    class BoundPanel(FieldPanel.BoundPanel):
+        template_name = "wagtailadmin/panels/exclusive_taxonomy/exclusive_taxonomy_panel.html"
+
+        def get_context_data(self, parent_context: "Optional[RenderContext]" = None) -> "Optional[RenderContext]":
+            context = super().get_context_data(parent_context)
+
+            # If the instance is not saved yet, we don't translations to check against.
```
zerolab, line 19:
```suggestion
            # If the instance is not saved yet, we don't have any translations to check against.
```

zerolab (changes requested):

zerolab:
> If I create an English theme or topic page, save it as a draft and then go to the Welsh version and change the topic, then they are now out of sync. If I publish the Welsh page, the topic chosen on the Welsh page will override the English version.

I think it is worth adding validation for this. If saving non-default locale draft, validate that it matches that of the default locale

> Another thing I found is that if I create a Welsh theme or topic page, save as draft, then I go to the English version, it says on the English version This page will use the topic selected in the [English version](http://127.0.0.1:8000/admin/pages/21/edit/#tab-taxonomy) of the page: People's attitudes to data and then if I go back to the Welsh version then it also says This page will use the topic selected in the [English version](http://127.0.0.1:8000/admin/pages/21/edit/#tab-taxonomy) of the page: People's attitudes to data, now I am locked out in terms of changing the topics in both versions. The only way around this is to publish one of the versions (Welsh or English).

This is on the complex side - we could try to prevent adding pages in the non-default locale to avoid disparities like this. Let's run this past DP.

MaciekBaron:
@sanjeevz3009 @zerolab 

I have now raised this issue in core:

https://github.com/wagtail/wagtail/issues/13090

kacperpONS (approved):
Happy for this to go in while we wait for the issues in Wagtail core to be fixed.

zerolab (approved):

sanjeevz3009 (commented):
Reviewed!

PR #172, opened 2025-04-28
Enable chart point annotations

nimasmi:

This updates previous preview placeholder code for point annotations to work with the updated Design System spec at https://github.com/ONSdigital/design-system/blob/main/src/components/chart/_macro-options.md#annotations

Ticket [CCB-54](https://jira.ons.gov.uk/browse/CCB-54)

### How to review

> [!NOTE]
> This contains the code from #169, so is best reviewed after that is merged. Failing that, view individual commits to review.

Add various different types of chart block.
For each, add an "Annotation".
Test that the annotation is displayed at the corresponding data point index (1-based) on the x-axis (or y-axis for horizontal bar charts) and at the corresponding value ("value" corresponds to height for most charts except bar charts).

Note there is a display bug for horizontal bar charts when "show value labels", is also enabled.

<p>

image
image
image
image
image
image

</p>
</details>


cms/datavis/blocks/annotations.py
```diff
@@ -0,0 +1,46 @@
...
+from cms.datavis.blocks.utils import TextInputFloatBlock, TextInputIntegerBlock
+
+
+class PointAnnotationStructValue(StructValue):
+    def get_config(self) -> dict[str, Any]:
+        # The SVG coordinate system measures from the top left. It makes more
+        # sense to users to measure from the bottom left, to match the chart
+        # coordinates.
+        if label_offset_y := self.get("label_offset_y"):
+            label_offset_y = label_offset_y * -1
```
MaciekBaron, line 15: Non-blocking nit-pick, would this read better?
```suggestion
            label_offset_y = -label_offset_y
```

MaciekBaron (approved):
LGTM, just one non-blocking suggestion

cms/datavis/blocks/annotations.py
nimasmi: That's a good suggestion, thanks. I have changed this in 46cd3ef.

PR #173, opened 2025-04-28
Feature/add more bundles tests cms 358

AWitcherONS:

When the created_by user of a bundle has been deleted this then causes an error 
a fix has been added to HTML code 
### How to review

Run the unit tests
run the functional tests

### Follow-up Actions

these tests need to be combined with the appropriate files


functional_tests/features/bundle_page.feature
```diff
@@ -0,0 +1,24 @@
...
+    Scenario: A Viewer can inspect Bundles details
+        When a Viewer logs into the admin site
+        And the viewer is in the preview team
+        Then the user can see the Bundle menu item
+        And the user can inspect Bundle details
+
+    Scenario: A Viewer can inspect Bundles details
+        Given a Publishing Admin logs into the admin site
+        And Bundle has the creator removed
+        When a Viewer logs into the admin site
```
MaciekBaron, line 20: This line seems unnecessary as you already logged in two lines above
```suggestion
```

functional_tests/features/bundle_page.feature
```diff
@@ -0,0 +1,24 @@
+Feature: Publishing Officer can draft, edit, and publish bundles
+# Viewer
+# Publishing Admin
+# Publishing Officer
+    Background:
```
MaciekBaron, line 5: The `Background` block gets executed for all of your scenarios. This is why your scenarios are failing, because you are already logged in as a "Publishing Officer" when you try to log in as other users.

I think what you meant to do here is make this a scenario:
```suggestion
    Scenario: A Publishing Officer can add Bundles
```

functional_tests/features/bundle_page.feature
```diff
@@ -0,0 +1,24 @@
...
+        When the user clicks the Bundles menu item
+        Then the user can add Bundles
+
+    Scenario: A Viewer can inspect Bundles details
+        When a Viewer logs into the admin site
+        And the viewer is in the preview team
+        Then the user can see the Bundle menu item
+        And the user can inspect Bundle details
+
+    Scenario: A Viewer can inspect Bundles details
```
MaciekBaron, line 17: This used the same scenario name, I think you meant to call it this:
```suggestion
    Scenario: A Publishing Admin can inspect Bundles details
```

MaciekBaron (changes requested):

functional_tests/features/bundle_page.feature
```diff
@@ -0,0 +1,18 @@
+Feature: Publishing Officer can draft, edit, and publish bundles
+
+
+    Scenario: A Publishing Officer can add Bundles with Publishing Officer as the creator.  The bundle can be seen and accessed.
```
MaciekBaron, line 4: This should be split into separate scenarios:
* A Publishing Officer can add Bundles with Publishing Officer as the creator
* The bundle can be seen and accessed
* Once the creator is removed from the bundle then the bundle can still be seen and accessed

If you really want you could probably combine the first two (e.g. we can confirm that the bundle was successfully created by viewing it), but the last one should definitely be a separate test.

functional_tests/steps/bundles.py
```diff
@@ -53,3 +54,30 @@ def the_user_opens_page_chooser(context: Context) -> None:
...
+    context.bundle.save(update_fields=["created_by"])
+
+
+@step('a "{user_type}" logs into the admin site with username "{username}"')
+def user_logs_in_with_username(context: Context, user_type: str, username: str) -> None:
+    context.user_data = create_user(user_type)
+    context.page.goto(f"{context.base_url}/admin/login/")
+    context.page.get_by_placeholder("Enter your username").fill(context.user_data[username])
+    context.page.get_by_placeholder("Enter password").fill(context.user_data["password"])
+    context.page.get_by_role("button", name="Sign in").click()
```
MaciekBaron, line 76: This step is not being used, and is already implemented in `cms_login.py`, albeit without the username parameter:

https://github.com/ONSdigital/dis-wagtail/blob/c4cd24cfe3e57d124d804cef51befdcf911e95f1/functional_tests/steps/cms_login.py#L32-L33

functional_tests/steps/bundles.py
```diff
@@ -53,3 +54,30 @@ def the_user_opens_page_chooser(context: Context) -> None:
...
+
+
+@step("the user can inspect Bundle details")
+def the_user_can_see_the_bundle_details(context: Context) -> None:
+    context.page.get_by_role("link", name="Bundles", exact=True).click()
+    context.page.get_by_role("button", name=f"More options for '{context.bundle.name}'").click()
+    context.page.get_by_role("link", name=f"Inspect '{context.bundle.name}'").click()
```
MaciekBaron, line 83: This is an assertion, so it should use `@then`. However you are not making any assertions in the code. I can see you are clicking on things, but you should do checks using `expect()` or `.is_visible()` etc. to see if certain things exist or contain the text or elements you need.

functional_tests/steps/bundles.py
```diff
@@ -53,3 +54,30 @@ def the_user_opens_page_chooser(context: Context) -> None:
...
+
+@step("the user can add Bundles created by user")
+def a_bundle_with_name_item_name_has_been_created_by_username(context: Context) -> None:
+    context.bundle = BundleFactory(created_by=context.user_data["user"])
+
+
+@step("Bundle has the creator removed")
+def created_by_has_been_deleted(context: Context) -> None:
+    context.bundle.created_by = None
+    context.bundle.save(update_fields=["created_by"])
```
MaciekBaron, line 67: This sounds like it should be a `@given()` to be used at a beginning of a test, as a step should not directly manipulate any data in a functional test.

functional_tests/steps/bundles.py
```diff
@@ -53,3 +54,30 @@ def the_user_opens_page_chooser(context: Context) -> None:
 def the_locale_column_is_displayed(context: Context) -> None:
     modal = context.page.locator(".modal-body")
     modal.get_by_role("columnheader", name="Locale").is_visible()
+
+
+@step("the user can add Bundles created by user")
+def a_bundle_with_name_item_name_has_been_created_by_username(context: Context) -> None:
+    context.bundle = BundleFactory(created_by=context.user_data["user"])
```
MaciekBaron, line 61: This step sounds like an assertion (the user can do something), so it should use `@then` rather than `@step`. However you are not making an assertion in the code, you are creating a bundle.

If this is a pre-condition, then this should use `@given` and the step should be reworded (e.g. "a user created a bundle").

Your functional tests should roughly follow this sequence:

**Given** _pre-conditions_ (data modification allowed here to setup the scenario) using `@given` and occasionally `@when`
**When** _user actions_ (data modification not allowed here, only user actions) using `@when` and `@step`
**Then** _assertions_ (we check the presence or lack-of different things) using `@then`

MaciekBaron (changes requested):

MaciekBaron:
Hey @AWitcherONS, if you have any questions about my comments, let me know, I'll be happy to explain further.

functional_tests/steps/bundles.py
```diff
@@ -53,3 +57,48 @@ def the_user_opens_page_chooser(context: Context) -> None:
...
     modal.get_by_role("columnheader", name="Locale").is_visible()
+
+@step('the user can inspect Bundle details with name "{bundle_name}" and creator "{creator}"')
+def the_user_can_see_the_bundle_details_with_creator(context: Context, bundle_name: str, creator: str) -> None:
+    expect(context.page.get_by_text("Created by")).to_be_visible()
+    expect(context.page.get_by_text(creator)).to_be_visible()
+
+@step('a bundle has been created with name "{bundle_name}" and creator "{creator}"')
+def a_bundle_has_been_created(context: Context, bundle_name: str, creator: str) -> None:
+    context.user = User.objects.create_user(username=creator,email= "email@example.com",password= "password",first_name= "first_Name",last_name= "Last_Name")
```
MaciekBaron, line 68: We also have a `UserFactory` that you can use to make things easier

functional_tests/steps/bundles.py
```diff
@@ -53,3 +57,48 @@ def the_user_opens_page_chooser(context: Context) -> None:
...
+    context.bundle = BundleFactory(name = bundle_name, created_by = context.user )
+
+@step('the bundle menu shows bundle with name "{bundle_name}" and creator "{creator}"')
+def bundle_menu_show(context: Context, bundle_name: str, creator: str) -> None:
+    context.page.goto(context.base_url + "/admin/bundle/")
```
MaciekBaron, line 73: I think going to the page should be a separate step (here and on line 97)

MaciekBaron (changes requested):

functional_tests/steps/bundles.py
```diff
@@ -67,3 +69,63 @@ def the_selected_datasets_are_displayed(context: Context) -> None:
...
         "Deaths registered weekly in England and Wales by region (Edition: Example Dataset 3, Ver: 1)"
     ).is_visible()
+
+
+# bundle create amend
+@step('a bundle has been created with name "{bundle_name}" and creator "{creator}"')
+def a_bundle_has_been_created_with_user(context: Context, bundle_name: str, creator: str) -> None:
+    context.user = UserFactory.create(
+        username=creator, email="email@example.com", first_name="First_Name", last_name="Last_Name"
+    )
```
MaciekBaron, line 79: We don't need to call `.create` or provide the name or email – those will be randomly generated for us anyway.

This can be simplified to:
```suggestion
    context.user = UserFactory(username=creator)
```

However, looking through your additions again I don't think you ever check of this exact username anyway, so do we need to provide it in the first place? I see that you grab the full name and username directly from the user object you create.

I think we can remove that parameter to keep things simple and just have:
```python
context.user = UserFactory()
```

MaciekBaron (changes requested):

functional_tests/steps/bundles.py
```diff
@@ -67,3 +69,52 @@ def the_selected_datasets_are_displayed(context: Context) -> None:
     context.page.get_by_text(
         "Deaths registered weekly in England and Wales by region (Edition: Example Dataset 3, Ver: 1)"
     ).is_visible()
+
+
+# bundle create amend
+@step('a bundle has been created with name "{bundle_name}" and creator')
```
kacperpONS, line 75: I don't think it's worth parametrising this step tbh, we can just rely on the BundleFactory to generate it. 

Similarly in `bundle_menu_contains_value` and `bundle_menu_does_not_contain_value` you don't need to pass the bundle name as an argument and simply do `bundle.context.name`

functional_tests/steps/bundles.py
```diff
@@ -67,3 +69,52 @@ def the_selected_datasets_are_displayed(context: Context) -> None:
...
+
+@step("the bundle has creator removed")
+def delete_bundle_creator(context: Context) -> None:
+    context.bundle_creator.delete()
+
+
+# bundle goto
+@step("the user goes to the bundle inspect page")
+def go_to_bundle_inspect(context: Context) -> None:
+    context.page.goto(context.base_url + "/admin/bundle/inspect/1/")
```
kacperpONS, line 89: We should be careful with hardcoding urls, probably better to do `reverse('bundle:inspect', args=[context.bundle.pk])`

kacperpONS (approved):

PR #175, opened 2025-04-29
Chore: charts code cleanup

nimasmi:

This makes two non-functional changes to the datavis app code:

1. Change a test case name: ~AppendTagTestCase~ → ExtendFunctionTestCase, since it tests the `extend` function, which isn't a tag.
2. Remove use of the `Chart.supports_FOO` pattern for chart subclasses. The methods instead just detect the presence or absence of a value in the data, and the subclasses instead add or remove fields to enable setting those values or not.

I expect a future refactor of the charts will move to making better use of `StructValue` classes.

### How to review

> [!NOTE]
> This contains the code from #169, so is best reviewed after that is merged. Failing that, view individual commits to review.

No functional testing is needed, and there is no ticket for this work. So long as existing tests pass, this should be just a refactor.

### DIFF
--- a/cms/core/tests/test_tags.py
+++ b/cms/core/tests/test_tags.py
@@ -3,7 +3,13 @@ from django.test import TestCase
 from cms.core.templatetags.util_tags import extend
 
 
-class AppendTagTest(TestCase):
+class ExtendFunctionTest(TestCase):
+    """Tests of the `extend()` function.
+
+    This isn't registered as a template tag, but made globally available to be used in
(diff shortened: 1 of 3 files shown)


zerolab:
@nimasmi could you rebase this one to help reviews?

nimasmi:
> @nimasmi could you rebase this one to help reviews?

I reset the branch and cherry-picked the relevant commits. It should be less noisy now.

zerolab (approved):

PR #177, opened 2025-04-29
Add configurable chart aspect ratios

nimasmi:

Add configurable chart aspect ratios.

Ticket: [CCB-76](https://jira.ons.gov.uk/browse/CCB-76)

### How to review

- Check out the branch, and create an Information Page.
- Add various different types of chart block (except horizontal bar chart)
- For each, add **options** > **Desktop aspect ratio**, or **options** > **Mobile aspect ratio**.
- See that the aspect ratio is changed
- Try again for a bar chart, and see an error message

### DIFF
--- a/cms/datavis/blocks/base.py
+++ b/cms/datavis/blocks/base.py
@@ -111,9 +111,7 @@ class BaseVisualisationBlock(blocks.StructBlock):
     def get_context(self, value: "StructValue", parent_context: Optional[dict[str, Any]] = None) -> dict[str, Any]:
         context: dict[str, Any] = super().get_context(value, parent_context)
 
-        # Add template and visualisation context to support rendering
-        context["config"] = self.get_component_config(value)
-        context["download"] = self.get_download_config(value)
+        context["chart_config"] = self.get_component_config(value)
         return context
 
     def get_highcharts_chart_type(self, value: "StructValue") -> str:
@@ -126,12 +124,22 @@ class BaseVisualisationBlock(blocks.StructBlock):
         headers: list[str] = value["table"].headers
         rows: list[list[str | int | float]] = value["table"].rows
 
-        return {
+        config = {
+            "chartType": self.get_highcharts_chart_type(value),
+            "theme": value.get("theme"),
+            "title": value.get("title"),
+            "subtitle": value.get("subtitle"),
+            "caption": value.get("caption"),
             "legend": value.get("show_legend", True),
             "xAxis": self.get_x_axis_config(value.get("x_axis"), rows),
             "yAxis": self.get_y_axis_config(value.get("y_axis")),
             "series": self.get_series_data(value, headers, rows),
+            "useStackedLayout": value.get("use_stacked_layout"),
+            "annotations_values": self.get_annotations_config(value),
+            "download": self.get_download_config(value),
         }
+
+        return config
 
     def get_x_axis_config(
         self,
--- a/cms/datavis/blocks/base.py
+++ b/cms/datavis/blocks/base.py
@@ -1,6 +1,7 @@
 from collections.abc import Sequence
 from typing import Any, ClassVar, Optional, cast
 
+from django.core.exceptions import ValidationError
 from django.forms.widgets import Media, RadioSelect
 from django.utils.functional import cached_property
 from wagtail import blocks
@@ -106,14 +107,16 @@ class BaseVisualisationBlock(blocks.StructBlock):
         required=False,
     )
 
+    desktop_aspect_ratio = "desktop_aspect_ratio"
+    mobile_aspect_ratio = "mobile_aspect_ratio"
     options_key_map: ClassVar[dict[str, str]] = {
-        "desktop_aspect_ratio": "percentageHeightDesktop",
-        "mobile_aspect_ratio": "percentageHeightMobile",
+        desktop_aspect_ratio: "percentageHeightDesktop",
+        mobile_aspect_ratio: "percentageHeightMobile",
     }
     options = blocks.StreamBlock(
         [
             (
-                "desktop_aspect_ratio",
+                desktop_aspect_ratio,
                 AspectRatioBlock(
                     required=False,
                     help_text='Remove this option, or set "Default", to use the default aspect ratio for desktop.',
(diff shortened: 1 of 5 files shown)


cms/datavis/tests/test_chart_blocks.py
```diff
@@ -46,7 +46,7 @@ def setUp(self):
                 "value_suffix": "",
                 "tooltip_suffix": "",
             },
-            "annotations": [{"type": "point", "value": {"label": "Peak", "x_position": "2", "y_position": "140"}}],
+            "annotations": [{"type": "point", "value": {"label": "Peak", "x_position": 2, "y_position": 140}}],
```
nimasmi, line 49: **Comment**: I wasn't sure why this didn't raise any errors on the `feature/datavis` branch, but surely enough it didn't. It may seem unrelated, but it's fixed here because it occurred here. :shrug:

cms/datavis/blocks/base.py
```diff
@@ -105,15 +107,46 @@ class BaseVisualisationBlock(blocks.StructBlock):
...
+        mobile_aspect_ratio: "percentageHeightMobile",
+    }
+    options = blocks.StreamBlock(
+        [
+            (
+                desktop_aspect_ratio,
+                AspectRatioBlock(
+                    required=False,
+                    help_text='Remove this option, or set "Default", to use the default aspect ratio for desktop.',
+                    widget=RadioSelect(),
```
nimasmi, line 124: **Comment**: I was unsure why these attributes (particularly `required` and `widget`) needed to be set here as constructor kwargs, rather than defined as block subclass attributes.

At [Block options and methods](https://docs.wagtail.org/en/stable/reference/streamfield/blocks.html#block-options-and-methods), the Wagtail docs say: _All block definitions accept the following optional keyword arguments or Meta class attributes:_, but at [ChoiceBlock](https://docs.wagtail.org/en/stable/reference/streamfield/blocks.html#wagtail.blocks.ChoiceBlock) they say _The following keyword arguments are accepted in addition to the standard ones:_. Perhaps there's some subtle extra precision in that which means these are keyword-only.

cms/datavis/blocks/base.py
```diff
@@ -126,13 +159,24 @@ def get_component_config(self, value: "StructValue") -> dict[str, Any]:
...
+            "title": value.get("title"),
+            "subtitle": value.get("subtitle"),
+            "caption": value.get("caption"),
             "legend": value.get("show_legend", True),
             "xAxis": self.get_x_axis_config(value.get("x_axis"), rows),
             "yAxis": self.get_y_axis_config(value.get("y_axis")),
             "series": self.get_series_data(value, headers, rows),
+            "useStackedLayout": value.get("use_stacked_layout"),
+            "annotations_values": self.get_annotations_config(value),
+            "download": self.get_download_config(value),
```
nimasmi, line 175: **Comment**: This refactor in 28ba8d3 is to enable better dictionary manipulation in Python, for the additional options below.

cms/datavis/blocks/chart_options.py
```diff
@@ -0,0 +1,27 @@
...
+    choices = AspectRatioChoices.choices
+
+    def get_field(self, **kwargs: Any) -> TypedChoiceField:
+        """Coerce choices to integers, or None.
+
+        The Design System expects aspect ratios as integer percentages. To
+        revert to the default behaviour, the key must be either omitted or set
+        to `null`.
+        """
+        return TypedChoiceField(coerce=int, empty_value=None, **kwargs)
```
nimasmi, line 27: **Praise**: Thanks to @JakubMastalerz for helping me debug the types here.

cms/datavis/blocks/base.py
```diff
@@ -246,6 +299,22 @@ def get_download_config(self, value: "StructValue") -> dict[str, Any]:
...
+        if self.get_highcharts_chart_type(result) == "bar":
+            for i, option in enumerate(result["options"]):
+                if option.block_type in aspect_ratio_keys:
+                    options_errors[i] = ValidationError("Bar charts do not support aspect ratio options.")
+
+        if options_errors:
+            raise blocks.StructBlockValidationError(
+                block_errors={"options": blocks.StreamBlockValidationError(block_errors=options_errors)}
+            )
+        return result
```
MaciekBaron, line 316: A minor suggestion to follow the "return early" pattern and reduce nesting:
```suggestion
    def clean(self, value: "StructValue") -> "StructValue":
        result = super().clean(value)
        
        if self.get_highcharts_chart_type(result) != "bar":
            return result

        options_errors = {}
        aspect_ratio_keys = [self.desktop_aspect_ratio, self.mobile_aspect_ratio]
        for i, option in enumerate(result["options"]):
            if option.block_type in aspect_ratio_keys:
                options_errors[i] = ValidationError("Bar charts do not support aspect ratio options.")

        if options_errors:
            raise blocks.StructBlockValidationError(
                block_errors={"options": blocks.StreamBlockValidationError(block_errors=options_errors)}
            )
        return result
```

cms/datavis/blocks/base.py
```diff
@@ -227,6 +272,14 @@ def get_series_data(
...
+        options = {}
+        for option in value.get("options", []):
+            key = self.options_key_map[option.block_type]
+            options[key] = option.value
+
+        return options
+
     def get_download_config(self, value: "StructValue") -> dict[str, Any]:
         return {
             "title": f"Download: {value['title'][:30]} ... ",  # TODO work on download metadata
```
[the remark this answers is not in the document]
nimasmi: Thanks. This is code to be worked on in future. I don't know whether we'll even be truncating in future. Is it okay to leave it for now?

cms/datavis/blocks/base.py
nimasmi: I expect in future there will be more validation here. There's an open question on the tick intervals ticket about whether the team wants us to validate that any axis maxima and minima are a multiple of the tick interval, for example. I appreciate the suggestion, but perhaps this is something to look at when we do code review for the charts feature as a whole.

cms/datavis/blocks/base.py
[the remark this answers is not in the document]
MaciekBaron: Of course, feel free to ignore for now.

MaciekBaron (approved):

PR #187, opened 2025-05-09
Add retry for restoring snapshot in behave_fixtures

MaciekBaron:

The functional tests are currently quite flaky, and most often fail due to the following exception:

```
psycopg2.errors.ObjectInUse: database "test_ons" is being accessed by other users
```

This gets thrown by the following line:

```python
restore_snapshot(find_snapshot(context.clean_snapshot_name))
```

The exception has a knock on effect on any scenario which is run after it, as the database is not in a good state.

The PR adds a retry to the tear down and restore segment, alongside a 1 second wait to try and avoid any race conditions.

Note that this very rarely happens locally, and most commonly occurs when running as a GitHub action.

### How to review

Look at the pipeline and see if it passed correctly.

### Follow-up Actions

As this is just a workaround, we need to determine what the real issue is behind this problem.

### DIFF
--- a/functional_tests/behave_fixtures.py
+++ b/functional_tests/behave_fixtures.py
@@ -1,4 +1,7 @@
+import time
+
 import django
+import psycopg2
 from behave import fixture
 from behave.runner import Context
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


kacperpONS (approved):
Godsend!

zerolab (approved):

sanjeevz3009 (commented):
Reviewed!

PR #188, opened 2025-05-12
Prevent copying of topic and theme pages

MaciekBaron:

This PR related to CMS-430 and resolves an issue where unique taxonomy requirements would raise an exception when being copied due to the fact that there cannot be two pages with the same topic assigned.

This change displays a message explaining this when an admin user attempts to copy/duplicate a theme or topic page.

Loom video: https://www.loom.com/share/565abe76084e40f097a7a74f8145d96a?sid=1c717ce4-77b5-418e-bcc2-b9456b0b642d

### How to review

1. Checkout branch
2. Create a theme page and a topic page
3. Try to copy a theme page and a topic page
4. Confirm that in each scenario you are prevented from doing so, and a message is displayed

### Follow-up Actions

Confirm that the message used is satisfactory.

### DIFF
--- a/cms/themes/tests/test_pages.py
+++ b/cms/themes/tests/test_pages.py
@@ -1,0 +1,56 @@
+from http import HTTPStatus
+
+from django.urls import reverse
+from wagtail.test.utils import WagtailPageTestCase
+
+from cms.topics.tests.factories import ThemePageFactory
+
+
+class TopicPageTests(WagtailPageTestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.page = ThemePageFactory()
+        cls.superuser = cls.create_superuser(username="admin")
+
+    def test_default_route(self):
+        self.assertPageIsRoutable(self.page)
+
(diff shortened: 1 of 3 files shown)


zerolab:
Note: to check how this impact when we disable the uniqueness enforcement

cms/themes/tests/test_pages.py
```diff
@@ -0,0 +1,56 @@
+from http import HTTPStatus
+
+from django.urls import reverse
+from wagtail.test.utils import WagtailPageTestCase
+
+from cms.topics.tests.factories import ThemePageFactory
+
+
+class TopicPageTests(WagtailPageTestCase):
```
zerolab, line 9: nitpick:
```suggestion
class ThemePageTests(WagtailPageTestCase):
```

zerolab (changes requested):
A small nitpick, other than that it does what it says on the tin

nehakerung (commented):
LGTM - does what it says.
Some functional-tests would be good to add.

PR #189, opened 2025-05-12
Ensure dockerignore matches gitignore

RealOrangeOne:

Particularly useful when building the image locally, where the checkout may not be pristine. 

A script (which runs in CI) checks the files are the same (besides some extra bits in `.dockerignore`.

I went with this option over just symlinking in part to allow for them to drift, but also to ensure Windows support.

### How to review

Ensure that any changes to `.dockerfile` or `.gitignore` are correct and intended.


MaciekBaron:
**Question**: Could we not simplify this by using the inbuilt `diff` command?

Something like:
```bash
diff <(sed '1d' .gitignore) <(sed '1,6d' .dockerignore)
```

or if we want to anticipate additional changes above the comment:

```bash
diff <(sed '1d' .gitignore) <(sed '1,/# The below must be in sync/d' .dockerignore)
```

This saves us from having to add an additional Python script, adding it to the mypy setup, and we could use a lightweight runner instead without Python (so we skip the `actions/setup-python@v5` step).

RealOrangeOne:
It could probably be done with `sed`, but this makes it much easier to debug, and a bit more versatile. Adding a few more files to the Python linting isn't much of an issue.

sanjeevz3009 (approved):
LGTM!

MaciekBaron (approved):
Makes sense - LGTM!

PR #190, opened 2025-05-13
Column chart with line

nimasmi:

Implement "column chart with line".

Ticket: [CCB-64](https://jira.ons.gov.uk/browse/CCB-64)

This permits an editor to select a series for customisation. There is only one possible customisation at the moment: "series as line overlay". Multiple series can be selected, but not all.

### How to review

- Create an Information Page.
- Add a Bar/Column chart.
- Set the orientation to Column
- Add more than one data series.
- Add a "Series customisation", and enter the number of one of your series.
- Try this with a stacked column chart.
- Try some of the various error states (you may need to save the page to see the messages):
    - Enter 0, or a value higher than the number of series
    - Enter the same series number twice
    - Configure every data series as a line overlay
    - Try this with a bar chart

<p>Successful state</p>

image

<p>One of various possible error messages.</p>

image

</details>

### DIFF
--- a/cms/datavis/blocks/charts.py
+++ b/cms/datavis/blocks/charts.py
@@ -116,20 +116,20 @@ class BarColumnChartBlock(BaseVisualisationBlock):
         seen_series_numbers = set()
         for i, block in enumerate(value.get("series_customisation", [])):
             if block.block_type == self.SERIES_AS_LINE_OVERLAY_BLOCK:
-                if block.value in seen_series_numbers:
-                    sub_block_errors[i] = ValidationError("Duplicate series number.", code=self.ERROR_DUPLICATE_SERIES)
-                seen_series_numbers.add(block.value)
-
                 if value.get("select_chart_type") == self.ChartTypeChoices.BAR:
                     sub_block_errors[i] = ValidationError(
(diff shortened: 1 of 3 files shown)


cms/datavis/blocks/charts.py
```diff
@@ -74,6 +87,64 @@ class BarColumnChartBlock(BaseVisualisationBlock):
...
+                if value.get("select_chart_type") == self.ChartTypeChoices.BAR:
+                    sub_block_errors[i] = ValidationError(
+                        "Horizontal bar charts do not support line overlays.", code=self.ERROR_HORIZONTAL_BAR_NO_LINE
+                    )
+
+                # Raise an error if the series number is not in the range of the number of series
+                if block.value < 1 or block.value > len(series):
+                    sub_block_errors[i] = ValidationError(
+                        "Series number out of range.", code=self.ERROR_SERIES_OUT_OF_RANGE
+                    )
```
MaciekBaron, line 132: **Suggestion**: I think these can be `elif`s as `sub_block-errors[i]` gets set inside each of the blocks, so it gets overwritten anyway (we would just need to move `seen_series_numbers.add(block.value)` to after both `elif`s)

PR #191, opened 2025-05-13
Force read only DB connection

RealOrangeOne:

Whilst Django's handling of database connections is pretty good, it's not perfect. This PR makes local development match production - where the `read_replica` connection is read-only. 

This lead to the discovery of https://code.djangoproject.com/ticket/36389, where pages can't be published because the code uses the replica connection for writes. The solution (for now) is to identify models which are the target of these generic relations and force them to use the default connection. 

At the time of writing, these models are:

```python
{<class 'taggit.models.TaggedItem'>, <class 'wagtail.search.models.IndexEntry'>, <class 'wagtail.models.WorkflowState'>, <class 'wagtail.models.Revision'>}
```

This will have a slight performance / load penalty when interacting with these models, however it will hopefully be minimal. This only happens for interactions from the "internal" environment - "external" will never do writes to these models so always use the replicas (where performance is most important anyway).

I tried modifying the existing tests for the publish management commands, but they didn't play nicely with `TransactionTestCase`.

Additionally, I removed the overrides for settings around database configuration. These replace the configuration in `base.py`, which diverges too far from what's used in production.

### How to review

With this branch, do some basic smoke tests, particularly around things which write to the database.

Similarly, test that your local setup isn't broken by the database changes.

### Follow-up Actions

Once https://code.djangoproject.com/ticket/36389 is released, the workaround can be removed. This is unlikely to be anytime soon.


cms/settings/functional_test.py
```diff
@@ -1,10 +1,3 @@
-import copy
-
 from .test import *  # noqa: F403  # pylint: disable=wildcard-import,unused-wildcard-import
 
-DATABASES = {
-    "default": dj_database_url.config(default="postgres://ons:ons@localhost:15432/ons"),  # noqa: F405
-}
```
MaciekBaron, line 7: NB: the port here is 15432 as opposed to 5432 used elsewhere, you might want to keep this or otherwise take this into account as this is why your functional tests are currently failing.
https://github.com/ONSdigital/dis-wagtail/blob/30380c1a236b3df0590f6acf3d3d959c08b3b94d/functional_tests/docker-compose.yml#L6-L7

RealOrangeOne: Oh wow, nice catch!

cms/core/models/base.py
```diff
@@ -132,6 +133,24 @@ def has_ons_embed(self) -> bool:
 
         return False
 
+    def publish(self, *args: Any, **kwargs: Any) -> Any:
```
zerolab, line 136: hmm... I wonder if this is enough. For example the page edit view doesn't call `page.publish()` directly

RealOrangeOne: Huh, so you're right: https://github.com/wagtail/wagtail/blob/47fc2008640ece76be97f3d03da224b903a99047/wagtail/admin/views/pages/edit.py#L593 :sob:

RealOrangeOne: After much pain, this is fixed in https://github.com/ONSdigital/dis-wagtail/pull/191/commits/2989218ffafbf29a62e8ee60b88220e3462b4b35

cms/core/db_router.py
```diff
@@ -25,8 +27,21 @@ def __init__(self) -> None:
...
+        # @see db_for_read
+        self.generic_target_models: set[type[Model]] = set()
+        for model in apps.get_models():
+            for field in model._meta.get_fields():
+                if isinstance(field, GenericRelation) and field.related_model != "self":
+                    self.generic_target_models.add(field.related_model)
+
     def db_for_read(self, model: type[Model], **hints: Any) -> Optional[str]:
         """Determine which database should be used for read queries."""
+        # Models which are targetted by a GenericRelation must use the write connection.
```
zerolab, line 40: nitpick: 
```suggestion
        # Models which are targeted by a GenericRelation must use the write connection.
```

or even simpler, "Models used in a GenericRelation must use the write connection"

zerolab (commented):
Left a non-blocking comment.
This looks good to me, and much cleaner than the previous iteration (or the other options we discussed)

MaciekBaron (approved):

cms/core/db_router.py
RealOrangeOne: Fixed in https://github.com/ONSdigital/dis-wagtail/pull/191/commits/31ffdfd22b4d11a270ce83c97908e8bca5dde2f2

sanjeevz3009 (approved):
LGTM!

PR #192, opened 2025-05-14
Add non-js equation generation

MaciekBaron:

This PR relates to CMS-502. It adds non-js support for equations by using matplotlib to generate an SVG fallback for each equation.

**Loom video:**
https://www.loom.com/share/5ba535b27fe240f08f94b2270ef5b797

### How to review

**Note**: This PR introduces new OS-level dependencies. You can either rebuild your docker images, or you can enter the shell using `docker shell` and install them using `sudo apt install -y texlive-latex-extra texlive-fonts-recommended cm-super`.

1. Start creating an information page, methodology page or a statistical article page
2. Add an equation block
3. Save a valid equation, e.g. `$$\sum_{i=0}^n i^2 = \frac{(n^2+n)(2n+1)}{6}$$`
4. Preview the page
5. Ensure that the equation is displayed
6. Disable JavaScript and reload the page (this can be done using the [Toggle Javascript extension](https://chromewebstore.google.com/detail/toggle-javascript/cidlcjdalomndpeagkjpnefhljffbnlo?hl=en) on Chrome)
7. Ensure that the fallback is displayed

### Follow-up Actions

N/A


.github/workflows/ci.yml
```diff
@@ -139,7 +139,7 @@ jobs:
       - name: Install dependencies
         run: |
           make install-dev
-          sudo apt-get install -y gettext
+          sudo apt-get install -y gettext texlive-latex-extra texlive-fonts-recommended cm-super
```
zerolab, line 142: cc @RealOrangeOne - these will probably needed on the infra TF side too

zerolab: on reflection, we do use Docker images, and that is captured...

cms/static_src/sass/components/_js-only-element.scss
```diff
@@ -0,0 +1,9 @@
+@use 'config' as *;
+
+.js-only-element {
+    display: none;
+}
+
+.ons-js-enabled .js-only-element {
+    display: block;
+}
```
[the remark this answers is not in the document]
MaciekBaron: Addressed in [f2ddbd0](https://github.com/ONSdigital/dis-wagtail/pull/192/commits/f2ddbd04c6de272fd6d3885d84a629a93d5cc817)

cms/static_src/sass/components/_equation.scss
```diff
@@ -0,0 +1,6 @@
+@use 'config' as *;
+
+.ons-no-js-equation-container {
+    text-align: center;
+    overflow: auto;
```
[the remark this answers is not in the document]
MaciekBaron: Addressed in [f2ddbd0](https://github.com/ONSdigital/dis-wagtail/pull/192/commits/f2ddbd04c6de272fd6d3885d84a629a93d5cc817)

cms/jinja2/templates/components/streamfield/equation_block.html
```diff
@@ -0,0 +1,6 @@
+<div class="ons-u-db-no-js_disabled">
+    {{ value.equation }}
+</div>
+<noscript>
+    <div class="ons-u-ta-center">{{ value.svg | safe }}</div>
```
RealOrangeOne, line 5: Question: Could this be done by serving the SVG via an `img` tag, rather than baking it into the content directly? It's a lot of body bloat.

It also has the benefit of only trying to render it if it's actually used, which likely has some performance benefits.

cms/core/utils.py
```diff
@@ -1,10 +1,18 @@
+import io
 from typing import TYPE_CHECKING, Any, Optional, TypedDict
 
+import matplotlib as mpl
```
RealOrangeOne, line 4: Issue: matplotlib is not thread safe: https://matplotlib.org/stable/users/faq.html#work-with-threads.

Is there a different library we could use instead for this? Alternatively, this should be wrapped in a thread lock.

cms/jinja2/templates/components/streamfield/equation_block.html
MaciekBaron: Personally I don't think the impact is that significant – gzipped it's probably just a few KBs of extra payload. However if necessary we could create a custom view which returns the SVG, for instance.

The browser won't attempt to render it unless JS is disabled due to the `<noscript>` tag (if that is what you were talking about).

cms/core/forms.py
```diff
@@ -98,3 +100,33 @@ def clean_corrections(self) -> StreamValue:  # noqa: C901
...
+                    "content",
+                    ValidationError("The equation is not valid LaTeX. Please check the syntax and try again."),
+                )
+
+    def clean_content(self) -> StreamValue:
+        content: StreamValue = self.cleaned_data["content"]
+        if not content:
+            return content
+
+        for block in content:
```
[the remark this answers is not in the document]
RealOrangeOne: Do we know any details about what caching it does? I'm not sure how much disk space the containers will get in production, and we may limit it, so running out of disk space in the container mid request could result in errors.

cms/core/utils.py
MaciekBaron: Thanks for flagging this, I wrapped it in a lock and refactored it slightly based on [this](https://matplotlib.org/stable/gallery/user_interfaces/web_application_server_sgskip.html).

[1eda5ab](https://github.com/ONSdigital/dis-wagtail/pull/192/commits/1eda5abd109e35a9c8609d3703d04f707c4e2f33)

cms/jinja2/templates/components/streamfield/equation_block.html
RealOrangeOne: It won't render it on the page, but it'll still need to download and parse it. I guess it depends how complex the SVGs are. Small SVGs might be ok. In general I tend to prefer offloading that to a separate request even if it does only save a little page size.

MaciekBaron: I understand where you are coming from. I'll raise this with others and see how much of a priority we see this to be 👍

cms/core/forms.py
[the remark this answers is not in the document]
RealOrangeOne: I suspect it's in `matplotlib`'s general cache: https://matplotlib.org/devdocs/api/matplotlib_configuration_api.html#matplotlib.get_cachedir.

Let's keep this implementation, and I'll make sure production can deal with it.

cms/jinja2/templates/components/streamfield/equation_block.html
zerolab: My vote is on keeping this as is.

Offloading it to a separate request adds enough complexity that we don't need.

cms/core/forms.py
```diff
@@ -98,3 +100,33 @@ def clean_corrections(self) -> StreamValue:  # noqa: C901
 
 class NoLocaleFilterInChoosersForm(SearchFilterMixin, BaseFilterForm):
     """A chooser filter form that deliberately excludes the locale filter."""
+
+
+class PageWithEquationsAdminForm(WagtailAdminPageForm):
```
zerolab, line 105: we will need a system check that ensures any Page model that has an equation block has an admin form that inherit from this

MaciekBaron: Addressed in [23af0db](https://github.com/ONSdigital/dis-wagtail/pull/192/commits/23af0db6f4f4a2b6076fb77ba86edb072340cec0)

cms/core/checks.py
```diff
@@ -29,3 +33,22 @@ def check_wagtail_settings(*args: Any, **kwargs: Any) -> Iterator[CheckMessage]:
                 hint=f"Ensure generic setting extends {BaseGenericSetting!r}",
                 obj=model,
             )
+
+
+@register(Tags.models)
+def check_wagtail_pages(*args: Any, **kwargs: Any) -> Iterator[CheckMessage]:
+    for model in apps.get_models():
+        if not issubclass(model, Page):
+            continue
```
zerolab, line 42: suggestion: use `get_page_models` https://github.com/wagtail/wagtail/blob/988a48a75adc5a8ac0b68da9e65062349bad15b2/wagtail/models/pages.py#L107

cms/core/forms.py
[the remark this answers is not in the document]
zerolab: Given:
- requests may go to different web pods
- the cache may or may not be primed
- we're adding one more thing to QA and test

I am still leaning on "only do the work if it is needed", but let's add this as a next step so we test with a bigger number of equations in sandbox/staging

cms/core/forms.py
```diff
@@ -98,3 +100,33 @@ def clean_corrections(self) -> StreamValue:  # noqa: C901
...
+    def _process_content_block(self, block: StreamValue) -> None:
+        if block.block_type == "equation":
+            equation = block.value["equation"]
+            # Remove $$ from the start and end of the equation
+            if equation.startswith(FORMULA_INDICATOR) and equation.endswith(FORMULA_INDICATOR):
+                equation = equation[2:-2]
+            try:
+                block.value["svg"] = latex_formula_to_svg(equation)
+            except RuntimeError:
```
zerolab, line 114: we should log the original exception to the system log

zerolab (changes requested):
Lef a couple of more notes, mostly tidy up.
Looking good otherwise

sanjeevz3009 (approved):
Other than the discussions that's already happening. Everything looks good to me testing it locally.

Nit: Following up on the slack, message, it would be good to have a UI test that checks if the equation is rendering on the page when JS is disabled, if possible.

cms/core/forms.py
MaciekBaron: Logging added in [ff3d727](https://github.com/ONSdigital/dis-wagtail/pull/192/commits/ff3d72796fbed52aade83a81b7ff427c7ebd643f)

MaciekBaron:
@sanjeevz3009 non-JS UI tests added here [6c1686f](https://github.com/ONSdigital/dis-wagtail/pull/192/commits/6c1686fb20f94e0644604a41b4856c3af1070d51)

Hopefully this will allow for more non-JS fallback tests to be added.

sanjeevz3009:
> @sanjeevz3009 non-JS UI tests added here [6c1686f](https://github.com/ONSdigital/dis-wagtail/pull/192/commits/6c1686fb20f94e0644604a41b4856c3af1070d51)
> 
> Hopefully this will allow for more non-JS fallback tests to be added.

Thanks, Maciej, for taking the time to look into this and implementing it. Will re-review sometime today =)

sanjeevz3009 (approved):
Thank you for adding the no JS UI tests. Approved!

cms/core/checks.py
MaciekBaron: Changed in [9afa0fb](https://github.com/ONSdigital/dis-wagtail/pull/192/commits/9afa0fb22387597083060e80d701f986151c2a6a)

zerolab (approved):

zerolab (approved):

sanjeevz3009 (approved):

PR #193, opened 2025-05-14
Accessibility: HTML Validation Fixes

janehughes:

Tickets: 
https://jira.ons.gov.uk/browse/CMS-493
https://jira.ons.gov.uk/browse/CMS-494
https://jira.ons.gov.uk/browse/CMS-496

- Removes unnecessary `<section>` and `<h2>` from the glossary streamfield component - the `value.title` was generating an empty string and causing an invalid id on the generated `<section>` element which is not needed anyway
- Updates the methodology & stat article hero template to remove the `{% set = "" %}` logic as this was causing an empty div to be generated. I've updated the `itemsList` so now we are checking for the `term` in each item and only adding values that exist
- Remove `height="auto"` from image tags as this is invalid HTML and has no effect on the image

### How to review

- Add a Glossary component and see that the parent `<section>` and `<h2>` elements have been removed and are no longer adding empty values
- Check methodology & stat article pages
  - test adding and removing meta data from the hero to make sure they display as expected
  - inspect element to see that no empty `<div>`s are added when an optional field (such as contact details) is empty
- Check a release page and see that the hardcoded logos no longer have a `height="auto"` attribute and are still displaying as expected

### Follow-up Actions

N/A


cms/jinja2/templates/pages/methodology_page.html
```diff
@@ -10,44 +10,48 @@
...
+    {# fmt:on #}
 
+    {# filter out any optional items which haven't been set #}
+    {# fmt:off #}
+    {% set itemsList = [
+        publication_item,
+        last_revised_item,
+        contact_item
+    ] | selectattr('term') | list %}
+    {# fmt:on #}
```
zerolab, line 53: thought:


```
"itemsList": [publication_item, last_revised_item, contact_item] | selectattr('term') | list,
```

directly in the component call.

Same on the stats article page

cms/jinja2/templates/components/streamfield/glossary_section_block.html
```diff
@@ -4,22 +4,16 @@
...
+    onsAccordion({
+        "id": "accordion-" ~ block_id,
+        "allButton": {
+            "open": show_all_label,
+            "close": hide_all_label
+        },
+        "itemsList": formatted_glossary_terms,
+    })
+}}
+{# fmt:on#}
```
zerolab, line 18:
```suggestion
{{-
    onsAccordion({
        "id": "accordion-" ~ block_id,
        "allButton": {
            "open": show_all_label,
            "close": hide_all_label
        },
        "itemsList": formatted_glossary_terms,
    })
-}}
{# fmt:on #}
```

cms/jinja2/templates/pages/statistical_article_page.html
```diff
@@ -58,7 +37,57 @@
...
+    {% endif %}
+    {# fmt:on #}
+
+    {# fmt:off #}
+    {% if page.next_release_date %}
+        {% set next_release_item = {
+                "term": next_release_label,
+                "descriptions": [
+                    {
+                        "description": page.next_release_date|date("DATE_FORMAT") or _("To be announced")
```
zerolab, line 62: "To be announced" won't be picked up by Djanogs makemessages (ie. extract strings for translation)

zerolab (changes requested):
A couple more notes, then LGTM

cms/jinja2/templates/pages/methodology_page.html
janehughes: Nice suggestion, much prefer this thank you! ⭐ Updated

cms/jinja2/templates/pages/statistical_article_page.html
janehughes: Ah good spot thank you 👍  updated

cms/jinja2/templates/components/streamfield/glossary_section_block.html
janehughes: Updated - curious to know the benefit of `{{`- vs `{{` here if both work?

cms/jinja2/templates/pages/methodology_page.html
```diff
@@ -10,43 +10,52 @@
...
-            {
-                "term": last_revised_label,
+    {% set contact_description_html -%}
+        <a href="mailto:{{ page.contact_details.email }}">{{ page.contact_details.name }}</a>
+    {%- endset %}
+
+    {# fmt:off #}
+    {% if page.contact_details %}
+        {% set contact_item = {
+                "term": _("Contact:"),
```
MaciekBaron, line 47: I think we might need to move this to a variable like we did before so that it can be localised (here and in other places where we use an object/dictionary)

cms/jinja2/templates/components/census/census-logo.html
```diff
@@ -1 +1 @@
-<img src="https://cdn.ons.gov.uk/assets/images/census-logo/logo-census-2021-purple-landscape.svg" alt="Census 2021 logo" width="171" class="{{ census_logo_classes }}" width="171" height="auto">
+<img src="https://cdn.ons.gov.uk/assets/images/census-logo/logo-census-2021-purple-landscape.svg" alt="Census 2021 logo" width="171" class="{{ census_logo_classes }}" width="171">
```
MaciekBaron, line 1: NP: Double `width`
```suggestion
<img src="https://cdn.ons.gov.uk/assets/images/census-logo/logo-census-2021-purple-landscape.svg" alt="Census 2021 logo" width="171" class="{{ census_logo_classes }}">

MaciekBaron (changes requested):

cms/jinja2/templates/pages/methodology_page.html
janehughes: Good shout thank you - I've moved these to the top with the other labels now 👍

zerolab (approved):

sanjeevz3009 (approved):
Tested it locally and checked everything using DevTools. All looks good to me =)

PR #194, opened 2025-05-14
Refactor tests and business logic to use named objects

nimasmi:

There is no ticket for this work. It is refactoring and code cleanup for consistency and to reduce fragility.

- change some comparisons in business logic to use Enum constants (rather than their value strings)
- use named strings as error codes
- use error codes (rather than the message strings) for assertions in unit tests
- use capitalised names for constants (for consistency with changes added in #190)
- move bar-chart-only cleaning from the generic `BaseVisualisationBlock` to the child `BarAndColumnChartBlock`

### How to review

No functional changes. This needs code review and a passing test suite only.

### DIFF
--- a/cms/datavis/blocks/base.py
+++ b/cms/datavis/blocks/base.py
@@ -103,17 +103,17 @@ class BaseVisualisationBlock(blocks.StructBlock):
         required=False,
     )
 
-    desktop_aspect_ratio = "desktop_aspect_ratio"
-    mobile_aspect_ratio = "mobile_aspect_ratio"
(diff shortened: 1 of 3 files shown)


cms/datavis/blocks/base.py
```diff
@@ -301,22 +300,6 @@ def get_download_config(self, value: "StructValue") -> dict[str, Any]:
             ],
         }
 
-    def clean(self, value: "StructValue") -> "StructValue":
```
nimasmi, line 304: **Comment**: This is moved to `BarColumnChartBlock.clean()`

MaciekBaron (approved):

PR #195, opened 2025-05-14
Merge main branch into feature/datavis

nimasmi:
### What is the context of this PR?

This brings the datavis branch up to date with the latest work from the main project branch.

Ticket [CCB-74](https://jira.ons.gov.uk/browse/CCB-74)

PR #196, opened 2025-05-16
Update poetry dependencies and address mypy errors

MaciekBaron:

This PR addresses CMS-511.

The poetry lock file has been updated using `poetry update`, and mypy complaints have been addressed. This was done to help #195 get merged.

The ignored lines (`# type: ignore[misc]`) are due to an issue in `django-stubs`. Relevant issues and pull requests:
https://github.com/typeddjango/django-stubs/issues/2665
https://github.com/typeddjango/django-stubs/pull/2646

### How to review

1. Run `poetry install`
2. Ensure that `make mypy` passes without errors

### Follow-up Actions

Remove type ignores once issue is fixed in dependency.


sanjeevz3009 (approved):
LGTM!

nimasmi (approved):
These changes look good to me.

kacperpONS (approved):

poetry.lock
```diff
@@ -1,11 +1,12 @@
-# This file is automatically @generated by Poetry 1.8.4 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.1.1 and should not be changed by hand.
```
zerolab, line 1: not a big one, but we need to ensure the same poetry version as the one in the Dockerfile is used to update the lockfile (currently 1.8.4 - https://github.com/ONSdigital/dis-wagtail/blob/main/Dockerfile#L76)

PR #202, opened 2025-05-20
Add healthcheck endpoint

RealOrangeOne:

https://jira.ons.gov.uk/browse/CMS-468

Adds a healthcheck endpoint (as per https://github.com/ONSdigital/dp-standards/blob/main/HEALTH_CHECK_SPECIFICATION.md)

### How to review

Compare the endpoints output and the document above.


cms/core/views.py
```diff
@@ -68,14 +76,97 @@ def liveness(request: "HttpRequest") -> HttpResponse:
...
+        checks.append(
+            {
+                "name": f"{connection.alias} database",
+                "status": "CRITICAL" if failed else "OK",
+                "status_code": 500 if failed else 200,
+                "message": message,
+                "last_checked": now.isoformat(),
+                "last_success": now.isoformat() if not failed else None,
+                "last_failure": now.isoformat() if failed else None,
+            }
```
MaciekBaron, line 139: To keep the code DRY how about we write a function to create an entry so we can reuse it:
```python
def create_check_entry(name, message, failed):
    return {
        "name": name,
        "status": "CRITICAL" if failed else "OK",
        "status_code": 500 if failed else 200,
        "message": message,
        "last_checked": now.isoformat(),
        "last_success": now.isoformat() if not failed else None,
        "last_failure": now.isoformat() if failed else None,
    }
```

MaciekBaron (changes requested):
Just one comment, plus I think the tests might need to consider execution time as they are currently failing?

cms/core/views.py
RealOrangeOne: oooh, I like it!

MaciekBaron (approved):
LGTM!

zerolab:
question: would we ever have a WARNING state?

zerolab:
Sorry for the noise! Updated the wrong ticket 🙈

AdamHawtin (approved):

PR #204, opened 2025-05-21
Streamfield DS integration & updates

janehughes:

Tickets:
https://jira.ons.gov.uk/browse/CMS-431
https://jira.ons.gov.uk/browse/CMS-344

Also relevant for https://jira.ons.gov.uk/browse/CMS-337

Updates related links & document streamfield blocks to use DS components and match designs (see ticket for link to designs).

Note that I have spoken to Joe and we've agreed to divert from designs a little and keep the document links styling as-is, to make sure the description is included as we don't have a pattern in the DS to match the current Figma designs exactly. 

### How to review

### Follow-up Actions


cms/core/blocks/related.py
```diff
@@ -31,6 +31,30 @@ def get_context(self, value: "ListValue", parent_context: dict | None = None) ->
...
+
+            link_data = {"title": {"text": text, "url": url}}
+
+            description = item.get("description")
+            if description:
+                link_data["description"] = description
+
+            related_links.append(link_data)
+
+        context["related_links"] = related_links
```
zerolab, line 56: `RelatedLinksBlock` is a list of `RelatedContentBlock` (which extends [`LinkBlock`](https://github.com/ONSdigital/dis-wagtail/blob/1dec7c3740c126136689ed7bb4c3347ad39f462c/cms/core/blocks/base.py#L40) that in turn has a helpful [`LinkBlockStructValue`](https://github.com/ONSdigital/dis-wagtail/blob/1dec7c3740c126136689ed7bb4c3347ad39f462c/cms/core/blocks/base.py#L14) which provied a `link` property that currently outputs data like `{'url': '/', 'text': 'The title', 'description': ''}`.

I suggest updating that to `{"title": {'url': the_relevant_url, 'text': the_title}` and so this then becomes:

context["related_links"] = [item.link for item in value]

zerolab (changes requested):
One suggestion for improvement. Otherwise, LGTM

cms/core/blocks/related.py
janehughes: If we update LinkBlockStructValue to output a different structure for this, is that going to have an impact on any other blocks using it? Or is that not a problem? apologies if it's an obvious q!

nehakerung (approved):
LGTM - everything looks good

MaciekBaron (approved):

sanjeevz3009 (approved):
LGTM!

PR #205, opened 2025-05-21
Scatter plot

nimasmi:

This adds ScatterPlot to the available chart options.

Ticket: https://jira.ons.gov.uk/browse/CCB-52

Data format:

|X|Y|Group name|
|---|---|---|
|1|2|First|
|3|4|Second|
|5|6|First|

<p>

To try to help editors, I have added default value for a new table:

image

Example plot, using the data from the Design System

- [GitHub source](https://github.com/ONSdigital/design-system/blob/main/src/components/chart/example-scatter-chart.njk)
- [rendered on Netlify](https://ons-design-system-preview.netlify.app/components/chart/example-scatter-chart)

image

</p>
</details> 

### How to review

Create a Statistical Article Page, and within the content streamfield, add a Scatter Plot block.


cms/datavis/blocks/charts.py
```diff
@@ -173,3 +182,57 @@ def clean_options(self, value: "StructValue") -> "StructValue":
...
+    annotations = blocks.StreamBlock(
+        # Use coordinate-based annotations for scatter plots
+        [
+            ("point", CoordinatePointAnnotationBlock()),
+        ],
+        required=False,
+    )
+
+    class Meta:
+        icon = "chart-line"
```
nimasmi, line 221: **Comment**: Sadly [Font Awesome 4.7 icons](https://fontawesome.com/v4/icons/) (which comes from [wagtailfontawesome](https://github.com/dablabs/wagtailfontawesome)) do not include a scatter plot icon. There are some in the non-free version of Font Awesome 6: https://fontawesome.com/search?q=scatter&o=r.

MaciekBaron (approved):
Just one non-blocking ignorable suggestion, LGTM otherwise!

cms/datavis/blocks/charts.py
```diff
@@ -173,3 +182,57 @@ def clean_options(self, value: "StructValue") -> "StructValue":
...
+            groups[group_name].append((x, y))
+
+        series = []
+        for group_name, data_points in groups.items():
+            item = {
+                "name": group_name,
+                "data": data_points,
+                "marker": True,
+            }
+            series.append(item)
```
MaciekBaron, line 237: **Suggestion:** Ignore me if this is too pedantic, but maybe we could do this using a list comprehension to make it more pythonic? It's a question of preference though.
```suggestion
        series = [
            {
                "name": group_name,
                "data": data_points,
                "marker": True,
            }
            for group_name, data_points in groups.items()
        ]
```

MaciekBaron (approved):
Just one non-blocking ignorable suggestion, LGTM otherwise!

cms/datavis/blocks/charts.py
nimasmi: Funnily enough while adding this feature I did at one point consider a list comprehension, and rejected it for readability or something. It may have been here, but I don't remember. Yours looks better, so I'll take it. Thanks!

PR #208, opened 2025-05-22
Slim down packages used for LaTeX

MaciekBaron:

This PR relates to CMS-515 and an issue originally raised by @kacperpONS.

`texlive-fonts-recommended` is included in `texlive-latex-extra`, and `cm-super` doesn't seem to be needed currently.

Reference: https://tex.stackexchange.com/questions/245982/differences-between-texlive-packages-in-linux/504566#504566

### How to review

Ensure that equation generation still works.

### Follow-up Actions

Reconsider whether `cm-super` is needed: https://ctan.org/pkg/cm-super

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,7 +139,7 @@ jobs:
       - name: Install dependencies
         run: |
           make install-dev
-          sudo apt-get install -y gettext texlive-latex-extra texlive-fonts-recommended cm-super
+          sudo apt-get install -y gettext texlive-latex-extra
 
       - uses: actions/download-artifact@v4
         with:
@@ -225,7 +225,7 @@ jobs:
       - name: Install dependencies
         run: |
           make install-dev
-          sudo apt-get install -y gettext texlive-latex-extra texlive-fonts-recommended cm-super
+          sudo apt-get install -y gettext texlive-latex-extra
 
       - name: Playwright Install
         run: poetry run python -m playwright install --with-deps ${{ matrix.browser }}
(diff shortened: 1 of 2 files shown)


kacperpONS (approved):

AdamHawtin (approved):

sanjeevz3009 (approved):

Dockerfile
```diff
@@ -54,10 +54,9 @@ RUN apt --quiet --yes update \
         jq \
         unzip \
         gettext \
-        texlive-latex-extra \
-        texlive-fonts-recommended \
         cm-super \
         postgresql-common \
+    && apt --quiet --yes install --no-install-recommends texlive-latex-extra \
```
RealOrangeOne, line 59: Question: Why is this on a separate line - The above already uses `--no-install-recommends`. Does it need to be installed in a separate transaction?

MaciekBaron: You're right, these can be combined, addressed in [ad2bb93](https://github.com/ONSdigital/dis-wagtail/pull/208/commits/ad2bb9349c176f2257181c1050284e9687450740)

.github/workflows/ci.yml
```diff
@@ -139,7 +139,8 @@ jobs:
       - name: Install dependencies
         run: |
           make install-dev
-          sudo apt-get install -y gettext texlive-latex-extra texlive-fonts-recommended cm-super
+          sudo apt-get install -y gettext cm-super
+          sudo apt-get install -y --no-install-recommends texlive-latex-extra
```
zerolab, line 143: question: could the CI not do the install in one go too?

PR #200, opened 2025-05-20
Feature/statistical article page chart ccb 103

nimasmi:
### What is the context of this PR?

Ticket: [CCB-103 – Enable chart component on Statistical Article page - Chart builder](https://jira.ons.gov.uk/browse/CCB-103)

This enables charts on StatisticalArticlePage, and MethodologyPage. The latter is a side effect, but this was approved once questioned.

As a precursor step, this PR also removes previous work to load chart JS media with VisualisationPageMixin, and an associated check to make sure that any pages which could contain charts would subclass that mixin. At the moment, the ONS Design System component onsChart is loading the JavaScript assets.

### How to review

Add a chart to a SectionBlock on either StatisticalArticlePage, or MethodologyPage.

### Follow-up Actions

There should be a follow-up ticket to remove charts from InformationPage. However, since there are several such charts configured on the charts beta site at the moment, this would not be helpful for the team. This will be done in [CCB-124 – Disable charts on InformationPage – Chart builder](https://jira.ons.gov.uk/browse/CCB-124).

PR #209, opened 2025-05-22
Scatter Plot on Statistical Article and Methodology pages only

nimasmi:

Following the recent merge of #200 and #205, enable Scatter Plot on the future planned page types only.

### DIFF
--- a/cms/core/blocks/stream_blocks.py
+++ b/cms/core/blocks/stream_blocks.py
@@ -17,7 +17,7 @@ from cms.core.blocks import (
     WarningPanelBlock,
 )
 from cms.core.blocks.section_blocks import SectionBlock
-from cms.datavis.blocks import BarColumnChartBlock, LineChartBlock, ScatterPlotBlock
+from cms.datavis.blocks import BarColumnChartBlock, LineChartBlock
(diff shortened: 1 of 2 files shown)


cms/core/blocks/stream_blocks.py
```diff
@@ -59,7 +59,6 @@ class CoreStoryBlock(StreamBlock):
 
     line_chart = LineChartBlock(group="DataVis", label="Line Chart")
     bar_column_chart = BarColumnChartBlock(group="DataVis", label="Bar/Column Chart")
-    scatter_plot = ScatterPlotBlock(group="DataVis", label="Scatter Plot")
```
nimasmi, line 62: **Comment**: Though we are removing Scatter Plot Block, we are leaving Line Chart and Bar/Column Chart for now.

This is because Scatter Plot Block is new, and there will be no existing scatter plots on pages. Conversely, other team members will be testing recent features (such as annotations, aspect ratios, or axis configuration) using line or bar charts on the ons-charts-beta server. We'll give them a week's grace or so.

sanjeevz3009 (approved):

PR #214, opened 2025-05-28
Configure mobile/desktop tick interval independently

nimasmi:

Ticket: https://jira.ons.gov.uk/browse/CCB-93

This adds UI to configure the separate mobile/desktop tick interval parameters for charts.

### How to review

Edit any chart on a page. See that where there was previously just a "tick interval" field, there are now two, for mobile and desktop.

> [!NOTE]
> Bar/column charts should not have configurable tick intervals on their category axis, this is the x-axis for vertical column charts, and the y-axis for horizontal bar charts. However, as documented elsewhere, Highcharts transposes the axis for bar charts, so this rule is for the "x axis" in both cases. Our workaround is to call this the "category axis" for bar and column charts.
> See https://confluence.ons.gov.uk/display/NWP/Feature+mapping for this


cms/datavis/blocks/charts.py
```diff
@@ -68,8 +68,12 @@ class ChartTypeChoices(TextChoices):
...
         [
             (
-                "tick_interval",
-                TextInputIntegerBlock(label="Tick interval", required=False),
+                # Temporarily remove options. Title needs to be added back, so
+                # there's no point removing this field entirely.
+                "note",
+                blocks.StaticBlock(
+                    admin_text="(Temporary change) No options currently available for the category axis.",
+                ),
```
nimasmi, line 76: **Comment**: `"tick_interval"` should not have been here beforehand, but tick intervals had not yet been worked on. According to https://confluence.ons.gov.uk/display/NWP/Feature+mapping, this axis should still have an editable title, but since we have explicitly created a ticket https://jira.ons.gov.uk/browse/CCB-97 for this, I will leave this field temporarily blank and confusing.

cms/datavis/blocks/base.py
```diff
@@ -65,8 +65,12 @@ class BaseVisualisationBlock(blocks.StructBlock):
...
             ("max", TextInputFloatBlock(label="Maximum", required=False)),
             (
-                "tick_interval",
-                TextInputFloatBlock(label="Tick interval", required=False),
+                "tick_interval_mobile",
+                TextInputFloatBlock(label="Tick interval (mobile)", required=False),
+            ),
+            (
+                "tick_interval_desktop",
+                TextInputFloatBlock(label="Tick interval (desktop)", required=False),
```
MaciekBaron, line 73: **Question**: If you add a value for mobile, do you also need to provide a value for desktop, and vice-versa? If yes, maybe we can add some custom validation to enforce that? If not, please ignore.

MaciekBaron (approved):
LGTM!

PR #216, opened 2025-05-29
Ensure request errors are logged

RealOrangeOne:

I noticed when debugging an error that a number of log lines were being shown as `{}`, rather than containing any information. After some digging, it seems the formatter library we're using falls back to that if there's a JSON encoder issue:

https://github.com/marselester/json-log-formatter/blob/1797acb219d6e972ef8931df6ca9a3f2bffba163/json_log_formatter/__init__.py#L96

After some digging, having the `request` object in the record is what's preventing it from being encoded. With that removed, it works as expected.

Additionally, rather than the default formatting of just encoding the stacktrace as a string, I've used the [logging standard](https://github.com/ONSdigital/dp-standards/blob/main/LOGGING_STANDARDS.md#error-event-data), which makes it easier to read once encoded.

In future, if a message fails to encode, an error log is printed, which includes the raw (un JSON serialized) message, which should make debugging easier in future.

### How to review

Force an error in a view (eg `1 / 0`) and confirm that the exception details are logged correctly.


MaciekBaron (approved):
LGTM - will be good to add the tests at a later date as mentioned 👍

cms/core/logs.py
```diff
@@ -1,23 +1,43 @@
...
+    """A log formatter which outputs JSON and structures log messages as required.
+
+    https://github.com/ONSdigital/dp-standards/blob/main/LOGGING_STANDARDS.md
+    """
+
+    def json_record(self, message: str, extra: dict[str, Any], record: logging.LogRecord) -> dict[str, Any]:
+        record_data = {
+            "created_at": datetime.fromtimestamp(record.created),
+            "namespace": record.name,
+            "severity": SEVERITY_MAPPING.get(record.levelno, 3),
```
zerolab, line 26: non-blocking suggestion:

```suggestion
            "severity": SEVERITY_MAPPING.get(record.levelno, 3),  # Default to INFO
```

zerolab (approved):
Thank you for the tests. Added a small, non-blocking suggestion about the defatul error level

cms/core/logs.py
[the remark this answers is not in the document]
zerolab: These are mandated by https://github.com/ONSdigital/dp-standards/blob/main/LOGGING_STANDARDS.md#severity-levels

cms/core/logs.py
```diff
@@ -1,23 +1,43 @@
...
-class JSONFormatter(json_log_formatter.JSONFormatter):
-    """A log formatter which outputs JSON and structures log messages as required."""
-
-    def json_record(
-        self, message: str, extra: dict[str, str | int | float], record: LogRecord
-    ) -> dict[str, str | int | float]:
-        record_data: dict = super().json_record(message, extra, record)
 
-        del record_data["time"]
+class JSONFormatter(json_log_formatter.JSONFormatter):
```
sanjeevz3009, line 16: Per our discussion on Slack, when I tested this locally, I wasn't seeing any changes and was getting back `{}` as @kacperpONS also validated in his comment.

I created a new view and navigated to the URL (configured separately) for testing purposes:

```
def break_in_peace(request):
    """A quick endpoint that always raises ZeroDivisionError, so we can confirm
    our JSONFormatter catches it and prints a JSON blob to the console.
    """
    # This line will throw ZeroDivisionError:
    1 / 0
    return HttpResponse("You will never see this")
```

This returned `{}`, which shouldn't be the expected behaviour here.

Another test I did was breaking the `page_not_found` view:

```
def page_not_found(
    request: "HttpRequest", exception: Exception, template_name: str = "templates/pages/errors/404.html"
) -> HttpResponse:
    1 / 0
    """Custom 404 error view to use our page not found template."""
    return defaults.page_not_found(request, exception, template_name)
```

Same issue as before, as it returned `{}`.

After some digging, I figured adding:

```
def format(self, record):
        record.__dict__.pop("request", None)
        return super().format(record)
```

To the`JSONFormatter` class fixes the issue.

I think before the library’s format(record) was handed a LogRecord that still contained record.request = <HttpRequest>. As soon as it tried to serialise it, the JSON‐encoder blew up, and returned `{}`.

Now with the `format` override, we explicitly do `record.__dict__.pop("request", None)` in our override of `format()`. That means by the time the underlying library does its json.dumps(...), `record.request` has already been removed. Nothing non-serialisable remains, so you get a full JSON log.

Now the expected output is:

```
{"created_at": "2025-06-02T15:57:15.668394", "namespace": "django.request", "severity": 1, "event": "Internal Server Error: /cmsffff", "data": {"status_code": 500}}
```

```
{"created_at": "2025-06-02T16:12:57.689494", "namespace": "django.request", "severity": 1, "event": "Internal Server Error: /cms/test-break/", "data": {"status_code": 500}, "errors": [{"message": "ZeroDivisionError('division by zero')", "stack_trace": [{"file": "/Users/sanjeevsrithevan/Library/Caches/pypoetry/virtualenvs/cms-DZI0rODw-py3.12/lib/python3.12/site-packages/django/core/handlers/exception.py", "function": "inner", "line": "response = get_response(request)"}, {"file": "/Users/sanjeevsrithevan/Library/Caches/pypoetry/virtualenvs/cms-DZI0rODw-py3.12/lib/python3.12/site-packages/django/core/handlers/base.py", "function": "_get_response", "line": "response = wrapped_callback(request, *callback_args, **callback_kwargs)"}, {"file": "/Users/sanjeevsrithevan/Library/Caches/pypoetry/virtualenvs/cms-DZI0rODw-py3.12/lib/python3.12/site-packages/django/views/decorators/vary.py", "function": "_view_wrapper", "line": "response = func(request, *args, **kwargs)"}, {"file": "/Users/sanjeevsrithevan/Library/Caches/pypoetry/virtualenvs/cms-DZI0rODw-py3.12/lib/python3.12/site-packages/django/views/decorators/cache.py", "function": "_view_wrapper", "line": "response = viewfunc(request, *args, **kw)"}, {"file": "/Users/sanjeevsrithevan/Desktop/dis-wagtail/cms/core/views.py", "function": "break_in_peace", "line": "1 / 0"}]}]}
```

sanjeevz3009 (dismissed):
Per our Slack discussion, I requested the changes above.

cms/core/logs.py
zerolab: Good catch.

The issue is that the `request` object is in `extra` which is used as is in the modified `json_record`

sanjeevz3009: Thanks Jake. Resolving!

AdamHawtin (approved):
LGTM 👍

PR #221, opened 2025-05-30
Range annotations

nimasmi:

Ticket: https://jira.ons.gov.uk/browse/CCB-56

### How to review

Editing a chart on a Statistical Article Page, select Annotations > Range

<p>

#### Chooser

image

#### Dynamic validation

"Axis" set to X:

image

A float is allowed if "Axis" is set to Y.

You cannot set the three custom fields if "Label inside" is checked.

image

#### Bar/column chart axis labelling

We use different labels for bar/column charts, because Highcharts transposes X and Y willy-nilly.

image

#### Rendered

image

#### Category numbering behaviour

Setting

- start value: 2
- end value: 4

image

Is this predictable, and what users would expect?

</p>
</details>

### DIFF
--- a/cms/datavis/tests/test_annotations_range.py
+++ b/cms/datavis/tests/test_annotations_range.py
@@ -1,0 +1,209 @@
+from wagtail import blocks
+
+from cms.datavis.blocks.annotations import (
+    RangeAnnotationCategoricalBlock,
+    RangeAnnotationLinearBlock,
+)
+from cms.datavis.constants import AxisChoices
+from cms.datavis.tests.test_annotations_base import BaseAnnotationTestCase
+
+
+class GenericRangeAnnotationTestCase(BaseAnnotationTestCase):
+    block_type = RangeAnnotationLinearBlock
+
+    def test_basic(self):
+        raw_data = {
+            "label": "Jam",
+            "axis": "x",
+            "start_position": 3,
+            "end_position": 4,
+            "label_inside": False,
+            "label_offset_x": None,
+            "label_offset_y": None,
+            "label_width": None,
+        }
+        try:
+            self.block.clean(self.get_value(raw_data))
+        except blocks.StructBlockValidationError as e:
+            self.fail(f"ValidationError raised: {e.block_errors}")
+
+        config = self.get_config(raw_data)
+        self.assertEqual("Jam", config["text"])
+        self.assertFalse(config["labelInside"])
+        self.assertEqual(None, config["labelOffsetX"])
+        self.assertEqual(None, config["labelOffsetY"])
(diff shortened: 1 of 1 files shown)


cms/datavis/blocks/annotations.py
```diff
@@ -1,47 +1,73 @@
...
 
-    def get_x_position(self) -> float:
-        raise NotImplementedError("Subclasses must implement get_x_position()")
+    def get_x_position(self, x_position: int | float) -> int | float:
+        match self.X_AXIS_TYPE:
+            case AxisType.CATEGORICAL:
+                # The Design System component expects 0-based indices. We have a 1-based UI.
+                return int(x_position - 1)
+            case AxisType.LINEAR:
+                return x_position
```
MaciekBaron, line 40: **Suggestion**: I like how we raise a `RuntimeError` when the subclass does not set `X_AXIS_TYPE`. How about we add something like this as well? Unless it's physically impossible for there to be additional types.

```suggestion
                return x_position
            case _:
                raise NotImplementedError(
                    f"X position calculation not implemented for X_AXIS_TYPE: {self.X_AXIS_TYPE}"
                )
```

nimasmi: It's impossible, until a developer starts working on it. Highcharts has concepts of `datetime` and `logarithmic` axes, and indeed a different backend could have other types too.

I wanted these errors to be seen by the dev so they knew how to use the classes. In practice, I think that dev is only ever going to be me, but…

I have added this change in e5dc8555047e4c2024ecd5db5c35d2aae386cddb.

nimasmi:
Apologies, I just found my tests file, which had not been included in the commits.

cms/datavis/blocks/annotations.py
```diff
@@ -1,47 +1,73 @@
...
     x_position = blocks.StaticBlock()
     y_position = blocks.StaticBlock()
     label_offsets = blocks.StaticBlock(admin_text="Offsets are measured in pixels")
-    label_offset_x = TextInputIntegerBlock(label="Offset X", required=False)
-    label_offset_y = TextInputIntegerBlock(label="Offset Y", required=False)
+    label_offset_x = TextInputIntegerBlock(label="Offset X", default=0)
+    label_offset_y = TextInputIntegerBlock(label="Offset Y", default=0)
 
+    class Meta:
+        icon = "crosshairs"
```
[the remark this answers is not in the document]
zerolab: You could add your own icon. Perhaps something https://www.flaticon.com/free-icon/range_6348466?term=range&page=1&position=11&origin=tag&related_id=6348466 ?

MaciekBaron (approved):

PR #224, opened 2025-06-03
Support AWS SES (v2) for email sending

RealOrangeOne:

Add support for sending emails using AWS SES. This is addative configuration - console and SMTP are still fully supported.

### How to review

See the README of https://github.com/django-ses/django-ses for its installation instructions. For additional context:

- Authentication is handled implicitly (_magic_ :sparkles: ) rather than using `$AWS_ACCESS_KEY_ID` etc
- We'll be using SES's v2 API
- The region name will be set as `$AWS_REGION`. `AWS_SES_REGION_ENDPOINT` defaults to being based off `AWS_SES_REGION_NAME`.
- Secure defaults (eg TLS requirements through configuration sets) will be handled on the SES side, rather than application-side.

### Follow-up Actions

Deploying the SES infrastructure, as well as configuring the Wagtail deployment to use it are all out of scope.

Once this PR is merged, the existing environments will need `EMAIL_BACKEND_NAME=SMTP` set.


cms/settings/base.py
```diff
@@ -544,32 +541,43 @@
...
 # Email settings
-# We use SMTP to send emails. We typically use transactional email services
-# that let us use SMTP.
-# https://docs.djangoproject.com/en/2.1/topics/email/
+# https://docs.djangoproject.com/en/stable/topics/email/
+
+# Use fixed strings in case import paths move
+match env.get("EMAIL_BACKEND_NAME"):
+    case "SES":
+        EMAIL_BACKEND = "django_ses.SESBackend"
```
zerolab, line 549:
```suggestion
        # django-ses uses AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY
        # See https://github.com/django-ses/django-ses
        EMAIL_BACKEND = "django_ses.SESBackend"
```

cms/settings/base.py
```diff
@@ -544,32 +541,43 @@
...
+        AWS_SES_REGION_NAME = env["AWS_REGION"]
```
zerolab, line 550: question: doesn't `AWS_SES_REGION_ENDPOINT` need to be set too? https://github.com/django-ses/django-ses mentions:

```python
# Additionally, if you are not using the default AWS region of us-east-1,
# you need to specify a region, like so:
AWS_SES_REGION_NAME = 'us-west-2'
AWS_SES_REGION_ENDPOINT = 'email.us-west-2.amazonaws.com'
```

My reading of that is that we also need to set:
```python
AWS_SES_REGION_ENDPOINT = f"email.{AWS_SES_REGION_NAME}.amazonaws.com"
```

but happy to be corrected

cms/settings/base.py
RealOrangeOne: This isn't right thought - it doesn't use those env vars. See PR description.

cms/settings/base.py
RealOrangeOne: I think it's poor documentation. The default _is_ that value:

https://github.com/django-ses/django-ses/blob/f80a89094771f15a527c61cba802e5ad61979fe2/django_ses/settings.py#L21-L22

cms/settings/base.py
```diff
@@ -544,32 +541,43 @@
 
 
 # Email settings
-# We use SMTP to send emails. We typically use transactional email services
-# that let us use SMTP.
-# https://docs.djangoproject.com/en/2.1/topics/email/
+# https://docs.djangoproject.com/en/stable/topics/email/
+
+# Use fixed strings in case import paths move
+match env.get("EMAIL_BACKEND_NAME"):
```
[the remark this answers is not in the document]
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/224/commits/2a195f060e303ae4f0d555f0877e831efd967b44

zerolab (approved):

cms/settings/base.py
```diff
@@ -544,32 +541,43 @@
...
 # Email settings
-# We use SMTP to send emails. We typically use transactional email services
-# that let us use SMTP.
-# https://docs.djangoproject.com/en/2.1/topics/email/
+# https://docs.djangoproject.com/en/stable/topics/email/
+
+# Use fixed strings in case import paths move
+match env.get("EMAIL_BACKEND_NAME", "").upper():
+    case "SES":
+        EMAIL_BACKEND = "django_ses.SESBackend"
```
kacperpONS, line 549:
```suggestion
        # See https://github.com/django-ses/django-ses
        EMAIL_BACKEND = "django_ses.SESBackend"
```
I think Dan suggested this in https://github.com/ONSdigital/dis-wagtail/pull/224#discussion_r2123930331 but it got lost. I think it's worth keeping it there to avoid make it clear what we're using and allow quick reference to the env vars required by this library.

nehakerung (approved):
Read the readme and code looks good to me.

MebinAbraham (approved):

kacperpONS (approved):

sanjeevz3009 (commented):
Reviewed!

PR #225, opened 2025-06-03
Include language in no translation warning message

MaciekBaron:

This PR addresses CMS-512 and includes the name of the language in the message which informs the user about the non-existence of a localised version of a page.

**Preview**
image

### How to review
Make sure you compile messages with `./manage.py compilemessages`.

1. Create an English page
2. Try to view it in Welsh
3. Confirm that the top black bar contains a message in Welsh rather than in English. It should read: _**"Nid yw'r dudalen hon ar gael yn y Gymraeg ar hyn o bryd. Fe'i harddangosir yn ei hiaith wreiddiol"**_.

### Follow-up Actions

- Consider using `pgettext` for the name of the language if better ([link to docs](https://docs.python.org/3/library/gettext.html#gettext.pgettext)).
- Review the Welsh translation and update it if necessary (I used _Google Translate_ for this).


cms/jinja2/templates/base_page.html
```diff
@@ -43,7 +43,17 @@
 {%- endblock -%}
 
 {% block pageContent %}
-    {%- set announcement_text = _("This page is currently not available in English. It is displayed in its original language.") -%}
+    {# Set as variables so that makemessages picks it up #}
+    {%- set english_label = _("in English") -%}
```
MaciekBaron, line 47: **Comment**: Here we run into a grammatical problem with Welsh.

"Welsh" in Welsh is Cymraeg, but when we say "in Welsh" the noun uses soft mutation and becomes "Gymraeg" and the resulting phrase reads "yn y Gymraeg". Therefore we can't just translate the name of the language. This is why I use the "in Language" form here.

As we are using both parameters and particular translations, I thought this approach would be the best, but I could also look into using `pgettext` for this.

cms/jinja2/templates/base_page.html
```diff
@@ -43,7 +43,17 @@
 {%- endblock -%}
 
 {% block pageContent %}
-    {%- set announcement_text = _("This page is currently not available in English. It is displayed in its original language.") -%}
+    {# Set as variables so that makemessages picks it up #}
```
zerolab, line 46:
```suggestion
    {#- Set as variables so that makemessages picks it up -#}
```

cms/jinja2/templates/base_page.html
```diff
@@ -43,7 +43,17 @@
...
+    {%- set english_label = _("in English") -%}
+    {%- set welsh_label = _("in Welsh") -%}
+    {%- set language_labels = {
+    "en": english_label,
+    "cy": welsh_label
+    } -%}
+
+    {%- set announcement_text -%}
+        {% trans trimmed in_language=language_labels[request.LANGUAGE_CODE] %}This page is currently not available {{in_language}}. It is displayed in its original language.{% endtrans %}
+    {%- endset -%}
```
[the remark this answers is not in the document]
zerolab: I suppose https://github.com/ONSdigital/dis-wagtail/pull/225/files#r2123812833 explains why.

sanjeevz3009 (approved):
I have tested everything locally: `View in Welsh -> Welsh is displaying rather than English`. LGTM! All is working fine!

MaciekBaron:
Hi @nehakerung you need to compile the messages first so that they get picked up.

nehakerung (approved):
LGTM - the welsh text is displayed

cms/jinja2/templates/base_page.html
zerolab: @MaciekBaron can you add a comment to that effect in the template? Our future selves will be grateful.

MaciekBaron: Added in [699c618](https://github.com/ONSdigital/dis-wagtail/pull/225/commits/699c6186bf4940e1d5fa7988cc1dce739568898c)

PR #222, opened 2025-05-30
Add docs for the integrations

zerolab:
### What is the context of this PR?

Adds documentation for our integrations

### How to review

Verify the information is accurate and has the right level of information. To view the rendered file, click the `...` in top right corner of the file box in the "files changed" tab and select "view file". Or:

- [docs index](https://github.com/ONSdigital/dis-wagtail/blob/d8d2638a88ecbe0a6d0ae355be1900e0059b3def/docs/index.md)
- [integrations/auth.md](https://github.com/ONSdigital/dis-wagtail/blob/d8d2638a88ecbe0a6d0ae355be1900e0059b3def/docs/integrations/auth.md)
- [integrations/dataset_api.md](https://github.com/ONSdigital/dis-wagtail/blob/d8d2638a88ecbe0a6d0ae355be1900e0059b3def/docs/integrations/dataset_api.md)
- [integrations/logging.md](https://github.com/ONSdigital/dis-wagtail/blob/d8d2638a88ecbe0a6d0ae355be1900e0059b3def/docs/integrations/logging.md)
- [integrations/notifications.md](https://github.com/ONSdigital/dis-wagtail/blob/d8d2638a88ecbe0a6d0ae355be1900e0059b3def/docs/integrations/notifications.md)
- [integrations/search_service.md](https://github.com/ONSdigital/dis-wagtail/blob/d8d2638a88ecbe0a6d0ae355be1900e0059b3def/docs/integrations/search_service.md)

### Follow-up Actions

List any follow-up actions (if applicable), like needed documentation updates or additional testing.

PR #228, opened 2025-06-04
Authenticate to Kafaka using IAM

RealOrangeOne:

This PR makes 2 notable changes:

- Opt-in support for authenticating to Kafka using IAM credentials
- Support multiple Kafka URLs. These aren't for unique connections, but to connect to Kafka in a high-availability style (there are multiple connection URLs in production).

### How to review

This PR should not break existing functionality.

### Follow-up Actions

Configuring the IAM policy is out of scope of this PR.


cms/settings/base.py
```diff
@@ -932,7 +937,8 @@
 
 # Configuration for the External Search service
 SEARCH_INDEX_PUBLISHER_BACKEND = os.getenv("SEARCH_INDEX_PUBLISHER_BACKEND")
-KAFKA_SERVER = os.getenv("KAFKA_SERVER")
+KAFKA_SERVERS = os.getenv("KAFKA_SERVERS", "").split(",")
+KAFKA_USE_IAM_AUTH = os.getenv("KAFKA_USE_IAM_AUTH", "false").lower() == "true"
 KAFKA_CHANNEL_CREATED_OR_UPDATED = os.getenv("KAFKA_CHANNEL_CREATED_OR_UPDATED")
 KAFKA_CHANNEL_DELETED = os.getenv("KAFKA_CHANNEL_DELETED")
 KAFKA_API_VERSION = tuple(map(int, os.getenv("KAFKA_API_VERSION", "3,5,1").split(",")))
```
zerolab, line 944: unrelated, but was something bugging me when working on #222: @sanjeevz3009 any reason this is not using the dot notation as one would expect from a version (i.e. `3.5.1`)

RealOrangeOne: Also, looks like the client is capable of inferring this if it's not provided. Perhaps we just shouldn't set it at all? To avoid it getting out of sync with the environments (Especially since this implies we're pinning Kafka to patch versions).

zerolab: Yes, let's do that. One less moving part to worry about

https://github.com/dpkp/kafka-python/blob/9f7384c2bf32f07e4ae1039fd87024fdd966c124/kafka/client_async.py#L261-L262 is where that check happens

RealOrangeOne: @zerolab can this be extracted out as a future ticket so this discussion can be resolved?

sanjeevz3009: @zerolab we can use the dot notation `KAFKA_API_VERSION = tuple(map(int,os.getenv("KAFKA_API_VERSION", "3.5.1").split(".")))`

Due to the Kafka version being used by the OpenSauce @MebinAbraham and I decided to explicitly set the Kafka version to `3.5.1`.

And Jake is correct, but again, the intention was to make this explicit:

```
api_version (tuple): Specify which Kafka API version to use. If set to
            None, the client will attempt to determine the broker version via
            ApiVersionsRequest API or, for brokers earlier than 0.10, probing
            various known APIs. Dynamic version checking is performed eagerly
            during __init__ and can raise NoBrokersAvailableError if no connection
            was made before timeout (see api_version_auto_timeout_ms below).
            Different versions enable different functionality.
```

sanjeevz3009:
> LGTM.
> 
> I have a non-blocking question about the version setting since we're in this context, more for @sanjeevz3009 (but do pitch in @RealOrangeOne if you have additional context)

@zerolab Replied!

PR #229, opened 2025-06-04
Remove unused code from the charts preview

nimasmi:

The current charts feature retained some code from the original charts preview, which was implemented differently. The code was kept around, as some of it was gradually reused as we implemented new features.

This removes anything remaining from that preview.


nimasmi:
> @nimasmi nice clean up!
> 
> I noticed that `cms.datavis.fields.NonStrippingCharField` is not used anywhere and would be good to remove while you're at it.
> 
> Other than that, LGTM

I intend to use NonStrippingCharField still for [CCB-73 Data label and axis title affixes (prefix/suffix)](https://jira.ons.gov.uk/browse/CCB-73), so editors can choose whether to add a space between the affix and the number.

Do you prefer to remove it now (after all, we'll know where to find it), or leave it in?

PR #232, opened 2025-06-06
Reference line annotation

nimasmi:

Ticket: https://jira.ons.gov.uk/browse/CCB-55

### How to review

On a chart block, under "annotations" select "reference line".

<p>

image

image

</p>
</details>  

> [!NOTE]
> There is a bug in the DS at the moment: x-axis (or "category axis" for bar/column charts) reference lines are only displayed if there is a y-axis (or "value axis") reference line. This is fixed in https://github.com/ONSdigital/design-system/pull/3619. For now, a workaround is to create a y-axis reference line but at a value that is not shown on the chart.

### Follow-up Actions

Update the DS version to fix the above bug, once an update is released.

### DIFF
--- a/cms/datavis/blocks/annotations.py
+++ b/cms/datavis/blocks/annotations.py
@@ -1,4 +1,4 @@
-from typing import Any, ClassVar
+from typing import Any, ClassVar, cast
 
 from django.core.exceptions import ValidationError
 from wagtail import blocks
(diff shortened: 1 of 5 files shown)


cms/datavis/blocks/annotations.py
```diff
@@ -62,7 +66,7 @@ class BasePointAnnotationBlock(blocks.StructBlock):
     label_offset_y = TextInputIntegerBlock(label="Offset Y", default=0)
 
     class Meta:
-        icon = "crosshairs"
+        icon = "location-crosshairs"
```
nimasmi, line 69: **Comment**: This is for consistency between the three annotation types, now they're all together.

Screenshot from 2025-06-06 12-02-39

PR #238, opened 2025-06-09
Fix S3 Storage Tests Failing Locally

sanjeevz3009:

**Issue**

Our S3 Storage tests in `cms/private_media/tests/test_storages.py` currently fail if we run our tests locally, and when we have the AWS config set locally.

By default, boto3 will read `~/.aws/credentials` and `~/.aws/config` (and even EC2 metadata) when you instantiate an S3 client, so our tests were accidentally talking to `real` AWS settings.

**Fix**

- Inject dummy AWS credentials via Django settings.
- This ensures boto3:
    - Sees the `testing` credentials first (env-var precedence).
    - Never falls back to `~/.aws/*` or `EC2 metadata`.

### How to review

`Users/testuser/.aws/credentials`:

```
[default]
aws_access_key_id = testing
aws_secret_access_key = testing
aws_session_token = testing
```

`Users/testuser/.aws/config`

```
[default]
region = us-east-1
```

### Follow-up Actions


cms/private_media/tests/test_storages.py
```diff
@@ -9,8 +9,15 @@
 from cms.private_media.storages import AccessControlledS3Storage
 
 
+@override_settings(
+    AWS_STORAGE_BUCKET_NAME="test-bucket",
```
RealOrangeOne, line 13: Thought: Should these be set in `cms/settings/test.py`, so it's caught for all tests rather than just these?

I'm also confused why this happens - `mock_aws` should be dealing with this completely.

It also shouldn't be necessary to set all these values. Just setting the credentials (access key, secret key, session token) ought to be enough.

RealOrangeOne: My guess is that the reason it's happening is because `boto3` is being instantiated before `mock_aws` is being called, so it's not properly stubbing things out - but even that is just a guess.

MaciekBaron (approved):
LGTM - noting the conversation with Jake about the settings.

cms/private_media/tests/test_storages.py
sanjeevz3009: Okay, once Auth is all complete, I'll come back to this to dig in a bit more to see what the actual underlying issue is, as currently I only know `It should be working without this workaround, but it's not`.

sanjeevz3009: @RealOrangeOne Digged into this, will outline what the core issue is (Jake was on point here) and possible solutions, one solution being what I have already done on this PR.

sanjeevz3009: Did some further digging and found this doc https://docs.getmoto.org/en/latest/docs/getting_started.html#recommended-usage, https://docs.getmoto.org/en/latest/docs/getting_started.html#recommended-usage.

The recommended approach is to have dummy environment variables set up.

Moto does not block boto3 from reading our real AWS config or credentials. Moto only mocks the AWS service endpoints and responses, if our boto3 loads credentials, region and or config from your envrionment or files, it will use those unless we override them.

**Solutions:**
- Set dummy environment variables for credentials and region in your test setup.
- Or we can patch the boto3 Session per the docs

@MebinAbraham @RealOrangeOne Let me know what we want to do moving forwards.

MebinAbraham (approved):

cms/private_media/tests/test_storages.py
RealOrangeOne: Just moving these to `test.py` sounds like it's probably fine (with a comment, of course).

PR #243, opened 2025-06-10
Allow resources endpoint in internal env

MebinAbraham:

We need to enable the `/v1/resources` endpoint in the internal instance to ensure that the search re-indexing can function properly. The pipelines and services that require access to `/v1/resources` are located within the internal network. For now, I have kept the endpoint in the external instance to match existing behaviour in the legacy CMS. Further discussions will determine whether we need to expose this endpoint to the public in the future. This will be decided later.

### How to review

Run the app and ensure that the `v1/resources` endpoint is available regardless of whether `IS_EXTERNAL_ENV` is true or false.

### Follow-up Actions

### DIFF
--- a/docs/integrations/search_service.md
+++ b/docs/integrations/search_service.md
@@ -9,7 +9,7 @@ and 'search-content-deleted' (deleted) events, aligning with the `StandardPayloa
 - [Search metadata contract](https://github.com/ONSdigital/dis-search-upstream-stub/blob/main/docs/contract/resource_metadata.yml)
 - [Search upstream service endpoints spec](https://github.com/ONSdigital/dis-search-upstream-stub/blob/main/specification.yml)
 
-The CMS also provides a paginated Resource API endpoint with all published pages at `/v1/resources/` in the public facing instance. This is
+The CMS also provides a paginated Resource API endpoint with all published pages at `/v1/resources/`. This is
 used by the search service for reindexing.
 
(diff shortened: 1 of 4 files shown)


kacperpONS (approved):
LGTM

docs/integrations/search_service.md
```diff
@@ -9,7 +9,7 @@ and 'search-content-deleted' (deleted) events, aligning with the `StandardPayloa
 - [Search metadata contract](https://github.com/ONSdigital/dis-search-upstream-stub/blob/main/docs/contract/resource_metadata.yml)
 - [Search upstream service endpoints spec](https://github.com/ONSdigital/dis-search-upstream-stub/blob/main/specification.yml)
 
-The CMS also provides a paginated Resource API endpoint with all published pages at `/v1/resources/` in the public facing instance. This is
+The CMS also provides a paginated Resource API endpoint with all published pages at `/v1/resources/`. This is
 used by the search service for reindexing.
```
RealOrangeOne, line 13: Question: If it's only used by Search, should it be enabled in the external env? Feels like something which only makes sense on internal?

zerolab (approved):
My one thought was to fold the check on the external env with the one above, but this is more explicit.

Happy for this to go in

sanjeevz3009 (commented):
@MebinAbraham LGTM!

PR #246, opened 2025-06-11
Compile translations as part of container build

RealOrangeOne:

Compile the locale files as part of the container build, so they can be used properly by the application.

### How to review

Check the container builds, and that the locale files are available.

I made the assumption that `--use-fuzzy` was correct - I'm not entirely sure, but given there's some mention of fuzzy in the `.po` file, I assume we want it.

### DIFF
--- a/cms/core/custom_date_format.py
+++ b/cms/core/custom_date_format.py
@@ -20,4 +20,11 @@ class ONSDateFormat(DateFormat):
 
 
 def ons_date_format(value: datetime, format_string: str) -> str:
-    return ONSDateFormat(value).format(get_format(format_string))
+    date_format = ONSDateFormat(value).format(get_format(format_string))
+
+    # Note: Django currently has a typo in the Welsh translation for "July" (Gorffenaf instead of Gorffennaf).
+    # This is a temporary fix to correct the typo in the date format.
+    # TODO: Once the Django translation is fixed, this can be removed.
+    # Reference:
+    # https://github.com/django/django/blob/c1fa3fdd040718356e5a3b9a0fe699d73f47a940/django/conf/locale/cy/LC_MESSAGES/django.po#L926
+    return date_format.replace("Gorffenaf", "Gorffennaf")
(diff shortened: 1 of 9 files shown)


MaciekBaron (approved):
Thanks Jake!

zerolab (changes requested):
Let's drop fuzzy. In all the workflows I've seen or used, it was never with `--use-fuzzy`

https://www.gnu.org/software/gettext/manual/html_node/Fuzzy-Entries.html also calls out that these usually need a revision from the translator

.github/workflows/ci.yml
```diff
@@ -150,6 +150,8 @@ jobs:
         run: make load-design-system-templates
       - name: collectstatic
         run: poetry run ./manage.py collectstatic --verbosity 0 --noinput --clear
+      - name: compilemessages
+        run: poetry run ./manage.py compilemessages --verbosity 0 --use-fuzzy
```
MebinAbraham, line 154: I don't recommend using strings marked as fuzzy in production. These are entries where the translator wasn’t fully confident. Ours exists because @MaciekBaron added a bunch for development purposes. It's not a big deal if the translations aren’t perfect right now, since a native Welsh speaker will review and finalise them before launch.

MebinAbraham: For now, worth dropping `fuzzy` from PO file so have as much in Welsh as possible. They will all be re-reviewed anyway, so even if the translfation is 100% correct, it's fine for now.

RealOrangeOne: Fixed in https://github.com/ONSdigital/dis-wagtail/pull/246/commits/ccb7874f64177380d0136eb7aac848b247ef0446

zerolab: I think that may need removing the related msgids too. 

https://stackoverflow.com/questions/2710733/djangos-makemessages-creates-a-lot-of-fuzzy-entries has the tool to use.

The diff should be like:

```diff
- #, fuzzy
- #~| msgid "Release date"
- #~ msgid "Releases:"
- #~ msgstr "Dyddiad y datganiad"
```

zerolab (approved):

MebinAbraham (approved):

sanjeevz3009 (commented):
Reviewed!

PR #251, opened 2025-06-13
Update show-data-labels code

nimasmi:

This updates the logic for show-data-labels to ensure it is set / not set in the right cases. Ticket: https://jira.ons.gov.uk/browse/CCB-39.

It also refactors `ChartBlock.get_series_data` to permit easier overriding of individual series (singular) processing without unnecessary iterations. Unit tests are included.

### How to review

This is more functional than user-facing.

### DIFF
--- a/cms/datavis/blocks/base.py
+++ b/cms/datavis/blocks/base.py
@@ -44,7 +44,7 @@ class BaseVisualisationBlock(blocks.StructBlock):
     )
     show_legend = blocks.BooleanBlock(default=True, required=False)
     show_data_labels = blocks.StaticBlock()
-    use_stacked_layout = blocks.BooleanBlock(default=False, required=False)
+    use_stacked_layout = blocks.StaticBlock()
     show_markers = blocks.BooleanBlock(
         default=False,
         required=False,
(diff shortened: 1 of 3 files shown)


cms/datavis/blocks/base.py
```diff
@@ -258,29 +258,33 @@ def get_series_data(
...
-            item = {
-                "name": column,
-                "data": data_points,
-                "animation": False,
+    def get_series_item(
+        self, value: "StructValue", series_number: int, series_name: str, rows: list[list[str | int | float]]
+    ) -> dict[str, Any]:
+        """Get the configuration for a single series."""
+        # Extract data points, handling None/empty values
+        data_points = [r[series_number] if r[series_number] != "" else None for r in rows]
```
MaciekBaron, line 270: **Non-blocking suggestion**: Since `""` is falsy we can probably do this
```suggestion
        data_points = [r[series_number] or None for r in rows]
````

MaciekBaron (approved):
LGTM! Just the usual small suggestion from me if you find it useful 👍

cms/datavis/blocks/base.py
nimasmi: This is addressed in 321bd89.

PR #253, opened 2025-06-17
RelatedLinksBlock and LinkBlock block improvements

MaciekBaron:
### What is the context of this PR?

This PR related to CMS-337 and adds improvements to `RelatedLinksBlock` and `LInkBlock`.

It mostly covers refactoring, but it also adds a type and optional release date.

**Loom video 1**
https://www.loom.com/share/f3f0e9b857a040c8b5e81cce6a21d687

**Loom video 2**
https://www.loom.com/share/ce556e99faa347619f43262602a0acc2

### How to review
<details>
<summary>Information page</summary>

- In the admin panel, go to an information page
- Add a related links content block
- Add some related links
- Confirm that the links display correctly, and that the "Related links" headline displays as well

</details>
<details>
<summary>Statistical article page / methodology page</summary>

- In the admin panel, go to a statistical article page or a methodology page
- Add a new section block to the page, and set its title to "Related links"
- Within that section, add a related links content block
- Add some related links
- Confirm that the links display correctly, and that the "Related links" headline displays once only

</details>
<details>
<summary>Index page</summary>

- In the admin panel, go to an index page
- Ensure that it has children
- Confirm that they get displayed correctly
- Remove children and add pages manually
- Confirm that they also get displayed correctly

</details>

### Follow-up Actions

- Update the list of types
- Revisit rendering of "related link" on Index pages, which currently use `onsRelatedContent` instead of `onsDocumentList`

PR #260, opened 2025-06-19
Integrate corrections & notices with DS component

janehughes:

Ticket CMS-539

### How to review

- Review statistical article and see that the corrections & notices block is displaying as expected
- Check the stat article template and see that we have no custom code remaining and it has been updated to use the DS component

### Follow-up Actions

### DIFF
--- a/docs/custom-features/README.md
+++ b/docs/custom-features/README.md
@@ -6,3 +6,5 @@ The following are project-specific features:
 - [Scheduled jobs](scheduled-jobs.md)
 - [Media privacy](media_privacy.md)
 - [Migration-friendly StreamFields](migration_friendly_streamfields.md)
+- [Deployment environments](deployment-environments.md)
+- [DB read replica](db-read-replica.md)
--- a/docs/custom-features/db-read-replica.md
+++ b/docs/custom-features/db-read-replica.md
@@ -1,0 +1,10 @@
+# Database Read-Replica
+
+Each deployed environment uses a primary-replica model for the database. Write queries are sent to the primary instance, whilst read queries are balanced between a number of replicas. Replica instances are read-only - any write queries sent to them will return an error. For this reason, database migrations are only run against the primary instance.
+
+To ensure a consistent view of the database, the write instance is always used if there is an open transaction, even for reads.
+
+Generally, a developer shouldn't need to worry about this configuration, or explicitly handle it - it is handled automatically by a database router (`cms.core.db_router.ReadReplicaRouter`).
+
+> [!WARNING]
+> For performance-critical areas, some of Django's built-in queries should be avoided: `get_or_create`, `update_or_create`. These methods result in queries which always use the primary database instance. Instead, it may be better to use multiple explicit queries (eg `.get` then `.create`) so that the optimal connection is used.
--- a/docs/custom-features/deployment-environments.md
+++ b/docs/custom-features/deployment-environments.md
@@ -1,0 +1,25 @@
+# Deployment Environments
+
+In the interest of security, the application is deployed to 2 separate environments: The "internal" and "external" environments.
+
+The `IS_EXTERNAL_ENV` environment variable (and setting) is used to switch the app to the "external" environment.
+
+## "Internal" environment
+
+The internal is private and only accessible to ONS staff.
+
+The Wagtail admin is only accessible on the "internal" environment.
+
+Management commands, database migrations, and other [scheduled jobs](./scheduled-jobs.md) are only executed in the internal environment.
+
+## External
+
+The "external" environment is the one used by the public.
+
+Writeable database access is restricted to a limited set of tables. This is enforced both at the database level and by using a custom database router (see `cms.core.db_router.ExternalEnvRouter`).
+
+The URLs and apps required for the Wagtail admin are disabled. This includes the middleware required for authentication.
+
+Only a subset of media stored in S3 are accessible to the external env. Notably the `images/*` prefix needed for Wagtail's rendition generation.
+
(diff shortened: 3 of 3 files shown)


cms/articles/models.py
```diff
@@ -426,3 +436,35 @@ def serve_preview(self, request: "HttpRequest", mode_name: str) -> "TemplateResp
...
+
+    def get_serialized_corrections_and_notices(
+        self, request: "HttpRequest"
+    ) -> tuple[list[dict[str, Any]], list[dict[str, Any]]]:
+        """Returns a list of corrections and notices for the page."""
+        corrections = (
+            [
+                serialize_correction_or_notice(
+                    correction,
+                    superseded_url=self.get_url(request)
```
zerolab, line 448: suggestion: move `self.get_url(request)` outside of the loop. No need recalculate the same thing multiple times

cms/articles/utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+from django.utils.formats import date_format
+from django.utils.translation import gettext_lazy as _
+
+if TYPE_CHECKING:
+    from wagtail.blocks.stream_block import StreamChild
+
+
+def serialize_correction_or_notice(entry: "StreamChild", *, superseded_url: str | None = None) -> dict:
+    """Serialize a correction or notice entry for the API.
```
zerolab, line 11:
```suggestion
    """Serialize a correction or notice entry for the Design System.
```

cms/jinja2/templates/pages/statistical_article_page.html
```diff
@@ -132,63 +133,31 @@
         })
     }}
     {# fmt:on #}
+    {% if has_corrections or has_notices %}
+        {% set openText = _("Show detail") %}
+        {% set closeText = _("Close detail") %}
```
zerolab, line 138: question: given the DS component defaults are "Show detail" / "Hide detail", can we not use those?

(aside from the fact the DS strings are not marked for translation https://github.com/ONSdigital/design-system/blob/530f3ec8277554b6a214a4e7c9decf2d95e1a6ad/src/components/details-panel/_macro.njk#L2-L3)

Note that "Hide detail" is the correct string as per designs

cms/jinja2/templates/pages/statistical_article_page.html
```diff
@@ -132,63 +133,31 @@
...
+        {% set correctionHeading = _("Corrections") %}
+        {% set noticesHeading = _("Corrections") %}
```
zerolab, line 140:
```suggestion
```

as they are unused, and furthermore `noticesHeading` says "Corrections" too

functional_tests/steps/information_page.py
```diff
@@ -58,7 +58,6 @@ def check_information_page_content(context: Context, default_language=True) -> N
         expect(context.page.get_by_text("1 Ionawr 2024")).to_be_visible()
 
     expect(page.get_by_role("heading", name="Some example rich text content")).to_be_visible()
-    expect(page.get_by_text("n∑i=0i2=(n2+n)(2n+1)")).to_be_visible()
```
zerolab, line 61: question: why remove this?

cms/articles/utils.py
```diff
@@ -0,0 +1,34 @@
...
+            if this is a correction. If provided, the item will be labeled as a correction.
+
+    Returns:
+        dict: A dictionary containing the serialized data.
+    """
+    content = {
+        "text": _("Correction") if superseded_url else _("Notice"),
+        "date": {
+            "iso": date_format(entry.value["when"], "c"),
+            "short": date_format(entry.value["when"], "DATE_FORMAT"),
```
zerolab, line 25: this is not quite right. Corrections need to have the time too (Date, time)

Screenshot 2025-06-30 at 11 43 39

zerolab (changes requested):

cms/articles/utils.py
MaciekBaron: Fixed in [164a0ba](https://github.com/ONSdigital/dis-wagtail/pull/260/commits/164a0bada18e0ebddb1dd4f70cd8106db3e3a560)

cms/jinja2/templates/pages/statistical_article_page.html
zerolab: My worry is that the DS has lots of strings that are not marked for translation and we will have an inconsistent UI.

Would be good to understand the DS plan for i18n/l10n
cc @MebinAbraham @AdamHawtin

functional_tests/steps/statistical_article_page.py
```diff
@@ -244,18 +244,18 @@ def user_adds_a_notice(context: Context):
...
+    expect(context.page.get_by_text("13 March 2025")).to_be_hidden()
     expect(context.page.get_by_text("Correction text")).to_be_hidden()
 
 
 @then("the user can expand and collapse {block_type} details")
 def user_can_click_on_view_detail_to_expand_block(context: Context, block_type: str):
     if block_type == "correction":
         text = "Correction text"
-        date = "13 March 2025 1:59pm"
+        date = "13 March 2025"
```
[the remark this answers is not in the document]
kacperpONS: Ah, the `ons_date_format`! 
Completely unrelated with this PR, but could you also replace it in the `get_document_metadata_date` function? 
We should be ok to sneak it in as it was missed only yesterday in https://github.com/ONSdigital/dis-wagtail/pull/253 and since it influences what the external user sees it'd be good to keep it consistent.

kacperpONS (approved):
Noice!

functional_tests/steps/statistical_article_page.py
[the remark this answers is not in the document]
MaciekBaron: Addressed in [9c85bb0](https://github.com/ONSdigital/dis-wagtail/pull/260/commits/9c85bb07fd71337c0db72fa3b106aca0388324e9)

PR #262, opened 2025-06-20
Bundle and page aliases tidy up

zerolab:

This PR:
- exclude aliases from the bundle release calendar page chooser (as they cannot be published)
- fixes a JS error with the RC toggle date text logic (noticed on non-RC pages)
- allows unscheduling bundles if they are still in "ready to publish" but for some reason have not published
- allow approving a dataset-only bundle
- removes some pre-Wagtail 7.0 code
- links to live pages from the published bundle inspect view
- prevents copying a page as an alias, or copying and publishing
- prevents editing the release calendar page once it is link to a bundle that is ready to be published

Screencast: https://www.loom.com/share/73d45edff165457280d0724d7cb21552

### How to review

- dataset-only bundles
   - create a bundle and add only one or more datasets. You should be able to mark the bundle as "in review"
- bundle linked with release calendar page, ready to be published, but the cron job doesn't run and the RC date passes. You can create a RC page with the date several minutes in the future, then create a dataset-only bundle linked to it.
  -  once the RC date passes, you should be able to "unschedule" the bundle, removing the RC page
- editing an RC page that linked with bundle that is ready to be published
   - RC page cannot be edited until unlinked
- aliases - the RC page chooser on the bundle should filter out "alias only" pages. For example, when we create a new RC page in English, we get an alias in the Welsh tree.
- copy page - there should be no option to copy as alias, or copy and publish

### Follow-up Actions

### DIFF
--- a/cms/bundles/viewsets/bundle.py
+++ b/cms/bundles/viewsets/bundle.py
@@ -167,7 +167,11 @@ class BundleInspectView(InspectView):
         return user_can_manage_bundles(self.request.user)
 
     def get_fields(self) -> list[str]:
-        """Returns the list of fields to include in the inspect view."""
+        """Returns the list of fields to include in the inspect view.
+
+        Note: values are inserted by methods following the get_FIELDNAME_display_value pattern.
+        See InspectView.get_field_display_value.
+        """
         if self.can_manage:
             return [
                 "name",


cms/bundles/forms.py
```diff
@@ -41,6 +48,18 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        has_datasets = False
+        for form in self.formsets["bundled_datasets"].forms:
+            if not form.is_valid() or form.cleaned_data["DELETE"]:
+                continue
+
+            if form.clean().get("dataset"):
+                has_datasets = True
+                break
+
+        return has_datasets
```
MaciekBaron, line 61: **Suggestion**: How about

```suggestion
        return any(
            form.clean().get("dataset")
            for form in self.formsets["bundled_datasets"].forms
            if form.is_valid() and not form.cleaned_data["DELETE"]
        )
```
A bit more concise.

zerolab: To be clear, what I am saying is "I appreciate the suggestion. Will keep the more verbose version"

kacperpONS (commented):
A thought before I go deep with reviewing, for Bundles we allow to set the status to `Published` manually in the dropdown, but we have recently disabled an analogous action for Release Calendar pages. 
Additionally, this creates two separate user journeys to achieve the the publication of a Bundle (this, and clicking the blueish action button at the bottom of the page), which could be confusing. 
Is there a justification for that that I'm not seeing? If not it would be good to disable this in a follow up card since this one is already quite rich in content.

zerolab:
> A thought before I go deep with reviewing, for Bundles we allow to set the status to `Published` manually in the dropdown, but we have recently disabled an analogous action for Release Calendar pages. Additionally, this creates two separate user journeys to achieve the the publication of a Bundle (this, and clicking the blueish action button at the bottom of the page), which could be confusing. Is there a justification for that that I'm not seeing? If not it would be good to disable this in a follow up card since this one is already quite rich in content.

I hear you. Will wait for further feedback from the DP team. Definitely out of scope for this though

cms/bundles/viewsets/bundle.py
```diff
@@ -234,29 +224,44 @@ def get_scheduled_publication_display_value(self) -> str:
...
+
             return format_html(
                 '<a href="{}" target="_blank" rel="noopener">{}</a>',
-                reverse("bundles:preview_release_calendar", args=[self.object.pk]),
+                url,
                 self.object.release_calendar_page.get_admin_display_title(),
             )
         return "N/A"
 
     def get_pages_for_manager(self) -> "SafeString":
```
kacperpONS, line 239: It would be nice to see a docstring here with some context to decrease the cognitive load.

cms/bundles/forms.py
```diff
@@ -91,7 +110,7 @@ def _validate_bundled_pages_status(self) -> None:
                     form.add_error("page", "This page is not ready to be published")
                     num_pages_not_ready += 1
 
-        if not has_pages:
+        if not has_pages and not self._has_datasets():
             raise ValidationError("Cannot approve the bundle without any pages")
```
kacperpONS, line 114: `...or datasets` - in the error message

cms/bundles/viewsets/bundle.py
```diff
@@ -206,17 +207,6 @@ def get_field_label(self, field_name: str, field: "Field") -> str:
             case _:
                 return super().get_field_label(field_name, field)  # type: ignore[no-any-return]
 
-    def get_field_display_value(self, field_name: str, field: "Field") -> Any:
```
kacperpONS, line 209: Since this has been removed, it would be good to explain the magical name resolution that's happening in the background thanks to Wagtail's [get_field_display_value](https://github.com/wagtail/wagtail/blob/a783b8c3e713b119c4f72d037d3b9489b80854e3/wagtail/admin/views/generic/models.py#L1105), perhaps even include a link. 
I think the appropriate place for it could be the `get_fields` function, with the docstring something along the lines of:
```
        """
        Returns the list of fields to include in the inspect view.
        Values are inserted by methods following the get_FIELDNAME_display_value pattern.
        """
```

kacperpONS (changes requested):

cms/bundles/forms.py
zerolab: 6eaf9d7

kacperpONS (approved):
LGMT!

cms/bundles/viewsets/bundle.py
```diff
@@ -234,29 +228,47 @@ def get_scheduled_publication_display_value(self) -> str:
     def get_release_calendar_page_display_value(self) -> str:
         """Returns the release calendar page link if it exists."""
         if self.object.release_calendar_page:
+            if self.object.status == BundleStatus.PUBLISHED:
```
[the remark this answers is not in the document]
zerolab: I think I prefer this to stay as is. The bundle is published so the RC should be published too.

sanjeevz3009:
A couple of things I noticed:

- We don't have to provide `scheduling/ date` via an RC page or a `publication date` for a bundle, and we can still get it to `preview`, `ready to publish` and manually `publish` it. I guess this was an intended behaviour from before. I am not sure.

- If we select an RC page for scheduling the bundle and it's in the status `ready to publish`, and then if I try to remove the RC page/ clear choice and save, then it all saves and succeeds as if the operation is valid. But in reality, when you go back to the bundle, the RC page is still selected and still there. This seems to me to be a misleading behaviour. To get around this, I have to `unschedule` or put it into `preview` or `draft`, `save` and then remove the RC page and `save again` to actually remove it.

- Another behaviour I noticed was that if I select an RC page to schedule a bundle, it's `ready to publish`, the cron job hasn't run, now the date has passed and RC release date is in the past, now if I try to remove/ clear choice in the bundle, it won't let me. If I go back to the RC page, now it's locked. The only way around this is for me to put the bundle back to `preview` or `draft`, `save` and then `remove` the RC page from the bundle, `save`, and now the RC page will be unlocked for editing. Or I would have to `unschedule` where it would put the bundle status back to `draft`, `save` and then `remove` the RC page. Again, odd behaviour, due to the fact that I have the option to remove/ clear the choice of the RC page throughout this journey, and it succeeds every time I clear the choice/ remove the RC page and save.

- I can also select an RC page to schedule a bundle while the status is `ready for published`, and when I do again, everything succeeds, but when I go back to the bundle, the RC page has not been selected. I have to put the status in `preview` or `draft`, `save` and then select the RC page, `save` for it to actually save.

  - The above are very unusual behaviours and will definitely cause users to be confused about the whole workflow. We would need better/ improved validations.

- Another issue can be chucked into the backlog as an improvement: There is an inconsistent chooser experience between the RC page and the bundle dataset choosers, where one is a multi-chooser and the other one is not. A ticket to chuck in the backlog for improvements/ consistency would be nice.


This problem might be worth covering in part of this PR:
- There seems to be an issue with bundles not actually publishing the RC page when `manually published` via the bundle `published` option. 
- To reproduce this. `Create a RC page (provisional or confirmed, it doesn't matter)`  ->  `set the date in the future` -> `create a new bundle` -> `select the RC page created to schedule bundle` -> `add some other pages to bundled pages` -> `go through the flow and eventually set the status to published in the bundle` -> `save` -> `bundle should be published` -> `you will notice the other pages added in the bundled pages are all published and live and not in draft` -> `but the RC page used of scheduling of the bundle is not published and still in draft`.
  - Now, if you go to the `inspect bundle` -> `click on the RC page` -> `You will get a page not found error` -> `as it's now trying to load a page that doesn't exist/ live as the RC page is still in draft`.
  - This wouldn't occur if the cron job runs/ if we run the management commands locally to publish, to simulate how it would get published automatically.
- If I now delete this broken bundle and then select the RC page, which failed to go live/ which is still in draft as it's not published yet, it won't show in the chooser for a new bundle.

@nehakerung suggested the code change above because she came across the same issue, I guess, when she manually published an RC page that's used for scheduling a bundle and when she clicked on the RC page in the inspect bundle page, it errored out to page not found. She didn't quite understand the technical details of why/ assumed the error was because of something else. Told her to resolve that comment, but that's the reason she suggested the code change. Nevertheless, the code change she suggested will be redundant if the issue of the RC page not getting published gets fixed.

sanjeevz3009:
@zerolab There are a couple of other things I also found (final bits promise =)), which I needed some time to reproduce/ replicate. Here they are.

I don't think this is intended behaviour:
- `Create an RC page` -> `set the release date in the future` -> `save draft` -> `create a bundle` -> `select RC page to schedule the bundle` -> `now add some datasets` -> `cycle the status all the way to ready to publish` -> `save` -> `now add some pages that are in draft and not in a workflow and not approved` -> `save` -> `now we are able to get a pages that are not in a workflow/ ready to publish, actually get published`. We can bypass the whole page workflow requirement this way, no approval, and the page doesn't need to be in `ready to publish status`.

This may be a question:
- When I create an RC page and have datasets added to it on the RC page editor -> use that RC page for scheduling a bundle -> then when I also add datasets in the bundle editor itself, the datasets I selected on the RC page editor gets overridden by the datasets selected in the bundle editor when it gets published? **Is this intended behaviour?**

Another issue/ observation:
- `Create an RC page` -> `put in confirmed status` -> `set the release date in the future (maybe a month later)` -> `save draft` -> `create a bundle` -> `select RC page to schedule the bundle` ->  `add some datasets` -> `now cycle the status to all the way to ready to publish` -> `save` -> `now manually publish the bundle` -> `go and create a new bundle and now we can select a live RC page (RC page in confirmed status) for this new bundle` ->  `due to the RC page release date is in the near future` or and `the RC page not updating to the published status when a manually triggered to publish` as it's still in the confirmed status when published manually without the management command? Again not sure if this is intended.

Final issue/ question:
- `If we have RC page and it's been published/ live (confirmed status)` -> `but now I go edit it and set a new release date` -> `save draft` -> `now this will show when trying to schedule a bundle with a RC page (due to now being a draft version of the RC page)` -> the issue is now the RC page doesn't show the new release date set in the draft version it shows the published/ live version's release date/ original release date version (RC page in confirmed status)

PR #267, opened 2025-06-24
Add endpoint for running management commands

RealOrangeOne:

https://jira.ons.gov.uk/browse/DIS-3186

As an alternative approach to cron job scheduling, we want to try triggering tasks using a tiny `curl` container in the cluster, which in turn calls out to Wagtail, rather than running a dedicated Wagtail container. This should avoid any potential resource contention issues which would delay the container starting up, but also avoids the container startup time (~3s) since it's already running (with warm caches).

The endpoint is only enabled in the internal environment (it'll 404 on external), and requires basic auth. Similarly, the commands do not take any arguments, reducing the risk of any kind of network privilege escalation.

This PR should be additive - using an external scheduler should continue to work. This just enables an additional method.

##### Why is `django-extensions` moved to a dev dependency? 

`django-extensions` has a lot of management commands, many of which can be dangereous to run. Dropping it down to a dev dependency removes them from the main container and makes it smaller. We're not using the features for the main app itself, so it's a dependency we can easily demote.

### Follow-up Actions

If this idea is successful, there is likely value in extracting this out into a separate package. However, that can come in future.

Configuring the cluster to use these endpoints is intentionally out of scope.

Should this experiment fail, and we drop the approach, the PR will be reverted.

### DIFF
--- a/cms/core/views.py
+++ b/cms/core/views.py
@@ -5,8 +5,10 @@ from typing import TYPE_CHECKING
 
 import redis
 import redis.exceptions
+from baipw.utils import authorize as basic_auth_authorize
 from django.conf import settings
 from django.core.cache import caches
+from django.core.management import CommandError, call_command, get_commands
 from django.db import DatabaseError, connections
 from django.http import HttpResponse, HttpResponseServerError, JsonResponse
 from django.shortcuts import render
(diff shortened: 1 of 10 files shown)


cms/core/views.py
```diff
@@ -162,3 +164,23 @@ def build_check(name: str, message: str, failed: bool) -> dict:
...
+    try:
+        get_commands()[command_name]
+    except KeyError:
+        return HttpResponse("Unknown command", status=404)
+
+    try:
+        call_command(command_name, skip_checks=True, no_color=True)
+    except CommandError:
+        logger.exception("Error executing management command")
+        return HttpResponse("Command failed", status=400)
```
MaciekBaron, line 184: **Question**:

- Do we want to log unknown commands using the logger?
- Do we want to include the attempted command in the logs when there's an error?

cms/settings/base.py
```diff
@@ -950,6 +949,8 @@
 TAG = env.get("TAG") or None
 START_TIME = datetime.datetime.now(tz=datetime.UTC)
 
+MANAGEMENT_COMMANDS_URL_SECRET = env.get("MANAGEMENT_COMMANDS_URL_SECRET", "")
```
kacperpONS, line 952: This would be more clear if named `..._PASSWORD` so that it better represents it's role (and match the parameter name in `baipw`) 

Otherwise (or additionally), I think it deserves a comment in the settings file.

RealOrangeOne:
Closing in favor of #272

PR #259, opened 2025-06-19
DAC Report - Update Official Statistics Badge alt text (Level A)

janehughes:
### What is the context of this PR?

Ticket: https://jira.ons.gov.uk/browse/CMS-541

### How to review

- Check the release calendar page, inspect the Official Statistics logo and see that the alt text has been updated and follows the DAC guidance

### Follow-up Actions

n/a

PR #270, opened 2025-06-26
Add url to heros with statistics badge

janehughes:

No ticket for this - small update off the back of Mebin's comment in a previous PR https://github.com/ONSdigital/dis-wagtail/pull/259#discussion_r2157345014

Adds the same URL used on the live site to all heros with the statistics badge.

### How to review

- Check all page types which include the stats badge in the hero and see that the badge links to the correct page, and it opens in a new window

### Follow-up Actions


MaciekBaron (approved):
LGTM - I also added an integration test to cover this

kacperpONS (changes requested):
Functionally works ok 👍 How about defining this URL as a variable in the base html template to avoid duplication?

MaciekBaron:
@kacperpONS I think let's discuss this separately as the variable would only be used in the release calendar but would have to be defined in a shared template.

MaciekBaron:
@kacperpONS I've moved this now to a variable [3fd24d8](https://github.com/ONSdigital/dis-wagtail/pull/270/commits/3fd24d83d878b3c106ebdf6a58248842ac9737a0)

sanjeevz3009 (commented):
Reviewed!

PR #272, opened 2025-06-26
Support early spin up for `publish_bundles`.

RealOrangeOne:

When starting the management command for publishing bundles, most of the time is spent in provisioning the underlying infrastructure, and in the process starting up.

This PR adds the ability to "hold" before publishing if there are some upcoming. For example, if a bundle is due for publish at `07:00`, and `cron` starts the publish process at `06:59` with a `--include-future` of 60s, the process will start up, notice the `07:00` bundle, and sleep for ~55s. This means as soon as the clock hits 7am, everything is already running and started ready for publish.

If the process dies whilst sleeping, the restarted process picks up where it left off. 

This also means we can more safely reduce the scheduling frequency to 2 minutes.

### How to review

This change is additive - If `--include-future` isn't used, there's no change in functionality.

I've intentionally used `sched` rather than `apscheduler` so that the latter can be removed once `scheduler.py` is removed, reducing the number of libraries.

### Follow-up Actions

Infrastructure changes are handled externally.


cms/bundles/management/commands/publish_bundles.py
```diff
@@ -100,32 +111,54 @@ def handle_bundle(self, bundle: Bundle) -> None:
...
-            for bundle in bundles_to_publish:
+            # Explicitly use `time.time` so enterabs can be called with absolute timestamps.
+            bundle_scheduler = sched.scheduler(timefunc=time.time)
+
+            def wrap_handle_bundle(bundle: Bundle) -> None:
                 try:
+                    # Refresh the bundle immediately before publishing, in case it's changed.
+                    bundle.refresh_from_db()
+
+                    # Use `contains` rather than `in` to force a new query
```
MaciekBaron, line 152: I'm confused by this comment as we don't use `contains` anywhere here?

docs/custom-features/scheduled-jobs.md
```diff
@@ -11,9 +11,16 @@ every minute, and [`publish_scheduled_without_bundles`](https://github.com/ONSdi
 
 ## `publish_bundles`
 
-Is a management command that publishes [bundles](bundles.md) that are scheduled and ready to publish.
+Is a management command that publishes [bundles](bundles.md) that are scheduled and ready to publish. Bundles are considered for publishing if their release date is in the past.
+
+### `--include-future`
+
+To reduce the impact of the process and infrastructure warm-up time, `publish_bundles` starts earlier than necessary and includes bundles which need to be published in the future. The command then waits until each bundle's release time, and then publishes it. Bundles due to be published in the past are still published immediately without wait.
```
MaciekBaron, line 18: Non-blocking nit-pick: I understand you likely mean `wait()` here but I think this just might read better.
```suggestion
To reduce the impact of the process and infrastructure warm-up time, `publish_bundles` starts earlier than necessary and includes bundles which need to be published in the future. The command then waits until each bundle's release time, and then publishes it. Bundles due to be published in the past are still published immediately without waiting.
```

MaciekBaron (approved):
LGTM - just one tiny non blocking suggestion regarding wording

AdamHawtin (approved):
LGTM 👍 

I was interested to test what would happen with a long lookahead time and making changes to bundles that are picked up during that time, and in my testing it all appears to be handled gracefully and as expected, e.g. if the bundle is reverted from ready to publish then the script will correctly log that fact and not publish it.

AdamHawtin (approved):

PR #273, opened 2025-06-27
Bundles UI tests

AWitcherONS:
Bundles User Interface Journey from creation to Approval

**Testing Strategy**
User/Auth
Bundle:- volume, and contents 
The feature tests should be completed in a timely manner avoiding duplication where possible 

 role, function (Search(volume is significant), Create, Edit/insert Edit/remove, send to moderation, preview, approve)
 for each of these function what text can be seen,  what fields to edit(insert/delete), what actions (save etc) 
 what happens when mandatory fields fail validations on actions

    User role Bundle Life Cycle Happy Path 1 bundle Authorisation summary
                                 | Bundles Search | Dashboard Search | Create | Edit | Preview | Approve |
        Publishing Admin         | Can            |                  | Can    | Can  | Can     | Can     !
            As Creator of Bundle | N/A            | N/A              | N/A    | N/A  | N/A     | Cannot  |
            Not Creator          | N/A            | N/A              | N/A    | N/A  | N/A     | Can     |
        Publishing Officer       | Can            | N/A              | Can    | Can  | Can     | N/A     |
            As Creator of Bundle | N/A            | N/A              | N/A    | N/A  | N/A     | Cannot  |
            Not Creator          | N/A            | N/A              | N/A    | N/A  | N/A     | Can     |
        Viewer                   | N/A            | N/A              | Cannot | N/A  | N/A     | N/A     |
            not in preview team  | Cannot         | N/A              | N/A    | N/A  | Cannot  | N/A     |
            in preview team      | Can            | N/A              | N/A    | N/A  | Can     | N/A     |
/

### How to review
The Python code should comply with current standards
The feature tests run without failures and do not impact on run time

### Follow-up Actions
A Viewer cannot use the Search in Bundles menu there isn't an error but no know bundles are not found 
review on approval Authorisation of bundles


functional_tests/features/bundle_ui.feature
```diff
@@ -0,0 +1,267 @@
...
+        Given there is a Publishing Admin user
+        And there are 3 Preview teams
+        And the user is a member of the Preview teams
+        And there are 3 release calendar pages
+        And there are 3 Statistical Analysis pages
+        And there are 3 bundles
+        When the Publishing Admin logs in
+        And the user can see the Bundles menu item
+        And the user clicks the Bundles menu item
+        Then the user can search for unknown bundle with response "There are no bundles to"
```
MaciekBaron, line 14: Should this also say "There are no bundles to display" like on line 39? If it can say the same thing, then we don't need to parameterise our step as well.

functional_tests/steps/bundles.py
```diff
@@ -121,3 +126,166 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+    search_bundle = context.bundles[0]
+    context.page.get_by_role("link", name="Add bundle").click()
+    context.page.get_by_role("textbox", name="Name*").fill(search_bundle.name)
+    context.page.get_by_role("button", name="Create").click()
+
+
+@step('the user can search for unknown bundle with response "{response}"')
+def search_for_new_bundle(context: Context, response: str) -> None:
+    unknown_bundle = "Unknown bundle"
+    if unknown_bundle not in get_bundles_names(context.bundles):
```
MaciekBaron, line 218: I don't think we should have an `if` statement like this here. This step will pass if the "Unknown bundle" is amongst the bundles in the context, without really doing what it is supposed to (i.e. check if the user can search and can see the adequate response).

functional_tests/steps/bundles.py
```diff
@@ -121,3 +126,166 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
 def bundle_inspect_show(context: Context) -> None:
     expect(context.page.get_by_text("Created by")).to_be_visible()
     expect(context.page.get_by_text(context.bundle_creator.get_username())).not_to_be_visible()
+
+
+# Bundles UI Data Setup
+@given("there are {no_bundles} bundles")
+def multiple_bundles_create(context: Context, no_bundles: str) -> None:
+    context.bundles = []
+    if no_bundles.isdigit():
```
MaciekBaron, line 135: The step will fail silently if `no_bundles` is not a digit.

I think it's worth failing if it's not:
```python
if not no_bundles.isdigit():
    raise ValueError(f"Invalid parameter: '{no_bundles}' must be a digit.")
```

Same problem on line 143, 155 and 170.

functional_tests/steps/bundles.py
```diff
@@ -121,3 +126,166 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+
+
+@given("there are {no_preview_teams} Preview teams")
+def multiple_preview_teams_create(context: Context, no_preview_teams: str) -> None:
+    context.teams = []
+    if no_preview_teams.isdigit():
+        for preview_team_index in range(int(no_preview_teams)):
+            context.teams.append(
+                Team.objects.create(
+                    identifier="preview-team-" + str(preview_team_index), name="Preview_Team_" + str(preview_team_index)
```
MaciekBaron, line 147: f-strings will be better here for readability:
```suggestion
                    identifier=f"preview-team-{preview_team_index}", name=f"Preview_Team_{preview_team_index}"
```

We could also perhaps drop the underscores?

```suggestion
                    identifier=f"preview-team-{preview_team_index}", name=f"Preview Team {preview_team_index}"
```

functional_tests/steps/bundles.py
```diff
@@ -121,3 +126,166 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+@given("there are {no_release_calendar} release calendar pages")
+def multiple_release_calendar(context: Context, no_release_calendar: str) -> None:
+    context.release_calendar_pages = []
+    if no_release_calendar.isdigit():
+        for release_calendar_index in range(int(no_release_calendar)):
+            nowish = timezone.now() + timedelta(minutes=5 * (release_calendar_index + 1))
+            context.release_calendar_pages.append(
+                ReleaseCalendarPageFactory(
+                    release_date=nowish,
+                    title="Release Calendar Page" + str(release_calendar_index),
```
MaciekBaron, line 161: Similar to above, f-string a better (and I think there's a missing space):

```suggestion
                    title=f"Release Calendar Page {release_calendar_index}",
```

Same for line 174.

functional_tests/steps/bundles.py
```diff
@@ -121,3 +126,166 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+                )
+            )
+
+
+@given("there is a {user_role} user")
+def creat_user_by_role(context: Context, user_role: str) -> None:
+    context.user_data = create_user(user_role)
+
+
+@given("the {user_role} is a member of the Preview teams")
```
MaciekBaron, line 184: You've defined a parameter here `{user_role}` but it doesn't seem to be used by the function itself, so I think it can be dropped. If we keep it as "the user is a member of the Preview teams", it should be clear from the context that it's the user we created.

functional_tests/steps/bundles.py
```diff
@@ -121,3 +126,166 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+def add_user_to_preview_teams(context: Context) -> None:
+    for team in context.teams:
+        user = context.user_data["user"]
+        user.teams.add(team)
```
MaciekBaron, line 188: This can be simplified:
```suggestion
    user = context.user_data["user"]
    user.teams.add(*context.teams)
```
`.add()` allows us to provide multiple objects for optimisation.

https://docs.djangoproject.com/en/5.2/ref/models/relations/#django.db.models.fields.related.RelatedManager.add

functional_tests/steps/bundles.py
```diff
@@ -121,3 +126,166 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+
+    # modify status
+    context.page.locator("#id_status").select_option("IN_REVIEW")
+
+    # add preview team
+    context.page.get_by_role("button", name="Add preview team").click()
+    context.page.get_by_role("textbox", name="Search term").click()
+    context.page.get_by_role("textbox", name="Search term").fill(context.teams[1])
+    context.page.get_by_text(context.teams[1]).click()
+    context.page.get_by_role("button", name="Confirm selection").click()
```
MaciekBaron, line 262: I think we should follow this with an assertion to confirm that we have added the preview team successfully.

MaciekBaron (changes requested):

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,267 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+@given("the {user_role} is a member of the Preview teams")
+def add_user_to_preview_teams(context: Context, user_role: str) -> None:
+    next((item[user_role]["user"] for item in context.users if item[user_role]), None).teams.add(*context.teams)
+
+
+@given("there are {number_of_bundles} bundles with {bundle_details}")
+def multiple_bundles_create(context: Context, number_of_bundles: str, bundle_details: str) -> None:
+    bundle_dets = {}
+    if json_str_to_dict(bundle_details):
+        bundle_dets = json.loads(bundle_details)
```
MaciekBaron, line 191: This looks a bit strange to me, the function you call does `json.loads` and then you catch the exception which returns false. You then attempt json loads again if false is not returned.

I don't think this approach works well. From my point of view this function would never receive invalid json, unless there's a mistake.

functional_tests/features/bundle_ui.feature
```diff
@@ -0,0 +1,130 @@
...
+        And there are <number_of_bundles> bundles with <Bundle_Details>
+        When the <Role> logs in
+        Then the user can preview a bundle
+
+      Examples: bundles
+           | number_of_bundles | Role                       | Creator Role        | Bundle_Details                                                                                                                                              |
+           | 1                 | Publishing Admin           | Publishing Admin    | {"Role": "Publishing Admin",   "Creator Role": "Publishing Admin",   "status": "In_Review", "preview_teams": true,  "add_rel_cal": true, "add_stat_page": true} |
+           | 1                 | Publishing Officer         | Publishing Admin    | {"Role": "Publishing Officer", "Creator Role": "Publishing Admin",   "status": "In_Review", "preview_teams": true,  "add_rel_cal": true, "add_stat_page": true} |
+#           | 1                 | Viewer                     | Publishing Admin    | {"Role": "Publishing Admin",   "Creator Role": "Publishing Admin",   "status": "In_Review", "preview_teams": true,  "add_rel_cal": true, "add_stat_page": true} |
```
MaciekBaron, line 75: You've marked the PR as ready for review but there's still commented out code here.

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,267 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+        context.users = []
+
+    context.users.append({user_role: create_user(user_role)})
+
+
+@given("there are {no_statistical_analysis} Statistical Analysis pages")
+def multiple_statistical_analysis(context: Context, no_statistical_analysis: str) -> None:
+    context.statistical_article_pages = []
+    context.article_series_page = ArticleSeriesPageFactory(title="PSF")
+    if no_statistical_analysis.isdigit():
```
MaciekBaron, line 148: Do we need to check if `no_statistical_analysis` is a digit? It seems to me that it should always be a digit.

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,267 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+
+
+@step("the user can preview a bundle")
+def can_preview_bundle(context: Context) -> None:
+    expect(context.page.locator("#bundles-for-preview-heading")).to_contain_text("Bundles ready for preview")
+    expect(context.page.locator("#bundles-for-preview-content")).to_contain_text(context.bundles[0].name)
+    context.page.get_by_title("View").click()
+
+    expect(context.page.get_by_text("Name")).to_be_visible()
+    # expect(context.page.get_by_role("term").filter(has_text=re.compile(r"^Status$"))).to_be_visible()
```
MaciekBaron, line 359: More commented out code here

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,267 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+    expect(context.page.get_by_role("button", name=f"More options for '{context.bundles[0].name}'")).not_to_be_visible()
+
+
+@step("the user can approve a bundle")
+def can_approve_bundle(context: Context) -> None:
+    expect(context.page.locator("#latest-bundles-heading")).to_contain_text("Latest active bundles")
+    expect(context.page.locator("#latest-bundles-content")).to_contain_text(context.bundles[0].name)
+    context.page.get_by_title("Edit this bundle").click()
+    context.page.get_by_role("button", name="More actions").click()
+    context.page.locator("#panel-status-content").click()
```
MaciekBaron, line 388: I've already sent a comment to you about this. It is strange to perform actions at the end of a function which is designed to make assertions. Please revisit this.

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,267 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+        context.page.get_by_role("checkbox", name=context.teams[0].name).check()
+        context.page.wait_for_timeout(100)
+        expect(context.page.get_by_role("button", name="Confirm selection")).to_be_visible()
+        context.page.get_by_role("button", name="Confirm selection").click()
+
+
+def add_article_page_in_edit(context: Context) -> None:
+    # add Article
+    if hasattr(context, "statistical_article_pages"):
+        expect(context.page.get_by_role("heading", name="Bundled pages").locator("span")).to_be_visible()
```
MaciekBaron, line 320: The `expect` calls here are redundant, you are performing an action, not making a check

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,267 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+    if hasattr(context, "teams"):
+        expect(context.page.locator("#panel-preview_teams-heading")).to_contain_text("Preview teams")
+        context.page.get_by_role("button", name="Add preview team").click()
+        expect(context.page.get_by_role("textbox", name="Search term")).to_be_visible()
+        context.page.get_by_role("textbox", name="Search term").click()
+        context.page.get_by_role("textbox", name="Search term").fill(context.teams[0].name)
+        expect(context.page.get_by_role("checkbox", name=context.teams[0].name)).to_be_visible()
+        context.page.get_by_role("checkbox", name=context.teams[0].name).check()
+        context.page.wait_for_timeout(100)
+        expect(context.page.get_by_role("button", name="Confirm selection")).to_be_visible()
```
MaciekBaron, line 313: You don't need to call `expect` before clicking a button

MaciekBaron (changes requested):

MaciekBaron (changes requested):

functional_tests/features/bundle_ui.feature
MaciekBaron: Sorry, where's the reason posted?

MaciekBaron: This information should really be in the PR description so that reviewers know what's up. If you could add it there, and then add a TODO, that would be grand, thanks.

functional_tests/steps/bundles.py
MaciekBaron: Could you show me in what scenario that would not be an integer?

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,272 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+            BundlePage.objects.create(parent=bundle, page=page)
+
+
+def add_teams(context: Context) -> None:
+    for team in context.teams:
+        for bundle in context.bundles:
+            BundleTeam.objects.create(parent=bundle, team=team)
+
+
+def json_str_to_dict(bundle_details):
```
nehakerung, line 242: Would it be better to have the helper functions in their own step_helpers file

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,272 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+    expect(context.page.locator("#latest-bundles-content")).to_contain_text("Add bundle")
+    expect(context.page.get_by_text("Latest active bundles")).to_be_visible()
+    expect(context.page.get_by_role("link", name=bundle_name)).to_be_visible()
+
+
+@then("the user cannot create a bundle")
+def the_user_cannot_add_bundles(context: Context) -> None:
+    expect(context.page.get_by_role("link", name="Add bundle")).not_to_be_visible()
+    expect(context.page.get_by_text("Bundles ready for preview")).to_be_visible()
+    expect(context.page.get_by_text("There are currently no")).to_be_visible()
```
nehakerung, line 280: Could check for the full error message text

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,272 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+        context.page.get_by_role("button", name="Approve").click()
+    else:
+        expect(context.page.locator("#latest-bundles-heading")).to_contain_text("Latest active bundles")
+        expect(context.page.locator("#latest-bundles-content")).to_contain_text("no documents")
+        context.page.get_by_role("row", name="no documents Actions 31").get_by_role("link").click()
+        context.page.get_by_role("button", name="More actions").click()
+        context.page.get_by_role("button", name="Approve").click()
+
+
+def create_user_by_role(context, user_role):
```
nehakerung, line 396: Missing type hints(!)

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,272 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+@step("the user cannot preview a bundle")
+def cannot_preview_bundle(context: Context) -> None:
+    context.page.get_by_role("link", name="Bundles", exact=True).click()
+    expect(context.page.get_by_role("button", name=f"More options for '{context.bundles[0].name}'")).not_to_be_visible()
+
+
+@step("the user cannot approve a bundle")
+def cannot_approve_bundle(context: Context) -> None:
+    context.page.get_by_role("link", name="Bundles", exact=True).click()
+    expect(context.page.get_by_role("button", name=f"More options for '{context.bundles[0].name}'")).not_to_be_visible()
```
nehakerung, line 377: These functions are the same-  if thats correct you can add the steps like so
```
@step("the user cannot preview a bundle")
@step("the user cannot approve a bundle")
def ...
```

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,272 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+        ).to_be_visible()
+        context.page.get_by_role(
+            "checkbox", name=f"{context.article_series_page.title}: {context.statistical_article_pages[0].title}"
+        ).check()
+        context.page.wait_for_timeout(100)
+        expect(context.page.get_by_role("button", name="Confirm selection")).to_be_visible()
+        context.page.get_by_role("button", name="Confirm selection").click()
+
+
+def add_reslease_calendar_in_edit(context: Context) -> None:
```
MaciekBaron, line 340: Typo
```suggestion
def add_release_calendar_in_edit(context: Context) -> None:
```

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,272 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+                context.bundles.append(bundle)
+
+
+def add_article_pages(context: Context) -> None:
+    for page in context.statistical_article_pages:
+        for bundle in context.bundles:
+            BundlePage.objects.create(parent=bundle, page=page)
+
+
+def add_release_calendar(context: Context):
```
MaciekBaron, line 230: Missing return type
```suggestion
def add_release_calendar(context: Context) -> None:
```

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,272 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+
+
+@step("the user can approve a bundle")
+def can_approve_bundle(context: Context) -> None:
+    if len(context.bundles) == 1:
+        expect(context.page.locator("#latest-bundles-heading")).to_contain_text("Latest active bundles")
+        expect(context.page.locator("#latest-bundles-content")).to_contain_text(context.bundles[0].name)
+        context.page.get_by_title("Edit this bundle").click()
+        context.page.get_by_role("button", name="More actions").click()
+        context.page.get_by_role("button", name="Approve").click()
```
MaciekBaron, line 387: You can take these two lines out of the if/else block so that you don't repeat yourself

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,272 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+                if bool(bundle_dets["preview_teams"]) and hasattr(context, "teams"):
+                    add_teams(context)
+
+                if bool(bundle_dets["add_rel_cal"]) and hasattr(context, "release_calendar_pages"):
+                    add_release_calendar(context)
+
+                if bool(bundle_dets["add_stat_page"]) and hasattr(context, "statistical_article_pages"):
+                    add_article_pages(context)
+
+                context.bundles.append(bundle)
```
MaciekBaron, line 221: I've been reading this code and trying to understand it.

I see on line 194 you enter a for loop for the range `number_of_bundles`.

Within the for loop, you create the bundle on line 206, and append it to `context.bundles` on line 221.

However the things which happen around that don't seem to belong to the for loop:

- Lines 195-196 you create a creator role if one doesn't exist, this should be outside of the loop
- Lines 212-213 you call `add_teams` which creates a `BundleTeam` for each team and bundle, but since you are running this in a loop you recall this function on every iteration, recreating `BundleTeam` entries for existing bundles and teams
- Ditto on lines 215-216, you create unnecessary `BundlePage`s
- Ditto on lines 218-219

It seems more logical to me to take this code out of the loop so it only runs once for all bundles and pages, since the functions you call (`add_teams` etc.) already loop through your objects.

functional_tests/steps/bundles.py
MaciekBaron: I would remove this function all together

MaciekBaron (changes requested):
Could you also flesh out the PR description and include that comment you made on Slack about some of the scenarios failing and still being work in progress (you marked them as TODO for now)

functional_tests/steps/bundles.py
MaciekBaron: Hi Ann, just to clarify what I mean here: your assertion functions shouldn't end with actions but rather assertions using `expect()` - you should define what you expect to see rather than try to click on something and see if it works. For instance, check if the "Edit this bundle" link is visible and active etc.. Clicking on it doesn't tell us much.

functional_tests/steps/bundles.py
MaciekBaron: I think you mean "unit tests" instead of "functional tests"? As these are functional tests.
These `expect` calls are not needed because your actions will fail if the element doesn't exist, so the calls are redundant. If you look at other code you'll see we don't follow this pattern.

MaciekBaron (changes requested):

MaciekBaron (changes requested):

functional_tests/features/bundle_ui.feature
```diff
@@ -0,0 +1,133 @@
+Feature: UI Bundle Happy Paths
+    """
+    User Role Bundle Life Cycle Happy Path
+                                 | Search | Create | Edit | Preview | Approve |
+        Publishing Admin         | ToDo   | Can    | Can  | Can     | N/A     |
+            As Creator of Bundle | N/A    | N/A    | N/A  | N/A     | Cannot  |
```
AdamHawtin, line 6: To clarify something, currently the user who created the bundle _can_ approve the bundle themselves, it is pages that require a different user approval.

AdamHawtin (commented):
A lot of the test cases in the feature file seem to be commented out still, what is happening with these?

MaciekBaron:
> A lot of the test cases in the feature file seem to be commented out still, what is happening with these?

@AdamHawtin here's some additional context: https://github.com/ONSdigital/dis-wagtail/pull/273#discussion_r2307495188

functional_tests/steps/bundles.py
AWitcherONS: There is not a scenario where this would not be an integer just instinct where input is so precarious the same with json.loads

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,268 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+    add_preview_team_in_edit(context)
+    context.page.get_by_role("button", name="Save").click()
+    add_release_calendar_in_edit(context)
+    context.page.get_by_role("button", name="Save").click()
+
+    context.page.get_by_role("button", name="More actions").click()
+    context.page.get_by_role("button", name="Save to preview").click()
+
+
+def add_preview_team_in_edit(context: Context) -> None:
```
MaciekBaron, line 290: I think it might be worth moving these helper functions to a separate file in `step_helpers` just to follow our existing convention.

functional_tests/features/bundle_ui.feature
```diff
@@ -0,0 +1,133 @@
...
+        And there are 1 Statistical Analysis pages
+        And there are 1 release calendar pages
+        And there is a preview team
+        And the <Role> is a member of the preview team
+        And there are <number_of_bundles> bundles with <Bundle_Details>
+        When the <Role> logs in
+        Then the user can see the Bundles ready for preview
+        Then the <Role> can preview a bundle
+
+      Examples: bundles
```
sanjeevz3009, line 73: am confused of what's going on here? And if this is needed?

As I been debugging line by line the behave steps, and using the playwight inspector to check for assertions etc, and so far, this is not making sense. Assertions seems to be confusing and some assertions are incomplete as it's checking for half of a message or text on a page etc.

functional_tests/features/bundle_ui.feature
```diff
@@ -0,0 +1,133 @@
...
+           | 1                 | Viewer                     | Publishing Admin    | {"Role": "Publishing Admin",   "Creator Role": "Publishing Admin",   "status": "In_Review", "preview_teams": true,  "add_rel_cal": true, "add_stat_page": true} |
+           | 1                 | Publishing Admin           | Publishing Officer  | {"Role": "Publishing Admin",   "Creator Role": "Publishing Officer", "status": "In_Review", "preview_teams": true,  "add_rel_cal": true, "add_stat_page": true} |
+           | 1                 | Publishing Officer         | Publishing Officer  | {"Role": "Publishing Officer", "Creator Role": "Publishing Officer", "status": "In_Review", "preview_teams": true,  "add_rel_cal": true, "add_stat_page": true} |
+           | 1                 | Viewer                     | Publishing Officer  | {"Role": "Viewer",             "Creator Role": "Publishing Officer", "status": "In_Review", "preview_teams": true,  "add_rel_cal": true, "add_stat_page": true} |
+
+
+      Scenario Outline: A User cannot preview a bundle
```
sanjeevz3009, line 83: Why is there a nested scenario in here?

sanjeevz3009, line 83: Also similar to some other scenarios above, it doesn't seem to be doing anything?

sanjeevz3009, line 83: I tried to run this specific scenario and it won't run because it's nested under another scenario. So this scenario is basically not running. Even if it does, it will not be doing anything.

https://cucumber.io/docs
https://behave.readthedocs.io/en/latest/

functional_tests/features/bundle_ui.feature
```diff
@@ -0,0 +1,133 @@
...
+          Then the user cannot preview a bundle
+
+         Examples: bundles
+             | number_of_bundles | Role                       | Creator Role        | Bundle_Details                                                                                                                                       |
+#             | 1                 | Viewer                     | Publishing Admin    | {"Role": "Viewer",             "Creator Role": "Publishing Admin", "status": "In_Review", "preview_teams": false,  "add_rel_cal": true, "add_stat_page": true}|
+#             | 1                 | Publishing Officer         | Publishing Admin    | {"Role": "Publishing Officer", "Creator Role": "Publishing Admin", "status": "In_Review", "preview_teams": false,  "add_rel_cal": true, "add_stat_page": true} |
+
+#----- Bundle Approve UI Tests -----
+
+      Scenario Outline: A user can approve a bundle
```
sanjeevz3009, line 100: This is not currently doing anything

functional_tests/features/bundle_ui.feature
```diff
@@ -0,0 +1,133 @@
...
+        When the <Role> logs in
+        Then the user can approve a bundle
+
+        Examples: bundles
+            | number_of_bundles | Role                       | Creator Role        | Bundle_Details                                                                                                                                                 |
+#            | 1                 | Publishing Admin           | Publishing Officer  | {"Role": "Publishing Admin",   "Creator Role": "Publishing Officer", "status": "In_Review", "preview_teams": true, "add_rel_cal": true, "add_stat_page": true} |
+#            | 1                 | Publishing Officer         | Publishing Admin    | {"Role": "Publishing Officer", "Creator Role": "Publishing Admin",   "status": "In_Review", "preview_teams": true, "add_rel_cal": true, "add_stat_page": true} |
+
+
+      Scenario Outline: A user cannot approve a bundle
```
sanjeevz3009, line 117: Not doing anything

sanjeevz3009 (changes requested):
@MebinAbraham @MaciekBaron @AdamHawtin @AWitcherONS 
Did an initial round of review. This PR is far from complete. This doesn't really align with the bundle UI card, requirements and what we want out of it, the ticket.

Considering the bundles app in itself is a very custom feature developed by TBX, we will need to write really good scenarios and make sure our design of the step files, scenarios and the whole test foundation for bundles UI tests are really well laid out.

We would need to design the step definitions and the scenarios really well so that it's maintainable, readable, robust and future proof. You can take look at other UI tests, particularly RC UI tests where, you can build it out like lego blocks, where it very precise in what it tests, we can expand, such as add any big features to the RC page or remove or any small changes from adding new validation error messages on the RC page and we can reflect that in our scenarios/ test cases with ease, without increasing the file size, introducing complexity and going on a unmaintable path. These tests shouldn't be written in a sequential brute-force way; they should be as DRY as possible (some tests, such as for a small page or functionality, it's a different case). Bundles should have less complexity than the RC page, as that RC page was more complex to design in terms of the features it had, and has been tested thoroughly.

Otherwise, we take a huge risk of:
- Failing to catch any crucial regressions
- Bugs that may arise
- Any regressions caused by other features/apps in the codebase as bundle links heavily to so many other pages, services, and apps
  - RC pages and other pages you can add to the bundle
  - Datasets
  - Notifications
  - It all goes through a workflow
  - A couple of other things
  
If anything regresses, changes happen intentionally or unintentionally, then it can have a cascading effect on the bundle's functionality and how it behaves.
  
This is one of our key Django apps we have which is tied to scheduling publications, releases, and 7 am releases, and is one of the biggest goals of our product right now.

I would suggest taking a look at the release calendar page UI tests to see how we designed the scenarios, step definitions, step helpers, etc.

On some pages, such as the release calendar page, statistical article, methodology pages, bundles app/ page, we cannot take risks by not writing good quality tests/ inadequate tests that are not designed well and future-proof.

At least, even if we want to test something basic, writing the tests sequentially/ brute force (which will not be maintainable in the long run). I would expect something basic/ simple:
The below should be user-driven (driven by user interactions, unless for specific cases you can use factories)
- A user visits the bundles page (PO, PA, Viewer)
- If it's a user with the correct permissions -> they can schedule a bundle with the RC page (create the RC page via factories, we already have one, so you can just use that/ instantitate wherever necessary as we aren't testing RC page features here, so we don't care about that) -> add some pages (statsistical article, info, methodologies), add datasets etc -> we can take the bundle through various statuses successfully -> publish it etc (there are different ways achieve this)
-> Check if the bundle can be seen by a viewer in a particular preview team, inspect the bundle, and view the pages in the preview
-> Check for some basic validations, such as pages in the bundle have to be ready to publish/ approved through page workflows -> if not, we shouldn't be able to schedule a bundle
-> In a nutshell, checking the bundles pages/ apps functionality from a user's perspective in a comprehensive manner

I would suggest understanding, reviewing the bundle's features, what we can do, what we can't do and writing some basic scenarios to start with (the requirement of the bundle's UI ticket is to make sure we test this thoroughly).

Also, the above examples are only to give an idea, but there are way more test cases/ scenarios you will need to design.

Then you will have to make sure to design the code so that the scenarios, step definitions, and step helpers are robust, maintainable and future-proof.

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,268 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+        expect(context.page.get_by_role("button", name="Confirm selection")).to_be_visible()
+        context.page.get_by_role("button", name="Confirm selection").click()
+
+
+def add_release_calendar_in_edit(context: Context) -> None:
+    # add Release Calendar
+    if hasattr(context, "search_release_calendar"):
+        expect(context.page.get_by_role("button", name="Choose Release Calendar page")).to_be_visible()
+        context.page.get_by_role("button", name="Choose Release Calendar page").click()
+        context.page.get_by_role("textbox", name="Search term").click()
```
MaciekBaron, line 334: We don't need to `click()` on elements to `fill()` them (here and elsewhere).

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,268 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+                    bundle_approved = True
+                if bundle_dets["status"] == "In_Review":
+                    bundle_status = BundleStatus.IN_REVIEW
+                bundle = BundleFactory(
+                    created_by=context.users[bundle_dets["Creator Role"]]["user"],
+                    status=bundle_status,
+                    approved=bundle_approved,
+                )
+
+                if bool(bundle_dets["preview_teams"]) and hasattr(context, "team"):
```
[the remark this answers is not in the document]
AWitcherONS: removed the json function
Thanks

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,268 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+
+
+@then("the user cannot create a bundle")
+def cannot_add_bundles(context: Context) -> None:
+    expect(context.page.get_by_role("link", name="Add bundle")).not_to_be_visible()
+    expect(context.page.get_by_text("Bundles ready for preview")).to_be_visible()
+    expect(context.page.get_by_text("There are currently no")).to_be_visible()
+
+
+@step("the user can find the bundle")
```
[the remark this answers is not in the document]
AWitcherONS: Done

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,268 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+
+
+def add_preview_team_in_edit(context: Context) -> None:
+    # add preview team
+    if hasattr(context, "team"):
+        expect(context.page.locator("#panel-preview_teams-heading")).to_contain_text("Preview teams")
+        context.page.get_by_role("button", name="Add preview team").click()
+        expect(context.page.get_by_role("textbox", name="Search term")).to_be_visible()
+        context.page.get_by_role("textbox", name="Search term").click()
+        context.page.get_by_role("textbox", name="Search term").fill(context.teams[0].name)
```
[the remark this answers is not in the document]
AWitcherONS: There is only 1 team now and users are added or not by the scenario lines 
```
        And there is a preview team
        And the <role> is a member of the preview team
```
hope that has answered your confusion it did mine

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,268 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+        context.page.get_by_role("row", name=context.release_calendar_pages[0].title).get_by_role("link").click()
+        context.page.get_by_role("link", name=context.release_calendar_pages[0].title).click()
+
+
+@step("the user can see the Bundles ready for preview")
+def can_access_preview_bundle(context: Context) -> None:
+    expect(context.page.locator("#bundles-for-preview-heading")).to_contain_text("Bundles ready for preview")
+    expect(context.page.locator("#bundles-for-preview-content")).to_contain_text(context.bundles[0].name)
+    context.page.get_by_title("View").click()
+    context.page.wait_for_timeout(200)
```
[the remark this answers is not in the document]
AWitcherONS: I think this was for the dashboard in which playwright was having difficulties with a bundle in latest and ready for review so the pseudo search has had to be removed

functional_tests/steps/bundles.py
```diff
@@ -121,3 +128,268 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+                ReleaseCalendarPageFactory(
+                    release_date=nowish,
+                    title=f"Release Calendar Page {release_calendar_index}",
+                    status=ReleaseStatus.CONFIRMED,
+                )
+            )
+
+
+@given("there is a preview team")
+def multiple_preview_teams_create(context: Context) -> None:
```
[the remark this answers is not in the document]
AWitcherONS: See Above

functional_tests/steps/bundles.py
```diff
@@ -121,3 +129,337 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+
+
+@then("the logged in user fails to create a bundle due to mandatory field")
+def fail_to_save_bundle_name(context: Context) -> None:
+    context.page.goto(context.base_url + "/admin/")
+    expect(context.page.get_by_role("link", name="Add bundle")).to_be_visible()
+    context.page.get_by_role("link", name="Add bundle").click()
+    context.page.get_by_role("button", name="Save as draft").click()
+    expect(context.page.get_by_role("status")).to_contain_text("The bundle could not be created due to errors.")
+    expect(context.page.locator("#panel-child-name-errors")).to_contain_text("This field is required.")
```
MaciekBaron: Ideally we would split this to two steps
* the logged in user saves a bundle
* the logged in user sees a validation error

This allows you to avoid repetition, and better structures your test with steps that do something and assertions that check something.

MaciekBaron (changes requested):

functional_tests/steps/bundles.py
```diff
@@ -121,3 +129,337 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+def a_user_exists_by_role(context: Context, user_role: str) -> None:
+    if not hasattr(context, "users"):
+        context.users = {}
+    context.users[user_role] = create_user(user_role)
+
+
+@given("there is a Statistical Analysis page")
+def statistical_analysis(
+    context: Context,
+) -> None:
```
MaciekBaron: This can be on one line.

functional_tests/steps/bundles.py
```diff
@@ -121,3 +129,337 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+            bundle.save()
+
+            if bool(bundle_dets["add_stat_page"]) and hasattr(context, "statistical_article_page"):
+                page = context.statistical_article_page
+                BundlePage.objects.create(parent=bundle, page=page)
+                context.bundlepages.append(BundlePageFactory(parent=bundle, page=page))
+            bundle.save()
+            context.bundles.append(bundle)
+
+        sorted(context.bundles, key=lambda x: x.name)
```
MaciekBaron: The `sorted()` function _returns_ a sorted list.

https://docs.python.org/3/library/functions.html#sorted

functional_tests/steps/bundles.py
```diff
@@ -121,3 +129,337 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+
+            bundle_status = BundleStatus.DRAFT
+            bundle_approved = False
+            if bundle_dets["status"] == "Approved":
+                bundle_status = BundleStatus.APPROVED
+                bundle_approved = True
+            elif bundle_dets["status"] == "In_Review":
+                bundle_status = BundleStatus.IN_REVIEW
+
+            if bundle_dets["creator_role"] != "":
```
MaciekBaron: Can `bundle_dets["creator_role"]` ever be an empty string?

functional_tests/steps/bundles.py
```diff
@@ -121,3 +129,337 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+                bundle.created_by = user
+                bundle.status = bundle_status
+                bundle.approved = bundle_approved
+
+            else:
+                bundle.status = bundle_status
+                bundle.approved = bundle_approved
+            bundle.save()
+
+            if bool(bundle_dets["preview_teams"]) and hasattr(context, "team"):
```
MaciekBaron: Why do you check `hasattr(context, "team")`? Isn't that guaranteed by `there is a preview team` which is called in every scenario using this step?

functional_tests/steps/bundles.py
```diff
@@ -121,3 +129,337 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+                BundleTeam.objects.create(parent=bundle, team=context.team)
+                context.bundleteams.append({"parent": bundle, "team": context.team})
+            bundle.save()
+
+            if bool(bundle_dets["add_rel_cal"]) and hasattr(context, "release_calendar_page"):
```
MaciekBaron: If your bundle details have `"add_rel_cal"` as true, but you forgot to add a step to add a release calendar, this indicates a bug in the test code. Don't check for `hasattr(context, "release_calendar_page")`, fail here so that you can see your mistake.

functional_tests/steps/bundles.py
```diff
@@ -121,3 +129,337 @@ def the_user_can_see_the_bundle_details_with_creator(context: Context) -> None:
...
+    user.teams.add(context.team)
+
+
+@given("there are {number_of_bundles} bundles with {bundle_details}")
+def multiple_bundles_create(context: Context, number_of_bundles: str, bundle_details: str) -> None:
+    bundle_dets = json.loads(bundle_details)
+    if bundle_dets:
+        context.bundles = []
+        context.bundlepages = []
+        context.bundleteams = []
```
MaciekBaron: For consistency can't we have:
```suggestion
        context.bundle_pages = []
        context.bundle_teams = []
```

MaciekBaron: Second point here, why save this in `context` if you don't use it anywhere else? Am I missing something?

MaciekBaron (changes requested):

functional_tests/features/bundle_ui.feature
```diff
@@ -0,0 +1,133 @@
...
+        Then the user can create a bundle
+        And the user can see the created bundle
+
+        Examples: bundles
+           | role               |
+#           | Publishing Admin   |
+#           | Publishing Officer |
+#
+#
+    Scenario Outline: A User cannot create a bundle
```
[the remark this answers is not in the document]
AWitcherONS: Please see above

functional_tests/steps/bundles.py
```diff
@@ -10,14 +11,15 @@
 
 from cms.articles.tests.factories import ArticleSeriesPageFactory, StatisticalArticlePageFactory
 from cms.bundles.enums import BundleStatus
-from cms.bundles.models import BundlePage, BundleTeam
-from cms.bundles.tests.factories import BundleFactory, BundlePageFactory
-from cms.release_calendar.enums import ReleaseStatus
+from cms.bundles.models import BundleTeam
+from cms.bundles.tests.factories import BundleFactory
```
MaciekBaron: This looks like a bad merge. You removed `BundlePageFactory`, `ReleaseStatus` etc. even though those are used.

This would have been [picked up by the linter](https://github.com/ONSdigital/dis-wagtail/actions/runs/19671912859/job/56342699989?pr=273). Please remember to run linters before pushing code.

MaciekBaron (changes requested):

MaciekBaron (changes requested):

MaciekBaron (changes requested):

functional_tests/features/bundle_ui.feature
```diff
@@ -0,0 +1,211 @@
...
+#        | 1                 | Publishing Admin           | Publishing Officer  | {"role": "Publishing Admin",   "creator_role": "Publishing Officer", "status": "In_Review", "preview_teams": true, "add_rel_cal": true, "add_stat_page": true} |
+#        | 1                 | Publishing Admin           | Publishing Admin    | {"role": "Publishing Admin",   "creator_role": "Publishing Admin",   "status": "In_Review", "preview_teams": true, "add_rel_cal": true, "add_stat_page": true} |
+#        | 1                 | Publishing Officer         | Publishing Officer  | {"role": "Publishing Officer", "creator_role": "Publishing Officer", "status": "In_Review", "preview_teams": true, "add_rel_cal": true, "add_stat_page": true} |
+#        | 1                 | Publishing Officer         | Publishing Admin    | {"role": "Publishing Officer", "creator_role": "Publishing Admin",   "status": "In_Review", "preview_teams": true, "add_rel_cal": true, "add_stat_page": true} |
```
MaciekBaron: If there is a reason why you have this commented out, it would be prudent to either leave a comment explaining why, or update the PR description so that this is clear.

Same on line 62, 63 and 211.

functional_tests/steps/bundles.py
```diff
@@ -284,3 +290,431 @@ def user_sees_updated_release_calendar_page_title_release_date_status(context: C
...
+                context.bundlepages.append(BundlePageFactory(parent=bundle, page=page))
+            else:
+                now = (datetime.now() + timedelta(hours=4)).strftime("%Y-%m-%d %H:%M")
+                bundle.scheduled_date = now
+            bundle.save()
+
+            if bool(bundle_dets["add_stat_page"]) and hasattr(context, "statistical_article_page"):
+                page = context.statistical_article_page
+                BundlePage.objects.create(parent=bundle, page=page)
+                # context.bundlepages.append(BundlePageFactory(parent=bundle, page=page))
```
MaciekBaron: Commented out code left here.

MaciekBaron (changes requested):

MaciekBaron (changes requested):
I've noticed that in your latest commit you included a list of failing tests. They are failing because there are mistakes in your code. I have added a few comments to help you fix one of them ("A user can approve a bundle"). It's possible that these changes will also fix your other tests. They may also get resolved once you fix the things I mentioned in my other comments.

MaciekBaron (changes requested):

functional_tests/steps/bundles.py
```diff
@@ -284,3 +291,431 @@ def user_sees_updated_release_calendar_page_title_release_date_status(context: C
...
+
+@then("the {role} can preview the Statistical Analysis page")
+def preview_statistical_analysis_page(context: Context, role: str) -> None:
+    context.page.goto(context.base_url + "/admin/pages/")
+    context.page.get_by_role("button", name="Show filters").click()
+    context.page.get_by_role("button", name="Page type").click()
+    context.page.get_by_text("Statistical article page").click()
+    context.page.get_by_role("button", name="Show filters").click()
+    context.page.get_by_role("button", name="Locale").click()
+    context.page.locator("#id_locale").get_by_text("English").click()
```
MaciekBaron: Just to repeat what I said a few times now: when you have a `@then` statement, you want to make some assertions. Your step says "the {role} can preview the Statistical Analysis page" _(shouldn't this say Statistical Article page anyway?)_ but all you do is you go to the page index and change a few filters.

How does that check if you can preview the page? Why are there no assertions here, to check any expectations?

`/admin/pages` is the root index of all pages. Here by default you see just a list of pages Home pages. You can check if "Statistical article page" is in the list, but the thing is, _all_ page types are listed there. You don't even check if the page you created actually appears there, you just click on the filter.

Please rethink and rewrite this step to match the requirement.

functional_tests/steps/bundles.py
```diff
@@ -284,3 +291,431 @@ def user_sees_updated_release_calendar_page_title_release_date_status(context: C
...
+    context.page.get_by_role("textbox", name="Search term").fill(context.bundles[-1].name)
+    context.page.get_by_role("link", name=bundle_edit).click()
+    context.page.get_by_role("button", name="More actions").click()
+    context.page.get_by_role("button", name="Approve").click()
+    expect(context.page.get_by_role("status")).to_contain_text(bundle_status)
+
+
+@then("the logged in user cannot approve a bundle due to lack of pages")
+def cannot_approve_bundle_edge_case(context: Context) -> None:
+    bundle_edit = f"Edit '{context.bundles[-1].name}'"
```
MaciekBaron: 1. This test fails here as it cannot find the edit link.
2. This step should be split into at least two or three steps, e.g. "the user goes to the bundle view", "the user clicks approve", "the user is informed they cannot approve the bundle"
3. You should check that the message "Cannot approve the bundle without any pages or datasets" is visible, not just that the page could not be saved, in case something else has prevented it from being saved - be explicit in your tests.

MaciekBaron (changes requested):

functional_tests/features/bundle_ui.feature
```diff
@@ -0,0 +1,208 @@
...
+    Then the logged in user fails to create a bundle due used bundle name
+
+    Examples: bundles
+       | number_of_bundles | role                       | creator_role        |  bundle_details                                                                                                                                                |
+       | 1                 | Publishing Admin           | Publishing Officer  | {"role": "Publishing Admin",   "creator_role": "Publishing Officer", "status": "Draft", "preview_teams": false, "add_rel_cal": false, "add_stat_page": false} |
+       | 1                 | Publishing Officer         | Publishing Admin    | {"role": "Publishing Officer", "creator_role": "Publishing Officer", "status": "Draft", "preview_teams": false, "add_rel_cal": false, "add_stat_page": false} |
+
+Scenario Outline: A User cannot save a bundle due to duplicate schedule
+    Given there is a <role> user
+    And there is a Release Calendar page approved by <creator_role>
```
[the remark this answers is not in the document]
AWitcherONS: opps 
Resolved in [612ac1c](https://github.com/ONSdigital/dis-wagtail/pull/273/commits/612ac1c3d43b3d9ee6ffdea5479d80ac4d7998e8)

functional_tests/steps/bundles.py
AWitcherONS: Resolved in [612ac1c](https://github.com/ONSdigital/dis-wagtail/pull/273/commits/612ac1c3d43b3d9ee6ffdea5479d80ac4d7998e8)

functional_tests/features/bundle_ui.feature
```diff
@@ -0,0 +1,208 @@
...
+    And the <role> is a member of the preview team
+    And there are <number_of_bundles> bundles with <bundle_details>
+    When the <role> logs in
+    Then the logged in user cannot approve a bundle
+
+    Examples: bundles
+        | number_of_bundles | role                       | creator_role        | bundle_details                                                                                                                                                 |
+        | 1                 | Viewer                     | Publishing Admin    | {"role": "Viewer",             "creator_role": "Publishing Admin",  "status": "In_Review", "preview_teams": false,  "add_rel_cal": true, "add_stat_page": true}|
+        | 1                 | Publishing Officer         | Publishing Officer  | {"role": "Publishing Officer", "creator_role": "Publishing Officer", "status": "In_Review", "preview_teams": true, "add_rel_cal": true, "add_stat_page": true} |
+        | 1                 | Publishing Admin           | Publishing Admin    | {"role": "Publishing Admin",   "creator_role": "Publishing Admin",   "status": "In_Review", "preview_teams": true, "add_rel_cal": true, "add_stat_page": true} |
```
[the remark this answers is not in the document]
AWitcherONS: This Test has now been removed as is no longer applicable 
resolved [6c68056ac](https://github.com/ONSdigital/dis-wagtail/pull/273/changes/6c68056ac75b8d23ebc3585620d740d168c10752)

functional_tests/features/bundle_ui.feature
```diff
@@ -0,0 +1,208 @@
...
+    Given there is a <role> user
+    And there is a <creator_role> user
+    And there is a Statistical Analysis page approved by <creator_role>
+    And there is a Release Calendar page approved by <creator_role>
+    And there is a preview team
+    And the <role> is a member of the preview team
+    And there are <number_of_bundles> bundles with <bundle_details>
+    When the <role> logs in
+    Then the logged in user can approve a bundle
+    And the bundle pages are not live
```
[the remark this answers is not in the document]
AWitcherONS: Yeah it didn't make sense has been remove
resolved [6c68056ac](https://github.com/ONSdigital/dis-wagtail/pull/273/changes/6c68056ac75b8d23ebc3585620d740d168c10752)

functional_tests/features/bundle_ui.feature
```diff
@@ -0,0 +1,208 @@
...
+       | Publishing Officer |
+
+#---- Bundle Cannot Create UI Tests -----
+Scenario Outline: A User cannot create a bundle due to authorisation
+    Given there is a <role> user
+    When the <role> logs in
+    Then the logged in user cannot create a bundle
+    Examples: bundles
+       | role     |
+       | Viewer   |
```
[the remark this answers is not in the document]
MaciekBaron: Sorry Ann, I don't understand your reply.

[Scenario Outlines](https://behave.readthedocs.io/en/latest/tutorial/#scenario-outlines) are used for parameterised scenarios, where the scenario is executed for each row in the examples table.

However, you only have one example. It's a lot more readable to do it this way:

```suggestion
Scenario: A User cannot create a bundle due to authorisation
    Given there is a Viewer user
    When the Viewer logs in
    Then the logged in user cannot create a bundle
```

It's clear what we are testing, who the user is and what they are trying to do, and we use fewer lines. It all reads better – you don't need to go to the table to check what the `<role>` parameter will be. There is no reason to "replicate" the table from the scenario outline which has multiple different roles – I see no benefit in doing that. It just makes the code clunkier and more difficult to read. If you can show me a benefit, I can definitely take it into consideration.

functional_tests/steps/bundles.py
MaciekBaron: Point 2 hasn't been addressed.

Consider your current scenario ("A user cannot approve a bundle due to lack of pages or datasets"). After your initial setup with the `Given` + `And` statements, you have essentially just two steps:

```feature
    When the <role> logs in
    Then the logged in user cannot approve a bundle due to lack of pages
 ```
 
 This is a functional/UI test. We should be explicit in what the user is trying to do. Look at existing tests in the `bundle.feature` file, like this example:
 
 https://github.com/ONSdigital/dis-wagtail/blob/48622f3e254afe5ddbd4039962a5a468bea84a34/functional_tests/features/bundle.feature#L63-L72
 
 Notice how we clearly outline all of the steps the user is taking: what they are clicking, what they are putting in etc.. Your scenario doesn't do that. Some of your other scenarios suffer from the same problem.

**Review by MaciekBaron — CHANGES_REQUESTED** [review]
Thank you Ann for replying to some of my comments. It makes it easier to review and verify that you have addressed the issues raised. Could you please continue replying to all of the comments so that we know that everything has been addressed.

File: functional_tests/features/bundle_ui.feature [hunk 53]
```diff
@@ -0,0 +1,235 @@
...
+    Given there is a <role> user
+    And there is a <creator_role> user
+    And there is a release calendar page approved by <creator_role>
+    And the <role> logs in
+    And the logged in user goes to the bundle page
+    And the logged in user can see the create button
+    And the logged in user can create a bundle
+    And the logged in user adds a Name to the bundle
+    When the logged in user tries to find a non existing bundle page
+    Then the logged in user gets the following message for bundle_page
```
MaciekBaron: 1. What "following message"? A message doesn't follow.
2. What's `bundle_page`? The steps in your feature file should be human readable and make sense after variable substitution (this is not a variable, however). Please reconsider your approach as the sentence doesn't make sense. How about you store the search query in the context in your previous step so you can refer to it later on?

Here and on line 101

MaciekBaron (changes requested):

MaciekBaron (changes requested):

functional_tests/steps/bundles.py
```diff
@@ -317,3 +327,315 @@ def user_sees_validation_error_preventing_cancellation(context: Context) -> None
...
+
+@then("the logged in user add a schedule date to the bundle")
+def add_schedule_date(context: Context) -> None:
+    now = (datetime.now() + timedelta(hours=4)).strftime("%Y-%m-%d %H:%M")
+    context.page.get_by_role("textbox", name="or Publication date").fill(now)
+
+
+@then("the logged in user goes to the bundle page")
+def go_to_bundle_page(context: Context) -> None:
+    context.page.get_by_role("link", name="Bundles", exact=True).click()
```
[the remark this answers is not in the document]
MaciekBaron: You seem to have done a find-and-replace which affected other sentences ("can adds").

functional_tests/features/bundle_ui.feature
AWitcherONS: Hi 
this has been replaced with bundled_page
I think its more meaningful 
resolution in [07cf10a](https://github.com/ONSdigital/dis-wagtail/pull/273/commits/07cf10afc240fd87b32f36dce6af2dd30360a77d)

functional_tests/steps/bundles.py
```diff
@@ -359,3 +366,404 @@ def the_page_is_in_the_given_bundle_with_status(
...
+        expect(context.page.get_by_text(context.team.name)).to_be_visible()
+
+    expect(context.page.get_by_text("Created at")).to_be_visible()
+    expect(context.page.get_by_text("Created by")).to_be_visible()
+    expect(context.page.get_by_text("Scheduled publication")).to_be_visible()
+    expect(context.page.get_by_text("Associated release calendar")).to_be_visible()
+    expect(context.page.get_by_text("Datasets", exact=True)).to_be_visible()
+
+
+@step("the logged in user cannot approve a bundle")
```
MaciekBaron: This step is unused and it seems to be very similar to "the logged in user cannot approve a bundle due to lack of pages" on line 598.

sanjeevz3009:
I have taken over the ticket `CMS-333`. I will be closing this PR and creating a new one.

PR #277, opened 2025-06-30
Disallow line overlays for non-stacked column charts

helenb:

See https://jira.ons.gov.uk/browse/CCB-113

This change brings the chart builder in line with the DS code, to ensure that line overlays cannot be added to clustered column charts - only to stacked column charts.

### How to review

Try adding a column chart with multiple series, and select one series to be a line overlay.
If the column chart is not stacked, it should give an error message.
Other validation relating to selecting a series as an overlay should all still function correctly - see the ACs on the ticket for details.

### Follow-up Actions


MaciekBaron (approved):
LGTM

PR #279, opened 2025-06-30
Featured article component, with featured chart

nimasmi:

This change adds a `featured_chart` field to StatisticalArticlePage, and updates the featured article template to use the ONS Design System `onsFeaturedArticle` component.

Ticket: https://jira.ons.gov.uk/browse/CCB-95

In a change from the spec, the Wagtail "Featured article" preview mode actually renders the containing Topic page, rather than a mostly-blank template.

### How to review

On a Topic page, select a featured article series.
On the latest article in the series, add a featured chart (in the Promote tab).
Preview the page using the "Featured article" preview mode.
Visit the containing Topic page.

<details><summary>Screenshot</summary>
<p>

image

</p>
</details>


nimasmi:
**Comment**: I debated whether to include validation. The DS `onsFeaturedArticle` component specifies that one of `chart` or `image` is required ([macro spec](https://github.com/ONSdigital/design-system/blob/main/src/components/featured-article/_macro-options.md). However, this likely should not be required for all StatisticalArticlePages.

One option is some conditional validation:

- on Topic page, don't permit selecting a featured article series that does not have a chart or listing image configured
- on StatisticalArticlePage, require one of `featured_chart` or `listing_image` to be set if this page is selected as the featured article on its containing Topic page

However, validation like this is not very robust, and can be sidestepped by any number of things such as restoring an old revision of a page, or scheduling a new one for publishing. Do we want this, or is good publishing discipline enough?

cms/articles/models.py
```diff
@@ -414,11 +452,60 @@ def related_data(self, request: "HttpRequest") -> "TemplateResponse":
...
-            return cast("TemplateResponse", self.related_data(request))
+        match mode_name:
+            case "related_data":
+                return cast("TemplateResponse", self.related_data(request))
+            case "featured_article":
+                from cms.topics.models import TopicPage  # pylint: disable=import-outside-toplevel
+
+                topic_page = TopicPage.objects.ancestor_of(self).first()
+                topic_page.featured_series = self.get_parent().specific
+                return cast("TemplateResponse", topic_page.serve(request))
```
nimasmi, line 510: **Comment**: This, and the associated Gherkin feature definition, are a deviation from the ticket, which said: _Then I see a mostly plain page containing just the featured article component_. It fudges the content of the Topic page in memory for the preview, but doesn't save it. I think it's an improvement, but it's also a fudge, so something to flag for discussion.

cms/articles/tests/test_models.py
```diff
@@ -555,6 +557,82 @@ def test_load_in_external_env(self):
...
+        self.assertNotIn("image", data)
+
+    def test_as_featured_article_macro_data_without_chart(self):
+        """Test that a listing image is used when there is no chart."""
+        request = RequestFactory().get("/")
+        data = self.page.as_featured_article_macro_data(request)
+
+        self.assertNotIn("chart", data)
+
+        self.assertEqual(data["image"]["src"], self.page.listing_image.get_rendition("width-1252").url)
```
nimasmi, line 632: **Comment**: The DS does not specify an image size. I derived this from the maximum of 626 px wide I saw while resizing the viewport, scaled × 2 for high pixel density displays.

cms/datavis/blocks/featured_charts.py
```diff
@@ -0,0 +1,52 @@
...
+)
+
+
+class FeaturedLineChartBlock(LineChartBlock):
+    """Featured chart variant of LineChartBlock with simplified fields."""
+
+    # Override fields to remove unwanted options for featured charts
+    subtitle = None
+    caption = None
+    # TODO: remove download configuration options
```
nimasmi, line 16: **Comment**: Download configuration is currently a non-configurable placeholder, but will become configurable here once this change is added.

cms/articles/models.py
```diff
@@ -166,6 +176,25 @@ class StatisticalArticlePage(BundledPageMixin, RoutablePageMixin, BasePage):  #
     dataset_sorting = models.CharField(choices=SortingChoices.choices, default=SortingChoices.AS_SHOWN, max_length=32)
     datasets = StreamField(DatasetStoryBlock(), blank=True, default=list)
 
+    # Featured chart for Topic page display
+    featured_chart = StreamField(
```
zerolab, line 180: suggestion: move these to a blocks module

cms/articles/models.py
```diff
@@ -414,11 +452,60 @@ def related_data(self, request: "HttpRequest") -> "TemplateResponse":
...
+            },
+            "description": self.main_points_summary,
+        }
+
+        if self.release_date:
+            data["metadata"]["date"] = {
+                "prefix": _("Release date"),
+                "showPrefix": True,
+                "iso": self.release_date.isoformat(),
+                "short": ons_date_format(datetime.combine(self.release_date, datetime.min.time()), "DATE_FORMAT"),
```
zerolab, line 473: note: `ons_date_format(self.release_date, "DATE_FORMAT")` should also work.
I am curious to know why this was needed and whether we're missing a trick other places (like the hero in the statistical article page template).

cms/articles/models.py
```diff
@@ -414,11 +452,60 @@ def related_data(self, request: "HttpRequest") -> "TemplateResponse":
...
+
+        if self.featured_chart:
+            chart_block = self.featured_chart[0]  # pylint: disable=unsubscriptable-object
+            block_instance = chart_block.block
+            block_value = chart_block.value
+
+            if isinstance(block_instance, BaseVisualisationBlock):
+                data["chart"] = block_instance.get_component_config(block_value)
+            else:
+                raise ValueError(f"Unsupported block type: {type(block_instance)}")
```
zerolab, line 484: question: should this not move to the Page.clean or the admin form clean? That is, we want to surface these issues before publication

zerolab (changes requested):
Left a couple of notes/questions.

cms/articles/models.py
nimasmi: That's how I was using it earlier, but MyPy didn't like it, so I had to add `.combine()`.

cms/articles/models.py
nimasmi: I'm not sure that an editor would really be able to add an unsupported block type. This is here more as a placeholder because locally I have these lines in a git stash:

```diff
+            elif isinstance(block_instance, ONSEmbedBlock):
+                ...  # TODO Return component data for iframe, see https://github.com/ONSdigital/design-system/pull/3641
```

Perhaps the error lines can be removed.

zerolab: If the intention is to prevent us (developers) from shooting our foot, I'd recommend a system check.

Otherwise it may be cleaner to have a sensible fallback as we do not want the topic page to explode for unexplained reasons

cms/articles/models.py
zerolab: ah, I see what's happening here. 

```python
# cms/core/custom_date_format.py

def ons_date_format(value: datetime, format_string: str) -> str:
   ...
```

should be `def ons_date_format(value: date | datetime, format_string: str) -> str:

nimasmi: I have updated this in fcf0bf9fbc4c325b1b3f52cbb56d3917ee8d325e. I hope that's not overstepping the scope of this PR.

cms/articles/models.py
nimasmi: I have done this in 60d1b260f06bce4c2a5d9269595b999cd2937a7c.

cms/articles/models.py
zerolab: Not at all. And thank you!

cms/articles/models.py
nimasmi: I have removed these lines in 852f3c9.

cms/articles/models.py
nimasmi: Oh, thank you so much! I needed to move on to other high priority tickets before Friday, so I had just recorded this as a bug for now: https://jira.ons.gov.uk/browse/CCB-153.

nimasmi:
I have force-pushed this branch after updating and cherry-picking, as `feature/datavis` has now changed.

cms/articles/models.py
nimasmi: Thanks for the suggestion. I have added this in 5035076b4d41ad3cb573a609d9e9c3ba57a0e2f8.

PR #280, opened 2025-07-01
Ensure that data labels are not allowed if there are more than 20 data points in a series

helenb:

See https://jira.ons.gov.uk/browse/CCB-148

The Design System code is quite process intensive for calculating the positions of data labels. The DS has been updated to disallow data labels on bar charts with more than 20 data points - this is the equivalent change for the chart builder.

### How to review

In your local build, try adding a bar chart with more than 20 rows in the table. When you select to show data labels, they should not display. The help text should explain this.

The rule that data labels are only allowed on bar charts with 1 or 2 series should still apply.

### Follow-up Actions

### DIFF
--- a/cms/datavis/blocks/charts.py
+++ b/cms/datavis/blocks/charts.py
@@ -60,6 +60,7 @@ class LineChartBlock(BaseVisualisationBlock):
 class BarColumnChartBlock(BaseVisualisationBlock):
     x_axis_type = AxisType.CATEGORICAL
     MAX_SERIES_COUNT_WITH_DATA_LABELS = 2
+    MAX_DATA_POINTS_WITH_DATA_LABELS = 20
 
     # Error codes
     ERROR_DUPLICATE_SERIES = "duplicate_series_number"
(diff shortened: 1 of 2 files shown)


MaciekBaron (approved):

cms/datavis/blocks/charts.py
```diff
@@ -247,13 +252,15 @@ def get_series_item(
...
-        # This is only supported for non-stacked horizontal bar charts with 1 or 2 series
+        # This is only supported for non-stacked horizontal bar charts with 1 or 2 series,
+        # and where the number of data points does not exceed 20.
         if (
             value.get("show_data_labels")
             and value.get("select_chart_type") == BarColumnChartTypeChoices.BAR
             and not value.get("use_stacked_layout")
             # +1 to allow for the categories in the first column
             and len(value["table"].headers) <= self.MAX_SERIES_COUNT_WITH_DATA_LABELS + 1
+            and len(value["table"].rows) <= self.MAX_DATA_POINTS_WITH_DATA_LABELS
```
nimasmi, line 263: **Question**: I suspect that `value["table"].rows` may include the first row, which is used for the headings in the legend. Do we have an off-by-one situation here? I see you have a test for 21 data points failing validation, but I suspect a test case with 20 data points will incorrectly fail. Can we update or add to the tests, so that we test that 20 passes but 21 fails?

If the 20-row test case passes with no code changes that's good news, and thanks for writing this validation correctly, but but at least then we'll be testing the two boundary cases.

nimasmi (changes requested):

cms/datavis/blocks/charts.py
helenb: It doesn't include the header row because I checked, but I will update the tests anyway as that sounds sensible.

helenb: Now updated.

PR #282, opened 2025-07-02
Address defect in logic with line overlay on single series non-stacke…

helenb:
…d column chart

Addresses defect raised in https://jira.ons.gov.uk/browse/CCB-154

If a column chart has only 2 series, it should be possible to mark one series as a line overlay without the need for the chart to be stacked.

### How to review

- Test adding a series overlay to a column chart with 2 series. You should be able to do it with no error, even if the chart is not stacked
- Test adding a series overlay to a column chart with 3 or more series. You should not be able to do it if the chart is not stacked.

### Follow-up Actions

### DIFF
--- a/cms/datavis/blocks/charts.py
+++ b/cms/datavis/blocks/charts.py
@@ -184,7 +184,7 @@ class BarColumnChartBlock(BaseVisualisationBlock):
                     and not value.get("use_stacked_layout")
                 ) and len(series) > series_count_when_validating_line_overlay:
                     stream_block_errors[i] = ValidationError(
-                        "Non-stacked column charts with more than two series do not support line overlays.",
+                        "To show a line overlay with multiple series, use a stacked layout",
                         code=self.ERROR_NON_STACKED_COLUMN_NO_LINE,
                     )
 


cms/datavis/blocks/charts.py
```diff
@@ -171,18 +171,20 @@ def validate_series_customisation(self, value: "StructValue") -> "StructValue":
...
-                elif value.get("select_chart_type") == BarColumnChartTypeChoices.COLUMN and not value.get(
-                    "use_stacked_layout"
-                ):
+                elif (
+                    value.get("select_chart_type") == BarColumnChartTypeChoices.COLUMN
+                    and not value.get("use_stacked_layout")
+                ) and len(series) > series_count_when_validating_line_overlay:
                     stream_block_errors[i] = ValidationError(
-                        "Non-stacked column charts do not support line overlays.",
+                        "Non-stacked column charts with more than two series do not support line overlays.",
```
nimasmi, line 187: **Comment**: When writing these in the past I found we were just stacking clauses and qualifications on top of each other. Is it clearer to say:


> Clustered column charts do not support line overlays.

?

Though 'clustered' isn't really a term used in the UI. What about:

> To show a line overlay with multiple series, use a stacked layout.

?

nimasmi (approved):
The tests and validation look good. I have left an optional invitation to think about the user-facing error text.

PR #285, opened 2025-07-02
Enable chart axis min/max configuration

nimasmi:

Enable configuration of axis min/max limits for some chart types, and add "start/end on tick" checkboxes.

Ticket: https://jira.ons.gov.uk/browse/CCB-117

### How to review

See the x-axis and y-axis (or "value" axis) configuration for a chart type.

I have removed the placeholder fields for suffixes, which were a legacy from the proof-of-concept version of this code.

I have also used this opportunity to standardise on all chart types explicitly declaring their axis structblocks, as the rest of the code was heading in this direction anyway.


MaciekBaron (approved):
LGTM

PR #287, opened 2025-07-03
Add audio description field

helenb:

See https://jira.ons.gov.uk/browse/CCB-150
Adds a required audio description field for all charts.

### How to review

Check that when you edit a statistical article page, and add a chart, the audio description field is present and required.
Todo: add the field to the featured chart once that work is merged.

### Follow-up Actions

### DIFF
--- a/cms/datavis/blocks/base.py
+++ b/cms/datavis/blocks/base.py
@@ -27,6 +27,9 @@ class BaseVisualisationBlock(blocks.StructBlock):
 
     title = blocks.CharBlock()
     subtitle = blocks.CharBlock()
+    audio_description = blocks.TextBlock(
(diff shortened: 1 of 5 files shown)


cms/datavis/blocks/base.py
```diff
@@ -27,6 +27,9 @@ class BaseVisualisationBlock(blocks.StructBlock):
 
     title = blocks.CharBlock()
     subtitle = blocks.CharBlock()
+    audio_description = blocks.TextBlock(
```
zerolab, line 30: I keep going back to this question in my head -- will the field ever be used in other contexts? Would `accessible_description` be more representative?

absolutely non-blocking

zerolab: Sounds good to me, thank you

PR #288, opened 2025-07-03
Changes for the new url structure

zerolab:

This PR makes changes to conform to the planned URL structure update

<details>
<summary>🏗️ structure (relevant bits)</summary>

Path | Notes | In scope
-- | -- | --
https://www.ons.gov.uk/ |   |  
↳ /browse | list of themes | Day 2
&nbsp;&nbsp;↳ /{theme(s)} | specific theme pages with links to topics | Day 2
&nbsp;&nbsp;&nbsp;&nbsp;↳ /{theme(s)} | sub-theme | Day 2
↳ /release-calendar | the release calendar showing list of releases | Handled by the Search service
↳ /releases | redirect to release calendar | Y
&nbsp;&nbsp;↳ /{page_slug} | specific release calendar entry page | Y
↳ /{topic_slug} | topic landing page | Y
&nbsp;&nbsp;↳ /articles | list (or redirect to list) of all articles | Y
&nbsp;&nbsp; ↳ /{series_slug} | series landing page with latest article |
&nbsp;&nbsp; ↳ /editions | list of editions of series available (previous /previous-releases) |
&nbsp;&nbsp; ↳ /{edition_slug} | latest version of the edition |
&nbsp;&nbsp; ↳ /versions | list of versions of edition available |
&nbsp;&nbsp; ↳ /{version_number} | specific version |
&nbsp;&nbsp;↳ /methodologies | list (or redirect to list) of all methodologies | Maybe
&nbsp;&nbsp;&nbsp;&nbsp;↳ /{methodology_slug} | the methodology page | Y

</details>

> [!NOTE]
> As per details on the ticket, this does not update the structure retroactively

### How to review

- Go to admin, explore > home. 
- Add a new child page - Topic page
- observe that once created, it has an articles index and methodologies index
- add a series, with one or more articles. Add a correction to one of them
- observe the paths

### Follow-up Actions


cms/topics/tests/utils.py
```diff
@@ -0,0 +1,29 @@
...
+from django.urls import reverse
+from wagtail.test.utils.form_data import inline_formset, nested_form_data, rich_text, streamfield
+
+from cms.taxonomy.models import Topic
+
+if TYPE_CHECKING:
+    from django.test import Client
+
+
+def create_simple_topic_page(client: "Client", homepage_id: int) -> None:
```
MaciekBaron, line 12: Nit: Could we add a docstring here explaining that this sends a post request etc..

Or alternatively we could make rename this to `send_request_to_create_simple_topic_page`, even if it is a mouthful, what do you think?

cms/articles/models.py
```diff
@@ -474,3 +440,113 @@ def get_context(self, request: "HttpRequest", *args: Any, **kwargs: Any) -> dict
...
+                return redirect(page_url)
+
+        if kwargs.pop("related_data", None):
+            view, _view_args, _view_kwargs = self.resolve_subpage("/related-data/")
+            serve_kwargs = {**kwargs, **_view_kwargs}
+            return cast("HttpResponse", super().serve(request, view=view, args=args, kwargs=serve_kwargs))
+
+        if version := kwargs.pop("version", None):
+            view, _view_args, _view_kwargs = self.resolve_subpage(f"/versions/v{version}/")
+            serve_kwargs = {**kwargs, **_view_kwargs}
```
MaciekBaron, line 549: **Suggestion**:
Since `_view_kwargs` is used, should we get rid of the prefix to avoid confusion (assuming we follow the convention of adding _ to indicate unused variables)?

```suggestion
            view, _view_args, view_kwargs = self.resolve_subpage(f"/versions/v{version}/")
            serve_kwargs = {**kwargs, **view_kwargs}
```

Here and in the block above.

MaciekBaron (changes requested):

MaciekBaron (approved):
Looks like the tests still need fixing but apart from that LGTM 👍

functional_tests/features/methodology_page.feature
```diff
@@ -1,16 +1,16 @@
 Feature: A general use of Methodology Page
```
sanjeevz3009, line 2: We can utilise `Background` in this feature file and reduce the repetition of `Given a topic page exists under the homepage` in this feature file.

```suggestion
Background:
    Given a topic page exists under the homepage
```

zerolab: @sanjeevz3009 while tempting, I refrained from making changes outside of the scope of the ticket as it already changes quite a few files.

sanjeevz3009: All good. We have many more UI testing tickets where this can be cleaned up. Will chuck this comment in one of the UI tickets in the backlog. https://jira.ons.gov.uk/browse/CMS-329, added in the comments.

Resolving, thanks @zerolab

AdamHawtin:
> This is the current behavious (e.g. https://www.ons.gov.uk/economy/grossdomesticproductgdp/bulletins/quarterlynationalaccounts/previousreleases), unless I misunderstood something?

My mistake then, it seemed on in the new hierarchy, but I see it still makes sense considering the way a user would reach that page is from a link on the article itself.

AdamHawtin (approved):
LGTM, just a couple of tests need fixing.

sanjeevz3009 (commented):
Some minor comments and questions. I am experiencing some issues with the breadcrumbs on all pages. It's happening after @AdamHawtin SEO PR went in. I don't think it's my local setup.

So I stopped reviewing and dug into that. Maybe it's broken or maybe it's meant to be, but the breadcrumbs URL went from `<a class="ons-breadcrumbs__link" href="/theme-page-1/">Theme Page 1</a>` to`<a class="ons-breadcrumbs__link" href="http://localhost/theme-page-1/">Theme Page 1</a>`

Will clear this up tomorrow and continue

AdamHawtin:
@sanjeevz3009 Now the breadcrumbs are using full URLs, it matters whether the port set for the wagtail site is correct, I had the same issue and @zerolab pointed out it can be changed in the "site" settings in the admin UI. If you change it to `8000` to match out dev default, the full URL links should all work again. I think a nice dev convenience update would be to do this programmatically from a make target which could be included in `make dev-init`.

cms/topics/tests/test_models.py
```diff
@@ -35,10 +35,12 @@ def setUpTestData(cls):
...
         cls.topic_page.featured_series = cls.article_series
-        cls.topic_page.save()
+        cls.topic_page.save(update_fields=["featured_series"])
 
-        cls.methodology = MethodologyPageFactory(parent=cls.topic_page, publication_date=datetime(2024, 6, 1))
-        cls.another_methodology = MethodologyPageFactory(parent=cls.topic_page, publication_date=datetime(2024, 11, 1))
+        cls.methodology = MethodologyPageFactory(parent__parent=cls.topic_page, publication_date=datetime(2024, 6, 1))
+        cls.another_methodology = MethodologyPageFactory(
+            parent=cls.methodology.get_parent(), publication_date=datetime(2024, 11, 1)
+        )
```
sanjeevz3009, line 43: Question: Why do you do `parent__parent` here `cls.methodology = MethodologyPageFactory(parent__parent=cls.topic_page, publication_date=datetime(2024, 6, 1))`

Then the next line `parent` here `cls.another_methodology = MethodologyPageFactory(parent=cls.methodology.get_parent(), publication_date=datetime(2024, 11, 1)`

sanjeevz3009 (commented):
I am having some issues but I feel like it might be me, so will check it out with @AdamHawtin tomorrow.

**Further findings:**

1. One issue is when I try go to the `live` page of `Methodology Index Page` and `Articles Index Page` it takes me to the topic page. Is this intended?

Will have a final look tomorrow before approving.


Tested everything, apart from the comments I left already and questions, LGTM =)

cms/topics/tests/test_models.py
zerolab: The methodology page parent is now a methodology index page, so `parent__parent=cls.topic_page` ensure that `cls.methodology` is created under a methodology index page that is created under `cls.topic`.
As for `cls.another_methodology`, we want to use the freshly created methodology index page as a parent, which is `cls.methodology.get_parent()`

kacperpONS:
> 1. One issue is when I try to go to the `live` page of `Methodology Index Page` and `Articles Index Page` it takes me to the topic page. Is this intended?

Currently the `serve` method in `MethodologyIndexPage` and `ArticlesIndexPage` is intercepting and returning a link to the parent topic page. There's a [FIXME note](https://github.com/ONSdigital/dis-wagtail/pull/288/files#diff-a5e9694fcee2526edb5aab945808bffb159e7ce7520fd6be25f346186c75c1ffR66) to remove this once the pages are implemented as mentioned in https://github.com/ONSdigital/dis-wagtail/pull/288#issuecomment-3078041125.

Another side effect of that is when trying to preview either of those pages you'll get the `TemplateDoesNotExist` in the preview. For consistency we could overwrite the `serve_preview` function of those pages, but I think it would make more sense to add a very simple placeholder template @zerolab (just so it doesn't make non-tech people freak out when they see the error)

Another issue related to this that I spotted is that it always redirects to the URL of the **live** `TopicPage`, but there's no guarantee that it has been published so it could result in a 404. We can introduce a check and only redirect to the live page if it's published, otherwise redirecting to the preview URL. 
```
    def serve(self, request: "HttpRequest", *args: Any, **kwargs: Any) -> "HttpResponse":
        # FIXME: redirect to the publications listing for the topic
        if self.get_parent().live:
            parent_url = self.get_parent().get_url(request=request)
        else:
            parent_url = reverse("wagtailadmin_pages:preview_on_edit", args=[self.get_parent().id])
        return redirect(parent_url)
```

sanjeevz3009:
> > 1. One issue is when I try to go to the `live` page of `Methodology Index Page` and `Articles Index Page` it takes me to the topic page. Is this intended?
> 
> @sanjeevz3009 Currently the `serve` method in `MethodologyIndexPage` and `ArticlesIndexPage` is intercepting and returning a link to the parent topic page. There's a [FIXME note](https://github.com/ONSdigital/dis-wagtail/pull/288/files#diff-a5e9694fcee2526edb5aab945808bffb159e7ce7520fd6be25f346186c75c1ffR66) to remove this once the pages are implemented as mentioned in [#288 (comment)](https://github.com/ONSdigital/dis-wagtail/pull/288#issuecomment-3078041125).
> 
> Another side effect of that is when trying to preview either of those pages you'll get the `TemplateDoesNotExist` in the preview. For consistency we could overwrite the `serve_preview` function of those pages, but I think it would make more sense to add a very simple placeholder template @zerolab (just so it doesn't make non-tech people freak out when they see the error)
> 
> Another issue related to this that I spotted is that it always redirects to the URL of the **live** `TopicPage`, but there's no guarantee that it has been published so it could result in a 404. We can introduce a check and only redirect to the live page if it's published, otherwise redirecting to the preview URL.
> 
> ```
>     def serve(self, request: "HttpRequest", *args: Any, **kwargs: Any) -> "HttpResponse":
>         # FIXME: redirect to the publications listing for the topic
>         if self.get_parent().live:
>             parent_url = self.get_parent().get_url(request=request)
>         else:
>             parent_url = reverse("wagtailadmin_pages:preview_on_edit", args=[self.get_parent().id])
>         return redirect(parent_url)
> ```

@kacperpONS I spotted all the points you raised above (`I am having some issues, but I feel like it might be me, so I will check it out with @AdamHawtin tomorrow.`):
- My main concern was around the redirect to the topic page, as that's misleading out of the box, hence the intended comment. Should have mentioned it's misleading.
- For the other issues, I guess users will see a 404 for now. It's probably something for follow-up PRs, maybe, not sure what @zerolab wants to do.

The latter issue, not sure if it's worth tackling on this PR or if we have to fix that as a whole across Wagtail, as there are other similar behaviours like this in Wagtail where we error out when a parent page is not published and in draft for example. Probably not in scope here, but just a thought.

kacperpONS:
> The latter issue, not sure if it's worth tackling on this PR or if we have to fix that as a whole across Wagtail, as there are other similar behaviours like this in Wagtail where we error out when a parent page is not published and in draft, for example. Probably not in scope here, but just a thought.

@sanjeevz3009 Yea, but in this case it's a result of us overwriting the `serve` function, nothing really to do with Wagtail internals here. 

Good thing is that  we could eliminate all those issues if we just added placeholder templates like we do for the `ThemePage` so that the users won't see any unexpected/erroneous behaviour. @zerolab what do you think?

sanjeevz3009:
> > The latter issue, not sure if it's worth tackling on this PR or if we have to fix that as a whole across Wagtail, as there are other similar behaviours like this in Wagtail where we error out when a parent page is not published and in draft, for example. Probably not in scope here, but just a thought.
> 
> @sanjeevz3009 Yea, but in this case it's a result of us overwriting the `serve` function, nothing really to do with Wagtail internals here.
> 
> Good thing is that we could eliminate all those issues if we just added placeholder templates like we do for the `ThemePage` so that the users won't see any unexpected/erroneous behaviour. @zerolab what do you think?

Or not have a preview for now, idk! Whatever works, I guess!

zerolab:
@kacperpONS @sanjeevz3009 removed the preview in f035651

MebinAbraham (changes requested):
Looking good, most things seem as expected. A few queries/concerns from me.

1. We have a  `Theme Index page` but we haven’t finalised `/browse` journey or whether the slug should be enforced and given theme isn’t in scope for Day 2, I suggest we remove it from the installed apps for the production setup or at the very least, prevent users from creating Theme pages. Otherwise, we risk allowing people to publish pages that were never signed off. The same applies to Information pages, though I'm more open to revisiting that. Theme, however, needs proper design and rework before being reintroduced.
    - Related note: the Theme index template currently throws Jinja errors when viewed. If we plan to keep it, those need fixing.
2. When creating a Topic page, users can still see the option to add a child page via the `+` button or when inside the page via "Add child page", but they then get the message `Sorry, you cannot create a page at this location`. Can we disable the button or the option altogether to avoid giving users a dead end?
3. The help panel for Article Series still references outdated paths like `/latest` and `/previous-releases`. It currently says: `This is a container for article series. It provides the /latest,/previous-releases evergreen paths, as well as the actual statistical article pages. Add a new Statistical article page under this container.`. This needs to be updated to reflect the new structure.
4. This only affects sandbox/Heroku environments, but we need to decide (and communicate) how we’ll handle older content structures and migrate people to the new structure. I want to merge this before promoting to Staging, so we don't get issues with outdated structs/data in Staging.
5. This might just be me, I can’t seem to recreate it (will try to), posting in case anyone else came across it. I had content built using the old structure, then created a new statistical article. When I clicked “View latest,” it unexpectedly took me to a completely different (old) article. It could be a slug collision, but slugs are supposed to be unique. I haven’t been able to reproduce this again so it may have resolved itself. I’ll re-raise it if it happens again.
6. Non-blocker as it would be a follow-up: The paper also suggests that visiting `/versions` should list all available versions. We don’t currently have a design or requirement for this, but flagging in case it needs to be added later. Awaiting James’ input before taking action.

Unrelated but flagging for awareness:
1. Currently, when previewing a Stats Article that is the latest draft, the banner says “Not the latest release.” This is because `is_latest` property only checks against live content. We probably need a conditional check for whether the request is in preview mode or have multiple preview modes, I believe we had a card for revisit preview modes as a whole?. Something to revisit.
2. Can you recall the decision around breadcrumbs for `/editions`, `/the-evergreen`, and `/related-data`? The live site shows the same breadcrumb structure for all three, but in our build, `/editions` includes the latest stats article (e.g. Articles > CPI) which makes sense but `/related-data` doesn’t follow the same pattern. Feels inconsistent and worth tidying up in a follow-up.

cms/articles/wagtail_hooks.py
```diff
@@ -50,3 +52,13 @@ def before_delete_page(request: "HttpRequest", page: "Page") -> HttpResponseRedi
...
+
+
+@hooks.register("after_create_topic_page")
+def after_create_topic_page(request: "HttpRequest", topic_page: "TopicPage") -> "HttpResponse | None":
+    articles_index = ArticlesIndexPage(title="Articles")
+    topic_page.add_child(instance=articles_index)
+    articles_index.save_revision().publish()
+
+    for fn in hooks.get_hooks("after_create_page"):
+        fn(request, articles_index)
```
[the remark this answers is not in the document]
zerolab: 461e98d3

cms/articles/models.py
```diff
@@ -106,6 +125,20 @@ def previous_releases(self, request: "HttpRequest") -> "TemplateResponse":
...
+        if not (edition := StatisticalArticlePage.objects.live().child_of(self).filter(slug=slug).first()):
+            raise Http404
+        return cast("HttpResponse", edition.serve(request, serve_as_edition=True, **kwargs))
+
+    @path("editions/<str:slug>/related-data/")
+    def release_related_data(self, request: "HttpRequest", slug: str) -> "HttpResponse":
+        return cast("HttpResponse", self.release(request, slug, related_data=True))
+
+    @path("editions/<str:slug>/versions/v<int:version>/")
+    def release_with_versions(self, request: "HttpRequest", slug: str, version: int) -> "HttpResponse":
```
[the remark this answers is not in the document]
zerolab: 461e98d3

zerolab:
@MebinAbraham all relevant feedback addressed in 461e98d3

MebinAbraham:
> @MebinAbraham all relevant feedback addressed in [461e98d](https://github.com/ONSdigital/dis-wagtail/commit/461e98d3435c7fb13ca1ef4434a5868d5e13b0e0)

Thanks @zerolab, what are your thoughts on the other comments?

cms/topics/tests/factories.py
```diff
@@ -14,7 +14,7 @@ class Meta:
 
     title = factory.Faker("sentence", nb_words=4)
     summary = factory.Faker("text", max_nb_chars=100)
-    parent = factory.SubFactory(ThemePageFactory)
+    parent = factory.LazyFunction(lambda: HomePage.objects.first())  # pylint: disable=unnecessary-lambda
```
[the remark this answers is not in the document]
zerolab: 93577ab

cms/topics/tests/test_models.py
sanjeevz3009: Thanks Dan, make sense!

sanjeevz3009 (commented):
✅

PR #291, opened 2025-07-04
[DRAFT] Release Calendar Page Date Change Log Validation

MaciekBaron:

This PR relates to CMS-457.

More TBC

### How to review

TBC

### Follow-up Actions
 TBC

### DIFF
--- a/cms/release_calendar/forms.py
+++ b/cms/release_calendar/forms.py
@@ -4,6 +4,7 @@ from django.core.exceptions import ValidationError
 from django.urls import reverse
 from django.utils.html import format_html
 from wagtail.admin.forms import WagtailAdminPageForm
+from wagtail.blocks.stream_block import StreamValue
 from wagtail.models import Locale
 
 from cms.bundles.permissions import user_can_manage_bundles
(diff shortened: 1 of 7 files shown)


cms/core/static/js/blocks/readonly-struct-block.js
```diff
@@ -0,0 +1,38 @@
...
+      deleteButton.style.display = 'none';
+      const duplicateButton = getActionButton('DUPLICATE');
+      duplicateButton.style.display = 'none';
+    }
+
+    const getFieldContainer = (field) =>
+      document.querySelector(`${parent} [data-contentpath="${field}"]:has(#${prefix}-${field})`);
+
+    const frozenCheckbox = getFieldContainer('frozen');
+    frozenCheckbox.style.display = 'none';
```
zerolab, line 29: suggestion: what happens if the frozen field is not there?

Granted we use it in focussed way. Perhaps add a docblock to the class outlining that the StructBlock must have help text and a `frozen = blocks.BooleanBlock(required=False, default=False)` block too?

cms/release_calendar/forms.py
```diff
@@ -13,6 +14,9 @@
 from .enums import LOCKED_STATUS_STATUSES, NON_PROVISIONAL_STATUS_CHOICES, ReleaseStatus
 from .utils import get_translated_string, parse_day_month_year_time, parse_month_year
 
+if TYPE_CHECKING:
+    from .models import ReleaseCalendarPage
+
 DATE_SEPERATOR = {
```
zerolab, line 20: not in scope of this PR, but would be nice to fix: `DATE_SEPERATOR` -> `DATE_SEPARATOR`

zerolab (changes requested):
Two minor suggestion from me.

It does what it says on the tin. I agree on the follow-ups, especially clarifying the frozen bit. With CMS-485, once we pre-populate the previous release date, we can make that read-only ("frozen")

zerolab (approved):

cms/release_calendar/forms.py
sanjeevz3009: We definitely have to fix this part of this PR. Unacceptable 🫣

sanjeevz3009 (commented):
Thank you for the UI tests here =) Going to check one more thing before approval.

AdamHawtin (approved):
LGTM 👍

PR #306, opened 2025-07-14
Rendering LaTeX Equations Locally

sanjeevz3009:

Currently, when we try to utilise the equation block in Wagtail CMS outside of the container, when running the application locally, it will break due to the equivalent packages for rendering equations are not installed locally for MacOS development. This additionally causes the tests to fail when running them locally.

### How to review

**Before**
- Go to an information page -> add an example equation in -> save draft/ publish -> the page would throw an error, not save. 

**After**

- Follow the updated README instructions to install using `MacPorts`
- Go to an information page -> add an example equation in -> save draft/ publish -> the page should save successfully  -> go to the full preview or live page -> turn off JS -> the equations should be rendering successfully.

**Bonus**
- After dev feedback from a couple of members in the team on time wasted, going in circles due to the required `Node packages` not being installed, and the `npm build` not taking place (as we assume everything should be good to go after running the `make dev-init` command), I have decided to add these as part of the `make dev-init` target.

### Follow-up Actions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,18 @@ format-html:  ## Format the HTML code
 format-frontend:  ## Format front-end files (CSS, JS, YAML, MD)
 	npm run format
 
+.PHONY: nvm-use
+nvm-use: ## Use the Node Version Manager to set the Node version
+	nvm use
+	
+.PHONY: npm-install
+npm-install: nvm-use ## Install the Node.js dependencies
+	npm install
+
+.PHONY: npm-build
+npm-build: nvm-use ## Build the front-end assets
+	npm run build
+
 .PHONY: lint
 lint: lint-py lint-html lint-frontend lint-migrations ## Run all linters (python, html, front-end, migrations)
 
@@ -166,7 +178,7 @@ runserver: ## Run the Django application locally
 	poetry run python ./manage.py runserver 0:8000
 
 .PHONY: dev-init
-dev-init: load-design-system-templates collectstatic compilemessages makemigrations migrate load-topics createsuperuser  ## Run the pre-run setup scripts
+dev-init: load-design-system-templates collectstatic compilemessages makemigrations migrate load-topics createsuperuser npm-install npm-build ## Run the pre-run setup scripts
 
 .PHONY: functional-tests-up
 functional-tests-up:  ## Start the functional tests docker compose dependencies
--- a/README.md
+++ b/README.md
@@ -57,8 +57,9 @@ Ensure you have the following installed:
    using the Docker setup.
 5. **[Node](https://nodejs.org/en)** and **[`nvm` (Node Version Manager)](https://github.com/nvm-sh/nvm)** for front-end
    tooling.
-6. **[JQ](https://jqlang.github.io/jq/)** for the step in the build that installs the design system templates
-7. **Operation System**: Ubuntu/MacOS
+6. **[JQ](https://jqlang.github.io/jq/)** for the step in the build that installs the design system templates.
+7. **[MacPorts](https://www.macports.org/install.php)**: We need MacPorts so we can install the required packages to render equations using LaTeX strings via Matplotlib locally (outside of a docker container).
+8. **Operation System**: Ubuntu/ MacOS.
 
 ### Setup
 
@@ -461,6 +462,16 @@ and every `send_mail` call will appear instantly in the UI.
 > ```python
 > EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 > ```
+
+### Installing The Required LaTeX Packages For Local Development
+
+In order to generate the equations in Wagtail using LaTeX strings via Matplotlib (Non-JS Equations), we will need to use `MacPorts` to install the following packages `texlive-latex-extra`.
+
+As a prerequisite you will have to have `MacPorts` installed. See [Pre-requisites](#pre-requisites) step 7 for MacPorts installation.
+
+```bash
+sudo port install texlive-latex-extra
+```
 
 ### Django Migrations
 


README.md
```diff
@@ -57,8 +57,9 @@ Ensure you have the following installed:
    using the Docker setup.
 5. **[Node](https://nodejs.org/en)** and **[`nvm` (Node Version Manager)](https://github.com/nvm-sh/nvm)** for front-end
    tooling.
-6. **[JQ](https://jqlang.github.io/jq/)** for the step in the build that installs the design system templates
-7. **Operation System**: Ubuntu/MacOS
+6. **[JQ](https://jqlang.github.io/jq/)** for the step in the build that installs the design system templates.
+7. **[MacPorts](https://www.macports.org/install.php)**: We need MacPorts so we can install the required packages to render equations using LaTeX strings via Matplotlib locally (outside of a docker container).
```
RealOrangeOne, line 61: Nit: The requirement is `texlive-latex-extra` rather than MacPorts specifically, right? Perhaps worth changing this to be more explicit

```suggestion
7. `texlive-latex-extra`: Required by `matplotlib` to render LaTeX equations. See [below](#installing-the-required-latex-packages-for-local-development) for instructions on how to install on macOS.
```

Makefile
```diff
@@ -166,7 +178,7 @@ runserver: ## Run the Django application locally
 	poetry run python ./manage.py runserver 0:8000
 
 .PHONY: dev-init
-dev-init: load-design-system-templates collectstatic compilemessages makemigrations migrate load-topics createsuperuser  ## Run the pre-run setup scripts
+dev-init: load-design-system-templates collectstatic compilemessages makemigrations migrate load-topics createsuperuser npm-install npm-build ## Run the pre-run setup scripts
```
MebinAbraham, line 181: Nit: I don't think install should be here, just like poetry install isn't. Happy to keep npm-build. Worth updating readme accordingly around installing dependencies.

README.md
```diff
@@ -462,6 +464,16 @@ and every `send_mail` call will appear instantly in the UI.
...
 
+### Installing The Required LaTeX Packages For Local Development
+
+In order to generate the equations in Wagtail using LaTeX strings via Matplotlib (Non-JS Equations), we will need to use `MacPorts` to install the following packages `texlive-latex-extra`.
+
+As a prerequisite you will have to have [MacPorts](https://www.macports.org/install.php) installed.
+
+```bash
+sudo port install texlive-latex-extra
+```
```
MebinAbraham, line 475: Just a heads up, although installing the .pkg file is supposed to automatically update your PATH so that MacPorts can be found, this didn't happen for me (prob some local issue). I had to manually add it myself. So, if devs encounter issues with MacPorts not being recognised, they should add the following to their PATH.

```
export PATH=/opt/local/bin:/opt/local/sbin:$PATH
```

The documentation suggests that the installer should add it automatically when using the pkg installer https://guide.macports.org/#installing.shell.postflight.

Makefile
zerolab: I agree.

Also, I keep having this niggly feeling that perhaps we're abstracting simple commands in the makefile a bit too much. I get the ones that are unfamiliar perhaps, but isn't `npm install`/`npm run build` better than `make npm-install`/`make npm-build`? Not a blocker for this PR, but something to consider down the line

README.md
```diff
@@ -462,6 +464,16 @@ and every `send_mail` call will appear instantly in the UI.
...
 > ```
 
+### Installing The Required LaTeX Packages For Local Development
+
+In order to generate the equations in Wagtail using LaTeX strings via Matplotlib (Non-JS Equations), we will need to use `MacPorts` to install the following packages `texlive-latex-extra`.
+
+As a prerequisite you will have to have [MacPorts](https://www.macports.org/install.php) installed.
+
+```bash
+sudo port install texlive-latex-extra
```
MebinAbraham, line 474: For me, it still doesn't work, as it complains about fonts when trying to generate an equation.
I get `! LaTeX Error: File 'type1ec.sty' not found.`

Installing the following fixed it: `sudo port install texlive-fonts-recommended`

MebinAbraham (changes requested):

Makefile
MebinAbraham: Indeed, our Makefile has become quite large, and we will end up with many aliases for simple commands. I think we should make the README better if things are not clear. I'd encourage devs to customise their own environment by creating local aliases, using `nvm use` at shell startup, and so on. Let's revisit this topic soon.

README.md
MebinAbraham: Still no luck, the pkg installer already supposedly runs `sudo port -v selfupdate`, but I have tried it manually, then reinstalled `texlive-latex-extra` and still no joy. Can you try the following formula and see if it works `$$\sum_{i=0}^n i^2 = \frac{(n^2+n)(2n+1)}{6}$$`.

Makefile
AdamHawtin: Would the easiest and most consistent solution be to include `npm-install` in the `make install`/`make install-dev` targets? I agree we shouldn't be just abstracting single commands for the sake of it, but bundling them into sensible combinations to make dev workflows simpler seems useful to me.

sanjeevz3009: Adam and I had a discussion, and I'll be happy to go with leaving the `npm run build` within `make dev-init`. Happy to add `npm install` under `make install` and `make install-dev`.

README.md
sanjeevz3009: I'll do a fresh install from scratch and also got @AdamHawtin to test the instructions to see why we are experiencing different behaviours. Let's see comes out of it

sanjeevz3009:
@MaciekBaron @MebinAbraham @AdamHawtin @zerolab 

- I redid the whole thing from 0 to 100
- So installing MacPorts itself
    - It added the path to `.zprofile` as Adam found
- Didn't have to do `selfupdate ` this time as I tried without, last time I had to
- Ran `sudo port install texlive-latex-extra`
- The tests all pass, that's in `cms/core/tests/test_utils`
- It all works well in Wagtail

Adam, on the other hand, is having some issues as his tests aren't passing, but I think it works when he runs Wagtail. He also found ` $$\sqrt{3x-1} + (1+x)^2$$` and ` $$sqrt{3x-1} + (1+x)^2$$` don't seem to validate, as currently an error is thrown that it's not valid. This is valid LaTeX.

README.md
sanjeevz3009: Done! [6cd9231](https://github.com/ONSdigital/dis-wagtail/pull/306/commits/6cd9231266845164af17ba3e7b237e6fc6a924a2)

Makefile
sanjeevz3009: Done! @MebinAbraham @zerolab @AdamHawtin [ffbbb11](https://github.com/ONSdigital/dis-wagtail/pull/306/commits/ffbbb114d9f7f7615debf84cdb0834d52d9cd552)

PR #311, opened 2025-07-16
Disable only the global submitted notifications

zerolab:

Follow up to #307. Upon further testing, it is only "submitted notifications" that is global. The other ones are dependent on the user editing/interacting with the page, so they are relevant.

### How to review

### Follow-up Actions
N/A

<details>
<summary>command to run</summary>

```python
from wagtail.users.models import UserProfile

for up in UserProfile.objects.all():
   up.submitted_notifications = False
   up.approved_notifications = True
   up.rejected_notifications = True
   up.updated_comments_notifications = True
   up.save(update_fields=["submitted_notifications", "approved_notifications", "rejected_notifications", "updated_comments_notifications"])
```
</details>

### DIFF
--- a/cms/users/tests/test_signal_handlers.py
+++ b/cms/users/tests/test_signal_handlers.py
@@ -130,7 +130,9 @@ class UserProfileSignalTestCase(WagtailTestUtils, TestCase):
         cls.user = cls.create_superuser(username="admin", password="password")
         cls.profile = UserProfile.get_for_user(cls.user)
 
-    def test_user_profile_notifications_disabled_on_creation(self):
+    def test_user_profile_submitted_notifications_disabled_on_creation(self):
+        # only "submitted notifications" are disabled as they are global
+        # the rest are left as they are as they require interaction with the page/workflow-enabled snippet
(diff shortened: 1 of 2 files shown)


cms/users/tests/test_signal_handlers.py
```diff
@@ -132,20 +132,14 @@ def setUpTestData(cls) -> None:
 
     def test_user_profile_notifications_disabled_on_creation(self):
         self.assertFalse(self.profile.submitted_notifications)
-        self.assertFalse(self.profile.approved_notifications)
-        self.assertFalse(self.profile.rejected_notifications)
-        self.assertFalse(self.profile.updated_comments_notifications)
+        self.assertTrue(self.profile.approved_notifications)
```
MaciekBaron, line 135: NP: As this test function is called `test_user_profile_notifications_disabled_on_creation` should we add a comment above explaining which ones are actually disabled and which ones are enabled, based on your PR description, just for posterity?

MaciekBaron (approved):

sanjeevz3009 (approved):

PR #312, opened 2025-07-16
Add command to prevent concurrent migration execution

RealOrangeOne:

https://jira.ons.gov.uk/browse/CMS-95

Django's `migrate` command doesn't like being run concurrently, with multiple processes trying to lock the database.

This PR uses PostgreSQL's Advisory Locks to add a lock to running migrations, which prevents concurrency issues. This will slightly slow down deployments, however this will only be run in the internal environment, where the impact is significantly less.

Note: Because this is mostly made up of an external library and Django's migration state, testing it would be extremely difficult, and likely not worth the effort.

### How to review

### Follow-up Actions

Updating the infrastructure to run this command will be handled separately.


docs/custom-features/locked-migrate.md
```diff
@@ -0,0 +1,14 @@
...
+To work around this, there's a custom `locked_migrate` management command. This command uses a [PostgreSQL Advisory Lock](https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS) to ensure only a single `migrate` command can run.
+
+1. First, check if there are any migrations to run. This avoids the need to lock on or wait for the lock if there's nothing to do. This can be skipped with `--skip-unapplied-check`
+2. If the lock is free, acquire it and run `migrate`
+3. If the lock is held, wait for up to `--timeout`. If the lock frees during that time, the process acquires the lock and runs `migrate` as above. If the lock times out, the process terminates.
+
+Advisory locks are automatically released when the connections is terminated. If multiple processes are waiting for the lock, the first is given the lock (First in, first out).
+
+> [!WARNING]
+> Running `manage.py migrate` directly is still supported, however does not do any locking checks. Ensure no other migration commands are running before running this.
```
MebinAbraham, line 14: Q: To clarify, this will "lock" the call to "migrate," so essentially, either all or no migrations occur, and it is not at an individual migration file level?

docs/custom-features/locked-migrate.md
```diff
@@ -0,0 +1,14 @@
+# Locked Migrate
```
MebinAbraham, line 1: Can you update the links in https://github.com/ONSdigital/dis-wagtail/blob/main/docs/custom-features/README.md and https://github.com/ONSdigital/dis-wagtail/blob/main/docs/README.md

cms/core/management/commands/locked_migrate.py
```diff
@@ -0,0 +1,74 @@
+import sys
```
MebinAbraham, line 1: We probably want this for other commands like publish bundles, etc.

MebinAbraham: If the infra prevents parallel runs, then I'm happy as-is.

docs/custom-features/locked-migrate.md
RealOrangeOne: Done in https://github.com/ONSdigital/dis-wagtail/pull/312/commits/7f589a27bb6765ebcb3529af5d6995afba953b5e

docs/custom-features/locked-migrate.md
RealOrangeOne: We can. I'm not opposed to doing it that way, it just felt slightly cleaner having it as a separate command.

AdamHawtin (approved):
LGTM 👍

docs/custom-features/locked-migrate.md
zerolab: Fair enough.

sanjeevz3009 (approved):
LGTM

cms/core/management/commands/locked_migrate.py
```diff
@@ -0,0 +1,57 @@
...
+        if not options["skip_unapplied_check"] and not self._has_unapplied_migrations():
+            # If there are no migrations to run, don't touch the lock to
+            # let the process stop sooner
+            self.stdout.write("No migrations to run.", self.style.SUCCESS)
+            return
+
+        self.stdout.write("Acquiring lock...", self.style.MIGRATE_HEADING)
+
+        with pglock.advisory(lock_id=__name__, timeout=options["timeout"]) as acquired:
+            if acquired:
```
MebinAbraham, line 52: Q: Could we end up with stale locks? (edge case). What if the process exists after the lock was acquired?.
Do we want a max duration? What if the job hangs and the lock is never released?. Do we have a max runtime at infra level?.

MebinAbraham: > That said, there shouldn't be any reason that the migrations lock up - migrations are always run with --noinput. Similarly, in the case of data migrations, it might be expected that migrations take "a little" (eg up to a minute probably isn't a shock).

Yes, there shouldn't be; it's an edge case, and we would need a runbook to highlight what we would do if it did happen for some reason.

MebinAbraham (commented):
LGTM, not approval as I've contributed now.

cms/core/management/commands/locked_migrate.py
```diff
@@ -0,0 +1,57 @@
+import sys
```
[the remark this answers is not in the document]
zerolab: I think this is safe to resolve now

cms/core/management/commands/locked_migrate.py
RealOrangeOne: The default pod settings should be fine for our use case. I think the DR playbooks will need investigating at a later stage, as we think through everything we'd need to cover.

PR #313, opened 2025-07-16
Bundles - UX/user feedback tweaks, part 1

zerolab:

This is part 1 of a set of PRs for CMS-562.
Screencast - https://www.loom.com/share/4e8f921264b141e2b3481e8caec339c0

It:
- removes extraneous fields from bundle listings
- makes the main bundle listing ordered by name by default, and makes the remaining columns orderable
- hides published bundles by default, but allows surfacing them via the filters

### How to review

1. check the dashboard panels columns
2. check the bundle index view, apply status filters, order by different fields

### Follow-up Actions


cms/bundles/models.py
```diff
@@ -83,6 +86,34 @@ def previewable(self) -> Self:
     def annotate_release_date(self) -> "BundlesQuerySet":
         return self.annotate(release_date=models.F("release_date"))  # type: ignore[no-any-return]
 
+    def annotate_queryset_updated_at(self) -> "BundlesQuerySet":
```
zerolab, line 89: This is what the index view does (https://github.com/wagtail/wagtail/blob/60aca9201e78b592fb07227704d8139605afbabb/wagtail/admin/views/generic/models.py#L108-L131)

Added so we surface "updated at" on the dashboard.

I am half-tempted to transform this into its own datetime field on `Bundle`

cms/bundles/models.py
```diff
@@ -83,6 +86,34 @@ def previewable(self) -> Self:
...
+            log_model.objects.filter(
+                content_type=ContentType.objects.get_for_model(self.model, for_concrete_model=False),
+                object_id=Cast(models.OuterRef("pk"), models.CharField()),
+            )
+            .order_by("-timestamp", "-pk")
+            .values("timestamp")[:1]
+        )
+        return self.annotate(updated_at=models.Subquery(latest_log))  # type: ignore[no-any-return]
+
+    def annotate_status_label(self) -> "BundlesQuerySet":
```
zerolab, line 108: This is used for ordering by the status label on the index view, rather than the `BundleStatus` value, so that "Draft" (value: "DRAFT") is before "Ready to publish" (value: "APPROVED")

cms/bundles/viewsets/bundle.py
```diff
@@ -363,15 +364,40 @@ class BundleIndexView(IndexView):
...
+            queryset = queryset.exclude(status=BundleStatus.PUBLISHED)
+
+        return cast("BundlesQuerySet", super().filter_queryset(queryset))
+
+    def order_queryset(self, queryset: "BundlesQuerySet") -> "BundlesQuerySet":
+        if self.ordering in ["status", "-status", "scheduled_publication_date", "-scheduled_publication_date"]:
+            # ensures we have the _updated_at field used by core
+            queryset = self._annotate_queryset_updated_at(queryset)
+
+            match self.ordering:
```
zerolab, line 390: These are needed because `scheduled_publication_date` is a property.
And status needs to look at the label, rather than the value

sanjeevz3009:
For others' benefit:

The ACs from CMS-562 that are being covered in this PR
```
Given I am a member of the publishing team
When I view the bundles lists (on dashboard, and the bundles listing page)
Then I see the list sorted by bundle name ascending (A-Z)
And I do not see the added, added by, preview teams, approved at, appproved by columns

Given I am a member of the publishing team
When I view the bundles listing page
Then I do not see the published bundles by default, but can use the filter to show them
 
Given I am a member of the publishing team
When I view the bundles listing page
Then I have a column with buttons to edit and inspect the bundle
And all the columns are sortable

Given I am a member of the publishing team
When I view the bundles listing page
Then I have a button to see all bundles that are ready to publish
```

sanjeevz3009 (approved):
LGTM!

Nit: Currently the bundle index view table is as follows: `Name`, `Scheduled for`, `Status` and `Updated`

Meanwhile the dashboard panel columns: `Title`, `Status`, `Scheculed publication date` and `Updated`.

Some inconstiences between the order of the columns and the headers.

sanjeevz3009 (approved):
LGTM!

cms/bundles/models.py
MebinAbraham: > I am half-tempted to transform this into its own datetime field on Bundle
Would be my preference.

zerolab: ad5663b3

cms/bundles/migrations/0005_bundle_updated_at.py
```diff
@@ -0,0 +1,45 @@
...
+        ModelLogEntry.objects.filter(
+            content_type=ContentType.objects.get_for_model(Bundle, for_concrete_model=False).pk,
+            object_id=Cast(models.OuterRef("pk"), models.CharField()),
+        )
+        .order_by("-timestamp", "-pk")
+        .values("timestamp")[:1]
+    )
+    for bundle in Bundle.objects.annotate(_updated_at=models.Subquery(latest_log)):
+        bundle.updated_at = bundle._updated_at
+        bundle.save(update_fields=["updated_at"])
```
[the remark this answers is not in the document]
zerolab: Good catch/shout. 86e8b7a7

MaciekBaron (approved):

PR #314, opened 2025-07-17
Set Wagtail site ports in dev-init

AdamHawtin:
The Wagtail site configuration defaults to port `80` which means full URLs generated for pages in that site won't work out of the box in development environments running on port `8000`. This PR includes a command in `make dev-init` to change all Wagtail site ports to `8000` so they are correct in local dev environments.

### How to review
- Run the app locally _**without**_ running the updated `make dev-init` on this PR
- Observe that links on published pages generated with `.full_url()` don't work correctly, such as breadcrumbs
- Run the updated `make dev-init`
- Observe those same links should now include the correct port and work as expected

### DIFF
--- a/cms/core/blocks/__init__.py
+++ b/cms/core/blocks/__init__.py
@@ -1,3 +1,4 @@
+from .accordion import AccordionBlock, AccordionSectionBlock
 from .embeddable import (
     DocumentBlock,
     DocumentsBlock,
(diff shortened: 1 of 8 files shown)


sanjeevz3009 (approved):
Thanks @AdamHawtin for picking this up. LGTM!

PR #319, opened 2025-07-21
Exclude aliases from the bundle page chooser

zerolab:

This is a follow up to #262 which excludes aliases from the bundle pages chooser too

### How to review

- add/edit a bundle
- go to add pages - only editable pages show. All aliases are excluded.

### Follow-up Actions

### DIFF
--- a/cms/bundles/tests/test_viewsets.py
+++ b/cms/bundles/tests/test_viewsets.py
@@ -553,7 +553,9 @@ class BundlePageChooserViewsetTestCase(WagtailTestUtils, TestCase):
 
         cls.bundle = BundleFactory()
 
-        cls.page_draft = StatisticalArticlePageFactory(live=False, title="Article draft")
+        cls.page_draft = StatisticalArticlePageFactory(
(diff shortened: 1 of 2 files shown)


MaciekBaron (approved):

kacperpONS (commented):
Not something related to this work, but I just noticed a potentially confusing behaviour. 


In a scenario where: a Page gets added to a Bundle, and the Bundle is then saved. When you remove the Page in the UI without saving the Bundle and try to re-add the Page to the Bundle, it's not there in the `ChooserViewSet` widget until the Bundle is saved. 

Similarly, when you create a brand new Bundle, then you're able to add the same page to it multiple times before you save. This scenario seems to be handled by deduplicating the pages. (and this must be happening in core as I couldn't find anything in our `BundleEditView` that would be responsible for this)

This feels like something that must have been picked up by someone before me, but I didn't  find anything about it on the "Wagtail issues tracker" page, so I'm leaving this here for visibility.

kacperpONS (approved):
Other than the above LGTM

PR #320, opened 2025-07-22
Bundles - UX/user feedback tweaks, part 2

zerolab:

Builds on #313 and covers ACs: 4, 6, 6a, 7, 8, 9, 17, 18 and
- adds a "View 'ready to publish'" button in the bundles listing header
- moves the edit/inspect buttons on their own
- adds a proper action menu for the bundle add/edit view with varying options depending on the bundle status
- replaces the status field select with a help panel on the edit form
- tweaks the redirects on save

### How to review

See https://www.loom.com/share/1d2fd851732c4237993d05bcc7472b65

### Follow-up Actions


cms/bundles/tests/test_viewsets.py
```diff
@@ -147,28 +164,86 @@ def test_bundle_edit_view__redirects_to_index_for_published_bundles(self):
         response = self.client.get(self.published_bundle_edit_url)
         self.assertRedirects(response, self.bundle_index_url)
 
-    def test_bundle_edit_view__updates_approved_fields_on_save_and_approve(self):
-        """Checks the fields are populated if the user clicks the 'Save and approve' button."""
+    def post_with_action_and_test(self, action: str, expected_status: BundleStatus, redirects_to: str | None = None):
```
MaciekBaron, line 167: Nit-pick: Look like all function calls provide `redirects_to` so maybe we can simplify it?

cms/bundles/viewsets/bundle.py
```diff
@@ -134,6 +164,55 @@ def run_after_hook(self) -> None:
...
+                url = self.get_edit_url()
+        return cast(str, url)
+
+    def get_success_buttons(self) -> list:
+        # only include the edit button when not staying on the edit page.
+        match self.object.status:
+            case BundleStatus.IN_REVIEW:
+                buttons = cast(list, super().get_success_buttons())
+            case BundleStatus.APPROVED:
+                buttons = cast(list, super().get_success_buttons())
```
MaciekBaron, line 211: Both cases are the same, is this intentional?

cms/bundles/templates/bundles/wagtailadmin/action_menu/menu_item.html
```diff
@@ -0,0 +1,15 @@
+{% load i18n wagtailadmin_tags %}
+<button
+    type="submit"
+    name="{{ name }}"
+    value="{{ name }}"
+    class="button button-longrunning"
+    data-controller="w-progress{% if use_shortcut %} w-kbd{% endif %}"
+    data-action="w-progress#activate"
+    data-w-progress-active-value="{{ label_progress }}"
+    {% if use_shortcut %} data-w-kbd-key-value="mod+s" data-w-kbd-scope-value="global"{% endif %}
```
MaciekBaron, line 10: Super nit-pick: There's a new line before the if block so we can probably drop the whitespace?

```suggestion
    {% if use_shortcut %}data-w-kbd-key-value="mod+s" data-w-kbd-scope-value="global"{% endif %}
```

zerolab: 354bde99

cms/bundles/tests/test_viewsets.py
zerolab: 354bde99

PR #323, opened 2025-07-22
Bundles ux part 3 deduplication and locale width cms 562

zerolab:

This PR builds on #320 and covers ACs 10 and 12. It

- adds a base class to use for deduplicating items. Moves `DeduplicateTopicsAdminForm` to `cms.core` and uses that
- updates the locale column to massage the width so the text doesn't flow on 2 rows

### How to review

### Follow-up Actions


cms/core/forms.py
```diff
@@ -9,14 +9,55 @@
...
+            item = form.clean().get(target_field)
+            if item is None:
+                # tidy up in case the page reference is empty
+                self.formsets[formset].forms[idx].cleaned_data["DELETE"] = True
+                continue
+
+            if item in chosen:
+                # we saw this already, mark for removal to avoid duplicates
+                self.formsets[formset].forms[idx].cleaned_data["DELETE"] = True
+                continue
```
MaciekBaron, line 45: Both `if` blocks contain the same code, is this intentional?

MaciekBaron (dismissed):

sanjeevz3009 (commented):
LGTM =)

One of the things I noticed was that the `bundled pages` chooser behaves differently from the `datasets` and `preview teams` choosers, as in once a page has been selected and is already in the bundle, it won't show the same page again when using the `bundles pages` chooser, but this isn't the case for the `dataset` and `preview teams` choosers. But that's not in scope for this card.

Will approve first in the morning once JIRA is up on Monday =)

The multichooser functionality is awesome and consistent now for datasets, as before it differed for bundles and the release calendar page.

cms/core/forms.py
```diff
@@ -9,14 +9,50 @@
...
+            chosen.add(item)
+
+        return None
+
+
+class DeduplicateTopicsAdminForm(DeduplicateInlinePanelAdminForm):
+    def clean(self) -> dict[str, Any] | None:
+        cleaned_data: dict[str, Any] | None = super().clean()
+
+        self.deduplicate_formset("topics", "topic")
```
MaciekBaron, line 51: Minor non-blocking suggestion:

I think maybe it would be slightly more readable if we did:

```python
def deduplicate_formset(self, *, formset: str, target_field: str)
```

And then

```python
self.deduplicate_formset(formset="topics", target_field="topic")
```

This way it's immediately clear what those two strings are as it is difficult to tell from the method name alone.

cms/core/forms.py
```diff
@@ -9,14 +9,50 @@
...
 
+class DeduplicateInlinePanelAdminForm(WagtailAdminPageForm):
+    def deduplicate_formset(self, formset: str, target_field: str) -> None:
+        """Deduplicate InlinePanel chosen items.
+
+        Wagtail choosers currently do not have the ability to remove already selected values.
+        so the same item can be selected multiple times, so this method helps with deduplication.
+
+        Revisit once https://github.com/wagtail/wagtail/issues/10496 is fixed.
+        """
```
MebinAbraham, line 27: Nit:
```suggestion
        Wagtail choosers currently do not have the ability to remove already selected values.
        so the same item can be selected multiple times, so this method helps with deduplication.

        TODO: Revisit once https://github.com/wagtail/wagtail/issues/10496 is fixed.
        """
```

cms/core/forms.py
MebinAbraham: +1

cms/core/forms.py
```diff
@@ -9,14 +9,50 @@
...
 from cms.core.utils import FORMULA_INDICATOR, latex_formula_to_svg
-from cms.taxonomy.forms import DeduplicateTopicsAdminForm
 
 logger = logging.getLogger(__name__)
 
 
 LATEX_VALIDATION_ERROR = "The equation is not valid LaTeX. Please check the syntax and try again."
 
 
+class DeduplicateInlinePanelAdminForm(WagtailAdminPageForm):
```
MebinAbraham, line 19: ❤️

cms/core/forms.py
zerolab: e3ed553

cms/core/forms.py
zerolab: I agree. e3ed553

sanjeevz3009 (approved):
LGTM

cms/core/forms.py
kacperpONS: TIL- the ```*``` magick! 😃

PR #328, opened 2025-07-28
Fix cache key for kafka token

RealOrangeOne:

The Kafka client ignores all exceptions when generating a token.

The root of the issue is that `__qualname__` doesn't exist on class instances, only on the class itself.

### How to review

Confirm the `token` method is still cached, and has a reasonable cache key.

### DIFF
--- a/cms/search/tests/test_publishers.py
+++ b/cms/search/tests/test_publishers.py
@@ -7,7 +7,7 @@ from wagtail.test.utils import WagtailTestUtils
 from cms.articles.tests.factories import StatisticalArticlePageFactory
 from cms.methodology.tests.factories import MethodologyPageFactory
 from cms.release_calendar.tests.factories import ReleaseCalendarPageFactory
-from cms.search.publishers import BasePublisher, KafkaPublisher, LogPublisher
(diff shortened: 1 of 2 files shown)


cms/search/tests/test_publishers.py
```diff
@@ -134,6 +134,22 @@ def test_publish_deleted(self, mock_producer_class):
...
+
+    @patch("cms.search.publishers.MSKAuthTokenProvider.generate_auth_token")
+    def test_token_provider(self, mock_generate_auth_token):
+        mock_generate_auth_token.return_value = ("msk-token", None)
+        token_provider = IAMKafkaTokenProvider()
+
+        self.assertEqual(token_provider.token(), "msk-token")
+        self.assertEqual(token_provider.token(), "msk-token")
+
+        mock_generate_auth_token.assert_called_once()
```
MaciekBaron, line 151: Nit: Could we do this just for greater test clarity?
```suggestion
        mock_generate_auth_token.assert_not_called()

        # Retrieve the token twice to ensure caching works
        self.assertEqual(token_provider.token(), "msk-token")
        self.assertEqual(token_provider.token(), "msk-token")

        mock_generate_auth_token.assert_called_once()
```

PR #329, opened 2025-07-28
Bundles - UX/user feedback tweaks, part 4 - choosers, panels and overall tidy ups

zerolab:

This PR builds on #323 and its parents and covers ACs 5, 11, 13, 14, 15, 16, 18. It:
- makes the bundle page chooser return the same metadata as it does when accessing the edit page with a bundle page (include the full admin title and the workflow status)
- includes and "add to bundle" link on the bundle info panel if a bundleable page is not yet in a bundle
- makes the bundle edit form read-only when the bundle is "Ready to publish"
- locks page editing and removes the option to cancel workflow when the page is in a bundle that is ready to publish
- shows the dashboard panel with bundles ready for review to publishing staff regardless of the team membership
- ensure the dashboard bundle lists are sorted alphabetically, but within the latest 10.
- tidies up some of the bundle tests
- ensures manual publication updates the linked release calendar page (possible if for some reason the bundle was not published automatically and the release date <= now)

Commits are atomic

### How to review

See ACs and https://www.loom.com/share/9159a017f9f1468595cdb6f2f2231e6a
([superseded screencast](https://www.loom.com/share/c255c6aee10c4321ad2adbd9f1adf5ec))

### Follow-up Actions


cms/bundles/panels.py
```diff
@@ -38,37 +40,96 @@ def _content_for_instance(self, instance: "Model") -> Union[str, "SafeString"]:
...
-                        bundle.get_status_display(),
-                    )
-            else:
-                content = format_html("<p>{}</p>", "This page is not part of any bundles.")
-            return content
+                return format_html(
+                    "<p>This page is in the following bundle: {} (Status: {})</p>",
+                    bundle.name,
+                    bundle.get_status_display(),
+                )
```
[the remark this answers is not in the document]
zerolab: 7cbdba2

MaciekBaron (approved):

AdamHawtin (commented):
Trying to test the bundle moving to "In Preview" when a bundled page is cancelled, only the page creator seems to be able to cancel the workflow? Even for other publishing admin users, I'm seeing "Page Locked" in the interaction menu with the banner "This page is awaiting 'Ready to publish' in the 'Release review' workflow. Only reviewers for this task can edit the page.". Is this expected, or could I be missing a step somewhere? This is after adding the page to a bundle and approving it.

Also, does there need to be some sort of warning/speedbump when cancelling the page workflows if their bundle is ready to publish? This change seems to make it possible to cancel a ready bundle with one click, without making that particularly obvious to the user

AdamHawtin:
@zerolab The new banner makes sense and works well 👍 

Checking and debugging that `user_can_cancel` logic, it seems to be that only the user which created the page and the user which submitted for review can then cancel. I think we need to validate if this is suitable for our use case, or whether any publishing admin (for example) should be able to cancel page workflows.

cms/bundles/tests/test_viewsets.py
```diff
@@ -139,6 +141,39 @@ def test_bundle_add_view__without_permissions(self):
                 self.assertRedirects(response, "/admin/")
                 self.assertContains(response, "Sorry, you do not have permission to access this area.")
 
+    def test_date_placeholder(self):
+        """Test that the date input field displays date placeholder."""
+        response = self.client.get(self.bundle_add_url, follow=True)
+
+        datetime_placeholder = "YYYY-MM-DD HH:MM"
+        self.assertContains(
```
MaciekBaron, line 149: Non blocking nit: I think in other places we've been using `self.assertInHTML` if we want consistency, but this also works

cms/workflows/locks.py
```diff
@@ -0,0 +1,55 @@
...
+        lock_message = ""
+        if user_can_manage_bundles(user):
+            return format_html(
+                "This page is included in a bundle that is ready to be published. You must revert the bundle "
+                "to <strong>Draft</strong> or <strong>In preview</strong> in order to make further changes. "
+                '<span class="buttons">'
+                '<a type="button" class="button button-small button-secondary" href="{url}">Manage bundle</a></span>',
+                base_message=lock_message,
+                url=reverse("bundle:edit", args=(self.object.active_bundle.pk,)),
+            )
```
MaciekBaron, line 34: `lock_message` seems redundant here as it's an empty string and doesn't change

cms/articles/models.py
```diff
@@ -432,6 +433,11 @@ def dataset_document_list(self) -> list[dict[str, Any]]:
             dataset_documents = sorted(dataset_documents, key=lambda d: d["title"]["text"])
         return dataset_documents
 
+    @cached_property
+    def parent_for_choosers(self) -> Page:
+        topic_page_class = resolve_model_string("topics.TopicPage")
+        return topic_page_class.objects.ancestor_of(self).first().specific_deferred
+
```
kacperpONS, line 440:
```suggestion
    def parent_for_choosers(self) -> Page:
        """Display the ancestor Topic page instead of the Article Series page in choosers."""
        topic_page_class = resolve_model_string("topics.TopicPage")
        return topic_page_class.objects.ancestor_of(self).first().specific_deferred

```
😇

cms/workflows/locks.py
zerolab: ah, remnant from a different version. removed 49b3cfee

cms/articles/models.py
zerolab: 49b3cfee

cms/bundles/forms.py
```diff
@@ -52,6 +64,8 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
 
     def _has_datasets(self) -> bool:
         has_datasets = False
+        if "bundled_pages" not in self.formsets:
+            return False
```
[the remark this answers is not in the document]
zerolab: good catch. 49b3cfee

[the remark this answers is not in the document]
kacperpONS: That makes more sense now! 😄

cms/bundles/management/commands/publish_bundles.py
```diff
@@ -117,21 +56,19 @@ def _handle_bundle_action(self, bundle: Bundle) -> None:
...
             # Confirm the bundle is still approved
             if bundle.status != BundleStatus.APPROVED:
-                logger.error("Bundle no longer approved", extra={"bundle_id": bundle.id})
+                logger.error("Bundle no longer approved", extra={"bundle_id": bundle.pk})
                 return
 
             self.handle_bundle(bundle)
         except Exception:  # pylint: disable=broad-exception-caught
-            logger.exception("Publish failed", extra={"bundle_id": bundle.id, "event": "publish_failed"})
+            logger.exception("Publish failed", extra={"bundle_id": bundle.pk, "event": "publish_failed"})
```
kacperpONS, line 64: Out of scope, but it would be good to at least send an email to the relevant people when this happens. (preview team/ last editor/ person who approved?)

Potentially in the future also a visual notification within Wagtail (another panel on the the main dashboard?)

kacperpONS (approved):

sanjeevz3009 (commented):
Reviewed!

PR #330, opened 2025-07-29
Extend SESSION_COOKIE_AGE when Cognito disabled

MebinAbraham:

Most of us for local dev are testing with Wagtail core logins. This PR updates the `dev.py` to ensure that sessions do not expire at 15 minutes if Cognito is not enabled.

Note, this only works if you set `AWS_COGNITO_LOGIN_ENABLED` via env vars. If you don't, you can also override the value in your `local.py` file.

### How to review

Set `AWS_COGNITO_LOGIN_ENABLED` to False via env var, then ensure you are not logged out after 15 minutes of inactivity. If you don't want to wait around for 15 minutes, temporarily adjust the [base.py](https://github.com/ONSdigital/dis-wagtail/blob/main/cms/settings/base.py#L1006) value to like 10 seconds and validate that the `dev.py` value takes precedence. 

### Follow-up Actions

### DIFF
--- a/cms/search/publishers.py
+++ b/cms/search/publishers.py
@@ -63,7 +63,8 @@ class IAMKafkaTokenProvider(AbstractTokenProvider):
 
     # Generating the token does a request, so cache it for slightly less than the expiration.
     @memory_cache(
-        MSKAuthTokenProvider.DEFAULT_TOKEN_EXPIRY_SECONDS - 5, key_generator_callable=lambda self: self.__qualname__
+        MSKAuthTokenProvider.DEFAULT_TOKEN_EXPIRY_SECONDS - 5,
+        key_generator_callable=lambda self: f"{__name__}.{type(self).__qualname__}",
     )
     def token(self) -> str:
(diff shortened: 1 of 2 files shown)


cms/settings/dev.py
```diff
@@ -86,6 +86,8 @@
     WAGTAILADMIN_LOGIN_URL = env.get("WAGTAILADMIN_LOGIN_URL", f"{AUTH_STUB_URL}/florence/login")
     LOGOUT_REDIRECT_URL = env.get("LOGOUT_REDIRECT_URL", f"{AUTH_STUB_URL}/florence/logout")
     AWS_REGION = "eu-west-2"
+else:
+    SESSION_COOKIE_AGE = 60 * 60 * 24  # 1 day in seconds
```
RealOrangeOne, line 90: Question: Should this just live in `base.py` instead? Given we're already manually applying it as an env var anyway.

sanjeevz3009 (approved):
LGTM =)

kacperpONS (approved):

RealOrangeOne (approved):

PR #332, opened 2025-07-30
Add and use `make test-no-parallel` on GitHub

MaciekBaron:

This PR adds a a new `make` command called `test-no-parallel` which is exactly the same as `test` but executes the test without the `--parallel` flag. This is then used for the CI workflow on GitHub.

The reason behind this is to reduce test flakiness as there are signs of test isolation issues. The old `make test` command will still work as usual.

### How to review

Observe that tests pass.

### Follow-up Actions

Evaluate impact of change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -164,7 +164,7 @@ jobs:
       - name: Compile translations
         run: make compilemessages
       - name: Test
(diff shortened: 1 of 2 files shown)


MaciekBaron:
Superseded by #335

PR #333, opened 2025-07-31
GTM and GA support

AdamHawtin:
Quick Video Overview: https://www.loom.com/share/d47453df092648ca94da50b7f9e1a6dd?sid=4ed6a43f-6864-4abc-8552-40b0f5f68c48

To enable better Google Analytics support through Google Tag Manager, we need to include more metadata in the page datalayer and some page components, in line with the new specification.

This PR adds attributes/metdata to:
- The page level datalayer to add metadata to page view events
- Top and footer navigation components
- Table of contents components
- Related links components
- Document Download components
- Accordion components

### How to review
The spec we are working to is linked to in the associated Jira card, to verify the metadata added is the right metadata. 

Fully testing this will require setting up a GA environment and running the app configured to point to it, for this review it should be sufficient to run the app locally and check in the browser that the correct datalayer pushes and component attributes are being set.

**Note:** Some of the design system components are missing support for attributes in the places we require, the table of contents and navigation components don't yet support the per item attributes so these values we are passing to the component will currently be ignored.

### Follow-up Actions

This is a first pass MVP implementation, we expect follow-up tasks over time to support more of the requested spec.
There are some components which require GTM attributes which we have not yet implemented (chart and file downloads).


cms/core/blocks/base.py
```diff
@@ -82,9 +85,46 @@ def get_related_link(self, context: dict | None = None) -> dict[str, str | dict[
...
+        if (
+            str(type(target_page)) == "StatisticalArticlePage"
+        ):  # TODO better way to check type without circular imports?
+            attributes.update(
+                {
+                    "data-gtm-click-output-series": target_page.cached_analytics_values["outputSeries"],
+                    "data-gtm-click-output-edition": target_page.cached_analytics_values["outputEdition"],
+                    "data-gtm-click-release-date": target_page.cached_analytics_values["releaseDate"],
+                }
+            )
```
MaciekBaron, line 125: `str(type(target_page))` will not result in the name of the class:

```python
from cms.articles.models import StatisticalArticlePage
article = StatisticalArticlePage.objects.first()
str(type(article)) # "<class 'cms.articles.models.StatisticalArticlePage'>"
```

I'd do it this way (also simplifying access to `target_page.cached_analytics_values`):
```python
if target_page.__class__.__name__ == "StatisticalArticlePage":
    cav = target_page.cached_analytics_values
    attributes.update({
        "data-gtm-click-output-series": cav.get("outputSeries", ""),
        "data-gtm-click-output-edition": cav.get("outputEdition", ""),
        "data-gtm-click-release-date": cav.get("releaseDate", ""),
    })
```
_You're already doing it here:_
https://github.com/ONSdigital/dis-wagtail/blob/40c70c96b430071152235aa561370109d0cc7b46/cms/core/models/base.py#L246

cms/release_calendar/models.py
```diff
@@ -294,3 +297,12 @@ def get_lock(self) -> Optional["BaseLock"]:
             return PageInBundleReadyToBePublishedLock(self)
 
         return super().get_lock()
+
+    @cached_property
+    def cached_analytics_values(self) -> dict[str, str | bool]:
```
MaciekBaron, line 302: This gets repeated, should we consider a mixin?

AdamHawtin: Although a few implementations look very similar, they're not quite identical because the field name for `lastUpdatedDate` varies between models, which I think makes it more trouble than it's worth to try and factor the common shape out into a mixin.

cms/core/blocks/base.py
AdamHawtin: Done 👍

MaciekBaron (changes requested):

cms/topics/tests/test_models.py
```diff
@@ -381,6 +424,14 @@ def test_cannot_add_children_once_articles_and_methodologies_index_are_created(s
         self.assertContains(response, "Add child page", 3)
         self.assertContains(response, reverse("wagtailadmin_pages:add_subpage", args=[new_topic_page.pk]))
 
+    def test_get_analytics_values(self):
+        analytics_values = self.topic_page.get_analytics_values(get_dummy_request())
+        self.assertEqual(analytics_values["pageTitle"], self.topic_page.title)
+        self.assertEqual(analytics_values["contentType"], self.topic_page.analytics_content_type)
+        self.assertEqual(analytics_values["contentGroup"], self.topic_page.analytics_content_group)
+        self.assertEqual(analytics_values["contentTheme"], self.topic_page.analytics_content_theme)
+        self.assertEqual(analytics_values["contentTheme"], self.topic_page.analytics_content_theme)
```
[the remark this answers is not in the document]
AdamHawtin: Addressed in [5af6472](https://github.com/ONSdigital/dis-wagtail/pull/333/commits/5af6472276a5ffc4892c53b28c001065a0cb3d9d)

MaciekBaron (approved):

sanjeevz3009 (commented):
Reviewed! Had some queries with @AdamHawtin LGTM!

PR #337, opened 2025-08-01
Fix navigation tags for the temporarily removed page from ThemeLinkBlock

zerolab:

Fixes a regression in #288

### How to review

Requires a main menu that links to a theme page, created prior to #288 
Note: this is difficult to create a test for given we drop the "page" block in `ThemeLinkBlock.__init__`

### Follow-up Actions


MebinAbraham (approved):
Working as expected.

sanjeevz3009 (approved):

PR #341, opened 2025-08-04
Replace preview frame with page in charts functional test

MaciekBaron:

This PR changes the approach of the featured charts tests by using a preview page rather than relying on a preview iframe.

This is done in order to reduce flakiness as seen for example here: https://github.com/ONSdigital/dis-wagtail/actions/runs/16722800944/job/47331061044

Several things have been tested, such as waiting for the `networkidle` event, before moving to this solution.

### How to review

Ensure that functional tests pass during the headless pipeline run.

### Follow-up Actions

N/A

### DIFF
--- a/functional_tests/steps/statistical_article_page.py
+++ b/functional_tests/steps/statistical_article_page.py
@@ -570,6 +570,7 @@ def user_selects_featured_chart_preview_mode(context: Context):
     preview_mode_select.select_option(value="featured_article")
     context.page.wait_for_timeout(250)
 
+    # Open preview in new tab for more reliable testing
     browser_context = context.playwright_context
     with browser_context.expect_page() as preview_page:
         context.page.get_by_role("link", name="Preview in new tab").click()


AdamHawtin (approved):
LGTM 👍

functional_tests/steps/statistical_article_page.py
```diff
@@ -568,18 +568,22 @@ def user_selects_featured_chart_preview_mode(context: Context):
     context.page.wait_for_timeout(500)
     preview_mode_select = context.page.locator("#id_preview_mode")
     preview_mode_select.select_option(value="featured_article")
+    context.page.wait_for_timeout(250)
+
+    browser_context = context.playwright_context
+    with browser_context.expect_page() as preview_page:
+        context.page.get_by_role("link", name="Preview in new tab").click()
```
zerolab, line 575: could use a comment as to why we're doing this, orther than that LGTM

zerolab (changes requested):

functional_tests/steps/statistical_article_page.py
MaciekBaron: Added here [ae7dfa8](https://github.com/ONSdigital/dis-wagtail/pull/341/commits/ae7dfa861bbb5492936c0f606b491822d5c7841c)

PR #346, opened 2025-08-08
Dataset blocks clean up - block templates + duplicated dataset validation

kacperpONS:
# TYPO IN THE BRANCH NAME: it's **CMS-721** ‼️🚨⚠️

The main changes:
1. Align the usage of the `DatasetStoryBlock` between Topic and Release Calendar Pages. Store a formatted list of datasets to be used with `onsDocumentList` as a cached property on the page model. That is instead of using the `template` assigned to the `Struct` and `StreamBlocks`.

2. Change the validation in `DatasetStoryBlock` to check for duplicates based on the entire URL of the dataset - rather than just the `path` part of it.

3. Extract validation of ONS urls into an util function and reuse it in Dataset and Time Series link blocks. The function will enforce `https://`, and a list of allowed domains.

4. Extract URL normalisation into an util function and reuse it in Dataset and Time Series `StreamBlocks` when checking for duplicate urls. This function will  ignore the `www` and `https://`  prefixes and the trailing slash to consider functionally equivalent urls as duplicates.

5. Extract parts of `cms/core/utils` into separate modules for DS-related formatting, and URL utils.

### How to review

### Follow-up Actions


cms/datasets/blocks.py
```diff
@@ -65,35 +62,32 @@ def clean(self, value: StreamValue, ignore_required_constraints: bool = False) -
...
-            url_path = (
-                block.value.url_path if block.block_type == "dataset_lookup" else urlparse(block.value["url"]).path
-            )
-            url_paths[url_path].add(block_index)
+            if block.block_type == "dataset_lookup":
+                # For dataset lookup blocks, use the website URL
+                url = block.value.website_url
+            else:
+                url = block.value["url"].rstrip("/")  # Treat URLs with and without trailing slashes as equivalent
+            urls[url].add(block_index)
```
zerolab, line 85:
```suggestion
                url = block.value["url"]  # Treat URLs with and without trailing slashes as equivalent
            urls[.rstrip("/")].add(block_index)
```

so the check is covers manually entered and chosen dataset URLs

cms/core/formatting_utils.py
```diff
@@ -0,0 +1,115 @@
...
+    """Returns a dictionary with formatted date information for the DS document component metadata."""
+    return {
+        "prefix": prefix or _("Released"),
+        "showPrefix": True,
+        "iso": date_format(value, "c"),
+        "short": ons_date_format(value, "DATE_FORMAT"),
+    }
+
+
+def get_document_metadata(
```
sanjeevz3009, line 97: I think we are missing a unit for this function

sanjeevz3009 (commented):
Really good PR and great initiative with this, as it has a lot of improvements/ heavy refactoring, making it easy to maintain and extend in the long run 🎉

We have good unit test coverage. It would be good to have some integration tests to see if **everything is working together**. 
But then we do have UI tests cards in the backlog to test these pages in a more comprehensive manner, so not a blocker here, I guess...

Speaking of the importance of integration tests:

A couple of crucial issues I found:
**RC Page**:
- When adding manually linked datasets, as long as I enter the valid URL, I can save, and it all works
- This seems odd as the title is marked as a required field, but I can save without having the title
- If I don't have the title filled in, then the links won't render on the page; it just says `dataset`, but they do save successfully
- You can also fill in the description without the title, and it will display `dataset` and the description without the hyperlink

Same behaviour:
- Statistical Article: Related Data
- Topic Page

Some other inconsistencies I noticed were with:
- Time Series Page Link block, the way we throw the validation errors is different from manually adding linked dataset block
    - Might be one for a follow-up card, this one
- iFrame block validation again is inconsistent with the Time Series Page Block validation.
   - Might be one for a follow-up card again

cms/core/formatting_utils.py
kacperpONS: [9ba70d6](https://github.com/ONSdigital/dis-wagtail/pull/346/commits/9ba70d69f02228b53026353684b530ff357e0bb0)

kacperpONS:
@sanjeevz3009 CMS-746 🫡

nehakerung (approved):

PR #352, opened 2025-08-27
Bug: Topic Page TS section - errors not appearing on mandatory fields

kacperpONS:

A defect has been raised where custom validation for the Time Series URL is called before the `StructBlock`'s default validation that checks for the presence of mandatory fields. This results that in a scenario where all the mandatory fields are empty, the errors are displayed only for the URL field, and it's only the errors for the custom validation, but not the ones that point out that the mandatory fields are empty.

We want to change the order of the validation to ensure that all the mandatory fields are filled in and only then kick in with the custom validation for the URL. 

The expected behavior from the bug card:
> Error messages should be displayed for all sections left blank and not the URL section only

### How to review

### Follow-up Actions


cms/topics/tests/test_blocks.py
```diff
@@ -221,3 +221,54 @@ def test_time_series_page_link_block_validation_fails_on_invalid_domain(self):
...
+
+        with self.assertRaises(StructBlockValidationError) as info:
+            block.clean(value)
+
+        self.assertEqual(info.exception.block_errors["title"].message, "This field is required.")
+        self.assertEqual(info.exception.block_errors["description"].message, "This field is required.")
+        self.assertNotIn("url", info.exception.block_errors)
+
+        # Now, run validation on non-empty title and description but (the same) invalid URL
+
```
sanjeevz3009, line 257: Nit: Blank line under the comment. It would be nice to close it.

cms/topics/blocks.py
```diff
@@ -195,8 +195,10 @@ class Meta:
 
     def clean(self, value: "StructValue") -> "StructValue":
```
sanjeevz3009, line 196: Don't want to block as I will be away soon:

**Important:**
- Things are mostly _working_ as expected but there are some issues

**Issues:**
- Write something for the title field
- Leave the description blank
- If you fill in a URL such as `https://www.economy/nationalaccounts/balanceofpayments/timeseries/jzdz/mret`

When you do this an error is only thrown for the description field being blank not for the URL.
Exact same behaviour if do the above but instead fill in the description field and leave the title field blank

cms/topics/tests/test_blocks.py
```diff
@@ -221,3 +221,54 @@ def test_time_series_page_link_block_validation_fails_on_invalid_domain(self):
...
+        value = {
+            "title": "Title",
+            "description": "Description",
+            "url": "https://invalid-domain.com/time-series",
+        }
+
+        with self.assertRaises(StructBlockValidationError) as info:
+            block.clean(value)
+
+        patterns_str = " or ".join(settings.ONS_ALLOWED_LINK_DOMAINS)
```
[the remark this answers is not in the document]
kacperpONS: Yeah, that's already been there and this name is used in a few other places, good shout [1b4fb46](https://github.com/ONSdigital/dis-wagtail/pull/352/commits/1b4fb469b9a56d39ba6162677068602c6a6e489a)

cms/topics/blocks.py
sanjeevz3009: Great!

sanjeevz3009 (approved):
UI tests would be good, but we have a topic page UI card in the backlog, so will leave it out.

LGTM! Thank you for the changes!

cms/topics/blocks.py
```diff
@@ -197,9 +197,9 @@ def clean(self, value: "StructValue") -> "StructValue":
         """Checks that the given time series page URL matches the allowed domain."""
         errors = {}
 
-        for field in ("title", "description", "url"):
-            if not value.get(field):
-                errors[field] = ValidationError("This field is required.")
+        for child_block in self.child_blocks.values():
```
sanjeevz3009, line 200: Kudos to this!

cms/topics/tests/test_blocks.py
```diff
@@ -215,9 +215,62 @@ def test_time_series_page_link_block_validation_fails_on_invalid_domain(self):
         with self.assertRaises(StructBlockValidationError) as info:
             block.clean(value)
 
-        patterns_str = " or ".join(settings.ONS_ALLOWED_LINK_DOMAINS)
+        allowed_domains = " or ".join(settings.ONS_ALLOWED_LINK_DOMAINS)
```
MaciekBaron, line 218: This might be a pit nitpicky, but I think what would be better here is to do the following:
* Override the setting for `ONS_ALLOWED_LINK_DOMAINS` using a decorator
* Explicitly state the expected string on line 222

This way we also test that the code builds the error message correctly, as opposed to recreating the logic which builds the error message in the test (so "for a given value of `ONS_ALLOWED_LINK_DOMAINS`, this is the error message I expect"). Hope this makes sense!

(Also around line 258)

MaciekBaron (changes requested):

MaciekBaron (approved):
Thanks Kacper!

PR #354, opened 2025-08-29
Add feature flag for v1/resources endpoint

MaciekBaron:

This PR relates to CMS 732 and adds a feature flag for the `v1/resources` endpoint.

The feature flag defaults to false in the base settings, and defaults to true in the dev settings.

### How to review

1. In settings, set `CMS_RESOURCES_ENDPOINT_ENABLED` to true
2. Try to reach the `v1/resources` endpoint
3. Confirm that it renders correctly
4. In settings, set `CMS_RESOURCES_ENDPOINT_ENABLED` to false
5. Try to reach the `v1/resources` endpoint
6. Confirm that it returns a 404

### Follow-up Actions

- [ ] Pre-deploy, we need to add this new env var to the sandbox and staging helm chart for the internal instance.

### DIFF
--- a/cms/datavis/blocks/charts.py
+++ b/cms/datavis/blocks/charts.py
@@ -735,9 +735,9 @@ class IframeBlock(BaseVisualisationBlock):
             is_hostname_in_domain(parsed_url.hostname, allowed_domain)
             for allowed_domain in settings.IFRAME_VISUALISATION_ALLOWED_DOMAINS
         ):
-            patterns_str = " or ".join(settings.IFRAME_VISUALISATION_ALLOWED_DOMAINS)
+            allowed_domains = " or ".join(settings.IFRAME_VISUALISATION_ALLOWED_DOMAINS)
             errors["iframe_source_url"] = ValidationError(
-                f"The URL hostname is not in the list of allowed domains: {patterns_str}"
+                f"The URL hostname is not in the list of allowed domains: {allowed_domains}"
             )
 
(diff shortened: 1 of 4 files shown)


cms/search/tests/test_api_views.py
```diff
@@ -80,8 +81,22 @@ def test_resources_available_in_external_env(self):
...
         self.assertEqual(response.status_code, 200)
 
+    @override_settings(CMS_RESOURCES_ENDPOINT_ENABLED=False)
+    def test_resources_endpoint_disabled_returns_404(self):
+        """When the CMS_RESOURCES_ENDPOINT_ENABLED setting is False, the endpoint should return 404."""
+        response = self.client.get(RESOURCE_ENDPOINT)
+        self.assertEqual(response.status_code, 404)
+
+    @override_settings(CMS_RESOURCES_ENDPOINT_ENABLED=False)
+    @override_settings(IS_EXTERNAL_ENV=False)
```
sanjeevz3009, line 91: We don't need to override the `IS_EXTERNAL_ENV=False` here as we already override this on the class level.

MaciekBaron: Thanks, this was actually supposed to be `True`, which I now changed.

sanjeevz3009: Make more sense now. Resolving!

nehakerung (approved):

PR #362, opened 2025-09-09
Add `slug` field to taxonomy Topics

kacperpONS:

This PR adds a mandatory `slug` field to the taxonomy Topic model. 

Note that even though the `slug` field isn't technically enforced in the Swagger spec in [dp-topic-api](https://github.com/ONSdigital/dp-topic-api/tree/develop), after talking to Linden I was able to confirm that we're ok to treat the `slug` as if it was mandatory - it's there for all the existing Topics and they're not expecting to introduce any new ones.

The migration added in this PR introduces the mandatory field with a placeholder value, which is meant to be replaced with the `slug` value from the API when the `sync_topics` management command is run after deploying the changes. 

The fixture loading the topics has been updated and the `slug` field has been added.

### How to review

Make sure the code looks ok, the custom migration file in particular.

### Follow-up Actions
- Ensure that the `sync_topics` command is run in each environment after the deployment is complete and the new migration has been applied
- Go back to https://github.com/ONSdigital/dis-wagtail/pull/359 and finish the job

### DIFF
--- a/cms/taxonomy/tests/test_management_commands.py
+++ b/cms/taxonomy/tests/test_management_commands.py
@@ -82,6 +82,7 @@ class SyncTopicsTests(TestCase):
             # When
             call_command("sync_topics")
 
+            # Then
             self.assertEqual(logs.records[0].message, "Cannot create or update topic: missing slug.")
             self.assertEqual(logs.records[0].topic, topic.id)
 


cms/taxonomy/management/commands/sync_topics.py
```diff
@@ -103,6 +104,9 @@ def _sync_with_fetched_topics(fetched_topics: Iterable[Mapping[str, str]]) -> No
     updated_count = 0
     created_count = 0
     for fetched_topic in fetched_topics:
+        if not fetched_topic.get("slug"):
+            logger.warning("Cannot create topic: missing slug.", extra={"topic": fetched_topic["id"]})
```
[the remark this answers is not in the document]
kacperpONS: That was intentional - we wouldn't want to save (or update) a Topic that doesn't have a slug, as that would raise an `IntegrityError` because the field is mandatory on the db level.

MaciekBaron (approved):

poetry.lock
```diff
@@ -1,11 +1,12 @@
-# This file is automatically @generated by Poetry 1.8.4 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.0.1 and should not be changed by hand.
```
sanjeevz3009, line 1: Non-blocker: Make sure our poetry version matches what's in the Dockerfile which is currently `1.8.5`

sanjeevz3009 (approved):

poetry.lock
sanjeevz3009: Don't want to block here, so not requesting any changes.

sanjeevz3009, line 1: Do re-generate with the correct version of Poetry, please.
```suggestion
# This file is automatically @generated by Poetry 1.8.5 and should not be changed by hand.
```

sanjeevz3009 (changes requested):

poetry.lock
sanjeevz3009: Our Dockerfile is on `1.8.5`?

PR #368, opened 2025-09-17
Add test for previous releases breadcrumbs

AdamHawtin:

This PR relates to a bug that appears to have been fixed in other changes since it was raised, as such it doesn't change any functionality, just adds a page test to protect against regression.

The bug was for incorrect breadcrumbs on the previous releases for statistical articles, which were apparently missing a breadcrumb for the latest article. This breadcrumb is now already present, so this PR just adds a page test verifying this. 

### How to review
Check the described bug is indeed fixed, check the new test makes sense.

### Follow-up Actions
None

### DIFF
--- a/cms/articles/tests/test_pages.py
+++ b/cms/articles/tests/test_pages.py
@@ -63,7 +63,7 @@ class ArticleSeriesPageTests(WagtailPageTestCase):
 
     def test_previous_releases_breadcrumbs(self):
         """Test that the previous releases page includes a breadcrumb for the latest article."""
-        _latest_article = StatisticalArticlePageFactory(parent=self.page, title="Latest Article")
+        StatisticalArticlePageFactory(parent=self.page, title="Latest Article")
         response = self.client.get(f"{self.page.url}/editions")
         self.assertEqual(response.status_code, HTTPStatus.OK)
 


cms/articles/tests/test_pages.py
```diff
@@ -61,6 +61,19 @@ def test_previous_releases_article_list(self):
         self.assertContains(response, first_article.title)
         self.assertContains(response, second_article.title)
 
+    def test_previous_releases_breadcrumbs(self):
+        """Test that the previous releases page includes a breadcrumb for the latest article."""
+        _latest_article = StatisticalArticlePageFactory(parent=self.page, title="Latest Article")
```
MaciekBaron, line 66: NP: We have prior art where we don't assign the result to an unused variable and just rely on the side effect, so we can stick to doing this:

https://github.com/ONSdigital/dis-wagtail/blob/283b88b84cc197440fe861438f3d2e9b3d61c449/cms/articles/tests/test_models.py#L167-L175

MaciekBaron (changes requested):

cms/articles/tests/test_pages.py
AdamHawtin: Good catch, that was a lazy fix to linting complaints, fixed in https://github.com/ONSdigital/dis-wagtail/pull/368/commits/b07eca106e9092a86139f459413db92c82e314ef

MebinAbraham (approved):
We need to do this same change but actually fix the breadcrumb for the `related-data` page. It doesn't link to the latest article like here but should. Will raise a new card as there are a couple of tweaks we need to make to the related data page.

sanjeevz3009 (approved):

PR #370, opened 2025-09-18
Remove duplicate breadcrumb logic in jinja template

AdamHawtin:
The `BasePage` `get_breadcrumbs` method was intended to be used as the default method of getting the list of breadcrumbs to render for a page, but the Jinja template component was missed and still has its own (mostly duplicate) logic, which is used as the default in the page base Jinja template. This PR changes that template component to use the page's `get_breadcrumbs` method to ensure the proper default behaviour, and any page specific override is used, and ensure we're not duplicating the logic.

### How to review
This shouldn't result in any functional changes, check that the breadcrumbs for all pages are not changed.

### Follow-up Actions
None


MaciekBaron (approved):
Tests pass with no regressions, LGTM 👍

sanjeevz3009 (approved):
LGTM

PR #377, opened 2025-09-29
Update Welsh translation furniture

AWitcherONS:

Add new welsh translations into django.po 
Amended current unit tests

### How to review

Check that the unit tests run 
and that the translations are now added to web

### Follow-up Actions
lint for django.po replace "'" with "’" (speech mark apostrophe with typographic apostrophe) but not in html links  where block starts with 'msgstr'

when running unit tests in docker you need to run make make messages before


MaciekBaron (approved):

MebinAbraham (changes requested):
Could you run `make makemessages` and commit the changes, the file references in the `.po` have changed and need updating. 

We can look to add a linter in future to validate this.

cms/locale/cy/LC_MESSAGES/django.po
```diff
@@ -19,28 +19,29 @@ msgstr ""
...
+#: cms/articles/models.py:407
 #: cms/jinja2/templates/components/contact_details/contact_details.html:6
 #: cms/methodology/models.py:188 cms/release_calendar/models.py:246
 msgid "Contact details"
 msgstr "Manylion cyswllt"
 
-#: cms/articles/models.py:435
-msgid "All data related to"
-msgstr ""
+#: cms/articles/models.py:445
```
MebinAbraham, line 37: It looks like this was generated before the final changes to the `models.py` file. The lines don't match (445 here but 444 in model.py). If you run `make makemessages` again, it will fix it. Then just commit the new changes.

MebinAbraham (approved):

PR #379, opened 2025-09-30
Chore: Use Kafka as the search backend in for UI tests and local development

sanjeevz3009:

We want Kafka to be the default search backend for functional tests and for the standard local development setup. Developers can still switch to `LogPublisher` if they prefer by overriding `local.py`. We have a Kafka broker container as part of our tech-stack (originally created for integration test purposes), but it is not being actively exercised.

### How to review

- Check the `functional_test.py` and `dev.py` settings are updated accordingly.
- Spin up Wagtail -> `Publish an information page` -> `See the logs where it should show the Kafka Publisher being used` -> `you will also see the response back from the Kafka broker container`.
-   Run the UI tests and check if it's using the Kafka search back-end/ functional tests kafka broker container.

### Follow-up Actions

### DIFF
--- a/functional_tests/docker-compose-dev.yml
+++ b/functional_tests/docker-compose-dev.yml
@@ -40,8 +40,8 @@ services:
     extends:
       file: docker-compose.yml
       service: broker
-    # environment:
-    #   KAFKA_CFG_ADVERTISED_LISTENERS: 'PLAINTEXT://broker:9092,EXTERNAL://localhost:19094'
+    environment:
+      KAFKA_CFG_ADVERTISED_LISTENERS: 'PLAINTEXT://broker:9092,EXTERNAL://localhost:19094'
     ports: !override
       - '19094:9094'
 
--- a/functional_tests/docker-compose.yml
+++ b/functional_tests/docker-compose.yml
@@ -17,8 +17,8 @@ services:
     extends:
       file: ../docker-compose-dev.yml
       service: broker
-    # environment:
-    #   KAFKA_CFG_ADVERTISED_LISTENERS: 'PLAINTEXT://broker:9092,EXTERNAL://localhost:19094'
+    environment:
+      KAFKA_CFG_ADVERTISED_LISTENERS: 'PLAINTEXT://broker:9092,EXTERNAL://localhost:19094'
     ports: !override
       - '19094:9094'
 


.github/workflows/ci.yml
```diff
@@ -213,6 +213,35 @@ jobs:
       PLAYWRIGHT_BROWSER: ${{ matrix.browser }}
       PLAYWRIGHT_TRACES_DIR: playwright_traces
       PLAYWRIGHT_TRACE: 'true'
+      KAFKA_SERVERS: localhost:9092
+
+    services:
+      broker:
+        image: bitnamilegacy/kafka:3.5.1
+        env:
```
MebinAbraham, line 221: We don't want a broker service here, we want it to be part of the compose stack. You should add the broker override to compose file:
- https://github.com/ONSdigital/dis-wagtail/blob/chore/use-kafka-backend-in-tests-dev-CMS-757/functional_tests/docker-compose.yml
- https://github.com/ONSdigital/dis-wagtail/blob/chore/use-kafka-backend-in-tests-dev-CMS-757/functional_tests/docker-compose-dev.yml

sanjeevz3009: This came to my mind of adding it to a compose file, then I somehow realised we actually have dedicated Docker-Compose files for `functional_tests`. Mind was blurry. Will get on it and update the comment above.

sanjeevz3009: I am such an idiot sandwich. Reading the makefile `functional-tests-up` and it being used in the CI step ` run: make functional-tests` should of gave this away.

sanjeevz3009: [10cace9](https://github.com/ONSdigital/dis-wagtail/pull/379/commits/10cace9b8d67c3ab9afea4fb6c198301b3285b55)

cms/settings/functional_test.py
```diff
@@ -1,5 +1,11 @@
+import os
+
```
MaciekBaron, line 2: NP: We usually do this
```python
env = os.environ.copy()
...
env.get("KAFKA_SERVERS", "localhost:19094")
```

MaciekBaron (changes requested):
Functional tests still seem to be broken after the change, will need updating.

MaciekBaron (approved):
LGTM 👍

PR #381, opened 2025-10-02
docs(streamfield): document orphaned data behaviour

MebinAbraham:

Adds a new section to the StreamField documentation explaining what happens when blocks are removed from models using the custom cms.core.fields.StreamField implementation.

### How to review

Does it make sense, is it accurate?

### Follow-up Actions


AWitcherONS (approved):
Understandable and as complete with current knowledge and requirements

MaciekBaron (approved):

PR #445, opened 2025-12-15
feat(perf): scripts to create test data

RealOrangeOne:

We need a way to create test data in bulk for performance testing etc.

https://officefornationalstatistics.atlassian.net/browse/CMS-948

Data creation is seeded, meaning it can be recreated reliably.

Data is identified by a prefix. To save complexity, it's only possible to delete all test data, rather than specific seeds.

### How to review

There are 3 new management commands. 

- `create_test_data` - creates data based on a seed.
- `delete_test_data` - removes all test data from the system (based on the prefix)
- `show_default_test_data_config` - displays the default test data generation config in the terminal for guidance

### Follow-up Actions

- Expand to cover more model types


MaciekBaron (changes requested):

cms/test_data/config.py
```diff
@@ -0,0 +1,47 @@
...
+        return self
+
+
+class ModelCreationConfig(BaseModel):
+    count: NonNegativeInt | RangeConfig = 1
+
+
+class PageCreationConfig(ModelCreationConfig):
+    published: FractionalFloat = 0.5
+    revisions: NonNegativeInt | RangeConfig = 0
```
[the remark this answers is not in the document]
RealOrangeOne: TIL. Thanks for looking into this. Yeah I agree this should be tweaked.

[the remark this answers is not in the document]
MaciekBaron: But that's precisely my point: when using the CMS you can't create a page with no revisions.

cms/test_data/management/commands/create_test_data.py
kevinhowbrook, line 1: **Suggestion**

Would be nice to have the commands in the `make` file

cms/test_data/management/commands/delete_test_data.py
kevinhowbrook, line 1: **Suggestion**

Would be nice to have the commands in the make file

cms/test_data/management/commands/create_test_data.py
```diff
@@ -0,0 +1,181 @@
...
+
+            for _ in range(self.get_config_count(self.config.topics.revisions)):
+                topic_page.specific.save_revision()
+
+            if (
+                not topic_page.live
+                and topic_page.latest_revision_id
+                and self.faker.boolean(int(self.config.topics.published_probability * 100))
+            ):
+                topic_page.specific.latest_revision.publish()
```
kevinhowbrook: **Issue**

Does this actually execute? TopicPageFactory inherits from wagtail_factories.PageFactory, which creates pages as live=True by default. So not topic_page.live is always False

cms/test_data/tests/test_commands.py
```diff
@@ -0,0 +1,208 @@
...
+from django.test import TestCase
+from django.test.testcases import SimpleTestCase
+from modelsearch.index import class_is_indexed
+from wagtail.models import ReferenceIndex
+
+from cms.datasets.models import Dataset
+from cms.images.models import CustomImage
+from cms.taxonomy.models import Topic
+from cms.test_data.config import TestDataConfig
+from cms.test_data.management.commands.create_test_data import SEEDED_DATA_PREFIX
```
kevinhowbrook: **Issue**

Looks like this is in `cms.test_data.constants`  now

kevinhowbrook (changes requested):
Had to merge in `main` as best as I could locally to test, but both commands work well. It would be nice to have the commands in the Makefile, nice one @RealOrangeOne 

I think the larger task here is rebasing or merging main but I'd love to see this go in then we can build it out further if we need/want to

cms/test_data/management/commands/create_test_data.py
```diff
@@ -0,0 +1,181 @@
...
+        parser.add_argument(
+            "--seed", nargs="?", default=4, type=int, help="Random seed to produce deterministic output"
+        )
+        parser.add_argument("--config", type=validate_config_file, default=TestDataConfig(), help="Config file")
+        parser.add_argument(
+            "--noinput",
+            "--no-input",
+            action="store_false",
+            dest="interactive",
+            help="Tells Django to NOT prompt the user for input of any kind.",
```
kevinhowbrook, line 171:
```suggestion
            help="Tells Django NOT to prompt the user for input of any kind.",
```

RealOrangeOne: This is copied from the help text of `migrate`.

cms/test_data/management/commands/delete_test_data.py
```diff
@@ -0,0 +1,147 @@
...
+        for model in models:
+            if class_is_indexed(model):
+                post_save.connect(post_save_signal_handler, sender=model)
+
+            if ReferenceIndex.model_is_indexable(model):
+                post_save.connect(update_reference_index_on_save, sender=model)
+
+
+class Command(BaseCommand):
+    help = "Delete all random test data"
```
kevinhowbrook:
```suggestion
    help = "Delete all random test data generate by create_test_data"
```

cms/test_data/management/commands/create_test_data.py
RealOrangeOne: The problem is you need to be able to pass in arguments to `create_test_data`, which isn't supported when calling through a `Makefile` (easily). Leaving it as a management command to run is probably fine, and not much extra. If we commit a standard config, then absolutely it should be.

cms/test_data/management/commands/delete_test_data.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/445/commits/6be15754a7778b6a75afb9c50ad413d70d128c1e

cms/test_data/tests/test_commands.py
RealOrangeOne: D'oh! https://github.com/ONSdigital/dis-wagtail/pull/445/commits/36b035cfe11cbf871be47c012b15c984b2b9c848

cms/test_data/management/commands/create_test_data.py
RealOrangeOne: No, no it doesn't :facepalm: https://github.com/ONSdigital/dis-wagtail/pull/445/commits/657ac3a39aedbcde89020dde311c8fb274b14004

cms/test_data/config.py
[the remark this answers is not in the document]
RealOrangeOne: Oh right :facepalm: https://github.com/ONSdigital/dis-wagtail/pull/445/commits/a3a1f5cf869eff12fd087465f036371bfa20a94e

cms/test_data/management/commands/delete_test_data.py
```diff
@@ -0,0 +1,147 @@
...
+        self.stdout.write("Deleting data...", self.style.NOTICE)
+
+        with disable_signals(list(collector.data.keys())), transaction.atomic():
+            for model, instances in collector.data.items():
+                # Use queryset delete methods to use any customized behaviour
+                model._default_manager.filter(pk__in=[instance.pk for instance in instances]).delete()  # pylint: disable=protected-access
+
+        # NB: Because the topic tree hides the root node in a number core method, it gets corrupted
+        # during deletion. Manually repair the tree afterwards.
+        Topic.fix_tree()
```
BJacksonONS: Think most of what I've noticed has already been addressed, but was wondering about this. Might be something to follow up on later rather than do here.

Would we be better off doing something to keep the tree clean and avoiding `fix_tree()`? If this is potentially being run against existing environments with data, this might scale badly as I think it rebuilds everything. Maybe it's not a big issue though.

BJacksonONS: Could use a second manager or flag in `get_queryset` to toggle the custom behaviour.

RealOrangeOne: The problem is that the root node is hidden in `get_queryset`. I tried removing that, but it broke a lot of other behaviour. Fixing the tree like this felt like the easiest solution. Especially since it'll only be "broken" for a small amount of time, only affect topics, and won't ever be run in an environment where the data matters.

With that said, the comment is clearly nonsense, so I'll tweak that.

RealOrangeOne: I've moved the tree fixing inside the delete transaction, which should prevent the "broken" tree from being seen by other connections.

https://github.com/ONSdigital/dis-wagtail/pull/445/commits/436b3997e870169100c41fdbdc3440d7af942603

BJacksonONS: I was wondering if we could do something like

```python
from asgiref.local import Local

_disable_topic_root_filtering = Local()

class TopicManager(MP_NodeManager):
    def get_queryset(self) -> TreeQuerySet:
        """Filter out the dummy root topic from all querysets unless disabled."""
        qs = TreeQuerySet(self.model).order_by("path")
        if getattr(_disable_topic_root_filtering, "value", False):
            return qs
        return qs.filter(depth__gt=1)

class Topic():
    ....
    @classmethod
    @contextmanager
    def disable_root_filtering(cls) -> Generator[None, None, None]:
        previous = getattr(_disable_topic_root_filtering, "value", False)
        _disable_topic_root_filtering.value = True
        try:
            yield
        finally:
            _disable_topic_root_filtering.value = previous
```

Then when we're doing the deletions we can just use the context manager

```python
with disable_signals(list(collector.data.keys())), transaction.atomic(), Topic.disable_root_filtering():
```

Should work correctly both in a standalone cron job or if it was called in a server process.

cms/taxonomy/tests/factories.py
```diff
@@ -1,20 +1,32 @@
...
         model = Topic
-        strategy = factory.BUILD_STRATEGY  # To prevent Factory from trying to save the node
+        django_get_or_create = ("id", "parent")
 
-    id: str = faker.Faker("random_number", digits=50, fix_len=True)  # Use a very long ID for uniqueness
-    slug: str = faker.Faker("slug")
-    title: str = faker.Faker("sentence", nb_words=3)
-    description: str = faker.Faker("sentence", nb_words=10)
+    parent = factory.LazyFunction(Topic.objects.root_topic)
+    id = factory.Faker("random_number", digits=50, fix_len=True)  # Use a very long ID for uniqueness
```
MaciekBaron, line 13: Question: Why not `factory.Faker("uuid4")`, would that work?

RealOrangeOne: That would also work, but this hasn't been changed from how it was before

cms/test_data/management/commands/delete_test_data.py
RealOrangeOne: That should work, and feels less risky than a broken tree. I'll see if it works.

RealOrangeOne: It seems even that results in a broken tree - the root node has `numchild` which doesn't match its actual children.

BJacksonONS: Are we creating items under it while the root is hidden? Maybe there are some writes we're doing that could also be wrapped in the context manager that would give a correct `numchild`?

RealOrangeOne: I tried wrapping the entire command in the decorator - same issue.

BJacksonONS: Fair, the change in https://github.com/ONSdigital/dis-wagtail/commit/436b3997e870169100c41fdbdc3440d7af942603 probably makes the most sense then.

pyproject.toml
```diff
@@ -47,6 +47,8 @@ wagtailtables = "^0.2.2"
 pyjwt = "^2.13.0"
 cryptography = "^48.0.1"
 django-pglock = "^1.8.0"
+wagtail-factories = "^4.2.1"
```
BJacksonONS: We still have `wagtail-factories` in dev deps pinned at a higher version ("^4.3.0"). Should clean that up and maybe bump this one to match if we're going to have it in the main group now.

BJacksonONS:
Was just chatting to @MebinAbraham and thought it would be good to add some kind of env var flag so we can disable the command in prod easily.

Given this promotes some dev dependencies to regular ones and unconditionally installs, would provide some peace of mind against human error when running tests later.

BJacksonONS:
Currently getting 

```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "taxonomy_topic_path_key"
DETAIL:  Key (path)=(00010001) already exists.
```
When running this against a freshly created db migrated with `make dev-init`, trying to determine why.

BJacksonONS:
> Currently getting
> 
> ```
> django.db.utils.IntegrityError: duplicate key value violates unique constraint "taxonomy_topic_path_key"
> DETAIL:  Key (path)=(00010001) already exists.
> ```
> 
> When running this against a freshly created db migrated with `make dev-init`, trying to determine why.

I think `make load-topics` is leaving treebeard in a broken state, the create command works fine if I run against a fresh db that has only had migrations run and no topic load.

This needs to run against other environments that have topics loaded, this might be an issue then if we aren't 100% on the cause.

cms/test_data/management/commands/create_test_data.py
```diff
@@ -0,0 +1,178 @@
...
+        return created
+
+    def create_from_factory(self, factory_class: type[Factory], factory_kwargs: dict) -> Model:
+        """Create a model instance from a given factory."""
+        return self.create_batch_from_factory(factory_class, 1, factory_kwargs)[0]
+
+    def random_model(self, model: type[Model]) -> Model:
+        """Select a random previously-generated model."""
+        try:
+            return self.faker.random_element(self.model_registry[model])
```
BJacksonONS: We get an error on this line if any of the model configs have a count of "0".

cms/test_data/management/commands/create_test_data.py
```diff
@@ -0,0 +1,178 @@
...
+    try:
+        return TestDataConfig.model_validate_json(config_path.read_text())
+    except ValidationError as e:
+        raise ArgumentTypeError(str(e)) from e
+
+
+class Command(BaseCommand):
+    help = "Create random test data"
+
+    def add_arguments(self, parser: ArgumentParser) -> None:
```
BJacksonONS, line 163: An edge case that probably won't matter, but you can pass a blank seed var to get a "None" seed, which means no user input can match and the command fails.

Image

RealOrangeOne:
>  thought it would be good to add some kind of env var flag so we can disable the command in prod easily

It might be, although there are plenty of other commands which cause issues which aren't disabled - `reset_db` for example.

cms/test_data/management/commands/create_test_data.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/445/commits/a4cf296cc86ec799802ee4ff7f45ac36a9839693

cms/test_data/management/commands/create_test_data.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/445/commits/e9d0a37cdcfb29613d4d24cb18234ff1a3d663d9

pyproject.toml
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/445/commits/97c1f9ed5dde7e646996e57340fb0e27d0159a7a

cms/test_data/random.py
```diff
@@ -0,0 +1,28 @@
...
+import factory
+import factory.random
+
+
+def get_default_locale() -> str:
+    return str(factory.Faker._DEFAULT_LOCALE)  # pylint: disable=protected-access
+
+
+def _set_default_locale(locale: str) -> None:
+    factory.Faker._DEFAULT_LOCALE = locale  # pylint: disable=protected-access
```
MaciekBaron, line 13: **Non-blocking, minor:** Maybe we could add `faker` to the name (`get_default_faker_locale` / `_set_default_faker_locale`) just for better clarity? Code completion might use this by accident for other purposes.

cms/test_data/constants.py
```diff
@@ -0,0 +1 @@
+SEEDED_DATA_PREFIX = "Z-RANDOM "
```
MaciekBaron: **Non-blocking, minor:** Perhaps we should move this to settings so it is configurable? In case anyone has a different preference for the generated data.

MaciekBaron (approved):
LGTM, just two non-blocking comments.

cms/test_data/random.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/445/commits/27a9e1ff350ec57fa653f380ec63f00e616a6c23

cms/test_data/constants.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/445/commits/6aa3e6c9c94b909d076d9be5fa93e2f5e54853b7

kevinhowbrook (approved):
lgtm, only spot is that we could add `DEFAULT_LOCALE = "en_GB"` but it's not a blocker and potentially not worth it at this point.
Great work picking this up and getting it over the line 🥳

MebinAbraham:

cms/topics/tests/factories.py
```diff
@@ -5,16 +5,47 @@
...
+
+
 class TopicPageFactory(wagtail_factories.PageFactory):
     """Factory for TopicPage."""
 
     class Meta:
         model = TopicPage
+        django_get_or_create = ("title", "parent")
 
     parent = factory.LazyFunction(lambda: HomePage.objects.first())  # pylint: disable=unnecessary-lambda
```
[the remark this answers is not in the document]
BJacksonONS: Addressed here

https://github.com/ONSdigital/dis-wagtail/pull/445/commits/9a0a0047574d7d6394330991b04fda9dff21c229

cms/test_data/management/commands/create_test_data.py
```diff
@@ -0,0 +1,196 @@
...
+        self.title_factory = factory.LazyFunction(self.random_title)  # type: ignore[no-untyped-call]
+
+        self.root_page = Site.objects.get(is_default_site=True).root_page
+
+        self.model_registry: dict[type[Model], list[Model]] = {}
+
+        self.get_config_count = partial(self.config.get_count, faker=self.faker)
+
+    def random_title(self) -> str:
+        return settings.CMS_TEST_DATA_PREFIX + self.faker.sentence(nb_words=3)
```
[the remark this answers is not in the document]
BJacksonONS: Addressed here

https://github.com/ONSdigital/dis-wagtail/pull/445/commits/005028b8d5b04340ecc55af31f3d8216befb6605

cms/taxonomy/models.py
```diff
@@ -17,15 +18,16 @@
...
 
 
-class TopicManager(models.Manager):
-    def get_queryset(self) -> QuerySet:
+class TopicManager(MP_NodeManager):
+    def get_queryset(self) -> TreeQuerySet:
         """Filter out the dummy root topic from all querysets."""
-        return super().get_queryset().filter(depth__gt=1)
+        # Reuse Wagtail's custom tree QuerySet for helpful utils
+        return TreeQuerySet(self.model).order_by("path").filter(depth__gt=1)
```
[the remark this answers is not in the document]
BJacksonONS: Not sure this has much impact, have fixed the _db and _hints via kwargs on TreeQuerySet.

https://github.com/ONSdigital/dis-wagtail/pull/445/commits/ba5096619f2ea23e948a025565828d9ef2d105bc

BJacksonONS:
Trying a rereview from copilot. Will see if it resolves it's own previous comments.

cms/test_data/management/commands/delete_test_data.py
```diff
@@ -0,0 +1,129 @@
...
+            action="store_true",
+            help="Just show the models to delete - don't delete anything",
+        )
+
+    def _get_lookups(self, model: type[Model]) -> Q:
+        """Query any field which might contain the prefix."""
+        lookups = Q()
+        for field in model._meta.fields:
+            if field.is_relation or not isinstance(field, CharField | TextField):
+                continue
```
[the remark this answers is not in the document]
BJacksonONS: This is valid as of python 3.10 and we're on 3.14. Would have caused failures in tests if this was an issue.

MebinAbraham (commented):
We need to document these. Let's add instructions to README.md.

Can we also follow up with a doc under docs/ for this.

cms/test_data/management/commands/create_test_data.py
```diff
@@ -0,0 +1,198 @@
+import itertools
```
MebinAbraham, line 1: Let's add these to Makefile to easy invocation.

sanjeevz3009 (approved):
LGTM!

cms/test_data/constants.py
sanjeevz3009: Resolving on behalf!

cms/test_data/random.py
sanjeevz3009: Resolving on behalf!

cms/taxonomy/tests/factories.py
sanjeevz3009: Resolving on behalf!

sanjeevz3009:
Apart from @MebinAbraham comment here: https://github.com/ONSdigital/dis-wagtail/pull/445#discussion_r3719830654, the branch needs updating with main, and the `poetry.lock` file merge conflicts need resolving.

This PR is good to go in.

.pylintrc
```diff
@@ -431,6 +431,7 @@ disable=raw-checker-failed,
         missing-function-docstring,
         fixme,
         too-many-public-methods,
+        cyclic-import,
```
MebinAbraham, line 434: 👀 👀 👀 👀 👀 👀 😄

BJacksonONS: That's from main, was introduced here

https://github.com/ONSdigital/dis-wagtail/pull/676/changes#diff-ba293004ae8f18975051e825bd8d5c70137a8b5d42d16e0f7c9354aa634c446cR434

cms/test_data/management/commands/create_test_data.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/445/commits/a85e58ebd2e79f312583395beb52bdc42af26996

Makefile
```diff
@@ -115,6 +115,25 @@ load-design-system-templates:  ## Load the design system templates
...
+.PHONY: delete-test-data-dry-run
+delete-test-data-dry-run:  ## Dry run for deleting test data
+	poetry run python ./manage.py delete_test_data --dry-run
+
+.PHONY: delete-test-data
+delete-test-data:  ## Delete test data. Use NOINPUT=1 to skip confirmation prompt.
+	poetry run python ./manage.py delete_test_data $(if $(NOINPUT), --noinput)
+
+.PHONY: show-default-test-data-config
+delete-test-data:  ## Show default config. Use SCHEMA=1 to show JSON schema.
```
MebinAbraham: How about prefixing `test-data` for easier completions? Thoughts?

MebinAbraham: Wrong command key, should be `show-default-test-data-config`.

BJacksonONS: Ooops, silly copy paste 🤦

https://github.com/ONSdigital/dis-wagtail/pull/445/commits/5739bbd7c7962dbc677ee3dfa1bd383d492db10a

BJacksonONS: Yeah I like that

https://github.com/ONSdigital/dis-wagtail/pull/445/commits/5739bbd7c7962dbc677ee3dfa1bd383d492db10a

BJacksonONS: My only other thought on that is maybe making it harder to accidentally hit the delete command rather than the dry run, but it requires user input to follow through by default anyway.

MebinAbraham: Ya, since it needs follow-up prompts, I think that's okay.

MebinAbraham (approved):

.pylintrc
MebinAbraham: We'll come back to this.

Makefile
```diff
@@ -115,6 +115,25 @@ load-design-system-templates:  ## Load the design system templates
 load-topics:  ## Load our fixture of taxonomy topics
 	poetry run python ./manage.py loaddata cms/taxonomy/fixtures/topics.json
 
+.PHONY: test-data-create
+test-data-create:  ## Seed test data
+	poetry run python ./manage.py create_test_data \
+		$(if $(SEED), --seed $(SEED)) \
+		$(if $(CONFIG), --config $(CONFIG)) \
+		$(if $(NOINPUT), --config) \
```
MebinAbraham:
```suggestion
		$(if $(NOINPUT), --noinput) \
```

sanjeevz3009 (approved):
LGTM!

sanjeevz3009 (approved):

PR #446, opened 2025-12-18
Expand activity logs

MaciekBaron:

This PR addresses ticket CMS 906 and expands the activity logs to cover more actions and content types.

~Note that this change introduces a new dependency, namely `django-crum`, in order to expose user details when using signals.~

### How to review

Perform one of the following actions and confirm that the activity was logged in site history:
- Edit a bundle
- Add a page to a bundle
- Remove a page from a bundle
- Add a team to a bundle
- Remove a team from a bundle
- Change bundle schedule
- Add a dataset
- Remove a dataset
- ~Add/edit/remove an image~
- ~Add/edit/update a document~
- Download a CSV from a preview
- Inspect a bundle
- Preview a bundle
- Preview any page
- Open a page for editting
- Open a bundle for editing

### Follow-up Actions

~Confirm that using the `django-crum` module is allowed for the project.~


pyproject.toml
```diff
@@ -49,6 +49,7 @@ pyjwt = "^2.10.1"
 cryptography = "^45.0.4"
 django-pglock = "^1.7.2"
 wagtail-schema-org = "^4.1.1"
+django-crum = "^0.7.9"
```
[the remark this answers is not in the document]
MaciekBaron: Currently, images and documents do create log entries, but only in specific scenarios - specifically, when using the non-multi version of the forms.

This means that the form at `/admin/images/multiple/add/` does not create a log entry when adding an image. Meanwhile, `/admin/images/add/` will (albeit you can only upload one image there). The default form you are taken to is the multi one.

Likewise if you mass delete images, that won't trigger the deletion log entries. It will however create the entries if you delete one image using the delete dropdown option.

I have removed `django-crud` for now and added tests to show the existing behaviour in: [58c6bb1](https://github.com/ONSdigital/dis-wagtail/pull/446/commits/58c6bb1690f4bb3fae92a2fb6e256433067200f2)

I have consulted this with the Wagtail team and they seem to believe that the fact that the log entries are created in the first place is because of a [refactor](https://github.com/wagtail/wagtail/commit/ed2f0c47206e1253b6db7091b055592111f1cbdb) done in 2014, following the work done on https://github.com/wagtail/wagtail/pull/7550 and performing a full audit log of all image/document actions wasn't part of a feature plan.

We have several things we could do here:
* Drop `django-crud` and instead write our own thread-local approach which we maintain (shouldn't be too complicated code-wide)
* Raise an issue in the Wagtail repo to look into implementing full logging for images and documents, so that we don't just have a partial implementation
* Force users to go through a single image upload form, and disable mass deletion of images via the admin panel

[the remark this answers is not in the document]
MebinAbraham: As discussed, if we have no other strong reason to add support for it apart from the multi-image/document upload path, we can consider disabling that functionality and revisit it in the future.

MebinAbraham (commented):
Have done an initial pass (not checked all events we need to log yet), and overall it looks good.

One key gap is the need for a ChooserView mixin or similar mechanism. We should ensure that any chooser displaying sensitive information elsewhere in the system is logged. While I don't have a comprehensive list immediately available, some areas that require attention include:
- Logging chooser activity for headline figures. For example, when headline figures are listed in a chooser on a Topic page, we should log an action since users are viewing unpublished values.
- Listing of images or documents, since their titles may contain sensitive information.

We should also review other choosers to identify any additional instances where sensitive information may be exposed and ensure appropriate auditing is in place.

cms/articles/views.py
```diff
@@ -103,4 +104,17 @@ def get(self, request: HttpRequest, page_id: int, revision_id: int, chart_id: st
...
 
-        return create_data_csv_download_response_from_data(data, title=chart_data.get("title", "chart"))
+        chart_title = chart_data.get("title")
+
+        # Log the chart download for audit trail
+        log(
+            action="content.chart_download",
+            instance=page,
+            data={
+                "chart_title": chart_title,
```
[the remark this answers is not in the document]
MaciekBaron: Chart title removed here: b7123912753dd5

AdamHawtin (approved):
LGTM, I can see the logs in both the site history and applications logs for all the actions I could test, with the expected data 👍

cms/articles/views.py
```diff
@@ -103,4 +104,17 @@ def get(self, request: HttpRequest, page_id: int, revision_id: int, chart_id: st
             )
             raise Http404 from e
 
-        return create_data_csv_download_response_from_data(data, title=chart_data.get("title", "chart"))
+        chart_title = chart_data.get("title")
+
+        # Log the chart download for audit trail
+        log(
```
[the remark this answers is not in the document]
MebinAbraham: Have checked, we don't need the request object as long as we have enough context for SoC to action.

pyproject.toml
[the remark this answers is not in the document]
MaciekBaron: Just for context/clarity, `django-crum` is no longer being used.

cms/core/wagtail_hooks.py
```diff
@@ -91,14 +91,8 @@ class ChartDownload(LogFormatter):  # pylint: disable=unused-variable
...
             """Returns the formatted log message."""
             try:
-                chart_title = log_entry.data.get("chart_title")
                 chart_id = log_entry.data.get("chart_id", "unknown")
-                message = "Downloaded chart CSV for chart "
-
-                if chart_title:
-                    message += chart_title
-                else:
-                    message += f"with ID {chart_id}"
```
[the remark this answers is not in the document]
MebinAbraham: Sounds sensible 👍

BJacksonONS (commented):
Currently having some issues with page preview logs, have discussed with @MaciekBaron and he's going to add some tests and investigate.

kevinhowbrook (commented):
Local testing notes here - (apologies in advance, not being able to add datasets and teams is likely down to my familiarity of the codebase so far)

- Edit a bundle - logged (as `edited`)
- Add a page to a bundle - logged but only when editing the bundle directly (see notes) logged as `Changed pages in bundle Added: Index page`
- Remove a page from a bundle - logged (as `Changed pages in bundle Removed: Index page`)
- Add a team to a bundle - (can't add a team)
- Remove a team from a bundle - (can't add a team)
- Change bundle schedule - logged as ` Changed publication date from '2026-05-31 10:51' to '2026-05-30 11:51' `
- Add a dataset - (locally adding a dataset raises a 404:  404 Client Error: Not Found for url: https://api.beta.ons.gov.uk/v1/dataset-editions?published=false&limit=1) not sure if we have stubs for this?
- Remove a dataset - n/a (cannot add one)
- Add/edit/remove an image - not logged
- Add/edit/update a document - not logged
- Download a CSV from a preview - can't see where to add CVs (assuming this means a preview of a page too)
- Inspect a bundle - logged (as `Viewed bundle details `)
- Preview a bundle - (can't preview an actual bundle, but previewing pages is logged as action  `Previewed page in mode:` )
- Preview any page logged as ` Previewed page in mode: ` - only logging on Index Page and Release Calander Index
- Open a page for editing - logged as `Viewed page editor` 
- Open a bundle for editing - logged as  `Viewed bundle editor`

cms/articles/models.py
```diff
@@ -658,6 +658,7 @@ def preview_modes(self) -> list[tuple[str, str]]:
         ]
 
     def serve_preview(self, request: HttpRequest, mode_name: str) -> TemplateResponse:
+        self._log_preview(request, mode_name)
```
BJacksonONS, line 687: `_log_preview` is called by the parent `serve_preview` when we call up on line 670, so this double logs.

cms/core/models/base.py
```diff
@@ -320,6 +325,28 @@ def _get_site_root_paths(self, request: HttpRequest | models.Model | None = None
...
+
+        log(
+            action="pages.preview_mode_used",
+            instance=self,
+            data={"preview_mode": mode_name},
+            user=request.user if request.user.is_authenticated else None,
+        )
+        cache.set(cache_key, True, timeout=self.audit_log_cooldown_seconds)
+
+    def serve_preview(self, request: HttpRequest, mode_name: str) -> TemplateResponse:
```
BJacksonONS, line 362: In my testing this never seems to get called for previews on InformationPages, so we never get a log. Trying to investigate why.

cms/articles/models.py
BJacksonONS: Nvm, double logging was caused by preview inside the edit page tab.

cms/core/models/base.py
BJacksonONS: Seems like this might be an inheritance issue, stepping through with a debugger shows we're calling `serve_preview` from the `RoutablePageMixin` rather than the base page.

kevinhowbrook: I had a look at this too as I was curious... I see a few ways to handle it so just adding my own thoughts.

One option would be to add a dedicated PreviewLoggingMixin embedded in NoTrailingSlashRoutablePageMixin. (Pages that don't inherit NoTrailingSlashRoutablePageMixin work fine)
The only job of a new mixin is calling _log_preview then super().serve_preview(), (placed before RoutablePageMixin in NoTrailingSlashRoutablePageMixin's inheritance

```
NoTrailingSlashRoutablePageMixin(PreviewLoggingMixin, RoutablePageMixin). 
```

Logging hits once in the right MRO position; I'm assuming cache handles the second call from BasePage though...

That being said, I prefer to avoid inheritance like that because it's just moving a problem elsewhere. Could be an unpopular opinion but I think specifically overriding `serve_preview` on each page model is better here - it's more specific and targeted and easier to read. But, we would need to remember to add it to future page models (or via an __init_subclass__ guard)

cms/core/tests/test_preview_audit.py
kevinhowbrook, line 1: **Suggestion**

related to [MRO comment raised](https://github.com/ONSdigital/dis-wagtail/pull/446#discussion_r3172861921)

We are only tests StatisticalArticlePage, which is actually one of the pages that works correctly because it explicitly overrides serve_preview.

Should we cover at least one representative of each inheritance pattern?  plain BasePage subclass, a routable page via NoTrailingSlashRoutablePageMixin, and a page with custom preview modes to ensure serve_preview triggers logging regardless of MRO.

cms/core/models/base.py
MaciekBaron: Thanks both, that's a good spot.

For the time being I added a specific override on the two models.

I also created a test to check for regressions. [ed6a30f](https://github.com/ONSdigital/dis-wagtail/pull/446/commits/ed6a30f3e0e3c755c5b69d9211b851f75940a5ae)

cms/core/tests/test_preview_audit.py
MaciekBaron: Added a test here: [ed6a30f](https://github.com/ONSdigital/dis-wagtail/pull/446/commits/ed6a30f3e0e3c755c5b69d9211b851f75940a5ae)

MaciekBaron:
> Local testing notes here - (apologies in advance, not being able to add datasets and teams is likely down to my familiarity of the codebase so far)
> 
> (...)

Thanks @kevinhowbrook - I have addressed one of the issues raised in your testing here [4c1f7a9](https://github.com/ONSdigital/dis-wagtail/pull/446/commits/4c1f7a9107ebb16096d7d175ed984c605280f6e0)

`Previewed page in mode: ` would show no mode when it was an empty string, so I made it default to `default`.

Regarding images and documents, this is a core Wagtail issue: the single file form correctly creates the log entry, but the default multifile form doesn't. This was discussed here: https://github.com/ONSdigital/dis-wagtail/pull/446#discussion_r2691277613

I have crossed those two things out, to avoid confusion.

BJacksonONS (approved):
LGTM

kevinhowbrook (approved):

cms/core/tests/test_preview_audit.py
```diff
@@ -211,3 +220,34 @@ def test_log_preview_does_not_log_unsaved_pages(self) -> None:
...
+            HomePage.objects.first(),  # Guaranteed to exist
+            InformationPageFactory.create(),
+            MethodologyPageFactory.create(),
+            TopicPageFactory.create(),
+            ThemePageFactory.create(),
+            ReleaseCalendarIndex.objects.first(),  # Guaranteed to exist
+            ReleaseCalendarPageFactory.create(),
+            StatisticalArticlePageFactory.create(),
+            CookiesPage.objects.first(),  # Guaranteed to exist
+        )
```
MebinAbraham, line 245: It would be good if we could do this dynamically for our Page types. What we want is, for this test to fail even if a new page is added but didnt log preview. We can have an exclusion for edge cases.

Currently, we are at the mercy of remembering to extend this each time.

cms/core/wagtail_hooks.py
```diff
@@ -60,6 +65,18 @@ def global_admin_css() -> str:
...
+@hooks.register("before_edit_page")
+def log_page_edit_view(request: HttpRequest, page: Page) -> None:
+    """Log when a user views the page edit view, with a cooldown to prevent duplicate entries."""
+    cache_key = f"page_edit_view_log:{page.pk}:{request.user.pk}"
+
+    if cache.get(cache_key):
+        return
+
+    log(action="pages.edit_view", instance=page)
+    cache.set(cache_key, True, timeout=PAGE_EDIT_VIEW_AUDIT_LOG_COOLDOWN_SECONDS)
```
MebinAbraham: Waiting on confirmation whether this is acceptable (cooldown).

cms/bundles/wagtail_hooks.py
```diff
@@ -312,3 +382,72 @@ def format_message(self, log_entry: ModelLogEntry) -> Any:
...
+
+    @actions.register_action("bundles.inspect")
+    class InspectBundle(LogFormatter):  # pylint: disable=unused-variable
+        """LogFormatter class for viewing the bundle inspect page."""
+
+        label = "View bundle details"
+
+        def format_message(self, log_entry: ModelLogEntry) -> Any:
+            """Returns the formatted log message."""
+            return "Viewed bundle details"
```
MebinAbraham: This could be interpreted as the bundle edit page. Since the page and buttons are called inspect, let's stick with it. I suggest: `Viewed bundle inspect page`

cms/bundles/wagtail_hooks.py
```diff
@@ -312,3 +382,72 @@ def format_message(self, log_entry: ModelLogEntry) -> Any:
...
+        """LogFormatter class for bundle publication date changes."""
+
+        label = "Change bundle schedule"
+
+        def format_message(self, log_entry: ModelLogEntry) -> Any:
+            """Returns the formatted log message."""
+            try:
+                old = log_entry.data.get("old", "Not set")
+                new = log_entry.data.get("new", "Not set")
+                return f"Changed publication date from '{old}' to '{new}'"
```
MebinAbraham, line 441: This seems to come out as `Changed publication date from 'x' to 'None'`. Maybe this should be:
```py
log_entry.data.get("new") or "Not set"
```

Same for old.

cms/bundles/viewsets/bundle.py
```diff
@@ -149,6 +132,7 @@ class BundleEditView(EditView):
     template_name = "bundles/wagtailadmin/edit.html"
     has_content_changes: bool = False
     start_time: float | None = None
+    audit_log_cooldown_seconds = 30
```
MebinAbraham: Ideally, this comes from a single setting

cms/bundles/utils.py
```diff
@@ -401,13 +401,6 @@ def publish_bundle(bundle: Bundle, *, update_status: bool = True) -> bool:
     Returns:
         True if all pages published successfully, False if any pages failed.
     """
-    logger.info(
-        "Publishing Bundle",
-        extra={
-            "bundle_id": bundle.pk,
-            "event": "publishing_bundle",
-        },
-    )
```
MebinAbraham, line 419: Is this now captured elsewhere?

cms/core/models/base.py
```diff
@@ -334,6 +339,29 @@ def full_edit_url(self) -> str | None:
...
+        self._log_preview(request, mode_name)
+        return cast("TemplateResponse", super().serve_preview(request, mode_name))
```
MebinAbraham: Nit: Let's avoid cast as per our style guide.
```py
response: TemplateResponse = super().serve_preview(request, mode_name)
return response
```

cms/core/wagtail_hooks.py
```diff
@@ -19,7 +21,10 @@
     from collections.abc import Sequence
 
     from django.http import HttpRequest, HttpResponsePermanentRedirect, HttpResponseRedirect
-    from wagtail.admin.views.bulk_action import BulkAction
+    from wagtail.admin.views.bulk_action import BulkAction, ModelLogEntry
+    from wagtail.log_actions import LogActionRegistry
+
+PAGE_EDIT_VIEW_AUDIT_LOG_COOLDOWN_SECONDS = 30
```
MebinAbraham: It would be good to drive via env var setting and a single global one, unless we need different ones for different actions.

Waiting on confirmation about this approach from a compliance perspective.

cms/core/tests/test_preview_audit.py
MaciekBaron: I actually looked into this already and decided against it due to complexity and messiness. We would need to do something like this to import all factories:

```python
for app_config in apps.get_app_configs():
    if app_config.name.startswith("cms."):
        try:
            importlib.import_module(f"{app_config.name}.tests.factories")
        except ModuleNotFoundError:
            pass
```
And then get the subclasses of `wagtail_factories.PageFactory`

And then ignore certain ones which already exist and ones we don't need to run (e.g. `BasePageFactory`).

It becomes quite unwieldily.

_However_, what we could do is add a test to check if a new type was added, something like:

```python
def test_covers_all_page_types(self):
    """Fail if a new page type is added but not exercised above."""
    # Assume here we added necessary imports
    all_page_types = {m for m in get_page_models() if issubclass(m, BasePage)}
    covered = {type(p) for p in self.pages}
    self.assertEqual(
        all_page_types - covered,
        set(),  # or a list of exceptions/edge cases you mentioned
        "New BasePage subclass(es) not covered by the preview audit test.",
    )
```

What do you think?

MebinAbraham: Happy with that `test_covers_all_page_types`. 👍

MaciekBaron: Added here: [c69a417](https://github.com/ONSdigital/dis-wagtail/pull/446/commits/c69a4176dd902648f38c72253d0adeb0bba1a0c7)

cms/core/models/base.py
MaciekBaron: Changed here: [6d6eb96](https://github.com/ONSdigital/dis-wagtail/pull/446/commits/6d6eb96af1a4c01f052c19669192452482e1a6e7)

cms/bundles/wagtail_hooks.py
MaciekBaron: Changed here: [6d6eb96](https://github.com/ONSdigital/dis-wagtail/pull/446/commits/6d6eb96af1a4c01f052c19669192452482e1a6e7)

cms/bundles/wagtail_hooks.py
MaciekBaron: Updated here per suggestion: [6d6eb96](https://github.com/ONSdigital/dis-wagtail/pull/446/commits/6d6eb96af1a4c01f052c19669192452482e1a6e7)

cms/bundles/utils.py
MaciekBaron: I can't remember why I removed that, must have been accidental or an oversight, reverted it: [517b186](https://github.com/ONSdigital/dis-wagtail/pull/446/commits/517b1862d8dc5f3e3587b99fdf7f354b5c962617)

MebinAbraham:
@MaciekBaron legal are happy with the cooldown approach. Could you here or as a follow up PR add a cooldown to Edit views for Pages/Snippets?. This should reduce the audit log building up due to auto save. Must not be used for bundle as we don't have revision history.

MaciekBaron:
@MebinAbraham I added a timeout for page edits here: [c2275f5](https://github.com/ONSdigital/dis-wagtail/pull/446/commits/c2275f5ffa32269c2174b02c277fc0f330416adb)

We don't touch snippets in this ticket, should I add them in a separate PR?

cms/core/wagtail_hooks.py
MaciekBaron: [c2275f5](https://github.com/ONSdigital/dis-wagtail/pull/446/commits/c2275f5ffa32269c2174b02c277fc0f330416adb)

cms/bundles/viewsets/bundle.py
MaciekBaron: [c2275f5](https://github.com/ONSdigital/dis-wagtail/pull/446/commits/c2275f5ffa32269c2174b02c277fc0f330416adb)

MebinAbraham:
> @MebinAbraham I added a timeout for page edits here: [c2275f5](https://github.com/ONSdigital/dis-wagtail/pull/446/commits/c2275f5ffa32269c2174b02c277fc0f330416adb)
> 
> We don't touch snippets in this ticket, should I add them in a separate PR?

Yes please. We need to audit snippets also.

MebinAbraham (approved):
LGTM 🎉 

Please follow up with snippets support. 👍

sanjeevz3009 (commented):
Reviewed!

PR #505, opened 2026-02-18
feat: frontend cache invalidation

zerolab:

CMS-724

This PR adds front-end cache invalidation with bespoke logic

To-Do:
- [x] Add dummy backend that logs
- [x] Expand logic for chart/table downloads after #471
- [x] Page slug change (cascading to all children)

### How to review

See tests or add:

```python
INSTALLED_APPS.append("cms.frontend_cache")
WAGTAILFRONTENDCACHE = {"default": {"BACKEND": "cms.frontend_cache.tests.backends.DummyFrontEndCacheBackend"}}
```

to your local.py and observe the logs after publishing a page

### Follow-up Actions

N/A

---


cms/frontend_cache/signal_handlers.py
```diff
@@ -0,0 +1,94 @@
...
+    ReleaseCalendarIndex = apps.get_model("release_calendar", "ReleaseCalendarIndex")
+    # pylint: enable=invalid-name
+
+    # TODO: remove HomePage from the list when serving it from Wagtail
+    # The Release Calendar, articles and methodology indexes are served by the Search service and has a short TTL
+    excluded_types = {ArticlesIndexPage, HomePage, MethodologyIndexPage, ReleaseCalendarIndex}
+    return {model for model in apps.get_models() if issubclass(model, BasePage) and model not in excluded_types}
+
+
+def _get_tracked_snippets() -> set[Model]:
```
RealOrangeOne: Question: What makes a snippet "tracked"?

cms/settings/base.py
```diff
@@ -644,7 +644,7 @@
 # restricted access.
 
 if "FRONTEND_CACHE_CLOUDFLARE_TOKEN" in env or "FRONTEND_CACHE_CLOUDFLARE_BEARER_TOKEN" in env:
-    INSTALLED_APPS.append("wagtail.contrib.frontend_cache")
```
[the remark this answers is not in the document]
zerolab: I thought of it. The two places that check that are the API and the redirects app - https://github.com/search?q=repo%3Awagtail%2Fwagtail+apps.is_installed%28%22wagtail.contrib.frontend_cache%22%29&type=code - we don't use one, and handle the other.

My thinking was our requirements are specific enough that it is easier to not include the app. Happy to include, and disconnect the various signal handlers if that is the consensus

cms/frontend_cache/signal_handlers.py
```diff
@@ -0,0 +1,94 @@
...
+    purge_page_containing_snippet_from_cache(instance)
+
+
+def snippet_deleted(sender: Model, instance: Model, **kwargs: Any) -> None:  # pylint: disable=unused-argument
+    from .cache import purge_page_containing_snippet_from_cache  # pylint: disable=import-outside-toplevel
+
+    purge_page_containing_snippet_from_cache(instance)
+
+
+def _get_indexed_page_models() -> set[Page]:
```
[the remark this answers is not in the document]
zerolab: renamed to tracked. as in tracking for front-end purging purposes. expanded the comments as to why we exclude certain types

3992e49c

cms/frontend_cache/signal_handlers.py
```diff
@@ -0,0 +1,94 @@
...
+from django.apps import apps
+from django.db.models.signals import post_delete
+from wagtail.signals import page_published, page_slug_changed, page_unpublished, published, unpublished
+
+if TYPE_CHECKING:
+    from django.db.models import Model
+    from wagtail.models import Page
+
+
+def page_published_signal_handler(instance: Page, **kwargs: Any) -> None:
```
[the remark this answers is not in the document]
zerolab: heh, yeah, naming is hard :)
3992e49c

AdamHawtin (commented):
Quick note from local testing, when publishing a page I'm seeing logs for purging three paths, including one with the `en-gb` locale explicitly in the path (e.g. `.../en-gb/my-page`), is this expected?

cms/frontend_cache/tests/test_signal_handlers.py
```diff
@@ -0,0 +1,597 @@
...
+            urls.add(self.topic_page_translation_url)
+
+        return urls
+
+    def test_page_publish__statistical_article(self, patched_purge_urls):
+        self.statistical_article.save_revision().publish()
+
+        patched_purge_urls.assert_called_once_with(self._get_base_expected_statistical_page_urls_to_purge())
+
+    def test_page_publish__statistical_article__with_related_terms(self, patched_purge_urls):
```
sanjeevz3009: Naming things is hard.

How about `test_page_publish_cache_purged_for_statistical_article_and_all_linked_topics`?

sanjeevz3009 (commented):
Did the initial round of reviews. So far looks good. I will do another round review tomorrow. I got a few more suggestions for possibly more tests and around the code itself, which I'll submit tomorrow.

cms/frontend_cache/cache.py
```diff
@@ -0,0 +1,228 @@
...
+
+    for page in (
+        Page.objects.filter(Q(id__in=page_ids) | Q(alias_of__in=page_ids)).specific(defer=True).live().iterator()
+    ):
+        urls.update(get_page_cached_urls(page))
+
+    return urls
+
+
+def get_related_topic_page_urls(page: Page, topic_ids: list[str] | None = None) -> set[str]:
```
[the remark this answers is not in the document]
zerolab: I would need to indeed. Didn't get to post an answer yesterday. In expanding the tests I noticed it is not quite covered. Almost done with the fix

cms/frontend_cache/tests/test_signal_handlers.py
```diff
@@ -0,0 +1,597 @@
...
+        )
+
+    @patch("cms.frontend_cache.signal_handlers.purge_page_from_frontend_cache")
+    def test_excluded_page_types__dont_react(self, patched_purge_page, patched_purge_urls):
+        self.home_page.save_revision().publish()
+
+        patched_purge_page.assert_not_called()
+        patched_purge_urls.assert_not_called()
+
+    def _get_base_expected_statistical_page_urls_to_purge(self, with_translation=True):
```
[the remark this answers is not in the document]
zerolab: fb163e1c

cms/frontend_cache/tests/test_signal_handlers.py
```diff
@@ -0,0 +1,597 @@
...
+                    "value": [{"page": self.information_page.pk}],
+                },
+            ]
+            self.statistical_article.content = [{"type": "section", "value": {"content": content, "title": "Section"}}]
+            self.statistical_article.save()
+
+            self.statistical_article.save_revision().publish()
+
+        self.information_page.save_revision().publish()
+        patched_purge_urls.assert_called_with(
```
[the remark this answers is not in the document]
zerolab: because we're publishing the information page, not the statistical article page =)

cms/frontend_cache/cache.py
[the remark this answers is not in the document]
zerolab: ef8a50d

cms/frontend_cache/tests/test_signal_handlers.py
```diff
@@ -0,0 +1,597 @@
...
+
+        self.statistical_article.save_revision().publish()
+
+        expected_urls = self._get_base_expected_statistical_page_urls_to_purge() | {
+            self.another_topic_page.get_full_url(self.request),
+            self.another_topic_page_translation.get_full_url(self.request),
+        }
+        patched_purge_urls.assert_called_once_with(expected_urls)
+
+    def test_page_publish__statistical_article__and_parent_topic_with_proper_translation(self, patched_purge_urls):
```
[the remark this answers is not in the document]
sanjeevz3009:
```
(renamed to without_translation_alias) 
```

I don't see this change?

cms/frontend_cache/tests/test_signal_handlers.py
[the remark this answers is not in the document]
sanjeevz3009: Late night brain =(

sanjeevz3009 (commented):
Thanks @zerolab for the changes. I resolved most of my comments; a few are left to address, and I left a few new minor ones. Looks good. Will do a functional test locally today afternoon and then if all is good, approval is on the way.

AdamHawtin (approved):
LGTM and works well testing locally by the log output 👍

MaciekBaron (approved):
One non-blocking nit and one question for Mebin, but apart from that LGTM

cms/frontend_cache/cache.py
```diff
@@ -0,0 +1,257 @@
...
+    source_page_ids = {page_old.pk}
+
+    for descendant in page.get_descendants().live().defer_streamfields().specific().iterator():
+        source_page_ids.add(descendant.pk)
+        descendant.url_path = page_old.url_path + descendant.url_path[url_path_length:]
+
+        urls.update(get_page_cached_urls(descendant))
+
+    if include_welsh_aliases:
+        urls |= _get_welsh_alias_urls(source_page_ids)
```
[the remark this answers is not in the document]
MebinAbraham: Ideally, it would be agnostic and loop over non-default locales. Though supporting a new language does require some thought, as not all locales will have sub-domains, etc. If we were using paths that are likely down the road, making this agnostic would mean less change and it should just work.

MaciekBaron:
Following the merger of #564 one of the tests is failing, I'm currently investigating the issue.

cms/frontend_cache/cache.py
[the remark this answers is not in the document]
MaciekBaron: Made it more flexible in terms of locales here: [35496de](https://github.com/ONSdigital/dis-wagtail/pull/505/commits/35496dec4341ef9bfc8d843c69b07d5d56486df9)

cms/frontend_cache/tests/test_signal_handlers.py
```diff
@@ -0,0 +1,597 @@
...
+        )
+
+    def test_unpublish__definition(self, mocked_purge_urls):
+        self.definition.unpublish()
+
+        mocked_purge_urls.assert_called_once_with(
+            {self.statistical_article_url, self.statistical_article_related_data_url}
+        )
+
+    def test_delete(self, mocked_purge_urls):
```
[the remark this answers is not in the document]
MaciekBaron: Addressed in [35496de](https://github.com/ONSdigital/dis-wagtail/pull/505/commits/35496dec4341ef9bfc8d843c69b07d5d56486df9)

cms/articles/models.py
```diff
@@ -111,6 +116,28 @@ def summary(self) -> str:
...
+    @transaction.atomic
+    def save(  # type: ignore[override]
+        self, clean: bool = True, user: User | None = None, log_action: bool = False, **kwargs: Any
+    ) -> ArticleSeriesPage | None:
+        instance: ArticleSeriesPage | None = super().save(  # type: ignore[call-arg]
+            clean=True, user=None, log_action=False, **kwargs
+        )
+
+        if self.pk:
+            original_values = ArticleSeriesPage.objects.values("title", "slug").filter(pk=self.pk).first()
```
kevinhowbrook: **Issue?**

Shouldn't original_values come _before_ the save on line 123 here?

cms/frontend_cache/cache.py
```diff
@@ -0,0 +1,258 @@
...
+    urls: set[str] = set()
+
+    if not statistical_article.is_latest:
+        return urls
+
+    topic_page_ids = []
+    # Get the series topic path and exclude it as we handle the parent topic separately.
+    # url_path is in the form: /home/topic/articles/series/ so this becomes /home/topic/
+    parent_topic_path = "/".join(article_series.url_path.rstrip("/").split("/")[:-2]) + "/"
+    featured_on_topics_qs: PageQuerySet = article_series.featured_on_topic.exclude(path=parent_topic_path)
```
kevinhowbrook: **Issue**

Shouldn't the filter use `url_path` rather than path here?
Page.path will give the treebeard path e,g `00010001`

cms/frontend_cache/signal_handlers.py
```diff
@@ -0,0 +1,121 @@
...
+    purge_old_page_paths_from_cache_after_move(
+        instance.specific, parent_page_before.specific_deferred, parent_page_after.specific_deferred, url_path_before
+    )
+
+
+def purges_series_children_from_frontend_cache(instance: ArticleSeriesPage, **kwargs: Any) -> None:
+    purge_series_children_from_cache(instance)
+
+
+def purge_pages_containing_the_published_snippet_from_frontend_cache(instance: Model, **kwargs: Any) -> None:
```
kevinhowbrook, line 68: **Nit**
Doesn't look like this one is connected up?

kevinhowbrook (changes requested):

cms/frontend_cache/signal_handlers.py
zerolab: Good spot. `published.connect(purge_pages_containing_the_unpublished_snippet_from_frontend_cache, sender=model)` should be `published.connect(purge_pages_containing_the_published_snippet_from_frontend_cache, sender=model)` on L114

cms/frontend_cache/cache.py
zerolab: It should indeed

cms/settings/base.py
```diff
@@ -674,8 +674,9 @@
 # The backend can be configured to use an account-wide API key, or an API token with
 # restricted access.
 
+WAGTAILFRONTENDCACHE_LANGUAGES: list[str] = []  # handled by cms.frontend_cache
```
BJacksonONS: Nit: I feel like this comment could be a bit more clear.

[The documentation](https://docs.wagtail.org/en/latest/reference/contrib/frontendcache.html) around `WAGTAILFRONTENDCACHE_LANGUAGES` could probably be more clear as to how it works (eg. says it defaults to an empty list, but actually pulls in all configured languages), so it's not obvious at a glance why we set this empty here.

BJacksonONS (commented):
Could we add some tests to make it clear how caching is handled when we're using subdomain based localisation?

The using `get_full_url` looks like it should work correctly for path based locales, but it's not 100% obvious what happens with subdomain based. My first instinct is to assume we'd clear both welsh and english for changes to either, which probably isn't the end of the world but it'd be helpful to be explicit.

cms/settings/base.py
zerolab: FWIW, we use `wagtail.contrib.frontend_cache.utils.purge_urls_from_cache` which under the hood ends up defaulting to the configured languages (https://github.com/wagtail/wagtail/blob/f0b2c5bed43843efde1673c498290ed3d3a65fd4/wagtail/contrib/frontend_cache/tasks.py#L33-L35). 

We handle the translations ourselves

BJacksonONS:
> Could we add some tests to make it clear how caching is handled when we're using subdomain based localisation?
> 
> The using `get_full_url` looks like it should work correctly for path based locales, but it's not 100% obvious what happens with subdomain based. My first instinct is to assume we'd clear both welsh and english for changes to either, which probably isn't the end of the world but it'd be helpful to be explicit.

Turns out I'd been using path based routing after testing something else so assumed that was the default in test. It's doing subdomain by default right now, so maybe we should add tests overriding to use path based to check that works correctly if we move to that later on.

cms/settings/base.py
BJacksonONS: Yeah I think the thing that confused me initially was this bit in the docs.

> Set WAGTAILFRONTENDCACHE_LANGUAGES to a list of languages (typically equal to [l[0] for l in settings.LANGUAGES]) to also purge the urls for each language of a purging url. This setting needs settings.USE_I18N to be True to work. **Its default is an empty list.**

Without reading the wagtail source the docs would make it seem it's not necessary.

BJacksonONS:
Okay did a bit more digging, and right now this fails under path based locales.

`get_full_url` in the cache purge function doesn't get passed a request currently, so for welsh pages it's returning the url for the welsh site (eg. cy.ons.gov.uk), rather than the default. This means we're trying to purge incorrect urls.
 
`https://cy.ons.gov.uk/cy/the-topic-page`
Rather than
`https://ons.gov.uk/cy/the-topic-page`

zerolab:
> Okay did a bit more digging, and right now this fails under path based locales.
> 
> `get_full_url` in the cache purge function doesn't get passed a request currently, so for welsh pages it's returning the url for the welsh site (eg. cy.ons.gov.uk), rather than the default. This means we're trying to purge incorrect urls.
> 
> `https://cy.ons.gov.uk/cy/the-topic-page` Rather than `https://ons.gov.uk/cy/the-topic-page`

@BJacksonONS it works as expected. With multiple site entries, the Site host is used in the URL matching and generation. With the subdomain locale setup, we have 2 `Site` entries: the default site (en) and the Welsh site (cy, pointing to the Welsh page tree). In a path based scenario, we'd only have the default site, so the full URL would have the expected format.

Locally, delete the Welsh site and it will behave as expected

cms/settings/base.py
zerolab: 9436e6cb (includes test for path-based locales)

cms/frontend_cache/signal_handlers.py
zerolab: 9436e6cb

it was working correctly because the handler for publish and unpublish do the same thing currently. They are separate because explicit is better than implicit as they say

cms/frontend_cache/cache.py
zerolab: 9436e6cb

this is more of a nice to have as the parent topic is already included in the URLs to purge.

cms/articles/models.py
zerolab: 9436e6cb

this worked, flawed as it was because the transaction was not committed. Amended to be clearer

BJacksonONS (approved):

BJacksonONS:
An issue around CSV downloads has been raised, we might need to ensure we're also taking download urls into account when we publish revisions of pages that contain them.

zerolab:
> An issue around CSV downloads has been raised, we might need to ensure we're also taking download urls into account when we publish revisions of pages that contain them.

I would rather we do it as a follow up -- this is a big enough PR that has been open for quite some time now.

kevinhowbrook (approved):
Thanks @zerolab 

I would say before this merges we should raise a ticket for the follow-up task of addressing the CSV issue mentioned https://github.com/ONSdigital/dis-wagtail/pull/505#issuecomment-4517568461

MebinAbraham:
Just need to confirm a few things before merge.

cms/frontend_cache/cache.py
```diff
@@ -0,0 +1,258 @@
...
+
+    See https://github.com/wagtail/wagtail/issues/13920
+    """
+    if (page_url := page.get_full_url(request=cache_object)) is None:
+        # nothing to be done if the page has no routable URL
+        return []
+
+    return [
+        f"{page_url.rstrip('/')}/{path.lstrip('/')}".rstrip("/") for path in page.specific_deferred.get_cached_paths()
+    ]
```
MebinAbraham, line 40: Should be conditional on `WAGTAIL_APPEND_SLASH`?

cms/standard_pages/models.py
```diff
@@ -74,6 +74,9 @@ def get_context(self, request: HttpRequest, *args: Any, **kwargs: Any) -> dict:
         context["table_of_contents"] = self.table_of_contents
         return context
 
+    def get_cached_paths(self) -> list[str]:
+        return ["/", *self.get_downloadable_block_paths()]
```
MebinAbraham, line 82: This would need to account for the index page the info page sits under.

cms/frontend_cache/signal_handlers.py
```diff
@@ -0,0 +1,121 @@
...
+    return {model for model in get_page_models() if issubclass(model, BasePage) and model not in EXCLUDED_PAGE_TYPES}
+
+
+def disconnect_signal_handlers() -> None:
+    """Disconnect the core front-end cache signal handlers as we handle them with specific logic."""
+    for model in get_page_models():
+        page_published.disconnect(page_published_signal_handler, sender=model)
+        page_unpublished.disconnect(page_unpublished_signal_handler, sender=model)
+        post_page_move.disconnect(autocreate_redirects_on_page_move, sender=model)
+        page_slug_changed.disconnect(autocreate_redirects_on_slug_change, sender=model)
```
MebinAbraham, line 101: We don't do this ourselves and want to keep it auto-creating redirects. I can't recall why the redirect ones were disconnected. I suspect that it also does cache purging, but we need a way to retain the redirect creation.

A worst-case solution that might be acceptable is that for slug changes which requires redirect creation, which don't happen often, we accept there are duplicate requests if we can't handle this nicely without a lot of effort.

cms/frontend_cache/signal_handlers.py
```diff
@@ -0,0 +1,121 @@
...
+    instance: Page,
+    parent_page_before: Page,
+    parent_page_after: Page,
+    url_path_before: str,
+    url_path_after: str,  # pylint: disable=unused-argument
+    **kwargs: Any,
+) -> None:
+    purge_old_page_paths_from_cache_after_move(
+        instance.specific, parent_page_before.specific_deferred, parent_page_after.specific_deferred, url_path_before
+    )
```
MebinAbraham, line 61: We would also want to purge new paths?. Otherwise, we could be serving cached 404s.

If a topic page moved, we purge the topic page but not the children.

cms/articles/models.py
```diff
@@ -857,3 +895,12 @@ def word_count(self) -> int:
...
         return title_word_count + summary_word_count + content_word_count
+
+    def get_cached_paths(self) -> Generator[str]:
+        yield "/"
+        yield "/related-data"  # always include, should a correction remove related datasets
+        if self.datasets:
+            for page_number in range(1, ceil(len(self.datasets) / settings.RELATED_DATASETS_PER_PAGE) + 1):
+                yield f"/related-data?page={page_number}"
+
+        yield from self.get_downloadable_block_paths()
```
MebinAbraham, line 934: This covers the current cached paths nicely, but I think it misses paths that existed before the edit and were removed by the new revision.

For example, removing a table/chart block or reducing related-data pagination means the old `/download-table/...`, `/download-chart/...`, or `/related-data?page=N` URL no longer appears in `get_cached_paths()`, so it will not be purged.

We should capture the previous cached paths before save/publish and union them into the purge set?

I think we need to be careful to check that we are always purging all new and all old URLs.

cms/articles/models.py
```diff
@@ -857,3 +895,12 @@ def word_count(self) -> int:
         summary_word_count = len(strip_tags(self.summary).split())
 
         return title_word_count + summary_word_count + content_word_count
+
+    def get_cached_paths(self) -> Generator[str]:
+        yield "/"
+        yield "/related-data"  # always include, should a correction remove related datasets
+        if self.datasets:
```
MebinAbraham, line 930: The comment says `/related-data` should always be included in case a correction removes related datasets. I think the same applies to `/related-data?page=1`.

Currently `?page=1` is only yielded when `self.datasets` is truthy, so removing all related datasets could leave the paginated URL stale.

cms/frontend_cache/cache.py
```diff
@@ -0,0 +1,258 @@
...
+    return urls
+
+
+def get_related_topic_page_urls(page: Page, topic_ids: list[str] | None = None) -> set[str]:
+    if page.alias_of_id is not None:
+        # skip if the given page is an alias. Aliases are not published directly, but
+        # via their source page, which then accounts for related topics and their aliases.
+        return set()
+
+    parent_topic = TopicPage.objects.ancestor_of(page).only("pk", "url_path").first()
```
MebinAbraham, line 111: Q: Is it possible for this not to exist?. Not without the current structure, right?

cms/frontend_cache/cache.py
```diff
@@ -0,0 +1,258 @@
...
+    urls = set(get_page_cached_urls(parent_topic))
+
+    topic_terms = topic_ids or getattr(page, "topic_ids", [])
+    related_topic_pages = (
+        TopicPage.objects.filter(topic__in=topic_terms).exclude(pk=parent_topic.pk).live().only("pk", "url_path")
+    )
```
MebinAbraham, line 117: Q: Any reason not to filter to just live public pages only?. We don't want purge requests for draft pages.

cms/frontend_cache/cache.py
```diff
@@ -0,0 +1,258 @@
...
+    elif isinstance(page, InformationPage):
+        urls.update(get_page_cached_urls(page.get_parent().specific_deferred))
+
+    urls.update(get_urls_featuring_object(page))
+    purge_urls_from_cache(urls)
+
+
+def purge_old_page_slugs_from_frontend_cache(page: Page, page_old: Page) -> None:
+    if urls := get_old_page_slugs(page, page_old):
+        purge_urls_from_cache(urls)
```
MebinAbraham, line 212: We need to purge "new" urls?. We do in some cases, but for scenarios such as a topic page slug being changed, I don't see any purges for the child pages?

cms/articles/models.py
```diff
@@ -203,6 +233,14 @@ def download_table_with_version(self, request: HttpRequest, slug: str, version:
         response: HttpResponse = self.release(request, slug, version=version, table_id=table_id)
         return response
 
+    def get_cached_paths(self) -> Generator[str]:
+        yield "/"
+        yield "/editions"
```
MebinAbraham, line 257: Article Series also has `/related-data`, which is a canonical URL. It also has pagination so we need to handle that similarly to `/editions`.

cms/frontend_cache/signal_handlers.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def purge_published_page_from_frontend_cache(instance: Page, **kwargs: Any) -> None:
+    purge_page_from_frontend_cache(instance)
+
+
+def purge_unpublished_page_from_frontend_cache(instance: Page, **kwargs: Any) -> None:
+    # note: this also covers page deletion
+    # https://github.com/wagtail/wagtail/blob/228058b56775b126a91ecc4e4338366869513ea8/wagtail/signal_handlers.py#L31
+    purge_page_from_frontend_cache(instance)
```
MebinAbraham, line 44: Im not sure this could be identical for publish and unpublish, the paths we clear depend on whether it is a publish or unpublish?

cms/frontend_cache/signal_handlers.py
```diff
@@ -0,0 +1,121 @@
...
+    for model in get_page_models():
+        page_published.disconnect(page_published_signal_handler, sender=model)
+        page_unpublished.disconnect(page_unpublished_signal_handler, sender=model)
+        post_page_move.disconnect(autocreate_redirects_on_page_move, sender=model)
+        page_slug_changed.disconnect(autocreate_redirects_on_slug_change, sender=model)
+
+
+def register_signal_handlers() -> None:
+    for model in _get_tracked_page_models():
+        page_published.connect(purge_published_page_from_frontend_cache, sender=model)
```
MebinAbraham, line 106: Q: Is `page_published` fired `on_commit` or before commit?. If not on commit, I guess we could have a small race condition between purge and old state repopulating cache?.

cms/frontend_cache/cache.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+def purge_page_from_frontend_cache(page: Page) -> None:
+    # get the page urls
+    urls = set(get_page_cached_urls(page))
+
+    # expand to custom logic
+    if isinstance(page, StatisticalArticlePage):
+        series_page = page.get_parent().specific_deferred
+        urls.update(get_page_cached_urls(series_page))
```
MebinAbraham, line 198: I think we have some edge cases for unpublish. Unpublishing the last article on `/editions?page=3` means the series now only has two pages. This code only purges the current series paths, meaning the removed last page may not be invalidated.

cms/articles/models.py
MebinAbraham: Since this is hit by both publish and unpublish, I think we have some edge cases. If we only account for currently version, we could have stale paths. i.e., the latest published has might have max `/related-data?page=2` but on unpublish the previous version that becomes latest may have `page=5` which we don't invalidate hence we would be serving 404s.

Our lives and logic would be much simpler in some cases if we looked to see if prefix purge was viable. (Needs discussion if it were to be persued)

cms/settings/base.py
```diff
@@ -673,8 +673,11 @@
 # The backend can be configured to use an account-wide API key, or an API token with
 # restricted access.
 
+# Deliberately set to an empty list as cms.frontend_cache uses purge_urls_from_cache which will
+# default to the list of defined content languages if this is not defined.
+WAGTAILFRONTENDCACHE_LANGUAGES: list[str] = []
 if "FRONTEND_CACHE_CLOUDFLARE_TOKEN" in env or "FRONTEND_CACHE_CLOUDFLARE_BEARER_TOKEN" in env:
-    INSTALLED_APPS.append("wagtail.contrib.frontend_cache")
+    INSTALLED_APPS += ["wagtail.contrib.frontend_cache", "cms.frontend_cache"]
```
MebinAbraham, line 698: A comment to suggest the order is important and intentional would be good (signal disconnects require the correct order)

cms/frontend_cache/cache.py
```diff
@@ -0,0 +1,258 @@
...
+        urls.update(get_page_cached_urls(topic_page_alias))
+
+    return urls
+
+
+def _get_other_language_alias_urls(source_page_ids: set) -> set[str]:
+    urls = set()
+    # Include other language aliases in this too.
+    # Also, since we don't have a request here, get a dummy one for the other language pages.
+    other_locale_ids = Locale.objects.exclude(pk=Locale.get_default().pk).values_list("pk", flat=True)
```
MebinAbraham, line 159: In theory, we could have the same locale alias, which is one of the use cases of alias pages?

cms/frontend_cache/signal_handlers.py
MebinAbraham: One consequence of this approach is that, on the first publish of a page, we do not clear the cache for the Welsh locale alias. This happens because the cache purge is triggered before the locale aliases have been created, meaning the Welsh alias is not yet known at the point the purge runs.

This becomes more significant if we increase the caching duration of 404 responses which is the plan. Caching 404s is only safe if we can guarantee that all relevant paths are purged whenever content becomes available. In this edge case, a user could request the Welsh URL before the page is first published, causing a 404 to be cached. When the page is subsequently published and the Welsh alias is created, the cached 404 would remain in place because that path was never purged.

Currently, the impact is limited because our 404 cache duration is relatively short (Cloudflare's default TTL is only 3 minutes), but it would become a more noticeable problem if we increase 404 caching as planned. Regardless, this would breach our 59-second compliance, hence we need to have a solution for this.

MebinAbraham (commented):
Was hoping to get this in, but there are some fundamental things that need discussion before this can merge. The likely best way forward, given where we are, is probably to get this merged into a feature branch and follow up with the various fixes we have identified.

It's unfortunate that this was parked for so long.

MaciekBaron:
@MebinAbraham I have changed the merge target to `feature-frontend-cache-invalidation` which was created using latest `main`

cms/frontend_cache/cache.py
MaciekBaron: Yep we can: [f26b4b0](https://github.com/ONSdigital/dis-wagtail/pull/505/commits/f26b4b09b0740367e5c28396bda93636a2f5b7c4)

cms/articles/models.py
MaciekBaron: Thanks, I analysed this a bit more and I think there is a wider issue in which shrinking pagination orphans the trailing page. So if there were 3 pages but there are now 2, then we would also need to yield `?page=3`.

I think the least complicated and most pragmatic solution would be to over-purge a trailing buffer. We could have an `PAGINATION_OVER_PURGE` setting (e.g. `2` or `3`) and always purge `n + PAGINATION_OVER_PURGE`. Purging a non-existent `?page=99` is harmless (it just clears a key that may hold a 404).

What do you think?

cms/frontend_cache/cache.py
MaciekBaron: No, it's not possible. Currently this branch is only executed for specific page types which are guaranteed to have a topic ancestor, based on our current rules.

cms/frontend_cache/cache.py
MaciekBaron: `get_old_page_slugs` calls `page.get_descendants()`, so child pages are handled (unless I misunderstood your quesiton)

cms/standard_pages/models.py
MaciekBaron: Already handled here:
https://github.com/ONSdigital/dis-wagtail/blob/f26b4b09b0740367e5c28396bda93636a2f5b7c4/cms/frontend_cache/cache.py#L203-L204

cms/frontend_cache/signal_handlers.py
MaciekBaron: Regarding the first point I believe this is already happening in that function, so maybe it's just a case of finding a better name for it?

https://github.com/ONSdigital/dis-wagtail/blob/f26b4b09b0740367e5c28396bda93636a2f5b7c4/cms/frontend_cache/cache.py#L233-L234

Regarding the second point, I'm still looking into it, I think you may be right that we don't update descendants, I think we might need to call `purge_series_children_from_cache` on the moved page (and perhaps rename that function)

cms/articles/models.py
MaciekBaron: I think this requires a wider discussion (we discussed parts of this briefly during out tech meeting). Especially the pagination side of things.

cms/frontend_cache/cache.py
MaciekBaron: We are using `live()` there, unless I don't understand your question?

cms/articles/models.py
MaciekBaron: Missing `/related-data` path added [f25c865](https://github.com/ONSdigital/dis-wagtail/pull/505/commits/f25c865b3d6366213c83047d219460488354560f). Editions paginated already handled, apart from the wider issue discussed earlier.

cms/frontend_cache/signal_handlers.py
MaciekBaron: I'm trying to understand what you mean: when we publish `/foo/bar` it goes from 404 to live content, when we unpublish `/foo/bar` it goes from live content to 404. Could you explain which paths differ? I might be missing something.

cms/frontend_cache/signal_handlers.py
MaciekBaron: Re. first question, it's fired after the page is published i.e. everything has been committed and the page is live.

Re. your second comment, when a page gets published, all of its aliases always get purged as well, so not sure if this applies?

cms/frontend_cache/cache.py
MaciekBaron: This is discussed in other places (replying for clarity)

MebinAbraham:
Can you temporarily update ci.yml to include the feature branch: `
feature-frontend-cache-invalidation` https://github.com/ONSdigital/dis-wagtail/blob/main/.github/workflows/ci.yml#L8

MaciekBaron:
@MebinAbraham added here: [40245ee](https://github.com/ONSdigital/dis-wagtail/pull/505/commits/40245ee6c6718c561f80dda671506e30afcf8d69)

cms/frontend_cache/cache.py
MebinAbraham: Totally missed the `live()` 🙈

cms/frontend_cache/signal_handlers.py
MebinAbraham: Re: 1, why are we not called `get_page_cached_urls` for methodology page?

> Regarding the first point I believe this is already happening in that function, so maybe it's just a case of finding a better name for it?

👍 

> If a topic page moved, we purge the topic page but not the children.

Needs handling in a follow up.

cms/articles/models.py
MebinAbraham: Ideally, we prefix purge, but since Wagtail doesn't officially support that yet, an overpurge seems like an appropriate pragmatic solution.

cms/frontend_cache/cache.py
MebinAbraham: What I mean is, this is only purge the old url when slugs change, we need to purge new urls?. I don't see a call to purge the new urls (post slug change?). on slug page we only call this function?.

cms/frontend_cache/signal_handlers.py
MebinAbraham: > Re. your second comment, when a page gets published, all of its aliases always get purged as well, so not sure if this applies?

I think this is okay for our use case. I think the issue is visible if you create a page and publish straight away; then no alias exists at the point of purge, as it is a race condition. Given we don't allow direct publishing, this is not an issue for us but technically an issue in core.

MebinAbraham (approved):
Once main is merged in (into base and here), this can be merged to the feature branch, and any outstanding comments can be worked on as follow-up PRs. 👍

cms/articles/models.py
MaciekBaron: Addressed in #780

cms/frontend_cache/cache.py
MaciekBaron: If I understand what you're asking: any new slugs will get purged by the page publish signal, so we don't need to do any extra work

cms/frontend_cache/signal_handlers.py
MaciekBaron: Addressed here: https://github.com/ONSdigital/dis-wagtail/pull/785

cms/frontend_cache/signal_handlers.py
MaciekBaron: Address here: https://github.com/ONSdigital/dis-wagtail/pull/785

cms/settings/base.py
MaciekBaron: Added here: https://github.com/ONSdigital/dis-wagtail/pull/785

cms/frontend_cache/cache.py
MaciekBaron: Broadened the scope here: https://github.com/ONSdigital/dis-wagtail/pull/785

cms/articles/models.py
MebinAbraham: The issue still exists for `get_downloadable_block_paths`, as if the content is removed, the path is no longer cleared. Might not be able to do anything nice without prefix purging. We can only know if we check "previously published version". We should make a note in code this as a known gap.

cms/frontend_cache/signal_handlers.py
MebinAbraham: Non issue, all good.

cms/articles/models.py
zerolab: Given this, it is best if the team helps review and test https://github.com/wagtail/wagtail/pull/13773 so that we can get prefix purging in

MebinAbraham: Sounds like a plan. I have created a card in the backlog to dedicate some time to this. This is currently not in scope (prefix bit for our next release but will try to get it sooner rather than later, due to release lead time etc)

PR #525, opened 2026-03-06
Bundle-aware private media serving

zerolab:

This enables previewers to view private media when previewing content in a bundle

### How to review

- create two preview teams
- create one or more pages (information, statistical article) that with newly uploaded images and docs (1x image, 1x doc, new for each page)
- submit the page(s) for review
- add a new bundle, add the page(s), add one of the teams, and save the bundle for review
- log in with a reviewer from the bundle team - you should be able to view the assets. Make a note of their URLs.
- log in with a reviwer that is not in the bundle team, try to access the asset URLs

[Screencast](https://www.loom.com/share/ad8b3eb34b934268abef8db783c55823)

### Follow-up Actions


cms/bundles/views/preview.py
```diff
@@ -97,6 +98,14 @@ def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> TemplateRespon
             data={"type": "page", "id": page_id, "title": getattr(page, "display_title", page.title)},
         )
 
+        # store the bundle id + page id in the session for private media checks
+        # TODO: should this be short-lived cache instead?
+        request.session["bundle-preview"] = {
```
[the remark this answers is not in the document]
RealOrangeOne: What's the purpose of the session side-channel? My concern is that the image requests are no longer stateless, and could behave strangely if a link is shared directly.

Cache doesn't feel quite right. Perhaps a signed cookie instead?

AdamHawtin (commented):
Testing this locally I can see it working as expected with draft, never published pages in bundle preview. However, when I add an image to an already live page and add that update page to a bundle, it's still not showing to the previewer and raising `PermissionDenied`. In debug I can see that the `referencing_pages` queryset is coming back empty for the already live pages.

cms/jinja2/templates/components/streamfield/image_block.html
```diff
@@ -2,25 +2,24 @@
 
 {# fmt:off #}
 <div>
-    {# Figure Title #}
```
sanjeevz3009: Why was this removed, and the below `{# Image #}` and `{# Figure Subtitle #}`? If there is a good reason, it would be good to be consistent and remove: `{# Notes #}`, `{# Download #}`, etc.?

sanjeevz3009: I just realised `{# Figure Title #}` and `{# Figure Subtitle #}` shifted in line next to the conditionals. Is this intentional?

sanjeevz3009 (commented):
Looks good so far. Will do another round of review later today.

cms/jinja2/templates/components/streamfield/image_block.html
```diff
@@ -2,25 +2,24 @@
...
+{%- if value.figure_title or value.figure_subtitle -%}
     <hgroup>
-        {% if value.figure_title %}
+        {%- if value.figure_title %}{# Figure Title #}
```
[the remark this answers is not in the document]
zerolab: https://jinja.palletsprojects.com/en/stable/templates/#whitespace-control

AdamHawtin (approved):
LGTM 👍

MebinAbraham (changes requested):
Mostly all are working as expected, some minor comments and an edge case that needs further consideration.

cms/bundles/views/preview.py
```diff
@@ -97,7 +98,27 @@ def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> TemplateRespon
...
+        response.set_signed_cookie(
+            "bundle-preview",
+            json.dumps(
+                {
+                    "bundle": bundle.pk,
+                    "page": page_id,
+                }
+            ),
+            max_age=settings.BUNDLE_PREVIEW_COOKIE_MAX_AGE,
+            salt=f"previewer-{request.user.pk}",
```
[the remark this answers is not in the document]
zerolab: Not really. Good call. 7852c764

cms/private_media/utils.py
```diff
@@ -19,3 +28,52 @@ def get_private_media_models() -> list[type[PrivateMediaMixin]]:
...
+    # the user can access the given asset if:
+    # - they can preview the bundle the page is in
+    # - and the given page
+    #   - references the asset, or
+    #   - the page is live, but also has unpublished changes. Once a page is live,
+    #     the reference index is only updated on publication, so any unpublished draft would not be tracked.
+    return (
+        in_referencing_pages
+        or Page.objects.filter(pk=preview_data["page"], live=True, has_unpublished_changes=True).exists()
+    ) and user_can_preview_bundle_by_id(user, preview_data["bundle"])
```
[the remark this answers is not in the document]
zerolab: 7852c764

MaciekBaron (approved):

cms/bundles/views/preview.py
```diff
@@ -97,7 +98,26 @@ def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> TemplateRespon
             data={"type": "page", "id": page_id, "title": getattr(page, "display_title", page.title)},
         )
 
-        return TemplateResponse(request, page.get_template(request), context)
+        response = TemplateResponse(request, page.get_template(request), context)
+
+        # Set a short-lived, signed cookie with the bundle and page information to be used by the
+        # private media serve view.
+        response.set_signed_cookie(
+            settings.BUNDLE_PREVIEW_COOKIE_NAME,
```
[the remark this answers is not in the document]
MaciekBaron: This is a good point, albeit we can't append the asset id (we don't know it at this point), we only know the bundle id and page id.
Meanwhile on the receiving end we don't know the page id (we check if the asset belongs to a page which we are allowed to preview).

The solution I came up with is to store a list of entries in the cookie:
* When you visit preview A, it stores bundle A and page X
* When you visit preview B, it stores bundle B and page Y

On the receiving end, we check if the asset it referenced by any of those pages (X, Y) and whether the user can access the bundles (A, B).


[c393dbd](https://github.com/ONSdigital/dis-wagtail/pull/525/commits/c393dbdb0a90153b3978d0e7327ee06c31fa5667)

cms/private_media/utils.py
```diff
@@ -19,3 +28,65 @@ def get_private_media_models() -> list[type[PrivateMediaMixin]]:
...
+            to_object_id=asset.pk,
+        )
+        .annotate(page_id=Cast("object_id", output_field=IntegerField()))
+        .filter(page_id=preview_data["page"])
+        .exists()
+    )
+    if in_referencing_page:
+        return True
+
+    can_access = False
```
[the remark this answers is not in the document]
MaciekBaron: Addressed as part of [c393dbd](https://github.com/ONSdigital/dis-wagtail/pull/525/commits/c393dbdb0a90153b3978d0e7327ee06c31fa5667)

cms/private_media/utils.py
```diff
@@ -19,3 +28,65 @@ def get_private_media_models() -> list[type[PrivateMediaMixin]]:
...
+def user_can_access_asset(
+    *,
+    request: HttpRequest,
+    user: User | AnonymousUser,
+    asset: PrivateMediaMixin,
+    permission_policy: CollectionOwnershipPermissionPolicy,
+) -> bool:
+    from cms.bundles.permissions import user_can_preview_bundle_by_id  # pylint: disable=import-outside-toplevel
+
+    if asset.is_public:
```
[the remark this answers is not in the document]
MaciekBaron: Seems impossible to circumvent this, _but_ probably good to return early to avoid any unnecessary calls, changed here: [37479a0](https://github.com/ONSdigital/dis-wagtail/pull/525/commits/37479a04724ac852da0c5063e58b54dba2555d5d)

cms/private_media/utils.py
```diff
@@ -19,3 +28,65 @@ def get_private_media_models() -> list[type[PrivateMediaMixin]]:
...
+    in_referencing_page = (
+        ReferenceIndex.objects.filter(
+            base_content_type__model=Page._meta.model_name.lower(),
+            base_content_type__app_label=Page._meta.app_label.lower(),
+            to_content_type__model=asset._meta.model_name.lower(),  # type: ignore[union-attr]
+            to_content_type__app_label=asset._meta.app_label.lower(),
+            to_object_id=asset.pk,
+        )
+        .annotate(page_id=Cast("object_id", output_field=IntegerField()))
+        .filter(page_id=preview_data["page"])
```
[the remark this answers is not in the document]
MaciekBaron: IMHO we should assume that the signed cookie is safe to accept as correct and valid.

cms/private_media/utils.py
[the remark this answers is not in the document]
MaciekBaron: Had to revert this change because I realised `cms.private_media.tests.test_serve_views.TestImageServeView.test_serve_private_image` tests if you can still view it if you are logged in as a super user on an external instance - AFAIK this should be impossible, so I need to clarify internally before implementing this.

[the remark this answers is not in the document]
MaciekBaron: Confirmed with the team, the pre-existing tests were incorrect. Updated them. [eacc405](https://github.com/ONSdigital/dis-wagtail/pull/525/commits/eacc405932da07e4e7205b83851016b696d82862)

BJacksonONS:
@MaciekBaron I'm having issues where the access cookie isn't getting set properly in safari. First thought it was part of the tracking and fingerprinting protection in private browsing but seems to be occurring in regular browsing mode too.

Can't replicate on other browsers though.

BJacksonONS:
> @MaciekBaron I'm having issues where the access cookie isn't getting set properly in safari. First thought it was part of the tracking and fingerprinting protection in private browsing but seems to be occurring in regular browsing mode too.
> 
> Can't replicate on other browsers though.

Sorry wasn't super familiar with Safari, turns out it doesn't set 'Secure' cookies on localhost.

cms/bundles/views/preview.py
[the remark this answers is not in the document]
BJacksonONS: Is there an argument for having secure be a setting? Possibly just `not settings.DEBUG`, as not all browsers play nice with secure cookies on localhost. Safari in particular but it's possible this could change.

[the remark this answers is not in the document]
MaciekBaron: @BJacksonONS I think that's fair, added this here: [3293911](https://github.com/ONSdigital/dis-wagtail/pull/525/commits/3293911d4006f9776ed86882361f7b1628ee9a55)

BJacksonONS (approved):
Looks good to me

cms/core/models/base.py
```diff
@@ -287,6 +287,17 @@ def analytics_content_theme(self) -> str | None:
...
+        ids = set()
+        stream_value = getattr(self, self.content_field_name)
+        # note: extract_references() is also used to populate the ReferenceIndex
+        references = stream_value.stream_block.extract_references(stream_value)
+        for model, object_id, _model_path, _content_path in references:
+            if model is asset_model:
+                ids.add(object_id)
+
+        return ids
+
```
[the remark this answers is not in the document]
MebinAbraham: Go for it?

cms/private_media/utils.py
```diff
@@ -19,3 +28,75 @@ def get_private_media_models() -> list[type[PrivateMediaMixin]]:
...
+            to_object_id=asset.pk,
+        )
+        .annotate(page_id=Cast("object_id", output_field=IntegerField()))
+        .filter(page_id__in=page_ids)
+        .exists()
+    )
+    if in_referencing_page:
+        return True
+
+    # Fall back to checking latest revisions for draft content
```
MebinAbraham: Can we make it clear that this is used only for Drafts of an already published page? If it was a draft but was never published, it would be in the reference index.

cms/private_media/utils.py
```diff
@@ -19,3 +28,75 @@ def get_private_media_models() -> list[type[PrivateMediaMixin]]:
...
+    cookie_data = request.get_signed_cookie(
+        settings.BUNDLE_PREVIEW_COOKIE_NAME,
+        default=None,
+        salt=f"previewer-{request.user.pk}",
+        max_age=settings.BUNDLE_PREVIEW_COOKIE_MAX_AGE,
+    )
+    if cookie_data is None:
+        return False
+
+    parsed = json.loads(cookie_data)
```
MebinAbraham: Non-blocker: Would benefit from defensive programming, in the event that the cookie was ever malformed, etc.
```py
try:
    parsed = json.loads(cookie_data)
except (TypeError, ValueError):
    return False
```

cms/private_media/utils.py
```diff
@@ -19,3 +28,75 @@ def get_private_media_models() -> list[type[PrivateMediaMixin]]:
...
+    # The cookie may contain multiple preview entries (one per tab/page) to avoid
+    # clobbering when multiple preview tabs are open simultaneously.
+    preview_entries: list[dict[str, int]] = parsed if isinstance(parsed, list) else [parsed]
+
+    # Filter to entries where the user can actually preview the bundle
+    authorized_page_ids = [
+        entry["page"] for entry in preview_entries if user_can_preview_bundle_by_id(user, entry["bundle"])
+    ]
```
MebinAbraham: Not going to block as I'm happy to revisit, but thought I'd raise.

This introduces repeated bundle lookups on every asset request, since we re-check each entry in `preview_entries` and `user_can_preview_bundle_by_id` does a DB lookup each time. In practice, this is probably only 1 to 2 entries, so I do not think it is a major performance problem, but it is still avoidable work in a hot path.

Do we need to re-authorise the bundle here at all? The cookie entry is only created after preview access has already been verified, the cookie is signed and user-scoped, and it is short-lived.

That said, because the cookie is rewritten on each preview and can contain multiple entries, removing this check entirely could let an older entry stay alive if the user keeps previewing other bundles within the expiry window.

So I think the options are:
1. bulk-fetch the bundles once here and keep the current re-authorisation behaviour
2. add a per-entry timestamp so entries expire independently, which would make it safer to rely on the cookie alone
3. Do nothing and accept it scales N+1, but might be okay for our usage. We should prob cache the perm check for a low ttl.

If we are comfortable with bounded stale access, option 2 may be enough. That would still be in line with our wider auth model, where Cognito tokens can remain valid until expiry, which for us is up to 15 minutes.

MebinAbraham (approved):
🎉

cms/private_media/utils.py
MaciekBaron: Added in [e0ea90a](https://github.com/ONSdigital/dis-wagtail/pull/525/commits/e0ea90ada4ce79da2ca1d42c12793bbf29d0c028)

~`TypeError` would only be raised if `cookie_data` was `None`, so I removed the check above and consolidated it within the `try`/`except` block.~ Actually mypy didn't like this so I refactored it slightly [80b56bf](https://github.com/ONSdigital/dis-wagtail/pull/525/commits/80b56bf0fdae2f81b812faaeca954d54974adf2b)

cms/private_media/utils.py
MaciekBaron: I think this is a good shout, and solution 2 seems sensible so that we can have unique lifetime for each entry - otherwise someone could keep regenerating the cookie to reset its signed lifetime.
Extreme edge case though.

cms/core/models/base.py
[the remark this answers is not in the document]
MaciekBaron: Done [1132f64](https://github.com/ONSdigital/dis-wagtail/pull/525/commits/1132f64dcf6edd519f9c30778edd04de7127adbc)

cms/private_media/utils.py
MaciekBaron: [72c27a3](https://github.com/ONSdigital/dis-wagtail/pull/525/commits/72c27a3d7e1cab85fb1bfb7487988df2e8c227a0)

cms/private_media/utils.py
MebinAbraham: Happy with option 2 also. Happy to tackle here or follow up?

MaciekBaron: I think I'll tackle it here so that it's done straight away.

cms/bundles/views/preview.py
```diff
@@ -57,6 +58,47 @@ def get_pages_in_bundle(self, bundle: Bundle) -> list[Page]:
 class PreviewBundlePageView(BundleContentsMixin, TemplateView):
     http_method_names: Sequence[str] = ["get"]
 
+    @staticmethod
+    def _update_preview_cookie(request: HttpRequest, response: TemplateResponse, bundle: Bundle, page_id: int) -> None:
```
MaciekBaron, line 62: I moved this to a static method to avoid `too-many-locals`

cms/private_media/utils.py
MaciekBaron: @MebinAbraham I added that logic here with a test: [51a94db](https://github.com/ONSdigital/dis-wagtail/pull/525/commits/51a94db9cd3c5c0946b6d0877d2172396294f21b)

cms/bundles/views/preview.py
```diff
@@ -57,6 +58,47 @@ def get_pages_in_bundle(self, bundle: Bundle) -> list[Page]:
...
+
+        now = int(time.time())
+        preview_entries = [e for e in preview_entries if e.get("page") != page_id and e.get("expires_at", 0) > now]
+        preview_entries.append(
+            {
+                "bundle": bundle.pk,
+                "page": page_id,
+                "expires_at": now + settings.BUNDLE_PREVIEW_COOKIE_MAX_AGE,
+            }
+        )
```
MebinAbraham, line 93: In theory, how big can this be before we hit the cookie value limit of 4096 bytes?

cms/private_media/utils.py
```diff
@@ -96,9 +97,14 @@ def user_can_access_asset(
...
-    # Filter to entries where the user can actually preview the bundle
+    # Filter to entries that haven't passed their per-entry expiry and where the user
+    # can actually preview the bundle. The per-entry expiry prevents the cookie from
+    # being refreshed indefinitely — each grant is pinned to when it was issued.
+    now = int(time.time())
     authorized_page_ids = [
-        entry["page"] for entry in preview_entries if user_can_preview_bundle_by_id(user, entry["bundle"])
+        entry["page"]
+        for entry in preview_entries
+        if entry.get("expires_at", 0) > now and user_can_preview_bundle_by_id(user, entry["bundle"])
```
MebinAbraham: Do we still need `user_can_preview_bundle_by_id`?. It would suggest it is safe to rely on the cookie itself, saves a lookup for each bundle for each asset. It means there is a 30 second residual access even when access is removed, which is consistent with our auth approach and low risk.

cms/bundles/views/preview.py
MaciekBaron: Good question:
```python
preview_entry = {'bundle': 1234, 'page': 4321, 'expires_at': 1776680670}

len(json.dumps(preview_entry).encode('utf-8'))  # 56

4096 / 56  # 73.14285714285714
```

We can fit roughly ~70 entries. But probably worth handling this gracefully, I introduced a form of "rate-limiting" here to prevent abuse: [781659d](https://github.com/ONSdigital/dis-wagtail/pull/525/commits/781659d3008a3ae842b09f8b413f7b1f30d7e8d0)

(Note I am not returning HTTP 429 so that Wagtail can handle this better)

cms/private_media/utils.py
MaciekBaron: I'm happy to remove it. It is possible to exploit but it's an extreme edge-case: there's a cooldown period of about 30 seconds when a user is removed from a bundle when they can still preview the page, but if we are happy with this, we can remove the check.

cms/bundles/views/preview.py
BJacksonONS: Incredibly minor nit-pic, we need to include the size of the signature in the maths.

MaciekBaron: That's a fair point; I set the max to 20 though so should be well within available space.

cms/private_media/utils.py
MaciekBaron: Removed in [37dce16](https://github.com/ONSdigital/dis-wagtail/pull/525/commits/37dce16e1c74169fcc0ed4f6a5537950d48ec2cd)

BJacksonONS:
Re tested with the newest changes.

Rate limiting worked when I set it to a lower value. Wasn't able to hit 20 entries even when intentionally navigating as quickly as I could between pages on a bundle of 21, so think the current default is reasonable.

cms/core/models/base.py
```diff
@@ -287,6 +287,13 @@ def analytics_content_theme(self) -> str | None:
         parent_theme = page_topic.get_base_parent()
         return cast(str, parent_theme.title)
 
+    def get_referenced_asset_ids(self, asset_model: models.Model) -> set[str]:
```
MebinAbraham: Here we are passing in the model class not instance:
```suggestion
    def get_referenced_asset_ids(self, asset_model: type[models.Model]) -> set[str]:
```

cms/private_media/utils.py
```diff
@@ -19,3 +29,80 @@ def get_private_media_models() -> list[type[PrivateMediaMixin]]:
...
+    except json.JSONDecodeError:
+        return False
+
+    # The cookie may contain multiple preview entries (one per tab/page) to avoid
+    # clobbering when multiple preview tabs are open simultaneously.
+    preview_entries: list[dict[str, int]] = parsed if isinstance(parsed, list) else [parsed]
+
+    # Filter to entries that haven't passed their per-entry expiry and where the user
+    # can actually preview the bundle. The per-entry expiry prevents the cookie from
+    # being refreshed indefinitely — each grant is pinned to when it was issued.
```
MebinAbraham: Comment now out of date but also worth mentioning the design choice of not checking the bundle access.

```suggestion
    # Trust non-expired preview grants from the signed, user-scoped cookie.
    # We intentionally do not re-check current bundle membership here: access is
    # granted based on the original preview authorisation and remains valid until
    # the entry's per-item expiry. The per-entry expiry prevents unrelated previews
    # from extending older grants indefinitely.
```

MebinAbraham (approved):

MebinAbraham (approved):

sanjeevz3009 (commented):
LGTM!

PR #533, opened 2026-03-13
Workflow tweaks: admin dashboard panel updates

zerolab:

This PR updates the "Awaiting your review" dashboard panel to exclude objects that are ready to publish, and adds a new panel that show the latest 10 pages that are ready to publish

[Relevant discussion](https://officefornationalstatistics.atlassian.net/browse/CMS-1043?focusedCommentId=731127)

### How to review

- add 3 pages and take them through the moderation workflow
  - one that is only in review, one that is ready to publish (but not in a bundle), and another that is ready to publish (but in a bundle)
- go to the admin dashboard
  - "awaiting your review" should only contain the first one
  - there should be a new panel titled "Pages ready to publish" that contains the other two. The first one has an action menu, the other doesn't but has a "in a bundle" label

Screenshot 2026-03-13 at 15 27 57

### Follow-up Actions

Get final decision on this.

### DIFF
--- a/cms/workflows/action_menu.py
+++ b/cms/workflows/action_menu.py
@@ -1,6 +1,7 @@
 from typing import Any
 
 from wagtail.admin.action_menu import ActionMenuItem
+from wagtail.admin.action_menu import SubmitForModerationMenuItem as CoreSubmitForModerationMenuItem
 
 
 class UnlockWorkflowMenuItem(ActionMenuItem):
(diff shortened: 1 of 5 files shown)


MaciekBaron (changes requested):

MebinAbraham:
Will come back to this, currently pending requirement calarifications.

PR #549, opened 2026-03-25
Add workaround for live button URL in admin panel

MaciekBaron:

This PR addresses the bug raised in CMS-1140 and adds a workaround for the issue.

When `CMS_USE_SUBDOMAIN_LOCALES` is disabled (the default for local development), the admin "Live" button on pages pointed to the wrong host (e.g. http://cy.web.ons.orb.local:8000/) because `_get_site_root_paths` fell back to Wagtail's default `Site.get_site_root_paths()`, which returned all `Site` records from the database, including those with production/staging hostnames loaded from `CMS_HOSTNAME_LOCALE_MAP` in `.env`. Note that other "Live" buttons would work correctly.

The workaround routes the non-subdomain code path through `get_mapped_site_root_paths(default_site_only=True)` instead, which filters to the default site and uses `WAGTAILADMIN_BASE_URL` as the root URL. This ensures the Live button points to http://localhost:8000 (or whatever you use) in development, regardless of what hostnames are stored in the database. This works if and only if one domain is being used to access the site (which should be true when `CMS_USE_SUBDOMAIN_LOCALES` is disabled).

### How to review

1. Ensure that `CMS_USE_SUBDOMAIN_LOCALES` is disabled
2. Go to an admin view such as `/admin/pages/3/`
3. Click or inspect the "Live" button in the top right corner
4. Confirm that the correct URL is being used (e.g. "localhost:8000", not the subdomain)
5. Change your local settings to enable `CMS_USE_SUBDOMAIN_LOCALES` and reload the server
6. Return to the same admin view (or refresh the page)
7. Confirm that the correct URL is being used (e.g. "web.ons.orb.local:8000", the subdomain)

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

Ensure this change introduces no regressions.

### DIFF
--- a/cms/core/models/base.py
+++ b/cms/core/models/base.py
@@ -292,9 +292,6 @@ class BasePage(PageLDMixin, ListingFieldsMixin, SocialFieldsMixin, Page):  # typ
 
         Note: while the method signature implies we get a request object, in reality it can be a model too.
         """
-        if not settings.CMS_USE_SUBDOMAIN_LOCALES:
-            return cast(list["SiteRootPath"], super()._get_site_root_paths(request=request))
-
         cache_object = request if request is not None else self
         try:
             # pylint: disable=protected-access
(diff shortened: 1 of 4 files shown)


sanjeevz3009 (approved):
Tested it locally, works well! Thanks, @MaciekBaron

MaciekBaron:
Closing for now, might re-open at a later date

PR #553, opened 2026-03-26
Header: Main Menu Translation Config Bugfix - CMS-1056

sanjeevz3009:

The `Menu` button in the page header is not currently translating when a website user toggles to the Welsh language.

We do have a `Menu` label correctly marked for translation in our Jinja template; however, we have a mistake in our header component config here: [dis-wagtail/cms/jinja2/templates/base.html at d73dca99e491afd4f6a637560e64cb7a235be3d8 · ONSdigital/dis-wagtail](https://github.com/ONSdigital/dis-wagtail/blob/d73dca99e491afd4f6a637560e64cb7a235be3d8/cms/jinja2/templates/base.html#L63) 

Currently, it is falling back on the default button text, which has no translation.

The key `toggleNavButton` is incorrect; it should be `toggleMenuButton`. This has been corrected on this PR, so our value will be passed through as expected, and translation will work the same as all other page furniture.

This PR also makes sure we are aligning with the ONS DS defaults for the menus, as there are no requirements to change them otherwise. ONS DS docs don't show the defaults, see: https://github.com/search?q=repo%3AONSdigital%2Fdesign-system%20toggleMenuButton&type=code

I also made a tiny update to the `markup.py` in how we mark computed values for translations:
- Without this change, `Download CSV...` is being removed from the `.po` file when we run `make makemessages` command.

I also fixed this test failure: https://github.com/ONSdigital/dis-wagtail/actions/runs/23548629557/job/68556380726

To understand the changes, see the Django docs on:
<details>
<summary>Named-string interpolation</summary>

```
The strings you pass to _() or gettext() can take placeholders, specified with Python’s standard named-string interpolation syntax. Example:

def my_view(request, m, d):
    output = _("Today is %(month)s %(day)s.") % {"month": m, "day": d}
    return HttpResponse(output)

This technique lets language-specific translations reorder the placeholder text.
For example, an English translation may be "Today is November 26.", while a Spanish
translation may be "Hoy es 26 de noviembre." – with the month and the day placeholders swapped.

For this reason, you should use named-string interpolation (e.g., %(day)s) instead of
positional interpolation (e.g., %s or %d) whenever you have more than a single parameter.
If you used positional interpolation, translations wouldn’t be able to reorder placeholder text.

Since string extraction is done by the xgettext command, only syntaxes supported by gettext
are supported by Django.
Python [f-strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings)
cannot be used directly with gettext functions because f-string expressions are evaluated
before they reach gettext. This means _(f"Welcome {name}") will not work as expected,
as the variable is substituted before translation occurs. Instead, use named-string interpolation:
```

</details>

https://docs.djangoproject.com/en/6.0/topics/i18n/translation/

### How to review

- Check the JIRA card and make sure all the changes are in place.
- Cross-check with the ONS DS docs: https://service-manual.ons.gov.uk/design-system/components/header to see that the config is correctly used/ set.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/cms/core/blocks/markup.py
+++ b/cms/core/blocks/markup.py
@@ -226,7 +226,7 @@ class ONSTableBlock(TinyTableBlock):
 
         return {
             "title": _("Download this table"),
-            "itemsList": [{"text": f"{_('Download CSV')}{size_suffix}", "url": csv_url}],
+            "itemsList": [{"text": _("Download CSV %(size)s") % {"size": size_suffix}, "url": csv_url}],
         }
 
     @staticmethod
--- a/cms/core/static/js/blocks/headline-figures-item-block.js
+++ b/cms/core/static/js/blocks/headline-figures-item-block.js
@@ -5,7 +5,7 @@ class HeadlineFiguresItemBlock extends window.wagtailStreamField.blocks.StructBl
     try {
       const usedFigures = document.getElementById('figures-used-by-ancestor');
       this.usedFigures = JSON.parse(usedFigures.innerText) || [];
-    } catch (e) {
+    } catch {
       this.usedFigures = [];
     }
   }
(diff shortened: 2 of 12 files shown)


MaciekBaron (changes requested):

MaciekBaron (approved):

MebinAbraham (approved):

MebinAbraham:
TODO: Recreate against main once https://github.com/ONSdigital/dis-wagtail/tree/optimise-page-config merges.

PR #564, opened 2026-04-01
Sync slugs after page is published

MaciekBaron:

This PR relates to CMS-1155. After a page is published, if it has translations which are aliases, it ensures that the aliases use the same slug as the published page.

This is related to two Wagtail issues:
https://github.com/wagtail/wagtail/issues/7168
https://github.com/wagtail/wagtail/issues/14102

### How to review

1. Create a page in English
2. Ensure that an alias was created in the Welsh page tree
3. Change the slug of the original page
4. Ensure that the alias was not updated for the Welsh alias
5. Publish the English page
6. Ensure that the alias _was_ updated for the Welsh alias

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

Monitor changes in core and adapt our code appropriately.

### DIFF
--- a/cms/core/signal_handlers.py
+++ b/cms/core/signal_handlers.py
@@ -43,7 +43,11 @@ def reload_logging_config(*, setting: str, **kwargs: Any) -> None:
 
 
 def sync_alias_translation_slugs(sender: Any, instance: Page, **kwargs: Any) -> None:  # pylint: disable=unused-argument
-    if instance.locale.language_code != settings.LANGUAGE_CODE:
+    if not instance.locale_id or instance.locale.language_code != settings.LANGUAGE_CODE:
+        return
+
+    # Don't attempt to sync slugs for children of the root page
+    if instance.get_parent().is_root():
         return
 
     updatable_aliased_translations = (
--- a/cms/core/signal_handlers.py
+++ b/cms/core/signal_handlers.py
@@ -8,6 +8,8 @@ from django.utils.log import configure_logging
 from wagtail.models import DraftStateMixin, Page, Revision
 from wagtail.signals import page_published
 
+HOME_PAGE_DEPTH = 2
+
 
 def remove_go_live_seconds(
     sender: Any,  # pylint: disable=unused-argument
@@ -47,7 +49,7 @@ def sync_alias_translation_slugs(sender: Any, instance: Page, **kwargs: Any) -> 
         return
 
     # Don't attempt to sync slugs for children of the root page
-    if instance.get_parent().is_root():
+    if instance.depth == HOME_PAGE_DEPTH:
         return
 
     updatable_aliased_translations = (
@@ -55,6 +57,7 @@ def sync_alias_translation_slugs(sender: Any, instance: Page, **kwargs: Any) -> 
     )
     for translation in updatable_aliased_translations:
         translation.slug = instance.slug
+        # Use save() to trigger any related functionality
         translation.save()
 
 
--- a/cms/core/signal_handlers.py
+++ b/cms/core/signal_handlers.py
@@ -6,7 +6,7 @@ from django.core.signals import setting_changed
 from django.db.models.signals import pre_save
 from django.utils.log import configure_logging
 from wagtail.models import DraftStateMixin, Page, Revision
-from wagtail.signals import page_published
+from wagtail.signals import page_slug_changed
 
 HOME_PAGE_DEPTH = 2
 
@@ -68,6 +68,6 @@ def register_signal_handlers() -> None:
 
     pre_save.connect(remove_approved_go_live_seconds, sender=Revision)
 
-    page_published.connect(sync_alias_translation_slugs)
+    page_slug_changed.connect(sync_alias_translation_slugs)
 
     setting_changed.connect(reload_logging_config)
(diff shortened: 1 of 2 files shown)


AdamHawtin (approved):

AdamHawtin (approved):

cms/core/signal_handlers.py
```diff
@@ -6,6 +6,9 @@
 from django.db.models.signals import pre_save
 from django.utils.log import configure_logging
 from wagtail.models import DraftStateMixin, Page, Revision
+from wagtail.signals import page_slug_changed
+
+HOME_PAGE_DEPTH = 2
```
sanjeevz3009, line 12: A comment on what this is, why it's needed, etc., would be nice here? @MaciekBaron If you can kindly bless us with a short comment =)

cms/core/signal_handlers.py
```diff
@@ -41,11 +44,30 @@ def reload_logging_config(*, setting: str, **kwargs: Any) -> None:
         configure_logging(settings.LOGGING_CONFIG, settings.LOGGING)
 
 
+def sync_alias_translation_slugs(sender: Any, instance: Page, **kwargs: Any) -> None:  # pylint: disable=unused-argument
+    if not instance.locale_id or instance.locale.language_code != settings.LANGUAGE_CODE:
```
sanjeevz3009: A short comment here would be nice.

Some inspiration is below:
```
This guard exits unless the changed page is in the default language.

Why: the default-locale page is treated as the source of truth, and this avoids
recursive/ ping-pong updates when translated aliases are saved.
I.e., No ping-pong write back into EN or repeated re-sync cycle.
```

sanjeevz3009 (commented):
Did a code review, LGTM!

Left a couple of comments around adding _comments_... I am conscious of it recently since it's been you, me, Dan, Adam and Mebin, so I guess naturally, we weren't conscious of it. We drifted away from certain things in code reviews, including adding comments, because we know what we are doing + have a good overall context of the codebase and project.

Considering we are going to onboard permanent ONS devs and other devs in the future, it would be a nice time to focus more on the readability of our files for new, inexperienced permanent starters, students, etc.

sanjeevz3009:
@MaciekBaron, What are we planning to do with existing pages where there is an English page and a Welsh alias of that page, but the English page slug has been updated, and now it's not in sync with the Welsh alias?

- We can write a migration to clean it up.
- Is it worth the effort, considering we aren't in prod right now? Probably not.
- It might be nice to have.
- Happy for it to be a follow-up if we think it's suitable/ or want to do it while we are at it.

I think we can leave it for now, I don't see a huge reason we need to re-sync it, except it would be _nice_.

What are your thoughts @MaciekBaron?

sanjeevz3009:
**Out of scope:**

@MebinAbraham @AdamHawtin @MaciekBaron 

On another note, I may have called this out, or maybe we are already aware, and I forgot.

If we have a page in draft and then open the draft page in external preview -> switch to Welsh -> `404` -> as post subdomain integration (I think) -> it goes or tries to go to the live version when you try to switch to Welsh -> but the live version/ Welsh alias page doesn't exist yet.
- This is some strange UX as if the user navigates to an English page in external preview and wants to see how the Welsh alias page looks -> since the Welsh page is not live as the English page is not live also -> `404`
- An English page is already live -> so the Welsh alias page is also live -> now the user makes some changes, and the English page is in draft -> they open the page on external preview -> switch to Welsh -> now it will take them to the live Welsh page -> confusing, misleading UX.

cms/core/signal_handlers.py
MebinAbraham: Q: Does this make a query to get locale?. If so, doing the check below for the depth first would be preferred.

cms/core/signal_handlers.py
MaciekBaron: Added comment [bef2782](https://github.com/ONSdigital/dis-wagtail/pull/564/commits/bef27826293b1c09f00ed89ac8345ddac358f6cd)

cms/core/signal_handlers.py
MaciekBaron: @sanjeevz3009 Added comment

@MebinAbraham I moved the block and also changed it to use `instance.locale_id != Locale.get_default().pk` - the latter part should get cached after the first call

[bef2782](https://github.com/ONSdigital/dis-wagtail/pull/564/commits/bef27826293b1c09f00ed89ac8345ddac358f6cd)

cms/core/signal_handlers.py
```diff
@@ -41,11 +45,32 @@ def reload_logging_config(*, setting: str, **kwargs: Any) -> None:
...
+    if not instance.locale_id or instance.locale_id != Locale.get_default().pk:
+        return
+
+    updatable_aliased_translations = (
+        instance.get_translations().filter(alias_of__isnull=False).exclude(slug=instance.slug)
+    )
+    for translation in updatable_aliased_translations:
+        translation.slug = instance.slug
+        # Use save() to trigger any related functionality
+        translation.save()
```
MebinAbraham, line 64: Since only slug is changing?

```suggestion
        translation.save(update_fields=["slug"])
```

MebinAbraham (approved):

cms/core/signal_handlers.py
MaciekBaron: This won't work as intended as we want to trigger a full save (hence the comment above)

PR #568, opened 2026-04-08
CMS-859 Disable page privacy settings for CMS users

AdamHawtin:

We need to disable Wagtail's page privacy features. This PR achieves this by overriding the `can_set_view_restrictions` permission test in the `BasePagePermissionTester` to always return false, preventing any CMS admin UI user from setting view restrictions for pages.

It also includes a migration to remove any existing page view restrictions as we are not supporting them and CMS users will no longer be able to remove them themselves after this change.

### How to review

Run this branch, then try accessing the page privacy settings for any page in the Wagtail admin UI, as an admin, publishing admin and publishing officer. You should not be presented with the options to change page privacy settings.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


AWitcherONS (approved):
Unit tests pass
page privacy not seen

MebinAbraham (approved):

cms/core/permission_testers.py
```diff
@@ -125,6 +125,12 @@ def can_delete(self, ignore_bulk: bool = False) -> bool:
         can_delete: bool = super().can_delete(ignore_bulk=ignore_bulk)
         return can_delete
 
+    def can_set_view_restrictions(self) -> bool:
+        """Overrides teh core can_set_view_restrictions to always return False.
```
sanjeevz3009:
```suggestion
        """Overrides the core can_set_view_restrictions to always return False.
```

MaciekBaron (approved):
LGTM

Question: Should we consider adding a migration removing any existing restrictions, for cleanup purposes?

sanjeevz3009:
> LGTM
> 
> Question: Should we consider adding a migration removing any existing restrictions, for cleanup purposes?

@MaciekBaron @AdamHawtin It's mentioned in the JIRA card:

```
We may need to consider doing a mop up of any pages that
have been made private in exisitng environments, since users
would not be able to make them public again after this change.
```

Maybe it needs a separate card as a follow-up? If so good to have a card created and chucked into the backlog.

sanjeevz3009 (approved):
Tested it against all page types. Works well. LGTM!

AdamHawtin:
> LGTM
> 
> Question: Should we consider adding a migration removing any existing restrictions, for cleanup purposes?

Good point, I've added a migration in [d1da0aa](https://github.com/ONSdigital/dis-wagtail/pull/568/commits/d1da0aa213574920097311a60ec8534efcf1d31e).

cms/core/migrations/0015_delete_existing_page_view_restrictions.py
```diff
@@ -0,0 +1,19 @@
...
+    PageViewRestriction.objects.all().delete()
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("core", "0014_allow_officers_to_lock"),
+    ]
+
+    operations = [
+        migrations.RunPython(delete_existing_page_view_restrictions),
```
sanjeevz3009: Do we want?

If we make this change, it would imply: `The intent explicit: rollback is allowed through this migration step, but it will not restore deleted restrictions.`

Else: `No reverse specified:
Django treats it as irreversible.
Trying to migrate backwards past it will raise an IrreversibleError.`

- 0015 must be unapplied before Django can get to 0014, 0013, etc.
- If 0015 has no reverse, rollback stops at 0015 with IrreversibleError.
- If 0015 uses noop reverse, Django can step back to 0014, then continue to 0013/0012, etc., as long as each earlier migration is itself reversible.

```suggestion
        migrations.RunPython(delete_existing_page_view_restrictions), reverse_code=migrations.RunPython.noop,
```

sanjeevz3009 (changes requested):

MebinAbraham (approved):
Other than Sanjeev's comment. 👍

cms/core/permission_testers.py
```diff
@@ -125,6 +125,12 @@ def can_delete(self, ignore_bulk: bool = False) -> bool:
         can_delete: bool = super().can_delete(ignore_bulk=ignore_bulk)
         return can_delete
 
+    def can_set_view_restrictions(self) -> bool:
+        """Overrides the core can_set_view_restrictions to always return False.
+        This functionally disables access to page privacy features for all CMS users, regardless of permissions.
+        """
+        return False
```
MebinAbraham, line 136: Nit and non-blocker. If we make this a setting that defaults to false, it makes it easier to allow this in future while still allowing testing of some of the fixes Sanjeev raised, with overrides in tests.

AdamHawtin: Good shout, I'll change it to use a setting to toggle between returning `False` and calling super for the default so we can easily revert to default behaviour.

cms/core/permission_testers.py
```diff
@@ -126,10 +126,13 @@ def can_delete(self, ignore_bulk: bool = False) -> bool:
...
 
     def can_set_view_restrictions(self) -> bool:
-        """Overrides the core can_set_view_restrictions to always return False.
-        This functionally disables access to page privacy features for all CMS users, regardless of permissions.
+        """Overrides the core can_set_view_restrictions to always return False if the
+        CMS_DISABLE_USER_PAGE_PRIVACY_ACCESS setting is True, functionally disabling access to page privacy features
+        for all CMS users, regardless of permissions.
         """
-        return False
+        if settings.CMS_DISABLE_USER_PAGE_PRIVACY_ACCESS:
```
MebinAbraham: Few nitpicks (non-blocker):
- I would suggest something like `CMS_PAGE_PRIVACY_CONTROLS_ENABLED` (we currently use adjectives mostly)
- Do we need "user"?
- General, I find avoiding negative better for clarity but in this case maybe disable is better when the default behaviour is enable? Not fussed just thinking out loud.

We do need to review all our env vars at somepoint and tidy them up and prefix like you have done with `CMS_` 👍

MebinAbraham (approved):

cms/core/permission_testers.py
AdamHawtin: Good points, happy to stick to a positive toggle for clarity, addressed in [6c2c9c2](https://github.com/ONSdigital/dis-wagtail/pull/568/commits/6c2c9c2d9b58d9b1fa26405fefc317bd616c287e)

MebinAbraham (approved):

MaciekBaron (approved):

sanjeevz3009 (approved):
LGTM!

PR #574, opened 2026-04-09
Upgrade Wagtail to 7.3, update dependencies

MaciekBaron:

This PR addresses CMS-1126 which requested an upgrade of Wagtail, relevant python dependencies, and node.

* Wagtail has been upgraded to 7.3.1 (see [release notes](https://docs.wagtail.org/en/stable/releases/7.3.html))
* A few Python dependencies have been updated alongside this upgrade
* Updated `InformationPageImageBlockRenderingTests` tests to correctly use the `file.url` as indicated in the comments
* `poetry check` reports no issues
* `make lint` reports no issues
* `make mypy` reports no issues
* Node version has been upgraded to latest LTS (v24, see [release notes](https://nodejs.org/en/blog/release/v24.14.1))
* `npm audit` reports 0 vulnerabilities

**Note**: I have disabled Wagtail's new auto save feature until we had time to discuss and verify that we would like it enabled.

### How to review

Ensure that the code functions correctly with no regressions.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

* Look into enabling auto-save by setting `WAGTAIL_AUTOSAVE_INTERVAL` to our preferred value (the default is `500`)
* Consider using Django's inbuilt CSP functionality and drop the dependency


pyproject.toml
```diff
@@ -12,26 +12,26 @@ requires-poetry = ">=2.2"
 
 [tool.poetry.dependencies]
 python = "^3.14"
-django = "~5.2.11"
-wagtail = "~7.2.2"
+django = "~5.2.13"
```
MebinAbraham: Can we not go to 6 now since 7.3 adds official support for Django 6?

cms/core/tests/test_blocks.py
```diff
@@ -1130,8 +1130,8 @@ def test_renders_small_and_large_renditions_and_alt_text(self):
 
         # onsImage macro outputs src/srcset with both specific rendition URLs
         # Use the underlying file URLs to be agnostic of serve method
-        self.assertContains(response, self.small.url)
-        self.assertContains(response, self.large.url)
+        self.assertContains(response, self.small.file.url)
```
sanjeevz3009, line 1133: Do we know the deeper reason for this change? I ask because I remember playing around with using `.file` and not using it when I wrote these tests. So I wonder what this upgrade does that makes these tests fail?

pyproject.toml
BJacksonONS: Will need to remove the --parallel flag from tests, at least temporarily, for tests to run properly if we do update to v6 [CMS-1163](https://officefornationalstatistics.atlassian.net/browse/CMS-1163).

sanjeevz3009: The release notes do say `Formalized support for Django 6.0 (Pravin Kamble)`.

https://docs.wagtail.org/en/latest/releases/7.3.html#maintenance

sanjeevz3009: So technically speaking, we can go to 6 now. Unless I am missing something.

MaciekBaron: Sorry, I assumed we wanted to stick to the LTS version, I must have misunderstood/misread the request: "Django was upgraded to the compatible LTS or latest minor supported by Wagtail version" - so the latest minor would be 6.0.4, happy to upgrade to that

BJacksonONS: If updating to v6 would it also be worth adding some followup items to assess moving to Django's built in CSP module and dropping a dependency?

MebinAbraham: @MaciekBaron Good point on LTS, I'll get back to you after I have some discussions.

@BJacksonONS that was the plan on https://officefornationalstatistics.atlassian.net/browse/CMS-994

sanjeevz3009: > Will need to remove the --parallel flag from tests, at least temporarily, for tests to run properly if we do update to v6 [CMS-168](https://officefornationalstatistics.atlassian.net/browse/CMS-168).

Docs: https://docs.djangoproject.com/en/6.0/releases/6.0/#tests

cms/core/tests/test_blocks.py
MaciekBaron: I spent some time debugging this and couldn't pinpoint the exact change yet, but I think something has perhaps changed which have influenced what the `published` signal sees (i.e. the image reference via `ReferenceIndex`) but that's just a hunch for now. However the comment does seem to imply that `.file` was intended to be used there in the first place.

sanjeevz3009:
@MaciekBaron @MebinAbraham The release notes say:

```
Wagtail’s default image quality settings are now optimized
for the needs of a wider number of sites, and consistent between image formats.
The default settings are now lower for JPEG and AVIF.
```

So we may want to tweak the quality settings, but it's probably one for a follow-up PR/ card: https://docs.wagtail.org/en/stable/topics/images.html#image-quality

Maybe we might already have a related JIRA card, which we can chuck this in, around image quality and what the business may want. I don't remember image quality being called out when I did the `ImageBlock` work.

cms/core/tests/test_blocks.py
BJacksonONS: I think this is related to wagtail using `django-tasks` v12 rather than v9, and the old test was actually a false positive.

`self.small` and `self.large` are getting saved to the test class instance at create time with the privacy setting as private. Previously Django tasks was [waiting until a transaction was committed](https://github.com/RealOrangeOne/django-tasks/blob/37dfb4c30da96dcaf5cc4f2c7aae0ea0cde75b3c/django_tasks/backends/immediate.py#L113) to run the publish image signal, which would never actually happen during the test, but now [it always runs immediately](https://github.com/RealOrangeOne/django-tasks/blob/d3666f039378451a4359f2219e9ccde248901615/django_tasks/backends/immediate.py#L111).

So before we were testing `.url` of a private image object against the private image fetched from the database, but now we're testing our in memory private image vs the public image fetched from the DB in the route.

So looking at `.file.url` is probably asserting the correct behaviour (public url is shown), but also we're still sort of leaving a potential bug lurking by letting the old, un-updated image model instance stay attached to the class.

MebinAbraham:
> @MaciekBaron @MebinAbraham The release notes say:
> 
> ```
> Wagtail’s default image quality settings are now optimized
> for the needs of a wider number of sites, and consistent between image formats.
> The default settings are now lower for JPEG and AVIF.
> ```
> 
> So we may want to tweak the quality settings, but it's probably one for a follow-up PR/ card: [docs.wagtail.org/en/stable/topics/images.html#image-quality](https://docs.wagtail.org/en/stable/topics/images.html#image-quality)
> 
> Maybe we might already have a related JIRA card, which we can chuck this in, around image quality and what the business may want. I don't remember image quality being called out when I did the `ImageBlock` work.

No action is needed at the moment, but yes we need to ensure the defaults are sufficient for ONS use. The action wouldn't be to make image quality configurable by users; it would just be to increase the default if deemed necessary. The new default should give size improvements without sacrificing much, at least any perceived reduction in quality. 

- https://onswebsite.slack.com/archives/C07356R644A/p1770122358330719?thread_ts=1770121837.243949&cid=C07356R644A
- https://wagtail.org/blog/40-smaller-images-same-quality/

cms/core/tests/test_blocks.py
MaciekBaron: Thanks for digging into this. I think for now we can keep the test as it is with the intended behaviour, and perhaps create a follow-up card to investigate if there is a potential edge-case bug here.

pyproject.toml
MebinAbraham: @MaciekBaron we are fine to upgrade to Django 6.

cms/settings/base.py
```diff
@@ -114,6 +114,7 @@
     "django_extensions",
     "django.contrib.auth",  # Wagtail requires the auth app be installed, even if it's not used.
     "django.contrib.contenttypes",
+    "django.contrib.postgres",
```
MaciekBaron, line 117: See: https://docs.djangoproject.com/en/6.0/releases/6.0/#django-contrib-postgres

pyproject.toml
MaciekBaron: Upgraded to 6.0 here: [b2314cb](https://github.com/ONSdigital/dis-wagtail/pull/574/commits/b2314cb32ee7e279083bd5e282ff59187436433c)

pyproject.toml
```diff
@@ -12,7 +12,7 @@ requires-poetry = ">=2.2"
 
 [tool.poetry.dependencies]
 python = "^3.14"
-django = "~5.2.13"
+django = "~6.0"
```
MebinAbraham: I know the lock is using 6.0.4, but we need to be on a patch version by policy, so let's set this to `~6.0.4` to avoid confusion and prevent any changes back to 6.0.0

BJacksonONS: Was going to say this

MaciekBaron: Updated: [5f9e5c0](https://github.com/ONSdigital/dis-wagtail/pull/574/commits/5f9e5c07fc0320372402cc63ada75a72736e43ba)

sanjeevz3009:
> > @MaciekBaron @MebinAbraham The release notes say:
> > ```
> > Wagtail’s default image quality settings are now optimized
> > for the needs of a wider number of sites, and consistent between image formats.
> > The default settings are now lower for JPEG and AVIF.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So we may want to tweak the quality settings, but it's probably one for a follow-up PR/ card: [docs.wagtail.org/en/stable/topics/images.html#image-quality](https://docs.wagtail.org/en/stable/topics/images.html#image-quality)
> > Maybe we might already have a related JIRA card, which we can chuck this in, around image quality and what the business may want. I don't remember image quality being called out when I did the `ImageBlock` work.
> 
> No action is needed at the moment, but yes we need to ensure the defaults are sufficient for ONS use. The action wouldn't be to make image quality configurable by users; it would just be to increase the default if deemed necessary. The new default should give size improvements without sacrificing much, at least any perceived reduction in quality.
> 
> * https://onswebsite.slack.com/archives/C07356R644A/p1770122358330719?thread_ts=1770121837.243949&cid=C07356R644A
> * https://wagtail.org/blog/40-smaller-images-same-quality/

@MebinAbraham 

Yes, I didn't mean users. We need to configure more based on the user needs, which I don't think, if I remember correctly, we had any or were given any/ never came up in any conversations.

pyproject.toml
sanjeevz3009: I was also going to say this. Thanks, Maciej!

pyproject.toml
```diff
@@ -12,26 +12,26 @@ requires-poetry = ">=2.2"
 
 [tool.poetry.dependencies]
 python = "^3.14"
-django = "~5.2.11"
-wagtail = "~7.2.2"
+django = "~6.0.4"
+wagtail = "~7.3.0"
```
MebinAbraham: Same for this sorry `7.3.1`

MebinAbraham (approved):

pyproject.toml
MaciekBaron: [7588bc1](https://github.com/ONSdigital/dis-wagtail/pull/574/commits/7588bc11333367273f972b47a40f20815c9f471f)

sanjeevz3009 (approved):
LGTM! Tested some of the core features locally, including bundles, bundle publish, etc. All is good.

PR #575, opened 2026-04-09
feat(privacy): enable privacy settings integration with Florence SSO

sanjeevz3009:

This PR is a initative taken after reviewing #568 where I uncovered the below issues, which this PR aims to address.

**This PR addresses several issues:**
- First one being, currently when a user sets a page behind a password protection using the page privacy setting -> when the user visits the live URL -> `404`. -> when it's meant to show the `password_required.html` template. We have missed this during the `NonTrailingSlashRedirectMiddleware` work.
- I have also improved the UI/ UX for the `password_required.html` template and how it surfaces the error to the user etc.
  - This has been fixed on this PR, and the front-end template has been improved.
- The second enhancement is around when we use the privacy settings to have a page behind `Private, accessible to any logged-in users` or `Private, accessible to users in specific groups` options, then we render the `login_page.html` template, and the logic behind it would not work due to `WAGTAIL_CORE_ADMIN_LOGIN_ENABLED=False`/ Wagtail core login being turned off in the deployed environments. + This template UI design/ UX needs improving.
  - I have fixed the core issue where, when users use the options to have a page behind a login wall, such as a user must be logged in or logged in and be part of a certain group -> then it will redirect the user to the Florence SSO page, and the user can be authenticated using that.

### How to review

- Read the following doc: https://docs.wagtail.org/en/stable/advanced_topics/privacy.html
- Read the `page_privacy.md`
- Enable `CMS_PAGE_PRIVACY_CONTROLS_ENABLED`, `CMS_COLLECTION_PRIVACY_CONTROLS_ENABLED` and `AWS_COGNITO_LOGIN_ENABLED`.
- Disable `WAGTAIL_CORE_ADMIN_LOGIN_ENABLED`
- Make sure the above issues are resolved:
  - No 404 when visiting a page that's password-protected
  - When Congito is enabled, you can successfully log in and view a page behind `Private, accessible to any logged-in users` or `Private, accessible to users in specific groups` page privacy options. To do this, you will need to use the `dis-authentication-stub` locally to try to simulate the SSO user journey/ flow. See the PR description on how: https://github.com/ONSdigital/dis-wagtail/pull/163
- Play around with the `password_required.html` template form features and make sure the errors are surfaced properly now to the user.
  
  
### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


cms/auth/views.py
```diff
@@ -54,3 +57,21 @@
...
+        allowed_hosts={request.get_host()},
+        require_https=request.is_secure(),
+    ):
+        next_url = "/"
+
+    absolute_next = request.build_absolute_uri(next_url)
+    query = urlencode({"redirect": absolute_next})
+    separator = "&" if "?" in settings.WAGTAILADMIN_LOGIN_URL else "?"
+
+    return redirect(f"{settings.WAGTAILADMIN_LOGIN_URL}{separator}{query}")
```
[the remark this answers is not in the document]
sanjeevz3009: Fixed/ improved!

cms/urls.py
```diff
@@ -34,6 +34,7 @@
         authenticate_with_password,
         name="wagtaildocs_authenticate_with_password",
     ),
+    path("auth/frontend-login", frontend_login_redirect, name="frontend_login_redirect"),
```
MebinAbraham: I'd rather us not have routes registered if the feature is not supported. These shouldn't be needed when `CMS_PAGE_PRIVACY_CONTROLS_ENABLED` is False.

MebinAbraham (commented):
Happy to get this in but the features need a complete revisit before we would use it in prod.

cms/urls.py
MebinAbraham: Re: `documents/authenticate_with_password/<int:restriction_id>`, no one other than a super admin can create a private collection currently, either. I would gate it behind another feature flag which we can extend in future if needed.

MebinAbraham (commented):
Page privacy is currently not working for routable pages also. If you make an article private, you can still access it. Other pages are fine.

sanjeevz3009:
> Happy to get this in but the features need a complete revisit before we would use it in prod.

@MebinAbraham Will make the improvements you suggested and then create a follow-up card to track the future improvements if and when we want to turn on the page privacy feature. You're correct, we need a full business analysis on this if we want to enable it in prod/ for business to use it effectively.

We will ship this fast, and in the future, we shall iterate on it.

This is more to fix a completely broken functionality so when we do turn it on or need to demo to business etc, we have something to show that works.

JIRA card: [Will add the link to the card here]

cms/jinja2/templates/pages/wagtail/password_required.html
```diff
@@ -1,23 +1,60 @@
...
+{% from "components/button/_macro.njk" import onsButton %}
+{% from "components/error/_macro.njk" import onsError %}
 
 {% block html_class %}no-scroll-behaviour{% endblock %}
 
 {% block title %}Password required{% endblock %}
 
 {% block body_class %}template-password-required{% endblock %}
 
+{% block header_area %}
```
helenb: Just as an fyi this will need updating if https://github.com/ONSdigital/dis-wagtail/pull/702 is merged first - there are now separate blocks for `{% breadcrumbs %}` and `{% hero %}`

cms/jinja2/templates/pages/wagtail/password_required.html
```diff
@@ -1,23 +1,60 @@
...
+        {% set non_field_errors = form.non_field_errors() %}
+        {% set password_errors = form.password.errors %}
+
+        {# fmt:off #}
+        {% if non_field_errors %}
+            {% call onsError({
+            "id": "password-required-form-errors",
+            "text": _("There is a problem with your submission")
+            }) %}
+                <ul class="ons-list">
```
helenb: Could you directly use `onsList` here?

cms/urls.py
sanjeevz3009: Massive improvements here: https://github.com/ONSdigital/dis-wagtail/pull/575/commits/c67618727b7b649497af73ff44eaacb5e3917620.

I think now this does the job.

cms/jinja2/templates/pages/wagtail/password_required.html
sanjeevz3009: Will leave this up to @BJacksonONS or @MebinAbraham to sort out =)

cms/jinja2/templates/pages/wagtail/password_required.html
sanjeevz3009: https://github.com/ONSdigital/dis-wagtail/pull/575#discussion_r3546482426

sanjeevz3009:
@MebinAbraham @helenb @BJacksonONS Anything else that comes out of this PR + some future considerations I discussed with Mebin, I'll write it all up on a JIRA card, so we can revisit if business needs page privacy functionality in the future. I put it on my `ToDo` for when I return from leave.

cms/settings/base.py
```diff
@@ -1052,6 +1052,18 @@
 AWS_COGNITO_USER_POOL_ID = env.get("AWS_COGNITO_USER_POOL_ID")
 AWS_COGNITO_APP_CLIENT_ID = env.get("AWS_COGNITO_APP_CLIENT_ID")
 
+# Privacy feature flags. Page privacy gates the page view-restriction controls and routes;
+# collection privacy gates the document view-restriction password route.
+CMS_PAGE_PRIVACY_CONTROLS_ENABLED = env.get("CMS_PAGE_PRIVACY_CONTROLS_ENABLED", "true").lower() == "true"
+CMS_COLLECTION_PRIVACY_CONTROLS_ENABLED = env.get("CMS_COLLECTION_PRIVACY_CONTROLS_ENABLED", "true").lower() == "true"
```
MebinAbraham: These should be off.

```suggestion
CMS_PAGE_PRIVACY_CONTROLS_ENABLED = env.get("CMS_PAGE_PRIVACY_CONTROLS_ENABLED", "false").lower() == "true"
CMS_COLLECTION_PRIVACY_CONTROLS_ENABLED = env.get("CMS_COLLECTION_PRIVACY_CONTROLS_ENABLED", "false").lower() == "true"
```

cms/jinja2/templates/pages/wagtail/password_required.html
helenb: @BJacksonONS  Discussed in stand-up and it's fine for this to go in first as long as you just give me the heads up so I can update the DS 74.0.0 pr.

cms/auth/checks.py
```diff
@@ -36,6 +36,36 @@ def check_aws_cognito(app_configs: Iterable[AppConfig] | None, **kwargs: Any) ->
...
+    """Catches the config where WAGTAILADMIN_LOGIN_URL redirects to itself.
+
+    When WAGTAIL_CORE_ADMIN_LOGIN_ENABLED is False, urls.py mounts a RedirectView at the
+    local admin login path targeting WAGTAILADMIN_LOGIN_URL. If that setting is left at
+    its local default, the redirect points at itself and users loop forever instead of
+    reaching the SSO login.
+    """
+    errors: list[Error] = []
+
+    if not getattr(settings, "AWS_COGNITO_LOGIN_ENABLED", False):
```
BJacksonONS: There's an edge case here if we have both disabled and a default admin login url that still results in the infinite loop.

We can just take the cognito login conditional out of this as the urls don't depend on it, or flag if both are missing in internal.

cms/jinja2/templates/pages/wagtail/password_required.html
```diff
@@ -1,23 +1,60 @@
...
+    {{
+        onsBreadcrumbs({
+            "ariaLabel": breadcrumbs_aria_label,
+            "itemsList": page.get_breadcrumbs(request=request),
+            "variant": 'grey'
+        })
+    }}
+    {{
+        onsHero({
+            "title": page.title,
```
BJacksonONS, line 31: This would leak the page title and breadcrumb trail, as we're rendering it before the user has actually authenticated.

cms/core/utils.py
```diff
@@ -127,6 +128,34 @@ def redirect_to_parent_listing(
...
+    unless it is served through the hook chain, mirroring wagtail.views.serve.
+    """
+
+    def _serve(inner_page: Page, inner_request: HttpRequest, serve_args: Any, serve_kwargs: Any) -> HttpResponse:
+        target = serve_callable if serve_callable is not None else inner_page.serve
+        response: HttpResponse = target(inner_request, *serve_args, **serve_kwargs)
+        return response
+
+    serve_chain: Callable[..., HttpResponse] = _serve
+    for fn in reversed(hooks.get_hooks("on_serve_page")):
```
BJacksonONS, line 187: The `on_serve_page` hook can end up hitting wagtail's [`accept_request`](https://github.com/wagtail/wagtail/blob/c10e634636dae4164c2793d89f1a950a5f64517b/wagtail/models/view_restrictions.py#L38) which tries to access user objects which don't exist in the external environment. This throws an attribute error leading to a 500 error.

Given we call `get_latest` for article series, someone might not even need to know the exact url of restricted pages to hit this. I think partly this is due to us using `live()` rather than `public()` on some code paths when filtering pages.

BJacksonONS (changes requested):
Did some review on this but haven't had the time to do any work on it. If I don't get a chance before @sanjeevz3009 is back I've added comments for the main issues.

cms/articles/models.py
```diff
@@ -143,7 +143,9 @@ def latest_article(self, request: HttpRequest, *args: Any, **kwargs: Any) -> Htt
         if not (latest := self.get_latest()):
             raise Http404
 
-        return latest.serve(request, *args, serve_as_edition=True, **kwargs)
+        return serve_page_with_view_restrictions(
```
BJacksonONS, line 159: I think most of the calls to `serve_page_with_view_restrictions` from methods on this model will actually result in every `on_page_serve` hook getting called twice. They've already happened once before wagtail hits this method, then we re-run the entire chain again.

At the moment I don't think any have side effects, but it's extra latency/queries, and if we add observability hooks later on they might send out spurious logs.

cms/jinja2/templates/pages/wagtail/password_required.html
sanjeevz3009: Correct, but again we don't have requirements for any of this, so I didn't think too deeply so left it as is. We can come back to it in the future.

cms/auth/checks.py
sanjeevz3009: Good call, the following changes improves upon your suggestion and a but more... https://github.com/ONSdigital/dis-wagtail/pull/575/commits/ebd3718a457ee4203c73905d998c0001e0f50dfe

cms/core/utils.py
sanjeevz3009: Improvements: https://github.com/ONSdigital/dis-wagtail/pull/575/commits/ea40ab09bf71bfb766291484c3431b6a7baeb909

cms/articles/models.py
sanjeevz3009: Good spot, Ben: the chain is now only replayed when the edition carries restrictions of its own; the residual double-run on that rare path is deliberate, in exchange for not duplicating Wagtail's response-building logic.

https://github.com/ONSdigital/dis-wagtail/pull/575/commits/d9bc0a077b8c3749033b999f52fd1dbf1b3de513

sanjeevz3009:
@BJacksonONS @MebinAbraham Addressed the crucial comments from Ben. Once Ben re-reviews and approves, we can merge

cms/urls.py
```diff
@@ -29,17 +28,28 @@
...
     path("health", core_views.health, name="health"),
-    path(
-        "documents/authenticate_with_password/<int:restriction_id>",
-        authenticate_with_password,
-        name="wagtaildocs_authenticate_with_password",
-    ),
 ]
 
+if settings.CMS_COLLECTION_PRIVACY_CONTROLS_ENABLED:
+    from wagtail.documents.views.serve import authenticate_with_password  # pylint: disable=ungrouped-imports
```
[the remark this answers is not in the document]
BJacksonONS: I don't think this is valid because of our custom `DocumentServeView`.

cms/jinja2/templates/pages/wagtail/password_required.html
BJacksonONS: @sanjeevz3009 The DS update went in so this will need updating.

cms/auth/middleware.py
BJacksonONS, line 71: We still flush the session here which is something we avoid in `_handle_unauthenticated_user`.

BJacksonONS, line 60: Same as line 70

cms/core/utils.py
```diff
@@ -127,6 +128,65 @@ def redirect_to_parent_listing(
...
+        raise Http404
+
+    def _serve(inner_page: Page, inner_request: HttpRequest, serve_args: Any, serve_kwargs: Any) -> HttpResponse:
+        target = serve_callable if serve_callable is not None else inner_page.serve
+        response: HttpResponse = target(inner_request, *serve_args, **serve_kwargs)
+        return response
+
+    if routed_page is not None and page.is_descendant_of(routed_page) and not _has_direct_view_restrictions(page):
+        # The URL-level hook run against `routed_page` already enforced every restriction
+        # that applies to `page` (its own set minus direct ones is exactly its ancestors'),
```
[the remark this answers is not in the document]
BJacksonONS: This looks valid to me, if there's one or more pages between `routed_page` and `page`.

cms/urls.py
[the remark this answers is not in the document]
BJacksonONS: Though maybe `DocumentServeView` would need to be updated to use the new helper, think right now it just goes by `is_public`. Possibly one for a follow up PR.

BJacksonONS (commented):
Think a couple more points to address, but don't see anything else needing work now.

docs/custom-features/page_privacy.md
```diff
@@ -34,9 +34,88 @@ As we intend to disable page privacy settings entirely, we are leaving search in
...
 privacy settings for a page and it's descendants.
 
-## Statistical Article Page Privacy Issue
+## Feature flags
 
-We have also discovered a bug in the current handling of privacy settings on Statistical Article pages. Because the statistical article page can be served by
-editions page handler, the hooks which check for page privacy do not correctly detect and block acccess when a statistical article page is private. This means
-that if a statistical article page is set to private, the page is still visible publicly despite having privacy settings applied. If we were to support page
-privacy features, this would need to be fixed.
+Page privacy feature needs more analysis in the future, as what we have implemented is more to fix the current completely broken functionality. Hence the page privacy controls are disabled now along with setting page privacy itself.
```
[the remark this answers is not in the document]
sanjeevz3009: Going with my human-written doc

cms/jinja2/templates/pages/wagtail/password_required.html
sanjeevz3009: Done: https://github.com/ONSdigital/dis-wagtail/pull/575/commits/cc5f3a9decb16bced6c60fd63dfcd23f6a132d65

cms/jinja2/templates/pages/wagtail/password_required.html
sanjeevz3009: Done: https://github.com/ONSdigital/dis-wagtail/pull/575/commits/cc5f3a9decb16bced6c60fd63dfcd23f6a132d65

cms/urls.py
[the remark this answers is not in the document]
sanjeevz3009: Will add this to the follow-up card, as we don't want to spend longer on this PR as we still don't have good analysis on this.

cms/core/wagtail_hooks.py
```diff
@@ -49,6 +50,21 @@ def register_icons(icons: list[str]) -> list[str]:
...
+    request: HttpRequest,
+    serve_args: Any,  # pylint: disable=unused-argument
+    serve_kwargs: Any,  # pylint: disable=unused-argument
+) -> None:
+    """In the external environment there is no auth machinery (no sessions, users, or auth
+    backends), so page view restrictions can never be satisfied - evaluating them crashes
+    on the missing request attributes. Treat restricted pages as not existing instead.
+    """
+    if settings.IS_EXTERNAL_ENV and page.get_view_restrictions().exists():
+        raise Http404
```
MaciekBaron, line 65: Question: I was looking at related PRs and tickets (e.g. CMS-859) - is it stipulated anywhere that the error code should be 404?
One one hand it might be desirable by hiding the presence of restricted URLs. On the other, this is a white "lie" as the content exists, so a more appropriate code would be 403.
I'm sure this is defined somewhere, but was just wondering where to find it?

cms/jinja2/templates/pages/wagtail/password_required.html
```diff
@@ -20,10 +49,25 @@
...
+        {% if non_field_errors %}
+            {% set non_field_error_items = [] %}
+            {% for error in non_field_errors %}
+                {% do non_field_error_items.append({"text": error}) %}
+            {% endfor %}
+
+            {% call onsError({
+            "id": "password-required-form-errors",
+            "text": _("There is a problem with your submission")
+            }) %}
```
MaciekBaron, line 65:
```suggestion
            {% call onsError({
                "id": "password-required-form-errors",
                "text": _("There is a problem with your submission")
            }) %}
```

cms/core/wagtail_hooks.py
BJacksonONS: IMO 404 is appropriate here, as slugs could give away information potentially. It's an acceptable use case based on the RFC
https://www.rfc-editor.org/rfc/rfc9110.html#name-403-forbidden

MaciekBaron: I agree, just want to make sure we have a decision log to cover this

cms/jinja2/templates/pages/wagtail/password_required.html
sanjeevz3009: Whoah, good eyes: https://github.com/ONSdigital/dis-wagtail/pull/575/commits/1f669531f0733236303cfbf107d821829513b647

cms/auth/middleware.py
sanjeevz3009: Excellent find btw, I just dug deep into this now: https://github.com/ONSdigital/dis-wagtail/pull/575/commits/1f669531f0733236303cfbf107d821829513b647

sanjeevz3009: Excellent find btw, I just dug deep into this now: https://github.com/ONSdigital/dis-wagtail/commit/1f669531f0733236303cfbf107d821829513b647

For context for other devs:

**Context on the session-flush fix:**

So we were still flushing sessions on the invalid-token and consistency-check paths, the same thing we deliberately avoid in `_handle_unauthenticated_user`. In practice, the user-visible casualty is password-protected pages, because Wagtail stores "this visitor has passed restriction X" in the session, and that's essentially the only meaningful state an anonymous visitor's session holds today. The middleware now follows one rule everywhere: a session is only flushed when there is an authenticated session to terminate.

`Before the fix`: if a user who has been given the password for an article page just says. They logged into Florence last week, so their browser still carries the JWT cookies, now expired. When they first visit the article, the middleware sees cookies present, fails to validate them, and calls logout(request). Django's logout() flushes the entire session whether or not anyone is logged in. They are anonymous, so nothing visible happens yet, and they get the password form as expected. They enter the correct password: Wagtail writes the passed-restriction mark into her session and redirects them back to the article. But that redirect is a new request, so the middleware runs again first: expired cookies → logout() → session flushed, including the mark they just earned. Wagtail then finds no mark and serves the password form again. They retype the password and loop forever. Crucially, a wrong password still shows an error (that renders inside the POST response itself, before any next request), so the form looks functional; the correct password just never "sticks". The only escape is clearing their cookies. This is the same bug we already fixed for visitors with no JWT cookies: present-but-invalid cookies (expired, garbage, or minted for a different Cognito client).

`After the fix`: when a user visits the article, the middleware sees the expired cookies but, since there is no authenticated user, leaves the session alone; the user simply isn't authenticated by them. They get the password form, enter the password, the mark is written, and the redirect back to the article finds the mark intact: the page renders, and it stays unlocked as they browse the related-data and download routes.

What deliberately did not change: if a user is actually logged in and their tokens expire, mismatch across tokens, or belong to a different user than the session (the token-swap defence), that authenticated session is still terminated exactly as before, an authenticated session whose credentials can no longer be verified must die, and the full flush on logout is Django's session-fixation protection doing its job. The guard only distinguishes "nobody to log out, so don't destroy bystander session state" from "real authenticated session, terminate it". One deliberate posture change to be aware of: anonymous visitors with inconsistent (not merely expired) tokens now keep their session rather than having it destroyed; nothing in an anonymous session is derived from those tokens, so there was no state the old flush was protecting.

cms/core/wagtail_hooks.py
sanjeevz3009: @MaciekBaron @BJacksonONS if there aren't any action for now, happy to resolve this?

cms/articles/models.py
```diff
@@ -122,10 +123,21 @@ def summary(self) -> str:
         # TODO: update to include drafts when looking at previews holistically.
         return latest.summary if (latest := self.get_latest()) else ""
 
+    def _servable_editions(self) -> PageQuerySet:
```
zerolab, line 128: random thought: if you override the default manager for `StatisticalArticlePage` where you encapsulate the `.public()` logic?
https://docs.djangoproject.com/en/6.1/topics/db/managers/#modifying-a-manager-s-initial-queryset

I believe this will simplify things quite a bit as `StatisticalArticlePage.objects` will filter out pages with a view restriction.

cms/auth/tests/test_middleware.py
```diff
@@ -233,6 +284,7 @@ def test_session_without_jwt_key_triggers_authenticate_and_saves_key(self):
     def test_client_id_mismatch(self):
         req = self._request()
         req.COOKIES = {"access": "tokA", "id": "tokID"}
+        user = User.objects.create(username="test", email="test@example.com")
```
MaciekBaron, line 287: Perhaps a follow-up: we could switch to using `UserFactory` as we do in the rest of the codebase, albeit we have existing instances of using `User.objects.create`

MaciekBaron (approved):
LGTM apart from existing comments

cms/core/middleware.py
```diff
@@ -9,11 +9,7 @@
 NON_TRAILING_SLASH_METHODS = {"GET", "HEAD"}
 CLOUDFLARE_CACHE_TAG_HEADER = "Cache-Tag"
 
-ALLOWED_REQUEST_PATHS = {
-    settings.DJANGO_ADMIN_HOME_PATH,
-    settings.WAGTAILADMIN_HOME_PATH,
-    "/__debug__/",
-}
+ALLOWED_REQUEST_PATHS = {settings.DJANGO_ADMIN_HOME_PATH, settings.WAGTAILADMIN_HOME_PATH, "/__debug__/", "/_util/"}
```
BJacksonONS, line 12: Think this is just a nit: We can make this `"/_util/authenticate_with_password/"` so `"/_util/login/"` can't be exposed.

cms/core/wagtail_hooks.py
```diff
@@ -49,6 +50,21 @@ def register_icons(icons: list[str]) -> list[str]:
...
+    page: Page,
+    request: HttpRequest,
+    serve_args: Any,  # pylint: disable=unused-argument
+    serve_kwargs: Any,  # pylint: disable=unused-argument
+) -> None:
+    """In the external environment there is no auth machinery (no sessions, users, or auth
+    backends), so page view restrictions can never be satisfied - evaluating them crashes
+    on the missing request attributes. Treat restricted pages as not existing instead.
+    """
+    if settings.IS_EXTERNAL_ENV and page.get_view_restrictions().exists():
```
BJacksonONS, line 64: As it is this adds extra queries to every page serve in external environment.

We could add `@memory_cache` to something that checks the paths that actually have restrictions beforehand so we don't need to do extra DB work?

BJacksonONS (commented):
Think there's really only one potential blocker here if we're thinking about perf, will leave it to @MebinAbraham or @sanjeevz3009 to decide if it needs work here or in a follow up.

MebinAbraham:
FYI this is incompatible with https://github.com/ONSdigital/dis-wagtail/pull/799, as that doesn't add support for view restrited pages. 

Note, we need to be careful about adding extra queries to "check" a feature we are not using. To fully support "restricted" pages, if there was a need, would require a lot more changes than this.

cms/auth/middleware.py
```diff
@@ -56,8 +56,12 @@ def process_request(self, request: HttpRequest) -> None:
...
         id_payload = validate_jwt(id_token, token_type="id")  # noqa: S106
         if not access_payload or not id_payload:
-            logger.info("Invalid or expired JWT tokens; logging out user.")
-            logout(request)
+            # Only terminate authenticated sessions: logout() flushes the whole session,
+            # which would destroy unrelated anonymous state (e.g. passed page view
+            # restrictions) on every request for visitors carrying stale or garbage cookies.
+            if request.user.is_authenticated:
+                logger.info("Invalid or expired JWT tokens; logging out user.")
+                logout(request)
```
MebinAbraham, line 64: Just need to check we have as a result of this. Can you retest the auth integration test cases (doc on Confluence) with these changes.

cms/auth/middleware.py
```diff
@@ -146,10 +153,16 @@ def _handle_cognito_disabled(request: HttpRequest) -> None:
...
-            request.user.is_authenticated and request.user.is_external_user
-        ):
+        """Logs out the user if JWT tokens are missing and the session configuration is unsuitable.
+
+        Anonymous sessions are left untouched: logout() flushes the whole session, which would
+        destroy unrelated state such as passed page view restrictions (password-protected pages)
+        on every request.
+        """
+        if not request.user.is_authenticated:
+            return
```
MebinAbraham, line 163: This is misleading against current method name. `_handle_unauthenticated_user` which on first check just returns.

cms/auth/views.py
```diff
@@ -54,3 +56,28 @@ def extend_session(request: HttpRequest) -> JsonResponse:
...
     return JsonResponse({"status": "error", "message": "Invalid request method."}, status=405)
+
+
+def frontend_login_redirect(request: HttpRequest) -> HttpResponse:
+    """Redirect Wagtail frontend private-page login requests to the appropriate login page.
+
+    With Cognito enabled, users are sent to the Florence SSO login, which expects the
+    absolute post-login target in a `redirect` query parameter. Otherwise, users fall
+    back to Wagtail's built-in frontend login view, which reads `next`.
+    """
```
MebinAbraham, line 67: This is making the assumption that we are redirecting to the Florence login page. If this feature were to existing, it could be for the "public" not internal. Agains, we are implementing against assumptions.

I think we are adding a lot of complexity and queries for a feature ons doesn't need. If they need it, it needs a lot of thought because we just turn it on, as we haven't always been designing with this feature in mind.

We need to ensure that if we keep this, it has minimal impact on complexity and performance.

MebinAbraham:
This warrants a tech discussion.

PR #578, opened 2026-04-13
Chore: Update Makefile install-dev Target

sanjeevz3009:

This PR updates the `makefile` `install-dev` target to include the commands to install `pre-commit` and install the `hooks` within the project. This is to reduce the chance of a new onboarding dev missing out on installing the `pre-commit` package and setup.
- There have been many incidents in the ONS where `pre-commit` could have prevented the exposure of a secret, API key, etc.
- Even though we have other tooling and measures in place for when there is a security incident, this PR change will reduce the chance of anything catastrophic happening even more.

Credit: @BJacksonONS for reviewing the starter guide and proposing this improvement. The confluence page, which contains the starter guide, has also been updated to reflect this change.

### How to review

- Review the `makefile` changes
- Make sure everything is working locally as expected, when using the updated makefile command

### Deployment Safety

Please select one:

- [X] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,6 +92,8 @@ install:  ## Install the dependencies excluding dev.
 install-dev:  ## Install the dependencies including dev.
 	npm ci
 	poetry install --no-root
+	pip install pre-commit
+	pre-commit install
 
 .PHONY: megalint
 .PHONY: megalint
--- a/Makefile
+++ b/Makefile
@@ -92,8 +92,7 @@ install:  ## Install the dependencies excluding dev.
 install-dev:  ## Install the dependencies including dev.
 	npm ci
 	poetry install --no-root
-	pip install pre-commit
-	pre-commit install
+	pip install pre-commit && pre-commit install
 
 .PHONY: megalint
 .PHONY: megalint


Makefile
```diff
@@ -92,6 +92,8 @@ install:  ## Install the dependencies excluding dev.
 install-dev:  ## Install the dependencies including dev.
 	npm ci
 	poetry install --no-root
+	pip install pre-commit
+	pre-commit install
```
MaciekBaron: Nit: I think perhaps we should do the following:

```suggestion
        pip install pre-commit && pre-commit install
```

This way we run `pre-commit install` only if the installation succeeds (or it is already installed).

The user will see the error message anyway.

BJacksonONS: Also agree with this, less clutter in the console

BJacksonONS (approved):
LGTM but Maciek's change would also be nice

Makefile
sanjeevz3009: Sounds good =)

Here is the change: https://github.com/ONSdigital/dis-wagtail/pull/578/commits/2ccf35aca8c04cb5606fc51a50cd5921922d1aae

BJacksonONS (approved):
Looks good 👍

MaciekBaron (approved):
Thanks Sanjeev!

Makefile
joepercival: FYI: I've seen it done this way elsewhere too :)

PR #581, opened 2026-04-15
Update cookies management contents, layout and language handling

MebinAbraham:

Updates the cookies management page to better match the new service content and page layout, while fixing language handling in the cookies banner.

The main changes are:
- removing `LANGUAGE_CODE` from the global context processor and using `request.LANGUAGE_CODE` directly in the base template so the cookie banner reflects the active request language
- moving the cookies page onto `base_page.html` so it gets the standard page structure and breadcrumbs
- page content updates
- extending unit and functional coverage for breadcrumbs, revised headings, and English/Welsh banner behaviour

### How to review

See the how to review the guide on the ticket [CMS-1143](https://officefornationalstatistics.atlassian.net/browse/CMS-1143)

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


cms/jinja2/templates/pages/release_calendar/release_calendar_page.html
```diff
@@ -130,8 +130,8 @@ <h2>{{ _("About the data") }}</h2>
                     </div>
                     {% if page.is_accredited %}
                         <p>
-                            {% trans trimmed %}
-                                These are accredited official statistics. They have been independently reviewed by the Office for Statistics Regulation (OSR) and found to comply with the standards of trustworthiness, quality and value in the <a href="https://code.statisticsauthority.gov.uk/the-code/">Code of Practice for Statistics</a>. This broadly means that the statistics:
+                            {% trans trimmed stats_authority_url="https://code.statisticsauthority.gov.uk/the-code/" %}
+                                These are accredited official statistics. They have been independently reviewed by the Office for Statistics Regulation (OSR) and found to comply with the standards of trustworthiness, quality and value in the <a href="{{ stats_authority_url }}">Code of Practice for Statistics</a>. This broadly means that the statistics:
```
MebinAbraham, line 134: Unrelated but fixes a warning when running `make makemessages`.

`
./cms/jinja2/templates/pages/release_calendar/release_calendar_page.html.py:133: warning: Message contains an embedded URL.  Better move it out of the translatable string, see https://www.gnu.org/software/gettext/manual/html_node/No-embedded-URLs.html
`

cms/jinja2/templates/pages/cookies.html
```diff
@@ -26,7 +27,9 @@ <h2>{{ _("Your cookie settings have been saved") }}</h2>
         {# fmt:on #}
 
         <p>{{ _("Cookies are files saved on your phone, tablet or computer when you visit a website.") }}</p>
-        <p>{{ _("We use cookies to store information about how you use the ONS website, such as the pages you visit.") }}</p>
+        <p>{{ _("We use cookies to collect and store information about how you use the ONS website, such as the pages you visit.") }}</p>
+        {# :TODO: The URL should be updated post launch as the slug for the cookies page is changing to /help/cookies-information. #}
```
sanjeevz3009, line 40: On the doc on the ticket, it says: `change-cookies`

BJacksonONS: There's a comment from Mebin saying this will be post launch

sanjeevz3009 (commented):
LGTM! Other than the UI tests need fixing. I left a minor comment; the code looks good, and I tested it functionally, and it all works.

cms/jinja2/templates/pages/cookies.html
sanjeevz3009: I meant this doc, if we are talking about the same thing. I was going based on the Google Doc attached to the ticket as the source of truth.

Screenshot 2026-04-15 at 13 47 51

BJacksonONS: Ticket on Jira says to match Version 1 content, and if you can see the comments on the word doc this is a post launch change.

cms/jinja2/templates/base_page.html
```diff
@@ -88,20 +88,32 @@
...
                     <div class="ons-page__main ons-u-mb-5xl {{ mainColClasses }}">
-                        <h1 class="ons-u-fs-3xl ons-u-mb-xl common-header__heading">
-                            {{ page_heading or page.title }}
-                        </h1>
+                        {# If the header area was not overridden then set the page title as the heading for the page #}
+                        {% if not header_area_content|trim %}
+                            <h1 class="ons-u-fs-3xl ons-u-mb-xl common-header__heading">
+                                {{ page_heading or page.title }}
+                            </h1>
+                        {% endif %}
```
MebinAbraham, line 114: Open to other suggestions, no real nice way to do this from what I can see.

The `header_area` is overridden most of the time, so this logic is basically saying ensure we output a heading when the header area is not overridden.

Another way to do this that would be much simpler is to remove the default logic and lets template always define header area and a page heading block.

cms/jinja2/templates/pages/cookies.html
sanjeevz3009: Thanks, Ben. I realised I opened it on JIRA, and the view it gives doesn't show the comments. Now speaking to Ben, I see that it's in the comments, from Amelia, that it will be `/help/cookies-information`

sanjeevz3009: Resolving!

cms/jinja2/templates/pages/cookies.html
```diff
@@ -1,15 +1,16 @@
-{% extends "templates/base.html" %}
+{% extends "templates/base_page.html" %}
 {% from "components/radios/_macro.njk" import onsRadios %}
 {% from "components/button/_macro.njk" import onsButton %}
 {% from "components/panel/_macro.njk" import onsPanel %}
 {% from "components/details/_macro.njk" import onsDetails %}
 {% from "components/table/_macro.njk" import onsTable %}
 
+{%- set page_heading -%}
```
MaciekBaron, line 8: I noticed that the page title (`<title>`) is not translated for the Welsh version and defaults to "Cookies". We could set `page_title` here instead and use that below?

MaciekBaron (changes requested):

cms/jinja2/templates/base_page.html
zerolab: suggestion: leave the `header_area` block as-is, and add a `{% block content_header %}{% endblock %}` block here.

Then `cms/jinja2/templates/pages/cookies.html` does:

```jinja2
...
{% block header_area %}{% endblock %}{# override the default header area output #}

...

{% block content_header %}{# output the page title in the main content area #}
<h1 class="ons-u-fs-3xl ons-u-mb-xl common-header__heading">
   {{ page.title }}
</h1>
{% endblock %}
...
```

MebinAbraham: The only reason I didn't override in cookies.html was exactly that, to avoid an override and to let the base page HTML handle it. If we are overriding from every template, I would prefer to remove the default content in the base_page HTML, as it defeats the point. I believe we are currently overriding in every single template so having a default doesn't provide any value, given that you have to override it to be empty in cases where you don't need it.

MebinAbraham: Another thing is, this will all change slightly as our markup is incorrect compared to the DS. The breadcrumb and title are at the wrong place currently.

cms/jinja2/templates/pages/cookies.html
MebinAbraham: Given I need the site here and given https://github.com/ONSdigital/dis-wagtail/pull/526 is in flight, I wonder if it should be done there. I can fetch site title here, but that PR will need to account for it.

MebinAbraham: Added in [`f0e0efb` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/581/commits/f0e0efb9bf81d5760280e9faad280a8a228dcaa6), will make a note in the other PR.

MaciekBaron (approved):

cms/jinja2/templates/pages/cookies.html
MebinAbraham: Also tests: [`4fca7dd` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/581/commits/4fca7dde11066a488e466e46c4b162b0fe8d81b4)

BJacksonONS (approved):
All looks right visually, happy if you've made your mind up on the templating changes.

cms/jinja2/templates/base_page.html
MebinAbraham: @zerolab got any strong feelings?. Happy to make the above change if you think it would be less confusing to devs.

@BJacksonONS @MaciekBaron any preference from yourself?

BJacksonONS: I think if there's going to be another update to address issues with the templates matching the design system then this makes sense for now.

MebinAbraham: Decided to go with [`689adc0` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/581/commits/689adc00177357194aea4363401fbc8676dbbed5) instead which is a hybrid of the two.

zerolab: 689adc0 sounds like a good middleground.
I lack the latest context on what the different changes to the other tamplates need to be around breadcrums/page titles, so can't comment wider than that

sanjeevz3009 (approved):
Other than the comments made already and UI tests need fixing. LGTM!

MebinAbraham:
Waiting on final legal signoff.

MebinAbraham:
Not had full sign-off yet, but will follow up if any changes are required so that we can progress testing and translations, which have some turnaround time.

BJacksonONS:
@MebinAbraham 
Likely not a blocker for now, but I'm seeing the not available in your language banner, even though this page is getting it's translations from `.po` files, and isn't managed via the admin panel.

image

BJacksonONS (approved):
Looks good

MebinAbraham:
> @MebinAbraham Likely not a blocker for now, but I'm seeing the not available in your language banner, even though this page is getting it's translations from `.po` files, and isn't managed via the admin panel.

Good spot. I'll do a follow-up to sort that.

PR #583, opened 2026-04-15
Update Makefile recursion and clear cache on dev init

MebinAbraham:

Adds a `cache-clear` helper target and runs it as part of `dev-init` to reduce issues caused by stale Django cache data during local environment setup.

Also updates the Makefile to use `$(MAKE)` instead of `make` for nested targets so recursive calls inherit make flags and behave more reliably.

### How to review

1. Run `make lint-py` and confirm it still invokes the `mypy` target as expected.

2. Run `make cache-clear` and confirm it clears the Django cache without errors.

3. Run `make dev-init` in a local dev environment and verify the cache clear step runs after the Wagtail site port update.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


sanjeevz3009 (approved):
LGTM!

MaciekBaron (approved):

PR #585, opened 2026-04-16
Use RFC3339 datetime format for page release_date and previous_date

BJacksonONS:

Addresses CMS-1166.

We want to more strictly standardise release_date and previous_date fields when being sent to the search service.

The new approach should simplify handling of both date and datetime objects when being passed into search, as well as avoid extra complexity at the point of call when dealing with None.

### How to review

Review test cases and assess timestamp format expectations are correct.
Ensure `release_date` and `previous_date` are being returned as RFC3339 compliant strings.

The new function should
- Return None if passed None
- Default to midnight if given just a date
- Maintain timezone information if already present on datetime objects
- Default to UTC if there is no timezone already

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy


cms/core/formatting_utils.py
```diff
@@ -122,3 +123,26 @@ def get_document_metadata(
...
+            value = localtime(value)
+    else:
+        # set time to midnight for dates
+        value = datetime.combine(value, datetime.min.time())
+
+    formatted = value.replace(microsecond=0).isoformat(sep="T")
+
+    # we want the timezone offset, but using make_aware on the input date
+    # can add timezone conversions for daylight savings
+    if value.tzinfo is None:
```
BJacksonONS: Do we want to assume non aware date times are already in the local timezone or UTC?

cms/core/formatting_utils.py
```diff
@@ -122,3 +123,24 @@ def get_document_metadata(
...
+
+    # datetime is subclass of date so look at that first
+    if isinstance(value, datetime):
+        if is_aware(value):
+            value = localtime(value)
+    else:
+        # set time to midnight for dates
+        value = datetime.combine(value, datetime.min.time())
+
+    if not is_aware(value):
```
BJacksonONS, line 141: @MebinAbraham Do we want to default to UTC or the current local when dealing with non tz aware date times?

cms/core/formatting_utils.py
```diff
@@ -122,3 +123,24 @@ def get_document_metadata(
         metadata["date"] = get_document_metadata_date(date_value, prefix=prefix)
 
     return metadata
+
+
+def to_rfc3339_datetime(value: date | datetime | None) -> str | None:
+    """Converts a date, datetime or None to an RFC3339 compliant date string."""
```
MebinAbraham:
```suggestion
    """Converts a date or datetime to an RFC3339 date-time compliant string."""
```

cms/core/formatting_utils.py
```diff
@@ -122,3 +123,24 @@ def get_document_metadata(
         metadata["date"] = get_document_metadata_date(date_value, prefix=prefix)
 
     return metadata
+
+
+def to_rfc3339_datetime(value: date | datetime | None) -> str | None:
```
MebinAbraham, line 132: suggestion, would benefit from type [overloads](https://typing.python.org/en/latest/spec/overload.html#overloads).

cms/core/formatting_utils.py
MebinAbraham: UTC.

cms/search/utils.py
```diff
@@ -21,7 +22,7 @@ def build_standard_resource_dict(page: Page) -> dict:
     return {
```
MebinAbraham, line 22: Let's also update a test for these tests methods to validate the date scenario.

BJacksonONS: yes that's already in progress, wanted to put up the previous query while I finished that off

cms/core/formatting_utils.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/585/changes/4e2fb74e7bbcd8f987d0ac8840890e26522dc6db

cms/search/utils.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/585/changes/4e2fb74e7bbcd8f987d0ac8840890e26522dc6db

There's some weird flakiness on some tests when running the whole suite I wouldn't expect. They run fine on their own, and don't seem to hit the new function at all.

cms/core/formatting_utils.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/585/changes/4e2fb74e7bbcd8f987d0ac8840890e26522dc6db

cms/core/formatting_utils.py
```diff
@@ -122,3 +123,28 @@ def get_document_metadata(
...
+    # datetime is subclass of date so look at that first
+    if isinstance(value, datetime):
+        if is_aware(value):
+            value = localtime(value)
+    else:
+        # set time to midnight for dates
+        value = datetime.combine(value, datetime.min.time())
+
+    if not is_aware(value):
+        value = make_aware(value, timezone=UTC)
```
MaciekBaron: I'm wondering if, to flatten this slightly and avoid the double `is_aware` we could do the following:

```suggestion
    if not isinstance(value, datetime):
        # set time to midnight for dates
        value = datetime.combine(value, datetime.min.time())

    if is_aware(value):
        value = localtime(value)
    else:
        value = make_aware(value, timezone=UTC)
```

Might also read a bit better.

BJacksonONS: Yes I like this change, though I think it'll have to be 
`value = localtime(value) if is_aware(value) else make_aware(value, timezone=UTC)`
to appease the linter.

Also wondering if we should change `value` to `date_`?

BJacksonONS: https://github.com/ONSdigital/dis-wagtail/commit/0b4853bec08e1c5431bb73f84615087336766e78

MaciekBaron (approved):
LGTM!

cms/core/formatting_utils.py
```diff
@@ -122,3 +123,23 @@ def get_document_metadata(
...
+def to_rfc3339_datetime(value: date | datetime | None) -> str | None:
+    """Converts a date or datetime to an RFC3339 date-time compliant string."""
+    if value is None:
+        return None
+
+    if not isinstance(value, datetime):
+        # set time to midnight for dates
+        value = datetime.combine(value, datetime.min.time())
+
+    value = localtime(value) if is_aware(value) else make_aware(value, timezone=UTC)
```
MebinAbraham: Q: Why are we converting using localtime? This normalises to current Django timezone, not UTC?. Since we don't normalise elsewhere, I suggest just returning the formatted string with timezone info not normalised.

BJacksonONS: I was looking at the behaviour of some of the other formatting utils, but looks like the use case for search isn't quite the same.

BJacksonONS: f92a4d368644fcd74b70c75258a369bbb66fcbbf

cms/core/formatting_utils.py
```diff
@@ -122,3 +123,24 @@ def get_document_metadata(
...
+    """Converts a date or datetime to an RFC3339 date-time compliant string."""
+    if value is None:
+        return None
+
+    if not isinstance(value, datetime):
+        # set time to midnight for dates
+        value = datetime.combine(value, datetime.min.time())
+
+    if not is_aware(value):
+        make_aware(value, timezone=UTC)
```
MebinAbraham: Missing var assignment. It doesn't mutate existing but returns a new datetime.
```suggestion
        value = make_aware(value, timezone=UTC)
```

BJacksonONS: d879dfec8243f01e18db93e13f087ed212cb9a7c

cms/core/tests/test_formatting_utils.py
```diff
@@ -236,3 +242,55 @@ def test_get_document_metadata_with_date_and_custom_prefix(self):
...
+            {
+                "name": "A timezone aware datetime in BST UTC+1",
+                "timezone": "Europe/London",
+                "date": timezone.make_aware(datetime(2020, 7, 15, 1, 30)),
+                "formatted_date": "2020-07-15T01:30:00+01:00",
+            },
+            {
+                "name": "A timezone aware datetime in EST UTC-5",
+                "timezone": "America/New_York",
+                "date": timezone.make_aware(datetime(2020, 1, 1, 12, 0)),
```
MebinAbraham: These tests need updating. You would want to use `astimezone` to make them into actual new york datetimes etc

BJacksonONS: 2dc3eaa131d58f122cd64780e41f4e271b3d48b4

sanjeevz3009 (approved):
Other than the comments made, LGTM!

MebinAbraham (approved):

PR #586, opened 2026-04-16
Ensure that iframe footnotes data is passed through to the chart design system component

helenb:

The footnotes added in the CMS were not correctly being passed through to the design system component for the iframe, so they were not displaying as expected.

See https://officefornationalstatistics.atlassian.net/browse/DPD-2500

### How to review

- Edit a statistical article page, and choose to add an iframe visualisation in the content area.
- Example visualisation url to use: https://www.ons.gov.uk/visualisations/dvc3294/fig01/index.html
- Add some footnotes to the visualisation, with some rich text options selected
- Preview or publish the changes, and ensure the footnotes display below the iframe.

Screenshot 2026-04-16 at 14 04 00

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


cms/datavis/blocks/charts.py
```diff
@@ -820,6 +821,10 @@ def get_component_config(
             "caption": value.get("caption"),
             "description": value.get("audio_description"),
             "iframeUrl": value.get("iframe_source_url"),
+            "footnotes": {
+                "title": _("Footnotes"),
+                "content": str(value.get("footnotes")),
```
zerolab:
```suggestion
                "content": str(value.get("footnotes", "")),
```

Otherwise you will get "None" (as a string) if there's no footnotes data

helenb: Nice, thanks - I've updated that.

cms/datavis/blocks/charts.py
```diff
@@ -820,6 +821,10 @@ def get_component_config(
             "caption": value.get("caption"),
             "description": value.get("audio_description"),
             "iframeUrl": value.get("iframe_source_url"),
+            "footnotes": {
+                "title": _("Footnotes"),
```
MaciekBaron, line 825: Minor: we'll need to run `make makemessages` as this will update the `.po` file (it will add a reference to this line)

MaciekBaron (changes requested):

cms/datavis/blocks/charts.py
```diff
@@ -820,6 +821,10 @@ def get_component_config(
...
+                "content": str(value.get("footnotes", "")),
+            },
```
MebinAbraham, line 827: We'll need some tests for this.

cms/datavis/blocks/charts.py
```diff
@@ -632,7 +633,7 @@ def get_series_data(self, value: StructValue) -> tuple[list[list[str | int | flo
         rows: list[list[str | int | float]] = value["table"].rows
         groups = defaultdict(list)
 
-        for x, y, group_name, *_ in rows:
```
MebinAbraham, line 635: This is standard practice fyi.

helenb: It was flagged by `lint-py` as an issue because of importing `gettext_lazy` as `_` - which is also standard from what I can see. Is there a better way to resolve the conflict?

helenb: See https://github.com/ONSdigital/dis-wagtail/actions/runs/24506731102/job/71627034309

MebinAbraham: Right I see, in that case, happy to leave as is. You could obv, use `gettext_lazy` without aliasing, but given this was a change due to that conflict, happy with it. I thought it was just changed because it was unclear. 👍

cms/datavis/blocks/charts.py
helenb: Updated in 3fa803c8c226f892a32c749b43e51c370b0f8ace

cms/datavis/blocks/charts.py
helenb: I've added some tests for the footnotes configuration specifically in https://github.com/ONSdigital/dis-wagtail/pull/586/changes/61af7ffdbd5eb72c05323133fb39a2f2c038761f. We might want to also consider tests for the other config options available for iframe blocks as well.

MaciekBaron (approved):
LGTM!

MebinAbraham (approved):

MebinAbraham (approved):

sanjeevz3009 (commented):
LGTM!

PR #587, opened 2026-04-17
Use design system social block in header

BJacksonONS:

Addresses CMS-579

The design system page template already has a social block we are overriding with our own config. Ideally we want to be conforming to the design system as much as possible.

This PR removes the override and passes the extra values into the `pageConfig` to insert the correct social config.

### How to review

- Compare expected output from `_template.njk` with output in template.
- Test defaults from social media settings display correctly
- Test page specific images/share texts display correctly

### Deployment Safety

Please select one:

- [ x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/cms/core/tests/test_pages.py
+++ b/cms/core/tests/test_pages.py
@@ -239,14 +239,15 @@ class SocialMetaTests(WagtailPageTestCase):
     def setUpTestData(cls):
         cls.page = InformationPageFactory()
 
-    def test_social_meta_image_fallback(self):
-        """Test that the image meta tag falls back to the ONS logo from the CDN."""
+    def test_social_meta_image_url(self):
+        """Test that the image meta tag uses ONS logo from the design system CDN."""
         response = self.client.get(self.page.get_url(request=self.dummy_request))
         self.assertEqual(response.status_code, HTTPStatus.OK)
 
         self.assertContains(
             response,
-            f'<meta property="og:image" content="{settings.DEFAULT_OG_IMAGE_URL}" />',
+            '<meta property="og:image"'
+            'content="https://cdn.ons.gov.uk/sdc/design-system/73.4.1/favicons/opengraph.png" />',
         )
 
 
--- a/cms/jinja2/templates/base.html
+++ b/cms/jinja2/templates/base.html
@@ -23,6 +23,10 @@
 {%- endset -%}
 {%- set page_title = page_title | default(page_title_default) -%}
 
+{%- set social_settings = settings.core.SocialMediaSettings -%}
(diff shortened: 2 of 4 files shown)


cms/jinja2/templates/base.html
```diff
@@ -88,6 +96,10 @@
     "meta": {
         "hrefLangs": hreflangs,
         "canonicalUrl": canonical_url,
+    },
+    "social": {
+        "xImage": image(social_image, "width-1000").url if social_image else False,
+        "xSite:": "@"
```
BJacksonONS, line 102: The previous override used `SocialMediaSettings.twitter_handle`, but this doesn't actually exist on `SocialMediaSettings` at the moment.

MaciekBaron (changes requested):
Thanks Benedict, could we also add tests to check for regression?

I thought this ticket was blocked because of https://github.com/ONSdigital/design-system/issues/3802 (the issue is closed on GitHub but tracked internally as ONSDESYS-852).

With this change I can see we would lose the `og:image`. Is this intended?

cms/core/tests/test_pages.py
```diff
@@ -239,14 +239,15 @@ class SocialMetaTests(WagtailPageTestCase):
     def setUpTestData(cls):
         cls.page = InformationPageFactory()
 
-    def test_social_meta_image_fallback(self):
-        """Test that the image meta tag falls back to the ONS logo from the CDN."""
+    def test_social_meta_image_url(self):
```
BJacksonONS, line 242: I updated this to pass, but I'm not really sure it's all that helpful now. The design system url and version are hard coded into _template.njk, so we don't have a variable to pull this in from, it'll need to be manually updated each time and the content in the templates are only likely to change with updates to the design system.

BJacksonONS:
> Thanks Benedict, could we also add tests to check for regression?
> 
> I thought this ticket was blocked because of [ONSdigital/design-system#3802](https://github.com/ONSdigital/design-system/issues/3802) (the issue is closed on GitHub but tracked internally as ONSDESYS-852).
> 
> With this change I can see we would lose the `og:image`. Is this intended?

We keep `og:image` as it's baked into the design system template.
https://cdn.ons.gov.uk/sdc/design-system/73.4.1/favicons/opengraph.png

BJacksonONS:
Ah okay we want to override it without completely replacing `assetsUrl`.

MebinAbraham:
Let's come back to this once the DS side is fixed, especially given that the logic will need to be in Python anyway once https://github.com/ONSdigital/dis-wagtail/pull/526 merges.

PR #588, opened 2026-04-20
Improve test reliability

BJacksonONS:

Makes a handful of small changes to improve test reliability.

- Updates a test that used `unittest.Testcase` rather than `django.test.TestCase` but accessed the database
- Use a database query to fetch a page when ordering by `-release_date` was unpredictable. Pages with identical release dates returned by a model factory are not necessarily in the order the database fetches them in.
- Updates two functional tests that would fail when there were multiple element matches on the page

The main change is centralising management of url configuration using the `setting_changed` signal. The primary cause of test instability I've been able to find was due to url config being lazily loaded after a `reset_url_caches` while settings were being overridden. Due to the shuffling of tests this could unpredictably lead to failures.

- `IS_EXTERNAL_ENV`: Affects whether wagtail urls loaded. Some tests were not cleaning up urls after setting this `True`.
- `WAGTAILADMIN_HOME_URL`: Some tests were overriding this to `/admin/`, and the leading `/` would lead to django setting the url to `/%2fadmin/`, breaking future tests.

### How to review

Run the full test suite a few times with different seeds, verify that there are no unexpected failures.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/cms/bundles/tests/test_action_menu.py
+++ b/cms/bundles/tests/test_action_menu.py
@@ -1,6 +1,5 @@
-from unittest import TestCase
-
 from django.forms import Media
+from django.test import TestCase
 from wagtail.coreutils import get_dummy_request
 
 from cms.bundles.action_menu import (
@@ -20,6 +19,7 @@ from cms.bundles.tests.factories import BundleFactory
 class BundleActionMenuTests(TestCase):
     @classmethod
     def setUpClass(cls) -> None:
+        super().setUpClass()
         cls.draft_bundle = BundleFactory()
         cls.in_review_bundle = BundleFactory(in_review=True)
         cls.approved_scheduled_bundle = BundleFactory(approved=True)
--- a/cms/bundles/tests/test_action_menu.py
+++ b/cms/bundles/tests/test_action_menu.py
@@ -18,8 +18,7 @@ from cms.bundles.tests.factories import BundleFactory
 
 class BundleActionMenuTests(TestCase):
     @classmethod
-    def setUpClass(cls) -> None:
-        super().setUpClass()
+    def setUpTestData(cls) -> None:
         cls.draft_bundle = BundleFactory()
         cls.in_review_bundle = BundleFactory(in_review=True)
         cls.approved_scheduled_bundle = BundleFactory(approved=True)
--- a/cms/settings/test.py
+++ b/cms/settings/test.py
@@ -103,7 +103,7 @@ CMS_HOSTNAME_LOCALE_MAP = {
 }
 CMS_HOSTNAME_ALTERNATIVES = {"ons.localhost": "pub.ons.localhost", "cy.ons.localhost": "cy.pub.ons.localhost"}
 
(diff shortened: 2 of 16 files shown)


cms/bundles/tests/test_action_menu.py
```diff
@@ -20,6 +19,7 @@
 class BundleActionMenuTests(TestCase):
     @classmethod
     def setUpClass(cls) -> None:
+        super().setUpClass()
```
zerolab: 🤔 this should've been

```suggestion
def setUpTestData(cls) -> None:
```

BJacksonONS: e60db623aac376bc8932f60b05df858a1b789389

cms/settings/test.py
```diff
@@ -97,3 +102,41 @@
...
+        del sys.modules[ROOT_URLCONF]  # noqa: F405
+        # Also delete any submodules
+        modules_to_delete = [mod for mod in sys.modules if mod.startswith(ROOT_URLCONF + ".")]  # noqa: F405
+        for mod in modules_to_delete:
+            del sys.modules[mod]  # noqa: F405
+
+    is_language_prefix_patterns_used.cache_clear()  # type: ignore
+
+
+setting_changed.connect(_reset_url_caches_on_setting_changed_signal_handler)
```
MebinAbraham, line 142: Can we now not remove `reset_url_caches` and its usages, given this should now handle it?

BJacksonONS: I'm going through and removing the calls to check if all instances can be removed. I tried at some point and it was causing issues, though I don't think I had the admin home url fix at the time.

BJacksonONS: Yep looks like we can now
587043929e0816bc177783b8beec94d62423273d

cms/settings/test.py
```diff
@@ -97,3 +102,39 @@
     "cy.pub.ons.localhost": "cy",
 }
 CMS_HOSTNAME_ALTERNATIVES = {"ons.localhost": "pub.ons.localhost", "cy.ons.localhost": "cy.pub.ons.localhost"}
+
+URL_CONFIG_SETTINGS = (
```
MaciekBaron: Super minor nit-pick: We can probably use a set here instead of a tuple for quicker lookups

BJacksonONS: Good point, no reason not to.

cebcd269cf9a8bdf508cc043d09f9984f3bb7c50

MebinAbraham (approved):
Nice one

MaciekBaron (approved):
LGTM!

BJacksonONS:
Added a couple of explicit settings that are used in dev. If you run tests out of a docker container while these are true in your .env file they can cause some tests to fail.

MaciekBaron (approved):

sanjeevz3009 (commented):
LGTM!

PR #589, opened 2026-04-20
Redirect to main article when correction not found

MaciekBaron:

This PR is related to CMS-1103 and #546.

It changes the behaviour of corrections: if one doesn't exist, the user is redirected to the main article (i.e. the latest version of the article).

This is mainly to improve the behaviour of the site when switching between languages: switching to Welsh when an English correction is open would return a 404.

Note that if the user is trying to download a CSV file, the route returns a 404 when the correction does not exist.

### How to review

1. Create an article
2. Ensure it has one or more corrections
3. Open the English version of the article
4. Open a correction, confirm it displays correctly
5. Change the correction ID in the URL to a non-existent one e.g. 9999
6. Ensure that you are redirected to the main article
7. Return to a valid correction
8. Switch locale to Welsh by using the language switcher
9. Ensure that you were redirected to the main article

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

We still need to decide exactly how to handle correction in different languages. The above solution is considered an acceptable behaviour for now.


cms/articles/models.py
```diff
@@ -670,24 +670,33 @@ def serve_preview(self, request: HttpRequest, mode_name: str) -> TemplateRespons
...
     @path("versions/<int:version>/")
     def previous_version(self, request: HttpRequest, version: int, **kwargs: Any) -> TemplateResponse | HttpResponse:
-        if version <= 0 or not self.corrections:
+        if version <= 0:
             raise Http404
 
+        chart_id = kwargs.pop("chart_id", None)
+        table_id = kwargs.pop("table_id", None)
+
+        is_requesting_download = bool(chart_id or table_id)
```
joepercival, line 681: Is there a possibility of IDs having a value of `0`? In which case this may evaluate to false

BJacksonONS: I think every major relational database starts autoincrement at 1 by default, and it's not great practice to override. I think some applications use 0 to mark foreign keys that have had their related object deleted.

BJacksonONS: Also looks like wagtail generates UUIDs for block IDs in the streamfield json object.

BJacksonONS:
This currently fails and returns a 404 when trying to review revisions for a welsh page that's still an alias.

cms/articles/models.py
```diff
@@ -670,24 +670,33 @@ def serve_preview(self, request: HttpRequest, mode_name: str) -> TemplateRespons
...
+
+        if not self.corrections:
+            if is_requesting_download:
+                raise Http404
+            return redirect(self.get_url(request))
+
         # Find correction by version
-        for correction in self.corrections:
+        for correction in self.corrections:  # pylint: disable=not-an-iterable
```
BJacksonONS, line 689: It looks like `self.corrections` is the list of corrections for the main english page, but then line 696 filters on `self.revisions` which fails for welsh aliases.

BJacksonONS (changes requested):

cms/articles/models.py
MaciekBaron: Thanks, I forgot to consider aliases, fixed here: [36275b7](https://github.com/ONSdigital/dis-wagtail/pull/589/commits/36275b7ceedfd0d592aac244c2634816a00b97f7)

cms/articles/models.py
MaciekBaron: `chart_id` and `table_id` are strings, so it would only evaluate to false if they were empty - however, they cannot be empty, as that returns 404 by the routing.

And, as Ben mentioned, we use UUIDs

cms/articles/models.py
BJacksonONS: Redirect works now.
Just double checking that it's intended for aliases to also still list the revisions for the english page? The links are still there but they always redirect to the latest. You can't see the page with the welsh furniture but english content for a previous version.

image

BJacksonONS: When I publish the alias into it's own page the revisions are still listed but now the links 404 as the revision ID isn't on the converted alias.

MaciekBaron: @MebinAbraham I think we should hide the revisions list on aliases, what do you think?

@BJacksonONS we should do a cleanup after converting an alias to a page, but that might need to be covered in a separate PR. We still need to have a conversation about how to handle these things (as I mentioned in the follow-up section)

BJacksonONS: Yeah I also agree hiding the revisions is probably the best for now.

sanjeevz3009: @MaciekBaron @BJacksonONS @MebinAbraham  Finished reviewing and testing this locally. Apart from the issue Ben pointed out above, this looks good.

I was going to lay out some options, apart from hiding the list of revisions/ corrections for the Welsh alias article pages.

Now I realise we have a very sticky situation.

Follow Maciej's steps on the PR/ create a stats article page, create some corrections etc, get it published. Now go and convert the stats article page into an ordinary Welsh page, and now you can't save or publish that page, due to the validation error of `You cannot create a notice or correction for a page that has not been published yet`. This means we can't have a Welsh standalone version of the page, as it's unusable, as it can't be saved or published. We also can't just remove the corrections, as all the content, including corrections, gets copied, but we can't delete the corrections (intended behaviour). So this will need fixing now/ soon, as we need this to work for phase 2 one-off articles, otherwise we can't convert the Welsh alias of the stats article page and get it published.

MaciekBaron: @sanjeevz3009 yes I think we need to discuss this during the tech meeting - I think the easiest thing would be to remove all corrections automatically once a translation alias is converted to a real page.

sanjeevz3009: Yes, that's one option I was thinking of also. Let's take all this to a tech session for sure, so we don't spiral out of control. I'll add it to the agenda.

sanjeevz3009: @MebinAbraham

RealOrangeOne (approved):

cms/articles/models.py
```diff
@@ -670,24 +670,33 @@ def serve_preview(self, request: HttpRequest, mode_name: str) -> TemplateRespons
...
             raise Http404
 
+        chart_id = kwargs.pop("chart_id", None)
+        table_id = kwargs.pop("table_id", None)
+
+        is_requesting_download = bool(chart_id or table_id)
+
+        if not self.corrections or self.alias_of_id:
+            if is_requesting_download:
+                raise Http404
```
kevinhowbrook, line 685: **suggestion**

I think there is a gap in testing here in checking the 404 is raised here.

cms/articles/tests/test_pages.py
```diff
@@ -143,6 +145,43 @@ def test_correction_routes(self):
         self.assertPageIsRoutable(self.page, "versions/2")
         self.assertPageIsRoutable(self.page, "versions/3")
 
+    def test_nonexistent_correction_redirects_to_article(self):
+        """Requesting a correction version that doesn't exist redirects to the article."""
+        # No corrections exist on this page
+        response = self.client.get(f"{self.page.url}/versions/1")
+        self.assertEqual(response.status_code, HTTPStatus.FOUND)
+        self.assertTrue(self.page.url.endswith(response.url))
```
MebinAbraham: Nit: Use `self.assertRedirects`?

cms/articles/tests/test_pages.py
```diff
@@ -143,6 +145,43 @@ def test_correction_routes(self):
...
+        ]
+        self.page.save_revision().publish()
+
+        welsh_alias = self.page.copy_for_translation(
+            locale=Locale.objects.get(language_code="cy"), copy_parents=True, alias=True
+        )
+
+        response = self.client.get(f"{welsh_alias.url}/versions/1", headers={"host": "cy.ons.localhost"})
+        self.assertEqual(response.status_code, HTTPStatus.FOUND)
+        self.assertTrue(welsh_alias.url.endswith(response.url))
```
MebinAbraham: Nit: Use `self.assertRedirects`?

We should be expecting 307, not 302.

cms/articles/models.py
```diff
@@ -670,24 +670,33 @@ def serve_preview(self, request: HttpRequest, mode_name: str) -> TemplateRespons
...
+
         # Find correction by version
-        for correction in self.corrections:
+        for correction in self.corrections:  # pylint: disable=not-an-iterable
             if correction.value["version_id"] == version:
                 break
         else:
-            raise Http404
+            if is_requesting_download:
+                raise Http404
```
MebinAbraham, line 694: Are these branches tested?. Which tests cover this?

MebinAbraham (commented):
As discussed, happy for this to go in with the caveat that there are still some odd user journeys. We can revisit and improve the corrections journey once we’ve done some further analysis.

cms/articles/models.py
```diff
@@ -670,24 +670,33 @@ def serve_preview(self, request: HttpRequest, mode_name: str) -> TemplateRespons
...
 
+        chart_id = kwargs.pop("chart_id", None)
+        table_id = kwargs.pop("table_id", None)
+
+        is_requesting_download = bool(chart_id or table_id)
+
+        if not self.corrections or self.alias_of_id:
+            if is_requesting_download:
+                raise Http404
+            return redirect(self.get_url(request))
```
BJacksonONS, line 686: Related to @MebinAbraham's comment on the test, would be good to use the utils redirect rather than the django one.

cms/articles/tests/test_pages.py
MaciekBaron: [d975915](https://github.com/ONSdigital/dis-wagtail/pull/589/commits/d9759152ed2f2d4adcc3caf094f926be371cb514)

cms/articles/tests/test_pages.py
MaciekBaron: `assertRedirects` [d975915](https://github.com/ONSdigital/dis-wagtail/pull/589/commits/d9759152ed2f2d4adcc3caf094f926be371cb514)

Django uses 302 by default, are you saying we should do a 307 redirect instead? (If so, why? Method preservation won't matter here)

MaciekBaron: Changed to use our util function here, and asserting 307 now - if we decide to revert back to 302, we can change it: [d0a65ca](https://github.com/ONSdigital/dis-wagtail/pull/589/commits/d0a65ca0614238116cbb471fe0047034c48e754b)[d0a65ca](https://github.com/ONSdigital/dis-wagtail/pull/589/commits/d0a65ca0614238116cbb471fe0047034c48e754b)

cms/articles/models.py
MaciekBaron: [d0a65ca](https://github.com/ONSdigital/dis-wagtail/pull/589/commits/d0a65ca0614238116cbb471fe0047034c48e754b)

cms/articles/tests/test_pages.py
MebinAbraham: Once you use our own utils redirect, you will need to update this. The standard across all services is to use 307/308s only. It's mainly for external facing views but we made the change across the board, so the test would error once our own util is used. Going forward unless there is a reason to use 302/301 we should use our own wrapper: https://github.com/ONSdigital/dis-wagtail/pull/597

cms/articles/models.py
MaciekBaron: Added here [8e2ef1a](https://github.com/ONSdigital/dis-wagtail/pull/589/commits/8e2ef1a165b433cdb8bfc70b58d4e2f86755dc93)

cms/articles/models.py
MaciekBaron: Partially, more tests added here: [8e2ef1a](https://github.com/ONSdigital/dis-wagtail/pull/589/commits/8e2ef1a165b433cdb8bfc70b58d4e2f86755dc93)[8e2ef1a](https://github.com/ONSdigital/dis-wagtail/pull/589/commits/8e2ef1a165b433cdb8bfc70b58d4e2f86755dc93)

cms/articles/tests/test_pages.py
MaciekBaron: I see, makes sense, addressed in the commit above.

MebinAbraham (approved):

BJacksonONS (approved):

PR #590, opened 2026-04-21
feat(i18n): add feature flag for cookies page untranslated notice

MebinAbraham:

1. Moves the untranslated-page notice check out of the base page template and onto the page model, so the behaviour can be overridden per page type.
2. It also adds a feature flag to suppress the notice on `CookiesPage` when needed. By default, the existing behaviour is unchanged and Welsh cookies page aliases will still show the notice. When `CMS_COOKIES_PAGE_UNTRANSLATED_NOTICE_ENABLED` is set to `false`, the notice is hidden for `CookiesPage`.
3. Removes the translations for the Cookies page until we get the full translations back. We cannot have the page be partially translated. `CMS_COOKIES_PAGE_UNTRANSLATED_NOTICE_ENABLED` should be set to `False` once we add the full translation for CookiesPage in.

### How to review

Ensure:
- changes make sense
- untranslated banner behaviour is unchanged for all pages except the cookies page
- you can toggle the notice on the cookies page by setting `CMS_COOKIES_PAGE_UNTRANSLATED_NOTICE_ENABLED`
- No welsh translation of cookies page exist (page furniture/breadcrumb is fine)

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/cms/core/models/base.py
+++ b/cms/core/models/base.py
@@ -203,6 +203,10 @@ class BasePage(PageLDMixin, ListingFieldsMixin, SocialFieldsMixin, Page):  # typ
             return request.build_absolute_uri(request.get_full_path())
         return cast(str, canonical_page.get_full_url(request=request))
 
+    def show_localised_version_not_available_notice(self, request: HttpRequest) -> bool:
+        """Return whether to show a notice that this page is not available in the active language."""
+        return bool(self.alias_of and self.alias_of.locale.language_code != request.LANGUAGE_CODE)
+
     def get_url_parts(self, request: HttpRequest | None = None) -> tuple[int, str | None, str | None] | None:
         """Override get_url_parts to generate URLs without trailing slashes."""
         parts = super().get_url_parts(request)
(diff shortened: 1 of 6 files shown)


BJacksonONS:
I don't know if there's a ticket for this yet but maybe could add a followup item to implement this for the case where a welsh page inherits live status and loses the banner before any translated content can be published.

MebinAbraham:
> I don't know if there's a ticket for this yet but maybe could add a followup item to implement this for the case where a welsh page inherits live status and loses the banner before any translated content can be published.

Don't think so, would you mind raising one.

MebinAbraham:
There are some gaps that weren't previously clear due to how Jinja's chainable undefined works. Going to add logic to handle page object not existing, which currently happens for things like 404s. ~~Might also change the logic to be a list of page types instead of just cookies pages.~~

MaciekBaron (approved):
Change makes sense, acknowledging the skipped test

BJacksonONS (approved):
Looks good

sanjeevz3009 (approved):
LGTM!

PR #591, opened 2026-04-21
ci(github-actions): block merges for PRs with blocking labels

MebinAbraham:

Adds a GitHub Actions workflow that prevents PRs from being merged when any blocking label is present.

### How to review

Ensure workflow makes sense and whether any other changes are needed.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/.github/workflows/pr-blocked-labels.yml
+++ b/.github/workflows/pr-blocked-labels.yml
@@ -27,5 +27,5 @@ jobs:
             On Hold ✋
           add_comment: true
           message:
-            "This PR is being prevented from merging because you have added one of our blocking labels: `{{ applied }}`\n\n
+            "This PR is being prevented from merging because you have added one of our blocking labels: {{ applied }}\n\n
             You'll need to remove it before this PR can be merged."
--- a/.github/workflows/pr-blocked-labels.yml
+++ b/.github/workflows/pr-blocked-labels.yml
@@ -1,0 +1,31 @@
+name: Block merge on blocking labels
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, labeled, unlabeled, edited]
+
+jobs:
+  label-check:
+    name: label-check
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+
+    steps:
+      - uses: mheap/github-action-required-labels@v5
+        with:
+          mode: exactly
+          count: 0
+          labels: |
+            do not merge
+            status: Blocked
+            status: In-progress
+            status: Needs tests
+            status: Needs functional/UI tests
+            invalid
+            On Hold ✋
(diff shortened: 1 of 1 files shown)


BJacksonONS (approved):
Looks good 👍 Did a quick skim of the action source code and no red flags.

.github/workflows/pr-blocked-labels.yml
```diff
@@ -0,0 +1,31 @@
...
+            do not merge
+            status: Blocked
+            status: In-progress
+            status: Needs tests
+            status: Needs functional/UI tests
+            invalid
+            On Hold ✋
+          add_comment: true
+          message:
+            "This PR is being prevented from merging because you have added one of our blocking labels: `{{ applied }}`\n\n
```
MaciekBaron: Non-blocking: Do we need the backticks here? Semantically they are usually reserved for commands or inline code.

MaciekBaron (approved):

sanjeevz3009 (approved):
LGTM!

.github/workflows/pr-blocked-labels.yml
```diff
@@ -0,0 +1,31 @@
...
+          count: 0
+          labels: |
+            do not merge
+            status: Blocked
+            status: In-progress
+            status: Needs tests
+            status: Needs functional/UI tests
+            invalid
+            On Hold ✋
+          add_comment: true
```
RealOrangeOne, line 28: Issue: Adding a comment is quite noisy. It means it's going to be added whenever anything about the PR is "edited", even if it's unrelated to the above labels.

Assuming this action fails the action, is that enough? People can click in if they need more detail on _why_ it's rejected.

MebinAbraham: It edits the existing comment almost always as long as it hasn't been "altered". 

From docs:
> Add a comment to the PR if required labels are missing. If a comment already exists, it will be updated. When the action passes, the comment will be deleted

Let's try it out, and if it becomes noisy or not useful, let's toggle it off.

.github/workflows/pr-blocked-labels.yml
MebinAbraham: [`ae7091b` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/591/commits/ae7091b9fbe6c0864723261e0db2ebb41bec71ee) and [ONSdigital/dis-helm-cms-service-chart@`b2a2bb6` (#113)](https://github.com/ONSdigital/dis-helm-cms-service-chart/pull/113/commits/b2a2bb61c0c45fddd7501d5b5748f91329808436)

MaciekBaron (approved):

PR #593, opened 2026-04-21
Remove Announcement Panel Block

sanjeevz3009:

JIRA Card: https://officefornationalstatistics.atlassian.net/browse/CMS-811

Amelia Nilan highlighted that the `announcement panel` would not be used at the page level under any circumstance. This is supported by the service manual here: https://service-manual.ons.gov.uk/design-system/components/announcement-panel. It says in this area that `Use an announcement panel is used to display important messages that should be seen by all users but are not directly related to the page content. For example, issues with an entire service.`

### How to review

- Make sure the announcement panel is fully removed
- Make sure all the test units, integration and UI tests are all passing

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


BJacksonONS:
Is the template being kept in to use as a service level component? Do we want to move it out of the streamfield directory or delete it?

sanjeevz3009:
> Is the template being kept in to use as a service level component? Do we want to move it out of the streamfield directory or delete it?

@BJacksonONS Good spot: https://github.com/ONSdigital/dis-wagtail/pull/593/commits/7a3606be1e064e0bd32bc6a6140a2da6317d1cd7

BJacksonONS:
Do we need a data migration to handle this gracefully? Or are we assuming there aren't enough/any already there to cause an issue?

sanjeevz3009:
> Do we need a data migration to handle this gracefully? Or are we assuming there aren't enough/any already there to cause an issue?

@BJacksonONS

See https://github.com/ONSdigital/dis-wagtail/blob/main/docs%2Fcustom-features%2Fmigration_friendly_streamfields.md

From an end user's perspective, any pages that contained the announcement panel its totally gone now.

From a Wagtail publishing user point of view, if they had any pages where they had a section and only an announcement panel within the content field, then the page would refuse to validate, as you need something within the content field.

Switch to main -> create an info page -> have a section and only the announcement panel in the content field -> now switch to this branch -> you will see when you go back to the info page in Wagtail admin -> you have to now add something in the content field.

Another area will be again if there is a bundle ready to publish -> then when it tries to publish -> the info page that contains a section -> whose content only contains an announcement panel -> now that info page won't publish -> as it will fail the validation.

So these scenarios are unlikely, and aren't worth writing a data migration for as these scenarios above, would mean, we need to data migration where we need to locate content fields on stats article, methodology, and info page -> see if they have any section where its content field only has an announcement field -> then replace that field with something or placeholder.
- This is not ideal; leave it up to the publishing editors to correct this

Or we have to disable the content field being mandatory temporarily, etc.

Considering all options and this situation, it should be fine to merge this as is, and put a disclaimer to the non-techies and business in Staging when this merges for this rare edge case.

If we were in prod and announcement panels were being used there already (i.e., pages already created and it has announcement panels). Then yes, we would need to handle this gracefully, slowly. I.e., disabling the content field being mandatory, so making sure any pages due to publish via bundles or page-level schedule publishing go ahead as planned without breaking, then once all those scenarios are handled, making the content field mandatory again, etc.

But even in prod, it'll be weird to have an info page example where it has a section and its content field only has 1 announcement panel, haha. But if we were in prod and pages existed with a section and with 1 announcement panel, then yes, we would need to investigate and consider that scenario, and handle it gracefully.

BJacksonONS (approved):

MaciekBaron (approved):

MebinAbraham:
Although this is a breaking change, given we are not in prod yet (almost there), we will treat it as non-breaking, as it was an unused feature. It will be released as part of the initial v1 release to prod.

PR #594, opened 2026-04-22
Improve redis probes

RealOrangeOne:

Currently, if Redis is unavailable, the probe fails and results in the pod being restarted. However, the app still behaves correctly as we use `IGNORE_EXCEPTIONS` to do a cache miss / noop rather than error.

Additionally, I've made the `/health` endpoint check all cache backends rather than just "default".

### How to review

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


cms/core/tests/test_views.py
```diff
@@ -241,7 +216,7 @@ def test_broken_redis_connection(self):
 
         self.assertEqual(response.status_code, 500)
 
-        cache_check = next(check for check in response.json()["checks"] if check["name"] == "cache")
+        cache_check = next(check for check in response.json()["checks"] if check["name"] == "default cache")
```
MebinAbraham: We should add an extra backend and assert that also exists.

cms/core/tests/test_views.py
```diff
@@ -109,20 +98,6 @@ def test_unexpected_database_error(self):
...
-    @override_settings(
-        CACHES={
-            "default": {
-                "BACKEND": "django_redis.cache.RedisCache",
-                "LOCATION": "redis:///does-not-exist",
-                "OPTIONS": {},
-            }
-        }
-    )
-    def test_broken_redis_connection(self):
```
MebinAbraham, line 121: Do we have an app-level test to validate this? Ideally, we have a test that proves broken connections do not hang the app. Currently, we are relying on the underlying libs implementation details.

MebinAbraham (approved):

cms/core/tests/test_views.py
RealOrangeOne: It's not an implementation detail - we set `IGNORE_EXCEPTIONS` which is intended to do exactly this. I tried adding a test to confirm it's enabled, but since our tests override the cache configuration it's not possible.

MebinAbraham: I wonder if `IGNORE_EXCEPTIONS` should be a setting that is enforced in `production.py`

sanjeevz3009 (approved):
Apart from the comments already made, LGTM!

cms/core/tests/test_views.py
RealOrangeOne: It's enforced in `base.py`. We could add an `assert`, but it seems unnecessary?

sanjeevz3009 (approved):

MebinAbraham (approved):

PR #595, opened 2026-04-22
Update rules around deleting and unpublishing pages

MaciekBaron:

This PR relates to CMS-830 and introduces new rules regarding deleting and unpublishing pages.

Consult the ticket to read the exact acceptance criteria, but to sum things up:

* Both publishing admins and publishing officer are able to delete pages which are unpublished, but are not allowed to delete pages which were published
* Publishing admins are able to unpublish a page. However, an unpublished page which was previously published cannot be deleted - in short, only pages which were _never_ published can be deleted.
* Release calendar pages cannot be unpublished by anyone, only cancelled.

This PR also removed redundant checks regarding headline figures - only published headline figures can be used, and therefore the rule about published pages trumps this rule and fulfils it.

**Note**: The card said _"any cases that we cannot use the permission tester or it does not make sense to do so, ensure these are done via Signals and not hooks"_ - however, we want to display the message in those scenarios, so we had to use hooks.

### How to review

* Create pages in various states and of various types as described above
* Try to delete and/or unpublish them, and make sure the rules described above are fulfilled
* Attempt to perform batch actions (delete, unpublish) and ensure the rules described above are fufilled

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

N/A


BJacksonONS (commented):
I'm happy with the changes as-is.

Think we will need to put in some follow-up/discussion items around the edge cases with translated pages.

cms/core/wagtail_hooks.py
```diff
@@ -53,6 +59,90 @@ def global_admin_css() -> str:
...
+    Release Calendar pages are handled by `cms.release_calendar.wagtail_hooks`.
+    """
+    if page.specific_class in (ReleaseCalendarPage, ReleaseCalendarIndex):
+        return None
+
+    if not _page_blocks_deletion(page):
+        return None
+
+    message = format_html(
+        "<b>{}</b><br>This page cannot be deleted because it (or one of its descendants) has been "
```
RealOrangeOne: Issue: I think the format string here isn't quite right. if it's just putting in a string literal, it might as well be the same string with `mark_safe`.

MaciekBaron: Fair point, cleaned up here: [65cc262](https://github.com/ONSdigital/dis-wagtail/pull/595/commits/65cc262523e358e465a109b2ec033127c3d1ec75)

cms/core/wagtail_hooks.py
```diff
@@ -133,9 +133,8 @@ def prevent_bulk_delete_of_previously_published_pages(
         preview += f", and {len(blocked_titles) - _BLOCKED_TITLES_PREVIEW_COUNT} more"
 
     message = format_html(
```
BJacksonONS, line 146: Should this also be `mark_safe`?

MaciekBaron: This one injects the list so I would keep format_html

BJacksonONS: Yep ignore me

cms/core/wagtail_hooks.py
```diff
@@ -1,18 +1,25 @@
 from typing import TYPE_CHECKING
 
 from django.conf import settings
+from django.shortcuts import redirect
```
BJacksonONS: Can we use the internal redirect util?

BJacksonONS (changes requested):

cms/core/wagtail_hooks.py
MaciekBaron: [149ad4f](https://github.com/ONSdigital/dis-wagtail/pull/595/commits/149ad4fcf17c5586f9580335d915f61d83435657)

sanjeevz3009:
> I'm happy with the changes as-is.
> 
> Think we will need to put in some follow-up/discussion items around the edge cases with translated pages.

@BJacksonONS, are you able to share your edge cases/ findings? For posterity/ documentation purposes on this PR. A brief explanation and bullet points would do.

cms/core/wagtail_hooks.py
```diff
@@ -53,6 +60,90 @@ def global_admin_css() -> str:
...
+    )
+
+    messages.warning(
+        request,
+        message,
+    )
+    return redirect("wagtailadmin_pages:edit", page.pk, preserve_request=False)
+
+
+@hooks.register("before_bulk_action")
```
sanjeevz3009, line 120: @MaciekBaron Excellent. I was digging into pressing issues like 4-5 to flush our cards for it, as it should be prioritised for one-off articles/ phase 2. I think you might have fixed 1 or 2 of the issues with this, haha.

Will cross-check.

sanjeevz3009 (commented):
Did a code review. Looking good. Will submit my final round of review + cross-check with some bug cards/ enhancements I was going to create, as your implementation might have resolved those issues as well =) Or this PR with minimal additions/ tweaks might fix those, too.

BJacksonONS:
> > I'm happy with the changes as-is.
> > Think we will need to put in some follow-up/discussion items around the edge cases with translated pages.
> 
> @BJacksonONS, are you able to share your edge cases/ findings? For posterity/ documentation purposes on this PR. A brief explanation and bullet points would do.

I raised this with @MaciekBaron.

Currently there's a case where if you have drafts in both English and Welsh (alias converted to real page), you can publish > unpublish one and still delete the other. I believe this can go both ways, doesn't matter which one was published.

MaciekBaron:
> Currently there's a case where if you have drafts in both English and Welsh (alias converted to real page), you can publish > unpublish one and still delete the other. I believe this can go both ways, doesn't matter which one was published.

@BJacksonONS I would argue this is intended behaviour.

I understand that we have this note on the ticket:

> _Note: Delete and unpublish of a Welsh and English page is consistent. E.g. If you unpublish or delete an English or Welsh, it will affect both sides. (This is a system limitation, for this to change Wagtail would need to provide the option to convert the page back to an alias so that the pages can be unlinked)._

However, I think the intention of the note is to describe the current situation, even if it is not entirely correct. I don't interpret the note as "if you publish an English version, you cannot delete a Welsh version". I wonder what @sanjeevz3009 thinks.

BJacksonONS:
> > Currently there's a case where if you have drafts in both English and Welsh (alias converted to real page), you can publish > unpublish one and still delete the other. I believe this can go both ways, doesn't matter which one was published.
> 
> @BJacksonONS I would argue this is intended behaviour.
> 
> I understand that we have this note on the ticket:
> 
> > _Note: Delete and unpublish of a Welsh and English page is consistent. E.g. If you unpublish or delete an English or Welsh, it will affect both sides. (This is a system limitation, for this to change Wagtail would need to provide the option to convert the page back to an alias so that the pages can be unlinked)._
> 
> However, I think the intention of the note is to describe the current situation, even if it is not entirely correct. I don't interpret the note as "if you publish an English version, you cannot delete a Welsh version". I wonder what @sanjeevz3009 thinks.

The delete can cascade to the other page which this PR blocks from being deleted directly, which is my main reason for thinking this isn't desirable.

MaciekBaron:
> The delete can cascade to the other page which this PR blocks from being deleted directly, which is my main reason for thinking this isn't desirable.

Sorry now I understand. I can look into implementing a safeguard for that.

MaciekBaron:
@BJacksonONS issue addressed here: [208e377](https://github.com/ONSdigital/dis-wagtail/pull/595/commits/208e3771828ed6a90ed5bfc47878d649dfd4aafa)

BJacksonONS (approved):
Translation changes work locally, looks good.

cms/core/tests/test_wagtail_hooks.py
```diff
@@ -0,0 +1,225 @@
...
+        self.client.force_login(self.superuser)
+
+    @staticmethod
+    def _bulk_delete_url(page_ids, next_url="/admin/"):
+        base = reverse("wagtail_bulk_action", args=("wagtailcore", "page", "delete"))
+        query = "&".join(f"id={pid}" for pid in page_ids)
+        return f"{base}?{query}&next={next_url}"
+
+    @staticmethod
+    def _never_published(parent):
```
sanjeevz3009:
```suggestion
    def _make_never_published(parent):
```

cms/core/wagtail_hooks.py
sanjeevz3009: Resolving on behalf.

sanjeevz3009 (changes requested):
Thanks, @MaciekBaron works well. Good tests also.

There are some inconsistent behaviours here around how we throw the error to the user, and user journey:
- When I use the bulk delete mechanism to try to delete pages, and if I am not allowed to, then the error is thrown, and I stay on the same page. I.e., if I am on the home page, trying to delete the index page -> error thrown here -> I am still on the same page.
- But if I try to do the same thing from the index page actions menu (so I am at home but trying to delete the index page via this method) -> then I am taken to the index page itself -> then I see the error on that page. Ideally, when trying to delete the page via the index page actions menu, I should have the same behaviour as when trying to delete the page via the bulk delete mechanism.
  - The same behaviour exists, let's say I go to the index page -> now looking at the children/ info pages. If I use the actions menu on the top nav and click delete (trying to delete the index page via this method) -> I am taken back to the index page -> error is thrown.
- Ideally, this also should have the same behaviour as when trying to delete the page via the bulk delete mechanism.

cms/core/tests/test_wagtail_hooks.py
MaciekBaron: Good spot [f63b01b](https://github.com/ONSdigital/dis-wagtail/pull/595/commits/f63b01b156928c1c353e7e32689f5e05e9d5bd83)

MaciekBaron:
@sanjeevz3009 thanks for pointing that out. That's a valid concern.

To address it I changed the hook to respect the `?next=` parameter and redirect back to the provided path. That should make the experience more consistent.

[7b46707](https://github.com/ONSdigital/dis-wagtail/pull/595/commits/7b46707f148e2cd279a6f8b23547d395de30cd22)

BJacksonONS (approved):
Rechecked the redirect behaviour, looks good

sanjeevz3009:
> * The same behaviour exists, let's say I go to the index page -> now looking at the children/ info pages. If I use the actions menu on the top nav and click delete (trying to delete the index page via this method) -> I am taken back to the index page -> error is thrown.

@MaciekBaron The first scenario is fixed, but not the second one (the one above in this comment).

MaciekBaron:
> > * The same behaviour exists, let's say I go to the index page -> now looking at the children/ info pages. If I use the actions menu on the top nav and click delete (trying to delete the index page via this method) -> I am taken back to the index page -> error is thrown.
> 
> @MaciekBaron The first scenario is fixed, but not the second one (the one above in this comment).

This is the "intended" Wagtail behaviour at the moment (i.e. the delete link there does not include the `next` parameter) - it would require modifying Wagtail's default behaviour. We should discuss internally before deciding if this requires changing.

sanjeevz3009:
> > > * The same behaviour exists, let's say I go to the index page -> now looking at the children/ info pages. If I use the actions menu on the top nav and click delete (trying to delete the index page via this method) -> I am taken back to the index page -> error is thrown.
> > 
> > 
> > @MaciekBaron The first scenario is fixed, but not the second one (the one above in this comment).
> 
> This is the "intended" Wagtail behaviour at the moment (i.e. the delete link there does not include the `next` parameter) - it would require modifying Wagtail's default behaviour. We should discuss internally before deciding if this requires changing.

@MaciekBaron Hmmmm, now this reminds me of a conversation Adam, Mebin, and I had about this, where this exact/ or some behaviour similar to this was affecting the user journey when sending hyperlinks around for pages, copying links, and it was causing a weird issue when a user goes through Florence SSO also. I remember we discussed this, and the `next` parameter was part of the conversation. It was around a month ago, maybe.

Will jog mine and Mebin's brain on it. Then bring it to the tech-session.

Then happy to get this in. Will add to the tech-session agenda.

Thank you for the changes @MaciekBaron

cms/core/wagtail_hooks.py
sanjeevz3009: Checked my list of cards to create, this does solve one of the issues/ bug cards I was going to create, actually. There are a few other ones, but they are unrelated, which this doesn't fix.

@MaciekBaron

sanjeevz3009 (approved):
LGTM! Thanks @MaciekBaron

MebinAbraham (commented):
@MaciekBaron 

> Note: The card said "any cases that we cannot use the permission tester or it does not make sense to do so, ensure these are done via Signals and not hooks" - however, we want to display the message in those scenarios, so we had to use hooks.

The main reason we wanted to use the permission tester was to avoid showing the delete button when the user usually cannot do anything with it. From a UX point of view, it is not ideal to show an action that most users cannot perform, especially if it only results in an error or blocked action after they click it. I think the AC's weren't retrospectively updated, hence the "show message" AC vs not show button option.

PR #596, opened 2026-04-22
fix dockerfile dev target for running with colima

BJacksonONS:

Fixes an issue with dev images built with colima not having all correct dependencies.

### How to review

Try building docker image with colima as the docker engine, shelling into a container and seeing if `honcho start` works and other dependencies (node, npm etc) are installed.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

NA


charlotteromaniuk:
This has fixed the issue for me, `honcho start` now runs fine - thanks

Dockerfile
```diff
@@ -197,29 +197,11 @@ ENV GIT_COMMIT=${GIT_COMMIT} BUILD_TIME=${BUILD_TIME}
...
-    echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_20.x nodistro main" > /etc/apt/sources.list.d/nodesource.list
-    # Update lists again and install Node.js
-    apt --quiet --yes update
-    apt --quiet --yes install nodejs
-    # Tidy up
-    apt --quiet --yes autoremove
-    rm -rf /var/lib/apt/lists/*
-EOF
+COPY ./scripts/install-docker-dev-deps.sh ./install-docker-dev-deps.sh
+RUN chmod +x ./install-docker-dev-deps.sh && ./install-docker-dev-deps.sh && rm ./install-docker-dev-deps.sh
```
MaciekBaron: Question: Upon failure, the script would not be removed (`... $$ rm`). I wonder if that's a problem? Maybe cleanup should happen independently.

BJacksonONS: the current behaviour with `&&` will cause an exit on that line if the script fails. Given we previously had a silent failure that was a little tricky to track down I think this makes sense, but happy to change it if there's consensus against.

BJacksonONS: I've properly set the file permissions in git now though so we should be able to take out the `chmod`.

.docker/install-docker-dev-deps.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
RealOrangeOne, line 1: Question: What is the difference between this and what was previously in the `Dockerfile`? I'm not seeing a difference

BJacksonONS: It's the same, but there must be an issue in [colima](https://github.com/abiosoft/colima) that wasn't parsing the `<<EOF` syntax properly from inside the dockerfile and the script wasn't running.

RealOrangeOne: Weird! You'll want to add `set -eux -o pipefail` here so it behaves correctly and aborts if any part of the script fails.

BJacksonONS: Yep, good call

0b145bade8177b84cbd4863fdb430ce93868936c

.docker/install-docker-dev-deps.sh
```diff
@@ -0,0 +1,22 @@
...
+
```
MaciekBaron, line 3: NP: Now that we have this in the scripts folder, I think we should add a comment at the top of the file explaining that this is only expected to be used by Docker and is not a utility script.

Alternatively we could consider moving this to `.docker/` where we already have `bashrc.sh`

RealOrangeOne: Moving to `.docker` sounds like a good idea to me.

BJacksonONS: Yep makes sense
f5948a1b7f148c6e77f01d64776d61b87a862aef

MaciekBaron (approved):

joepercival (approved):
honcho start works for me now, LGTM

sanjeevz3009 (approved):
Sorted out some minor queries and issues with Ben in a huddle. All is good, LGTM. Functionally tested!

PR #597, opened 2026-04-23
Replace `django.shortcuts.redirect` with `cms.core.utils.redirect`

BJacksonONS:

Replaces uses of django's redirect util with our own.

Sets `preserve_request=False` for cases where post requests redirect to get.

Adds a test for the case where a `StatisticalArticlePage` redirects to edition url, asserting the correct status code.

### How to review

Run tests to ensure they pass.

Run the app and check behaviour is correct for code paths affected by changes.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

NA

### DIFF
--- a/cms/articles/models.py
+++ b/cms/articles/models.py
@@ -7,7 +7,7 @@ from django.core.exceptions import ValidationError
 from django.core.paginator import EmptyPage, PageNotAnInteger, Paginator
 from django.db import models
 from django.http import Http404
-from django.shortcuts import get_object_or_404, redirect
+from django.shortcuts import get_object_or_404
 from django.utils.functional import cached_property
 from django.utils.html import strip_tags
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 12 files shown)


RealOrangeOne (approved):
Change looks good to me.

There are a few cases of `RedirectView` in `cms/urls.py` which probably want the same `preserve_request` treatment.

MebinAbraham:
> Change looks good to me.
> 
> There are a few cases of `RedirectView` in `cms/urls.py` which probably want the same `preserve_request` treatment.]

@BJacksonONS let's do these also. Thanks

BJacksonONS:
Discussed with @MebinAbraham, keeping the current config in `cms.urls` as django's `RedirectView` doesn't currently have a `preserve_request` option and there's not currently a compelling enough reason to write our own override.

MaciekBaron (approved):

RealOrangeOne:
> django's RedirectView doesn't currently have a preserve_request

TIL - nice find! I've raised a ticket to Django to see about getting it added https://code.djangoproject.com/ticket/37062

sanjeevz3009 (commented):
Reviewed!

PR #599, opened 2026-04-24
Replace URLBlock with RelativeOrAbsoluteURLBlock

MaciekBaron:

This PR addresses CMS-1187 and replaces instances of `URLBlock` with `RelativeOrAbsoluteURLBlock`.

### How to review

1. Add a `ThemeLinkBlock` when editing a menu
2. Ensure that you are allowed to provide a relative path as the "external url" _(see comment below)_
3. Add a `ManualDatasetBlock` when adding datasets to a statistical article page
4. Confirm that you are allowed to provide a relative path
5. Add a `TimeSeriesPageStoryBlock` when editing a topic page
6. Confirm that you are allowed to provide a relative path

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions
We need to confirm the following:

* `DatasetStoryBlock` and `TimeSeriesPageStoryBlock` check for duplicates, but they both do it slightly differently. We need to clarify if we want to treat `https://ons.gov.uk/datasets/foo` and `/datasets/foo` as duplicates in `DatasetStoryBlock`, and the similarly in `TimeSeriesPageStoryBlock`. Currently after the change, the datasets would treat that as a duplicate but the time series wouldn't (because the latter uses `normalise_url`, while the former uses` extract_url_path`)
* `ThemeLinkBlock` uses `external_url` as the name for the field, but now accepts relative paths. We should confirm that this is okay.


MebinAbraham:
> ThemeLinkBlock uses external_url as the name for the field, but now accepts relative paths. We should confirm that this is okay.

This is true in other places also, so we'll come back to it and tidy up.

cms/core/url_utils.py
```diff
@@ -33,7 +33,11 @@ def validate_ons_url_struct_block(
 def validate_ons_url(url: str) -> ValidationError | None:
     """Checks that the given URL matches the allowed ONS domain,
     otherwise return a dict holding a ValidationError to be used in the clean method of a StructBlock.
+    Relative URLs (starting with "/") are treated as internal and are not checked against the domain whitelist.
     """
+    if url.startswith("/"):
+        return None
```
MebinAbraham: If I put `abc` so no leading `/` this will fall into:
```py
if not parsed_url.hostname or parsed_url.scheme != "https":
        error = ValidationError(
            "Please enter a valid URL. It should start with 'https://' and contain a valid domain name."
        )
```

The error message is misleading. You can't just add, 'or relative url', the function needs a flag to indicate if relative URLs are allowed, then you could edit the message accordingly.y

MebinAbraham (commented):
Looks good just a minor comment.

cms/core/url_utils.py
MaciekBaron: The function is only used in two places and both need the relative URL, so I updated the error message to reflect that: [f2fac38](https://github.com/ONSdigital/dis-wagtail/pull/599/commits/f2fac38c191dd3c7b846804ed51d30b231deb1a8)

Please let me know if this is correct

cms/datasets/blocks.py
```diff
@@ -24,7 +24,7 @@
 class ManualDatasetBlock(StructBlock):
     title = CharBlock(required=True)
     description = TextBlock(required=False)
-    url = URLBlock(required=True)
+    url = RelativeOrAbsoluteURLBlock(required=True)
```
BJacksonONS: We should probably add the helptext here too

BJacksonONS (changes requested):

cms/core/url_utils.py
MebinAbraham: That would work for now, but my point was more given that it is a general utility function that might be called in a place relative URLs are not allowed, the message would be misleading.

cms/datasets/blocks.py
MaciekBaron: [412532b](https://github.com/ONSdigital/dis-wagtail/pull/599/commits/412532bea1c39c19de0cc12ed37a7855656c846e)

BJacksonONS: This one only allows the ONS domain so we should probably have the same text as the timeseries block, or at least something to indicate.

cms/navigation/blocks.py
```diff
@@ -1,15 +1,16 @@
...
 from cms.core.blocks.base import LinkBlock
+from cms.core.blocks.struct_blocks import RelativeOrAbsoluteURLBlock
 
 
 class ThemeLinkBlock(LinkBlock):
     page = PageChooserBlock(required=False, page_type="themes.ThemePage")
 
     # TODO: day 2, remove external_url, title overrides when allowing Theme page creation
-    external_url = URLBlock(required=True, label="External Link")
+    external_url = RelativeOrAbsoluteURLBlock(required=True, label="External Link")
```
BJacksonONS, line 13: This one should probably have helptext indicating it does allow external outside of ONS.

cms/datasets/blocks.py
```diff
@@ -24,7 +24,10 @@
 class ManualDatasetBlock(StructBlock):
     title = CharBlock(required=True)
     description = TextBlock(required=False)
-    url = URLBlock(required=True)
+    url = RelativeOrAbsoluteURLBlock(
+        required=True,
+        help_text="Enter a relative URL (e.g. /some/path) or a full URL.",
```
MebinAbraham: Suggestion: Use the same help text as timeseries block as it has the same validation.
https://github.com/ONSdigital/dis-wagtail/pull/599/changes#diff-2a210916e040b57f4eb3174738fbf4702cd65d9fb7767c356215a84614682c0cR195-R198

cms/core/url_utils.py
MaciekBaron: Right, I see what you mean. I added a parameter here: [a20badd](https://github.com/ONSdigital/dis-wagtail/pull/599/commits/a20badd28c1efe823a98e5932a654561c70d1080)

cms/datasets/blocks.py
MaciekBaron: Actually it's different, only the time series and dataset blocks check for the ONS domain:

https://github.com/ONSdigital/dis-wagtail/blob/0a7289513543690b61aac96dd72d56500751617a/cms/core/tests/test_blocks_struct_blocks.py#L13-L17

cms/datasets/blocks.py
MaciekBaron: Unless my Friday brain is causing me to not see something, it does allow other domains: https://github.com/ONSdigital/dis-wagtail/pull/599#discussion_r3138542620

cms/datasets/blocks.py
MebinAbraham: My comment is against the dataset block, which also calls `validate_ons_url_struct_block`?. What am I missing? 👀

cms/datasets/blocks.py
BJacksonONS: You might be looking at the navigation block?

[ManualDatasetBlock has the ons domain verification.](https://github.com/ONSdigital/dis-wagtail/blob/a20badd28c1efe823a98e5932a654561c70d1080/cms/datasets/blocks.py#L36)

image

cms/datasets/blocks.py
MaciekBaron: Ah yes sorry, as indicated in my other comment it is my Friday brain 🧠 [4b88a42](https://github.com/ONSdigital/dis-wagtail/pull/599/commits/4b88a4248c5c56b60089aa2315bbd4a62c8d9462)

cms/datasets/blocks.py
MaciekBaron: D'oh! [4b88a42](https://github.com/ONSdigital/dis-wagtail/pull/599/commits/4b88a4248c5c56b60089aa2315bbd4a62c8d9462)

cms/core/url_utils.py
```diff
@@ -24,22 +24,30 @@ def validate_ons_url_struct_block(
...
-def validate_ons_url(url: str) -> ValidationError | None:
+def validate_ons_url(url: str, *, allow_relative_urls: bool = False) -> ValidationError | None:
     """Checks that the given URL matches the allowed ONS domain,
     otherwise return a dict holding a ValidationError to be used in the clean method of a StructBlock.
+
+    If allow_relative_urls is True, relative URLs (starting with "/") are treated as internal and are
+    not checked against the domain whitelist.
     """
+    if url.startswith("/") and allow_relative_urls:
+        return None
```
MebinAbraham: Nit: It would be good to tighten our validation here and the `ReleativeOrAbsoluteURLBlock` to not allow things like `///example`, currently it passes validation.

cms/core/url_utils.py
```diff
@@ -53,15 +61,6 @@ def validate_ons_url(url: str) -> ValidationError | None:
     return error
 
 
-def normalise_url(url: str) -> str:
-    """Normalise functionally equivalent URLs for comparison
-    when checking for duplicates across StreamValue entities.
-    """
-    url = url.lower().rstrip("/")  # Treat URLs with and without trailing slashes as equivalent
-    url = url.removeprefix("https://").removeprefix("www.")  # Normalise the URL
-    return url
```
MebinAbraham, line 62: One gap resulting from this is that `/abc/` and `/abc` won't be considered duplicates. We should handle that. The same applies to full URLs. Maybe this should be called in `validate_ons_url` instead.

cms/core/url_utils.py
MaciekBaron: Added a simple regex for this here: [1fff2ca](https://github.com/ONSdigital/dis-wagtail/pull/599/commits/1fff2cac32e60dc4a0477d6bbb1de818de57efa9)

cms/core/url_utils.py
MaciekBaron: This is already handled - I expanded the tests to show this: [1d988c5](https://github.com/ONSdigital/dis-wagtail/pull/599/commits/1d988c5851ecddecce2d2d39a6766f0de1606f38)

cms/core/url_utils.py
```diff
@@ -1,9 +1,12 @@
+import re
 from urllib.parse import urlparse
 
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from wagtail.blocks import FieldBlock, StructValue
 
+ROOT_RELATIVE_URL_RE = re.compile(r"^/([^/]+/?)+$")
```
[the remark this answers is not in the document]
MaciekBaron: Expanded the regex to reduce ambiguity and the backtracking that would cause [54e1399](https://github.com/ONSdigital/dis-wagtail/pull/599/commits/54e139966a440bfa95c1743fe9ea2c55867d9639)

cms/navigation/blocks.py
MaciekBaron: It wasn't there before when external URLs were already allowed so not sure if it is needed. @MebinAbraham ?

MebinAbraham: Easy either way, we have a card to look at help text as a whole in https://officefornationalstatistics.atlassian.net/browse/CMS-340

BJacksonONS: I'm happy then

BJacksonONS (approved):
Looks good

RealOrangeOne (approved):

MebinAbraham (approved):

sanjeevz3009 (commented):
Reviewed!

PR #601, opened 2026-04-24
fix: preserve request method on slash redirects

MebinAbraham:

Updates the non-trailing slash redirect middleware to use the shared CMS redirect helper for permanent redirects, returning `308` instead of `301`.

### How to review

Go to a route with a trailing slash and observe that you now are redirected using 308 not 301.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/cms/core/middleware.py
+++ b/cms/core/middleware.py
@@ -35,7 +35,7 @@ class NonTrailingSlashRedirectMiddleware(MiddlewareMixin):
                 if query:
                     path_without_slash = f"{path_without_slash}?{query}"
 
-                url: HttpResponsePermanentRedirect = redirect(path_without_slash, permanent=True)  # type:ignore
+                url: HttpResponsePermanentRedirect = redirect(path_without_slash, permanent=True)  # type: ignore
                 return url
         return None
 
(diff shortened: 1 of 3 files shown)


cms/core/middleware.py
```diff
@@ -32,7 +34,7 @@ def process_request(self, request: HttpRequest) -> HttpResponsePermanentRedirect
                 query = request.META.get("QUERY_STRING", "")
                 if query:
                     path_without_slash = f"{path_without_slash}?{query}"
-                return HttpResponsePermanentRedirect(path_without_slash)
+                return redirect(path_without_slash, permanent=True)
```
MaciekBaron: We will need to update the return type as `redirect` returns `HttpResponseRedirect | HttpResponsePermanentRedirect`

MaciekBaron (changes requested):

cms/core/middleware.py
MebinAbraham: [`e01f689` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/601/commits/e01f6891b6757673f1fd193ebc95867175011f18) Type ignored and overridden as we want this to always be permanent.

MaciekBaron (approved):

BJacksonONS (approved):

MaciekBaron (approved):

sanjeevz3009 (commented):
Reviewed!

PR #602, opened 2026-04-24
Override translations for cookies page

MebinAbraham:

Adds a Welsh translation for Yes/No, using English so that when the language is toggled, it is not translated. Currently, it is still translated because those simple strings are still in the compiled Django package translations.

This approach was taken since those strings are not used elsewhere. The other option is to not mark them for translation in cookies.html.

Adding tests didn't seem worthwhile, given they will be fully translated in the near future.

### How to review

1. On `main` run `make compilemessages` then switch language on the cookies page and verify Yes/No is in Welsh
2. Switch to this branch, run `make compilemessages` and verify it is in English
3. Verify there are no other cases like this on the cookies page.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/cms/locale/cy/LC_MESSAGES/django.po
+++ b/cms/locale/cy/LC_MESSAGES/django.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2026-04-23 12:47+0100\n"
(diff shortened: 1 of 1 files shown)


BJacksonONS (approved):
Re-ran `make compilemessages`, currently all content in english with welsh page furniture.

MaciekBaron (approved):
Makes sense 👍

sanjeevz3009 (commented):
Reviewed!

PR #603, opened 2026-04-27
CSP and security headers

RealOrangeOne:

This PR makes a few security improvements to the site:

- Add a fairly strict CSP
- The CSRF and session cookies are now marked `httpOnly` for improved security
- Add a `Permissions-Policy` which blocks annoying or invasive features
- Disable HSTS and HTTPS redirects in the application, letting them be handled by the infrastructure (which they already are - this would be duplicate handling never hit)

The site should now match A+ on https://securityheaders.com/ and meets the recommendations of https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Headers_Cheat_Sheet.html

### How to review

Consider any pages which have external resources. Images, external scripts, embeds etc. Confirm not only that they still work, but that no CSP errors appear in the logs.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

CSP can be trivially disabled with `$DISABLE_CSP`.

### Follow-up Actions

Lots of testing will be needed on this one.


cms/settings/base.py
```diff
@@ -1146,3 +1108,50 @@
...
+WAGTAIL_CSP = deepcopy(SECURE_CSP)
+
+for csp_directives in WAGTAIL_CSP.values():
+    try:
+        # Wagtail's scripts don't add a nonce
+        csp_directives.remove(CSP.NONCE)
+    except ValueError:
+        pass
+
+# Wagail (and 3rd-party packages) depend on inline scripts
```
MebinAbraham: Let's document which ones. Any reason why they don't use a `nonce`?. It would be good to raise an issue.

cms/settings/base.py
```diff
@@ -164,11 +166,17 @@
     MIDDLEWARE.insert(common_middleware_index, "django.contrib.sessions.middleware.SessionMiddleware")
     MIDDLEWARE.insert(common_middleware_index, "xff.middleware.XForwardedForMiddleware")
 
+
+if env.get("DISABLE_CSP", "false").lower().strip() != "true":
```
MebinAbraham: Suggestion: Avoid double negatives. Disable false doesn't read as well as enable true. Also, let's namespace this (something we want to do going forward, and something that we want to do retrospectively without causing 🔥 at some point.)

To be consistent with our other env var, I suggest `CMS_CSP_ENABLED`

cms/settings/base.py
```diff
@@ -1146,3 +1108,50 @@
 
 # Disable Wagtail's auto-saving feature
 WAGTAIL_AUTOSAVE_INTERVAL = 0
+
+# Content Security policy settings
+# https://docs.djangoproject.com/en/6.0/ref/csp/
+static_sources = [ONS_CDN_URL, "https://cdnjs.cloudflare.com"]
+SECURE_CSP: dict[str, list] = {
```
MebinAbraham, line 1129: Need to handle GTM/GA + Hotjar:
- https://developers.google.com/tag-platform/security/guides/csp
- https://help.hotjar.com/hc/en-us/articles/36820026388881-Content-Security-Policies

Not sure if we need GA4 links, as we use GTM directly, I would need to check.

cms/static_src/javascript/auth.js
```diff
@@ -48,15 +46,9 @@ function getCookieByName(name) {
...
-  const csrfToken = getCookieByName(csrfCookieName);
-  if (!csrfToken) {
-    console.error('[WAGTAIL] CSRF token not found.');
-    return null;
-  }
-
   const headers = {
     'Content-Type': 'application/json',
-    [csrfHeaderName]: csrfToken,
+    [window.wagtailConfig.CSRF_HEADER_NAME]: window.wagtailConfig.CSRF_TOKEN,
```
MebinAbraham: Nice. Any info on `wagtailConfig`?. Is it always available?

cms/settings/base.py
```diff
@@ -744,36 +752,18 @@
...
 
 
-# This is a setting activating the HSTS header. This will enforce the visitors to use
-# HTTPS for an amount of time specified in the header. Since we are expecting our apps
-# to run via TLS by default, this header is activated by default.
-# The header can be deactivated by setting this setting to 0, as it is done in the
-# dev and testing settings.
-# https://docs.djangoproject.com/en/stable/ref/settings/#secure-hsts-seconds
-DEFAULT_HSTS_SECONDS = 30 * 24 * 60 * 60  # 30 days
-SECURE_HSTS_SECONDS = int(env.get("SECURE_HSTS_SECONDS", DEFAULT_HSTS_SECONDS))
```
MebinAbraham, line 769: Is this currently configured at Cloudflare or Nginx?. If CF, we still want to keep this as the internal routes do not go via Cloudflare.

cms/core/tests/test_views.py
```diff
@@ -43,6 +44,34 @@ def test_google_tag_manager_script_present(self):
         self.assertIn(settings.GOOGLE_TAG_MANAGER_CONTAINER_ID, response.rendered_content)
 
 
+class CSPTestCase(TestCase):
```
MebinAbraham: We should test the effective policy, otherwise, someone could mutate the `SECURE_CSP` config without any tests failing. We currently only validate that it exists in the header and that the nonce is set, not that the expected hosts are set.

cms/settings/base.py
```diff
@@ -1146,3 +1108,50 @@
...
+
+for csp_directives in WAGTAIL_CSP.values():
+    try:
+        # Wagtail's scripts don't add a nonce
+        csp_directives.remove(CSP.NONCE)
+    except ValueError:
+        pass
+
+# Wagail (and 3rd-party packages) depend on inline scripts
+WAGTAIL_CSP["script-src"].append(CSP.UNSAFE_INLINE)
```
MebinAbraham: `SECURE_CSP["script-src"]` already includes unsafe inline which `WAGTAIL_CSP` is a deepcopy of.

cms/settings/base.py
MebinAbraham, line 1129: We want to add `manifest-src` as the design system uses it:

`Loading a manifest from 'https://cdn.ons.gov.uk/sdc/design-system/73.5.0/favicons/manifest.json' violates the following Content Security Policy directive: "default-src 'self' ws://localhost:3000". Note that 'manifest-src' was not explicitly set, so 'default-src' is used as a fallback. The action has been blocked.`

cms/settings/base.py
```diff
@@ -1146,3 +1108,50 @@
...
+WAGTAIL_CSP["style-src"].append(CSP.UNSAFE_INLINE)
+
+# Set a sensible permissions policy to disable privacy-invading or annoying features
+PERMISSIONS_POLICY: dict = {
```
MebinAbraham, line 1162: Since this is an experimental feature, we should only "disable" features that modern browsers support: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy#browser_compatibility

For example, `ambient-light-sensor` is not supported by modern browsers and currently returns a warning in the console. Let's start small with well-supported ones only.

MebinAbraham (changes requested):
This will need some thorough testing by the team to validate that there are no regressions.

cms/settings/base.py
RealOrangeOne: Wagtail support for CSP is still ongoing. Since CSP support was only added in 6.0, it'll be a while before it's there in first-party scripts, let alone 3rd-party. Listing them explicitly is bound to get out of sync, so I don't know it's worth noting them here. It'll be a while before this changes.

cms/settings/base.py
RealOrangeOne: Good point - I'll double check to be sure.

cms/core/tests/test_views.py
RealOrangeOne: I'm not sure those kinds of tests gain anything - it just adds a second place to update. Code review is better at catching unwanted changes.

cms/settings/base.py
RealOrangeOne: Interesting - i'm not sure how I missed that locally.

cms/settings/base.py
zerolab: https://github.com/wagtail/wagtail/issues?q=is%3Aissue%20state%3Aopen%20csp should come in handy

cms/static_src/javascript/auth.js
zerolab: https://github.com/wagtail/wagtail/blob/ccf5b8221b486faca71cb1eba81de130f9da98fe/client/src/entrypoints/admin/core.js#L38
https://github.com/wagtail/wagtail/blob/ccf5b8221b486faca71cb1eba81de130f9da98fe/wagtail/admin/templates/wagtailadmin/admin_base.html#L26

it is always available in the admin

cms/settings/base.py
RealOrangeOne: It's happening in both places - Cloudflare _and_ ingress.

cms/settings/base.py
RealOrangeOne: `ambient-light-sensor` is the only one which shows a warning in the console. Removing it avoids any warnings. Besides `magnetometer`, the rest are supported by latest Chrome

cms/settings/base.py
RealOrangeOne: Hotjar requires `script-src: unsafe-inline`, so I'm going to remove any nonce handling.

cms/settings/base.py
MebinAbraham: If it is not supported across all major browsers, we should drop it. Use this as guidance: https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices

cms/core/tests/test_views.py
MebinAbraham: Code review on its own isn’t a strong enough safeguard here. It relies on someone spotting a subtle change in a fairly dense config, which is easy to miss.

Right now, if something like our GTM domain were removed from the CSP, no tests would fail. The site would still render fine, but tracking would silently break in production. That’s exactly the kind of regression that slips through review and only shows up later.

Given that, it’s worth adding a some tests, even if it means a bit of extra maintenance. The cost of keeping those tests in sync is much lower than the cost of a silent production issue that we could have caught automatically.

cms/settings/base.py
MebinAbraham: Got a link to the source (ping in Slack), for my awareness.

cms/settings/base.py
RealOrangeOne: Firefox doesn't support `Permissions-Policy` at all, so should it be removed entirely?

cms/settings/base.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/603/commits/d03932bf8ecd1e6d111c7b42838c5c7dd93305c7

cms/settings/base.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/603/commits/89701dad6b5fb7e3ae1a8220a42aa5cfa7540dec

cms/settings/base.py
RealOrangeOne: I've removed the poorly supported ones in https://github.com/ONSdigital/dis-wagtail/pull/603/commits/381fc3072a8b584de187c10d6201931d7ec2935e

cms/settings/base.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/603/commits/725fd308ee0c04464a3cc64d85d8b15285fef70e

cms/core/tests/test_views.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/603/commits/b9b476bf1dd749e2c45e79999335fe8779fdf19a

cms/settings/base.py
```diff
@@ -1146,3 +1108,47 @@
 
 # Disable Wagtail's auto-saving feature
 WAGTAIL_AUTOSAVE_INTERVAL = 0
+
+# Content Security policy settings
+# https://docs.djangoproject.com/en/6.0/ref/csp/
+static_sources = [ONS_CDN_URL, "cdnjs.cloudflare.com"]
+SECURE_CSP: dict[str, list] = {
+    "default-src": [CSP.SELF],
+    "frame-src": [CSP.SELF, *IFRAME_VISUALISATION_ALLOWED_DOMAINS],
```
MebinAbraham, line 1131: Given all moderns browsers support CSP, we can drop Xframe. https://github.com/ONSdigital/dis-wagtail/blob/main/cms/settings/base.py#L151

zerolab: if we do that, then we need to define https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/frame-ancestors as per https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Frame-Options

MebinAbraham: Yup ☝️ Set it to `CSP.SELF` on internal instance and `none` on external.

RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/603/commits/034cbd6a546b0e8d5bc67914323ee47a6a885edf

cms/jinja2/templates/base.html
```diff
@@ -28,6 +28,7 @@
 
         <script>
             function loadGTM(){
+                // nonce-aware version of the GTM container
```
MaciekBaron, line 31: Question: Comment will be visible in public-facing code, do we want to keep it?

cms/settings/tests/test_settings.py
```diff
@@ -30,3 +34,122 @@ def test_default_wagtail_admin_login_url_uses_home_path(self):
...
+    def _parse_csp(self, header_value: str) -> dict[str, list[str]]:
+        directives = {}
+
+        for directive in header_value.split(";"):
+            parts = directive.strip().split(" ", 1)
+            directives[parts[0]] = parts[1].split()
+
+        return directives
+
+    def get_csp_expressions(self, policy: dict[str, list[str]], directive: str) -> list[str]:
```
MaciekBaron: Minor: this should probably be also prefixed with `_`

cms/settings/base.py
```diff
@@ -145,10 +146,10 @@
     # SecurityMiddleware.
     # http://whitenoise.evans.io/en/stable/#quickstart-for-django-apps
     "cms.core.whitenoise.CMSWhiteNoiseMiddleware",
+    "django_permissions_policy.PermissionsPolicyMiddleware",
     "cms.locale.middleware.SubdomainLocaleMiddleware",
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
-    "django.middleware.clickjacking.XFrameOptionsMiddleware",
```
BJacksonONS, line 151: Do we want to conditionaly enable this if `CMS_CSP_ENABLED` is false and we aren't using `django.middleware.csp.ContentSecurityPolicyMiddleware`?

BJacksonONS: nevermind, thought we might be revering back to old config if that was disabled, but we're just turning all csp settings on/off.

cms/jinja2/templates/base.html
RealOrangeOne: The snippet is directly from Google, so I don't think it's a concern that we're labelling what it is.

cms/settings/tests/test_settings.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/603/commits/5cccb075cc255ce4ebb7dd1f64ca7f018e421770

MaciekBaron (approved):

cms/settings/tests/test_settings.py
```diff
@@ -30,3 +34,122 @@ def test_default_wagtail_admin_login_url_uses_home_path(self):
...
+    def test_frame_ancestors(self):
+        for url in self.urls:
+            with self.subTest(url):
+                response = self.client.get(url)
+
+                csp = self._parse_csp(response.headers["Content-Security-Policy"])
+
+                self.assertIn(CSP.SELF, self._get_csp_expressions(csp, "frame-ancestors"))
+
+                self.assertNotIn("X-Frame-Origin", response.headers)
```
BJacksonONS: Think this might be a typo, should be "X-Frame-Options"?

cms/settings/base.py
```diff
@@ -1158,3 +1119,48 @@
...
+# https://docs.djangoproject.com/en/6.0/ref/csp/
+static_sources = [ONS_CDN_URL, "cdnjs.cloudflare.com"]
+SECURE_CSP: dict[str, list] = {
+    "default-src": [CSP.SELF],
+    "frame-src": [CSP.SELF, *IFRAME_VISUALISATION_ALLOWED_DOMAINS],
+    # UNSAFE_INLINE is required by mathjax
+    "style-src": [CSP.SELF, *static_sources, CSP.UNSAFE_INLINE, "*.hotjar.com"],
+    "img-src": [CSP.SELF, ONS_CDN_URL, "www.googletagmanager.com", "*.hotjar.com"],
+    # UNSAFE_INLINE is required by hotjar
+    "script-src": [CSP.SELF, *static_sources, "*.hotjar.com", CSP.UNSAFE_INLINE],
```
BJacksonONS: Does script-src also need to include tag manager?

cms/static_src/javascript/auth.js
BJacksonONS: Looks like this script is loaded in non-admin pages, so this will fail there.

https://github.com/ONSdigital/dis-wagtail/blob/8df98c1ab9aa935cf98e15646a31680d71f141b8/cms/jinja2/templates/base_page.html#L126-L135

cms/settings/dev.py
```diff
@@ -100,6 +95,7 @@
 # Set auth stub as a trusted origin
 CSRF_TRUSTED_ORIGINS = env.get("CSRF_TRUSTED_ORIGINS", AUTH_STUB_URL).split(",")
 
+SECURE_CSP.setdefault("connect-src", []).extend(["ws://localhost:3000"])  # noqa: F405
```
BJacksonONS, line 98: This also needs to be added to `WAGTAIL_CSP` otherwise websocket connections fail when using cognito auth in dev.

cms/settings/tests/test_settings.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/603/commits/702e9d90c66d63ce45e76d41d7828dfb1094ba3b

cms/settings/base.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/603/commits/25dd603932052addba07640823c2521a672aa595

cms/settings/dev.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/603/commits/1b72dbec8c3e306ca593ef719807fa09578d894f

cms/settings/base.py
BJacksonONS: There was already a check for tag manager in "connect-src", now that's duplicated.

https://github.com/ONSdigital/dis-wagtail/commit/25dd603932052addba07640823c2521a672aa595#diff-b8b0ccd856b070d3ad07bd60821584a2e8e9d9aac499fce9d67f207819617b94L75

BJacksonONS (approved):

cms/settings/base.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/603/commits/297ba990ad32b3b01ece6d36355fa904f061010d

cms/static_src/javascript/auth.js
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/603/commits/552d69697a69b2ba96bb207c6ad0f2f4d947807b

cms/settings/base.py
MebinAbraham: @RealOrangeOne

cms/settings/base.py
```diff
@@ -164,11 +165,17 @@
...
+if env.get("CMS_CSP_ENABLED", "true").lower().strip() == "true":
+    MIDDLEWARE.append("django.middleware.csp.ContentSecurityPolicyMiddleware")
+
+
 ROOT_URLCONF = "cms.urls"
 
 context_processors = [
     "django.template.context_processors.debug",
     "django.template.context_processors.request",
+    "django.template.context_processors.csp",
```
MebinAbraham: Suggestion: Conditionally add this if `CMS_CSP_ENABLED` is true.

RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/603/commits/18b41e8243c25a0cc43c84e01cabc75b2af9e656

sanjeevz3009 (approved):
Apart from the existing comments. LGTM!

cms/settings/base.py
```diff
@@ -1158,3 +1122,48 @@
...
 WAGTAIL_AUTOSAVE_INTERVAL = int(env.get("WAGTAIL_AUTOSAVE_INTERVAL", 500))
+
+# Content Security policy settings
+# https://docs.djangoproject.com/en/6.0/ref/csp/
+static_sources = [ONS_CDN_URL, "cdnjs.cloudflare.com"]
+SECURE_CSP: dict[str, list] = {
+    "default-src": [CSP.SELF],
+    "frame-src": [CSP.SELF, *IFRAME_VISUALISATION_ALLOWED_DOMAINS],
+    # UNSAFE_INLINE is required by mathjax
+    "style-src": [CSP.SELF, *static_sources, CSP.UNSAFE_INLINE, "*.hotjar.com"],
```
MebinAbraham, line 1133: We need to add `https://fonts.googleapis.com` as it is used to load matierial icons in the admin.

MebinAbraham (approved):

cms/settings/base.py
MebinAbraham: [`c20219b` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/603/commits/c20219bee3de762df8fe0b7356e4884007d3fa20)

cms/settings/base.py
```diff
@@ -1146,6 +1146,10 @@
     "manifest-src": [CSP.SELF, ONS_CDN_URL],
     "frame-ancestors": [CSP.NONE if IS_EXTERNAL_ENV else CSP.SELF],
 }
+if not IS_EXTERNAL_ENV:
```
MaciekBaron, line 1152: Could we just add a comment explaining why we need this for internal only?

MebinAbraham: [`183a373` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/603/commits/183a3736815f92ee0182b2731db6aca1f4f7777a)

MaciekBaron (approved):

PR #604, opened 2026-04-27
Make Article Series pages bundleable

MaciekBaron:

This PR addresses CMS-986 and makes the Article Series Page bundleable.

### How to review

1. Create an Article Series Page
2. Create or edit an existing bundle
3. Ensure that you are able to add the page from step 1 to the bundle

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

N/A


sanjeevz3009 (commented):
Looks good. There are some weird buggy issues here, specifically with article series and all pages in general, when it comes to adding them to the bundle.

On the surface, the functionality all works, of course. The implementation is what I would have also done.

I will reproduce this and write up here tomorrow morning.
- We can fix the issues as a follow-up  + track it with a card, as I am cautious not to tamper with bundles near the go-live date.
- Let's see if the fixes are straightforward and tackle them separately.
- We can jump on a huddle and see what's up @MaciekBaron 
- But it would be good to fix, but it shouldn't be too bad for phase 1, but for phase 2, I think we have to fix it

cms/bundles/tests/test_wagtail_hooks.py
```diff
@@ -156,16 +156,16 @@ def test_add_to_bundle_buttons(self):
 
     def test_add_to_bundle_buttons__doesnt_show_for_pages_in_bundle(self):
```
sanjeevz3009: I see this test has been repurposed, but it seems misleading. Looking at the logic and what the test is doing vs the test case title: `test_add_to_bundle_buttons__doesnt_show_for_pages_in_bundle`

The setup links `self.statistical_article_page` into a bundle, but the assertions are made against `articles_index_page` URLs. That means this test currently behaves more like `button not shown for this index page/context` than strictly `button hidden because this same page is already in a bundle`.

cms/bundles/tests/test_utils.py
```diff
@@ -448,3 +449,13 @@ def test_default_language_is_english(self):
         content = serialize_bundle_content_for_published_release_calendar_page(bundle)
 
         self.assertEqual(content[0]["value"]["title"], "Publications")
+
+    def test_article_series_page_excluded_from_release_calendar_content(self):
```
sanjeevz3009, line 453: Nice addition.

Info for other devs, as I did some poking around, this is testing if you add an article series page to the bundled pages within a bundle, it gets filtered out on the release calendar (if you use to schedule the bundle with) page content.
- Currently, the only pages that will end up in the release calendar page content is Statistical Article and Methodology Pages.
  - I.e., if you add Index, info, article series, statistical article, methodology page etc into the bundled pages in the bundle -> use a release calendar page to schedule it -> once the bundle publishes -> in the release calendar page content section, it will only display statistical article and methodology pages, the other pages will go live, but will not be part of the RC page content.

sanjeevz3009 (approved):
Left a minor comment. Other than that, LGTM!

sanjeevz3009:
> Looks good. There are some weird buggy issues here, specifically with article series and all pages in general, when it comes to adding them to the bundle.
> 
> On the surface, the functionality all works, of course. The implementation is what I would have also done.
> 
> I will reproduce this and write up here tomorrow morning.
> 
> * We can fix the issues as a follow-up  + track it with a card, as I am cautious not to tamper with bundles near the go-live date.
> * Let's see if the fixes are straightforward and tackle them separately.
> * We can jump on a huddle and see what's up @MaciekBaron
> * But it would be good to fix, but it shouldn't be too bad for phase 1, but for phase 2, I think we have to fix it

Spoke to @MaciekBaron offline. The issues I discovered are unrelated to this PR, as before I thought it might be something unique to the article series, but it's all the pages that are bundleable.

**Issue:**
- Currently, any page that is bundleable -> if the page is live with no draft versions of it existing -> when you go to a bundle/ create a bundle -> you can't choose that page from the bundled pages chooser, as you can only add pages that have a draft version -> _but you can_ however use the `Add to bundle` button on the page explorer, the individual page or the actions menu and get that page into the bundle -> this should not be the expected behaviour -> it's not a great UX as it's very misleading and has consquences.

I will be creating a ticket for this bug and proposing some solutions that Maciej and I discussed.

CC: @MebinAbraham

BJacksonONS (approved):
Tested locally, looks good

cms/bundles/tests/test_wagtail_hooks.py
MaciekBaron: Renamed test here: a206c45

sanjeevz3009: Thanks Lord Baron @MaciekBaron

sanjeevz3009 (approved):
LGTM!

PR #605, opened 2026-04-28
Ensure the heading level used for the 'footnotes' text below the table is correctly nested

helenb:

Currently if footnotes are added to a table block, the headings are nested incorrectly. The onsDetails component is used to render the footnotes, and defaults to h2. We need to use a h4 if there is a title for the table (as the title uses h3), and a h3 otherwise (as then the previous heading level is the section heading, which uses h2).

This is one of the issues from the review in https://officefornationalstatistics.atlassian.net/browse/CMS-1169

I have a created a ticket specifically for this issue at https://officefornationalstatistics.atlassian.net/browse/CMS-1204

### How to review

In your local build, create an information page. Add a few table blocks to it, testing both with and without table headings. Run an accessibility checker such as Axe to ensure the headings are nested correctly, or check manually by reviewing the source code for the page.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


cms/core/blocks/markup.py
```diff
@@ -194,11 +194,14 @@ def get_context(self, value: dict, parent_context: dict | None = None) -> dict:
         if block_id and parent_context:
             options["download"] = self._get_download_config(parent_context=parent_context, block_id=block_id, data=data)
 
+        footnotes_heading_level = "4" if value.get("title") else "3"
```
MaciekBaron: Nit-pick: the spec says that the `headingLevel` should be an `int` instead of a string

https://service-manual.ons.gov.uk/design-system/components/details#example.example-details.nunjucks-macro-options

MaciekBaron (changes requested):

cms/core/blocks/markup.py
helenb: Good spot - now updated.

MebinAbraham: We should update the test to ensure it renders a heading with the correct level. Our current test only validates that the config is passed, not that the DS can make sense of it. The above issue would be picked up by a test that validates the actual heading.

helenb: OK I can add that - I wasn't sure if that was needed as presumably the DS has tests for the actual rendering.

MebinAbraham: It does, but it proves we passed them a config that is appropriate for the component. Otherwise, we could pass any config and our test would still pass, even if it wouldn't render as the DS expected, due to a different key name, etc.

helenb: That's now added @MebinAbraham

BJacksonONS (approved):
LGTM

MebinAbraham (approved):

MaciekBaron (approved):

sanjeevz3009 (approved):
LGTM!

PR #606, opened 2026-04-28
Add caching for the Dataset API requests

MaciekBaron:

This PR addresses CMS 421 and introduces caching for requests made to the Dataset API.

Key changes:

* The actual `requests.get` call was moved to a module level function where we apply `@memory_cache`
* Parameters and base headers are sorted before the call to ensure the cache key is deterministic

### How to review

Ensure that the tests fully cover the functionality and pass.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

Adjust the default TTL to whatever we think is sensible.


cms/datasets/models.py
```diff
@@ -18,7 +19,32 @@
...
-# TODO: This needs a revisit. Consider caching + fewer requests
+@memory_cache(
+    settings.CMS_DATASETS_API_CACHE_TTL_SECONDS,
+    prefix="datasets_api",
+)
+def _fetch_datasets_api_response(
+    url: str,
+    params_items: tuple[tuple[str, str | int | float | None], ...],
+    base_headers_items: tuple[tuple[str, str], ...],
+    token: str | None,
```
MaciekBaron, line 30: Note that `memory_cache` generates the key using md5, so the raw token will not be stored in the cache backend:

https://github.com/peterbe/django-cache-memoize/blob/4f4c5e323abb7975a10b70f38a619db4a74d5823/src/cache_memoize/__init__.py#L146-L148

MebinAbraham (commented):
This is not something we want to do anymore, at least with in-memory caching. This needs further discussion and thought.

PR #607, opened 2026-04-28
feat: redirect Wagtail admin home without trailing slash

MebinAbraham:

Redirect the configured Wagtail admin home path without a trailing slash to the slash-suffixed route when `WAGTAILADMIN_HOME_PATH` includes one.

This also derives `WAGTAILADMIN_LOGIN_URL` from `WAGTAILADMIN_HOME_PATH` by default, so custom admin paths keep their login URL aligned unless explicitly overridden.

### How to review

- Visit `WAGTAILADMIN_HOME_PATH` without a slash, and verify you are redirected to `WAGTAILADMIN_HOME_PATH`
- Verify there are no regressions to the admin

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


cms/urls.py
```diff
@@ -77,6 +77,13 @@
         ]
 
     wagtail_admin_patterns += wagtailadmin_urls.urlpatterns
+    if settings.WAGTAILADMIN_HOME_PATH.endswith("/"):
+        private_urlpatterns.append(
+            path(
+                settings.WAGTAILADMIN_HOME_PATH.removesuffix("/"),
```
MaciekBaron: Nit: in other places in the codebase we use `rstrip` as it is fine for a single character

BJacksonONS (approved):
Looks good, works locally

cms/settings/base.py
```diff
@@ -865,8 +865,11 @@
 if "WAGTAILADMIN_BASE_URL" in env:
     WAGTAILADMIN_BASE_URL = env["WAGTAILADMIN_BASE_URL"]
 
+WAGTAILADMIN_HOME_PATH = env.get("WAGTAILADMIN_HOME_PATH", "admin/")
+DJANGO_ADMIN_HOME_PATH = env.get("DJANGO_ADMIN_HOME_PATH", "django-admin/")
+
 # https://docs.wagtail.org/en/latest/reference/settings.html#wagtailadmin-login-url
-WAGTAILADMIN_LOGIN_URL = env.get("WAGTAILADMIN_LOGIN_URL", "/admin/login/")
+WAGTAILADMIN_LOGIN_URL = env.get("WAGTAILADMIN_LOGIN_URL", f"/{WAGTAILADMIN_HOME_PATH}login/")
```
MaciekBaron, line 872: Nit: should we test for this, i.e. that it defaults to this combined value?

cms/urls.py
MebinAbraham: Updated as although `admin///` is not deemed appropriate, technically, it does work and then the path stripping would be as expected. [`109c339` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/607/commits/109c3397818c57dfa01eab5931e910566d9db064)

cms/settings/base.py
MebinAbraham: [`f923ba7` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/607/commits/f923ba73b1075c62bc238c3c86b2ee63f5597b38)

MaciekBaron (approved):

sanjeevz3009 (approved):
LGTM!

PR #608, opened 2026-04-29
bug: Enforce field validation on contact and definition snippets

kevinhowbrook:
###  What is the context?

Ticket - [CMS-1066](https://officefornationalstatistics.atlassian.net/browse/CMS-1061?atlOrigin=eyJpIjoiN2E5ZWFkYmEzMWZkNGJmMzg5MWRkM2EyNWQ5NWVlMmEiLCJwIjoiaiJ9)

ContactDetails and Definition snippets both use Wagtail's DraftStateMixin, which allows editors to save incomplete drafts. The problem is that mandatory fields (name, email, definition) were not being validated on draft saves, an editor could submit an empty form and a draft would be created silently.

The fix is to use FieldPanel(required_on_save=True), the mechanism [Wagtail explicitly provides](https://docs.wagtail.org/en/stable/reference/pages/panels.html#wagtail.admin.panels.FieldPanel.required_on_save) for this use case. Fields declared with this flag are excluded from defer_required_fields(), so required-field validation fires on both draft saves and publishes. Optional fields (phone,  owner) are left as-is so editors can still save partial drafts for those.

### How to review

  1. Navigate to Snippets → Contact details → Add (or Definitions → Add).
  2. Submit the form without filling in any fields — you should now see inline validation errors
  for the required fields.
  3. Check that saving a partial draft with only optional fields left blank still works.
  4. Run the tests: poetry run python manage.py test cms.core.tests.test_models
  --settings=cms.settings.test

### Deployment Safety

- This change is safe to deploy without any special steps.

No migrations required. No viewset changes. Publish-path behaviour is unchanged.

### Follow-up Actions

The Jira ticket mentions a database-level constraint. A CheckConstraint would be needed for true DB-level enforcement... That can be addressed but I wanted to double-check that's needed here.

Note: the force pushes are me configuring the signed commits


cms/core/tests/test_models.py
```diff
@@ -82,6 +82,15 @@ def test_publishing_officer_cannot_publish(self):
...
         self.assertContains(response, "Sorry, you do not have permission to access this area.")
 
+    def test_save_draft_without_required_fields_shows_errors(self):
+        # No action-publish in POST data means Wagtail treats this as a draft save.
+        self.client.force_login(self.publishing_admin)
+        response = self.client.post(self.add_url, data={})
+        self.assertEqual(response.status_code, 200)
+        self.assertFormError(response.context["form"], "name", "This field is required.")
+        self.assertFormError(response.context["form"], "email", "This field is required.")
+        self.assertEqual(ContactDetails.objects.count(), 1)
```
MaciekBaron, line 93: Non-blocking nit: Could we make this assertion at the top of the test as well to confirm that we started with `1` and not `0`?

MaciekBaron (approved):

cms/core/tests/test_models.py
kevinhowbrook: added to both, thanks 4c83f24dee6c9a8c981b72c4e08303135149f4e2

sanjeevz3009 (approved):
LGTM!

BJacksonONS (approved):
Looks good

PR #611, opened 2026-04-29
Fix dataset chooser ignoring published flag

MaciekBaron:

This PR addresses CMS-892 and CMS-1200.

It fixes an issue in the dataset chooser which made it ignore the published flag in certain scenarios. This was done by ensuring that `preserve_url_parameters` contains `for_bundle` in the `DatasetChooserViewSet`.

It also fixes a serialisation bug which would trigger a 500 error when no results were returned.

### How to review
---

_A way to test this locally involves taking advantage of functional tests: for example you can increase the timeout in `user_opens_bundle_datasets_chooser` and run the test in non-headless mode so that you can interact with the site while the test is running and utilise the mocked API. You can run the test in non-headless mode by setting `PLAYWRIGHT_HEADLESS=False`, so the full command would be e.g.:_
```bash
PLAYWRIGHT_HEADLESS=False poetry run behave functional_tests -n "A CMS user cannot see the published state dropdown when selecting datasets for a bundle"
```
---
1. Create or edit a bundle
2. Click on "add dataset"
3. Ensure that the modal opens correctly
4. Ensure that you don't see any published datasets
5. Use the search functionality
6. Confirm that you still don't see any published datasets

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

N/A


cms/datasets/views.py
```diff
@@ -308,6 +308,9 @@ class DatasetChosenMultipleView(DatasetChosenMultipleViewMixin, ChosenResponseMi
 class DatasetChooserViewSet(ChooserViewSet):
     model = Dataset
     icon = "tag"
+    # Carry `for_bundle=true` through search/pagination URLs so the bundle chooser
+    # keeps forcing published=false instead of falling back to the form default.
+    preserve_url_parameters: ClassVar[list[str]] = ["multiple", "for_bundle"]
```
kevinhowbrook: **Suggestion (non-blocking)**


```python
 preserve_url_parameters: ClassVar[list[str]] = [*ChooserViewSet.preserve_url_parameters,
  "for_bundle"]
```

If Wagtail adds more parameters to ChooserViewSet.preserve_url_parameters in a future upgrade, this override will silently drop them. Wagtail's own [image chooser uses the safer pattern](https://github.com/wagtail/wagtail/blob/ccf5b8221b486faca71cb1eba81de130f9da98fe/wagtail/images/views/chooser.py#L381)... plus we also get multiple for free from [ChooserViewSet](https://github.com/wagtail/wagtail/blob/ccf5b8221b486faca71cb1eba81de130f9da98fe/wagtail/admin/viewsets/chooser.py#L36)

cms/datasets/tests/test_views.py
```diff
@@ -296,6 +298,41 @@ def test_audit_logging_for_unpublished_datasets(self, mock_ons_dataset, mock_log
...
+
+    @patch("cms.datasets.models.ONSDatasetApiQuerySet.fetch_api_response")
+    def test_for_bundle_preserved_in_rendered_chooser_with_additional_parameters(self, mock_fetch):
+        mock_fetch.return_value = {"items": [], "count": 0, "total_count": 0}
+
+        response = self.client.get(self.chooser_url, {"abc": "def", "for_bundle": "true"})
+
+        self.assertEqual(response.status_code, 200)
+        # The abc parameter will not be preserved.
+        self.assertContains(response, f'{self.chooser_url}results/?for_bundle=true\\"')
```
kevinhowbrook: **Nit-pick (non-blocking)**


The \\" boundary makes this test sensitive to Wagtail's HTML quoting style (although unlikely to change). Perhaps split into two explicit assertions instead:

```python
self.assertContains(response, f"{self.chooser_url}results/?for_bundle=true")
self.assertNotContains(response, "abc=def")
```

cms/datasets/views.py
MaciekBaron: Great point, changed here: [5dacb85](https://github.com/ONSdigital/dis-wagtail/pull/611/commits/5dacb8501820aed01fa874ec21a6690cf4848b0b)

cms/datasets/tests/test_views.py
MaciekBaron: That's fair: [5dacb85](https://github.com/ONSdigital/dis-wagtail/pull/611/commits/5dacb8501820aed01fa874ec21a6690cf4848b0b)

kevinhowbrook (approved):

RealOrangeOne (approved):

MebinAbraham (approved):
👍

sanjeevz3009 (commented):
LGTM!

PR #614, opened 2026-04-30
Ensure live pages with non published revisions are published by a bundle publish

BJacksonONS:

Addresses CMS-1209

Logic around bundle publishing changed at some point that prevented pages that were live but with new revisions being published when a bundle was published.

This PR ensures that revisions to already live pages are published as part of a bundle, and adds a regression test for the cases where they are not. It also adds logging for cases where zero pages are published as part of a bundle with tests.

### How to review

1. Publish a page
2. Create a draft revision of the page
3. Add to a bundle
4. Publish the bundle
5. Ensure the new changes were published, and not ignored

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### DIFF
--- a/cms/bundles/tests/test_management_commands.py
+++ b/cms/bundles/tests/test_management_commands.py
@@ -111,7 +111,7 @@ class PublishBundlesCommandTestCase(TestCase):
 
         # Check that we have a log entry
         self.assertEqual(ModelLogEntry.objects.filter(action="wagtail.publish.scheduled").count(), 1)
-        self.assertEqual(PageLogEntry.objects.filter(action="wagtail.publish.scheduled").count(), 2)
+        self.assertEqual(PageLogEntry.objects.filter(action="wagtail.publish.scheduled").count(), 3)
 
     @override_settings(SLACK_NOTIFICATIONS_WEBHOOK_URL="https://slack.example.com")
     @patch("cms.bundles.notifications.slack.notify_slack_of_publication_start")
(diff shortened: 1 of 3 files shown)


cms/bundles/utils.py
```diff
@@ -395,8 +408,9 @@ def publish_bundle(bundle: Bundle, *, update_status: bool = True) -> bool:
     notifications.notify_slack_of_publication_start(bundle, url=bundle.full_inspect_url)
 
     failed_page_publishes: list[int] = []
+    successful_page_publishes = 0
 
-    for page in bundle.get_bundled_pages().specific(defer=True).select_related("latest_revision").not_live():
```
kevinhowbrook, line 399: Does removing .not_live()  introduce the possibility that any live page in a bundle where latest_revision == live_revision (no new draft) will now be republished? (triggering signals and creating redundant log entries and perhaps the incoming activity log?) There could be a genuine reason to republish in this case I'm missing though.

I've written a failing test to try and explain this, showing a new publish log entry being created for a page that had nothing new to publish:

```python
def test_publish_bundle_does_not_republish_live_page_with_no_new_revision(self):
    self.statistical_article.publish(self.statistical_article.get_latest_revision())
    self.statistical_article.refresh_from_db()
    self.assertEqual(
        self.statistical_article.live_revision,
        self.statistical_article.get_latest_revision(),
    )

    BundlePageFactory(parent=self.bundle, page=self.statistical_article)

    log_count_before = PageLogEntry.objects.filter(
        page=self.statistical_article, action="wagtail.publish.scheduled"
    ).count()

    self.call_command()

    log_count_after = PageLogEntry.objects.filter(
        page=self.statistical_article, action="wagtail.publish.scheduled"
    ).count()

    self.assertEqual(log_count_before, log_count_after)
```

We could do something like (or write it with Q but less readable):

```python
for page in bundle.get_bundled_pages().specific(defer=True).select_related("latest_revision"):
    if page.live and page.live_revision_id == page.latest_revision_id:
        successful_page_publishes += 1
        continue
```

cms/bundles/tests/test_management_commands.py
```diff
@@ -194,6 +194,67 @@ def test_publish_bundle_with_release_calendar(self):
...
+        mock_utils_logger.error.assert_any_call(
+            "No pages were successfully published",
+            extra={
+                "bundle_id": self.bundle.pk,
+                "event": "publish_failed_no_success",
+            },
+        )
+
+    def test_publish_bundle_with_zero_pages(self):
+        """Test that a bundle with zero pages is not marked as published."""
```
kevinhowbrook, line 249: **Praise**

Good edge case testing!

cms/bundles/utils.py
BJacksonONS: There's no scenario where this can happen. 

Except for a UI bug flagged by @sanjeevz3009, you can only add pages to a bundle that have unpublished revisions. When you go to approve a bundle it has checks to ensure that there are unpublished changes for all pages contained, and there's no way to add a page to a bundle once it's been approved. It's also not possible to remove any unpublished drafts once they've been created so it's not possible to revert a page back to it's published without changes status after adding it to an approved bundle.

BJacksonONS: I suppose we could write a regression test for this case though in case of future changes, I'd probably not want to increment `successful_page_publishes` in that case though and skip.

kevinhowbrook: Yea my bad, I missed the bundle approval guards that prevent that, this is redundant given those constraints. Agree a regression test for future-proofing could still be useful if the business logic ever changes.

kevinhowbrook (approved):

cms/bundles/utils.py
```diff
@@ -395,8 +408,9 @@ def publish_bundle(bundle: Bundle, *, update_status: bool = True) -> bool:
     notifications.notify_slack_of_publication_start(bundle, url=bundle.full_inspect_url)
 
     failed_page_publishes: list[int] = []
+    successful_page_publishes = 0
```
MebinAbraham, line 413: Non-blocker nit: Is it worth adding a comment that this is to log any regressions, as this should not be possible anymore until it regresses...

MebinAbraham (approved):

cms/bundles/utils.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/614/commits/db865cf5827ac6937415639156b08ddeea2852d5

MebinAbraham (approved):

sanjeevz3009 (commented):
Reviewed!

PR #618, opened 2026-05-01
Fixes manually published bundle setting unknown approval data

kevinhowbrook:

Addresses [CMS-929](https://officefornationalstatistics.atlassian.net/browse/CMS-929?atlOrigin=eyJpIjoiYjMxZmFmYmQxODgyNDQ5NTlmOGUwNDQ1ZTM0ZjFhNDgiLCJwIjoiaiJ9)

When a Publishing Admin published a bundle directly (without a separate approval step), the `approved_by` and `approved_at` fields were never populated. Additionally, when publishing from `APPROVED` status, the form's `clean()` method was actively clearing those fields. The inspect view then had no approval data to display and fell through to "Unknown approval data".

Screenshot(1)

#### Here are the exact conditions required to reproduce the bug as a PA user:

1. Create a bundle with no publication date and no linked release calendar page
2. Add a page that is ready to publish
3. Move the bundle to "In Preview" via Save to Preview
4. Approve the bundle
5. Manually click Publish in the UI (only visible because there is no publication date, normally publishing is handled by the management command on a schedule)
6. Navigate to the bundle inspect view via the Published status filter (published bundles are
hidden from the default listing)
7. Check the Approval status field shows "Unknown approval data" (like the screenshot above)

####  Why this is extremely unlikely in practice:

- Bundles almost always have either a publication date or a linked release calendar page, without one the bundle has no scheduled go-live time
- The management command handles publishing in production, meaning the manual publish button in the UI is rarely if ever used in a real workflow (although I can't be sure of that)

In short: it requires a deliberately incomplete bundle setup, a manual publish action that bypasses the normal scheduled workflow.

### How to review

1. Log in as a Publishing Admin
2. Create a bundle with valid mandatory data and no publication date (so you can publish the bundle)
3. Approve and publish the bundle directly
4. Go to the bundles listing, use the filter to show published bundles and find the one you published
5. Open your publlshed bundle in the Inspect view
6. Confirm the Approval status shows the publisher's name and timestamp, not "Unknown approval
  data"

  ### Deployment Safety

  Please select one:

  - [x] Safe to auto-deploy
  - [ ] Not safe to auto-deploy

  ### Follow-up Actions

  None.

### DIFF
--- a/cms/bundles/tests/test_viewsets.py
+++ b/cms/bundles/tests/test_viewsets.py
@@ -375,12 +375,7 @@ class BundleViewSetEditTestCase(BundleViewSetTestCaseBase):
         self.post_with_action_and_test("action-publish", BundleStatus.PUBLISHED, self.bundle_index_url)
 
     def test_bundle_edit_view__manual_publish__sets_approved_by_and_approved_at(self):
-        """Publishing directly should populate approved_by and approved_at.
-
-        When a Publishing Admin publishes a bundle directly (without a separate approval step),
-        the approved_by and approved_at fields are not set, causing the inspect view to show
-        'Unknown approval data' instead of the publisher's name and timestamp.
-        """
+        """Publishing directly should populate approved_by and approved_at."""
         self.bundle.status = BundleStatus.APPROVED
         self.bundle.save(update_fields=["status"])
         self.post_with_action_and_test("action-publish", BundleStatus.PUBLISHED, self.bundle_index_url)
@@ -847,22 +842,6 @@ class BundleViewSetInspectTestCase(BundleViewSetTestCaseBase):
         self.assertContains(response, "1 July 2025 1:37pm")
         self.assertContains(response, "1 July 2025 1:45pm")
         self.assertContains(response, "1 July 2025 2:00pm")
-
-    def test_inspect_view__approval_status__shows_unknown_when_published_directly(self):
-        """Inspect view shows 'Unknown approval data' when a bundle is published directly.
-
-        When a Publishing Admin publishes without a separate approval step, approved_by
-        and approved_at are never set, causing the inspect view to display
-        'Unknown approval data' rather than the publisher's details.
-        """
-        self.bundle.status = BundleStatus.PUBLISHED
-        self.bundle.approved_by = None
-        self.bundle.approved_at = None
-        self.bundle.save(update_fields=["status", "approved_by", "approved_at"])
-
-        response = self.client.get(reverse("bundle:inspect", args=[self.bundle.pk]))
-
-        self.assertNotContains(response, "Unknown approval data")
 
     def test_inspect_view__shows_bundle_api_bundle_id_when_exists(self):
         # Ensure the bundle has "bundle_api_bundle_id" set
--- a/cms/bundles/tests/test_viewsets.py
+++ b/cms/bundles/tests/test_viewsets.py
@@ -374,6 +374,21 @@ class BundleViewSetEditTestCase(BundleViewSetTestCaseBase):
         self.bundle.save(update_fields=["status"])
         self.post_with_action_and_test("action-publish", BundleStatus.PUBLISHED, self.bundle_index_url)
 
+    def test_bundle_edit_view__manual_publish__sets_approved_by_and_approved_at(self):
+        """Publishing directly should populate approved_by and approved_at.
+
+        When a Publishing Admin publishes a bundle directly (without a separate approval step),
+        the approved_by and approved_at fields are not set, causing the inspect view to show
+        'Unknown approval data' instead of the publisher's name and timestamp.
+        """
+        self.bundle.status = BundleStatus.APPROVED
+        self.bundle.save(update_fields=["status"])
+        self.post_with_action_and_test("action-publish", BundleStatus.PUBLISHED, self.bundle_index_url)
+
+        self.bundle.refresh_from_db()
+        self.assertIsNotNone(self.bundle.approved_at)
+        self.assertIsNotNone(self.bundle.approved_by)
+
     def test_bundle_edit_view__shows_release_calendar_page_details(self):
         """Release calendar page's title, status and release date are displayed when selected in bundles."""
         for title, status, expected_text in self.RELEASE_CALENDAR_PAGE_CASES:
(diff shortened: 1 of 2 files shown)


cms/bundles/tests/test_viewsets.py
```diff
@@ -374,6 +374,21 @@ def test_bundle_edit_view__manual_publish__happy_path(self):
         self.bundle.save(update_fields=["status"])
         self.post_with_action_and_test("action-publish", BundleStatus.PUBLISHED, self.bundle_index_url)
 
+    def test_bundle_edit_view__manual_publish__sets_approved_by_and_approved_at(self):
+        """Publishing directly should populate approved_by and approved_at.
+
+        When a Publishing Admin publishes a bundle directly (without a separate approval step),
```
MaciekBaron: Nit: Looks like this comment describes the situation before the fix. Might confuse future devs.

kevinhowbrook: removed, forced into last commit for tidiness

BJacksonONS (approved):

MaciekBaron (approved):

MebinAbraham:
FYI: `In short: it requires a deliberately incomplete bundle setup, a manual publish action that bypasses the normal scheduled workflow.`

This is a feature, not a bug. They want the ability to publish manually without a schedule.

MebinAbraham (commented):
Can this still happen? If not, we should update accordingly: https://github.com/ONSdigital/dis-wagtail/blob/main/cms/bundles/viewsets/bundle.py#L393

We should ensure that if it is approved/published, it is not possible for the approver to be absent.

kevinhowbrook:
> Can this still happen? If not, we should update accordingly: https://github.com/ONSdigital/dis-wagtail/blob/main/cms/bundles/viewsets/bundle.py#L393
> 
> We should ensure that if it is approved/published, it is not possible for the approver to be absent.

@MebinAbraham 

Yea, it's still reachable in two cases this PR doesn't cover:

1. approved_by is a FK with on_delete=SET_NULL, so if the approving user is deleted later, approved_by_id goes back to NULL on already-published bundles.
2. Bundles that were published via action-publish before this fix already have approved_by=NULL in the DB; this PR doesn't backfill them.

MebinAbraham:
> > Can this still happen? If not, we should update accordingly: https://github.com/ONSdigital/dis-wagtail/blob/main/cms/bundles/viewsets/bundle.py#L393
> > We should ensure that if it is approved/published, it is not possible for the approver to be absent.
> 
> @MebinAbraham
> 
> Yea, it's still reachable in two cases this PR doesn't cover:
> 
> 1. approved_by is a FK with on_delete=SET_NULL, so if the approving user is deleted later, approved_by_id goes back to NULL on already-published bundles.
> 2. Bundles that were published via action-publish before this fix already have approved_by=NULL in the DB; this PR doesn't backfill them.

Ah ofc, 👍 

I would like to revisit this in future as we may need to preserve the history, i.e. who approved. However, it might be okay as they could use the audit logs in the event the user was deleted.

kevinhowbrook:
> > > Can this still happen? If not, we should update accordingly: https://github.com/ONSdigital/dis-wagtail/blob/main/cms/bundles/viewsets/bundle.py#L393
> > > We should ensure that if it is approved/published, it is not possible for the approver to be absent.
> > 
> > 
> > @MebinAbraham
> > Yea, it's still reachable in two cases this PR doesn't cover:
> > 
> > 1. approved_by is a FK with on_delete=SET_NULL, so if the approving user is deleted later, approved_by_id goes back to NULL on already-published bundles.
> > 2. Bundles that were published via action-publish before this fix already have approved_by=NULL in the DB; this PR doesn't backfill them.
> 
> Ah ofc, 👍
> 
> I would like to revisit this in future as we may need to preserve the history, i.e. who approved. However, it might be okay as they could use the audit logs in the event the user was deleted.

Thanks @MebinAbraham 

I was curious on the security log to see how it handles this too... looks to me like it's the same problem. It also uses an FK to the user (which is fine if a user is just marked as 'not-active') but if a user is deleted the history log reflects that:

image

So, I'm not 100% sure on how to address the issue here, and the [activity logs](https://github.com/ONSdigital/dis-wagtail/pull/446) could be an  issue too). From what I see _here_ we can do 2 things:

#### 1. Snapshot the users name at approval time
Add a string field that captures the approver's name when approval happens, alongside the existing FK. Display reads from the string, so it survives hard delete.

#### 2. Only soft delete users
Change user deletion to set is_active=False rather than removing the row. Would need conversation/confirmation about how that fits with GDPR / right-to-erasure (handled via a separate anonymisation path that keeps the row but scrubs PII)

sanjeevz3009 (approved):
LGTM! Let's get this =)

sanjeevz3009:
> > > > Can this still happen? If not, we should update accordingly: https://github.com/ONSdigital/dis-wagtail/blob/main/cms/bundles/viewsets/bundle.py#L393
> > > > We should ensure that if it is approved/published, it is not possible for the approver to be absent.
> > > 
> > > 
> > > @MebinAbraham
> > > Yea, it's still reachable in two cases this PR doesn't cover:
> > > 
> > > 1. approved_by is a FK with on_delete=SET_NULL, so if the approving user is deleted later, approved_by_id goes back to NULL on already-published bundles.
> > > 2. Bundles that were published via action-publish before this fix already have approved_by=NULL in the DB; this PR doesn't backfill them.
> > 
> > 
> > Ah ofc, 👍
> > I would like to revisit this in future as we may need to preserve the history, i.e. who approved. However, it might be okay as they could use the audit logs in the event the user was deleted.
> 
> Thanks @MebinAbraham
> 
> I was curious on the security log to see how it handles this too... looks to me like it's the same problem. It also uses an FK to the user (which is fine if a user is just marked as 'not-active') but if a user is deleted the history log reflects that:
> 
> image
> So, I'm not 100% sure on how to address the issue here, and the [activity logs](https://github.com/ONSdigital/dis-wagtail/pull/446) could be an issue too). From what I see _here_ we can do 2 things:
> 
> #### 1. Snapshot the users name at approval time
> Add a string field that captures the approver's name when approval happens, alongside the existing FK. Display reads from the string, so it survives hard delete.
> 
> #### 2. Only soft delete users
> Change user deletion to set is_active=False rather than removing the row. Would need conversation/confirmation about how that fits with GDPR / right-to-erasure (handled via a separate anonymisation path that keeps the row but scrubs PII)

@MebinAbraham @kevinhowbrook Let's get this in as is, and we can revisit in the future.

Spoke to Kev offline:
- He said he is happy to create a ticket so we can have it tracked somewhere, as we need to dig into this further.
- We can talk it through in the next tech session.
- Not a blocker for this bug fix and the scope of this bug card.

Thanks Kev =)

kevinhowbrook:
Thanks all - follow-up ticket created here [CMS-1223](https://officefornationalstatistics.atlassian.net/browse/CMS-1223)

sanjeevz3009:
> Thanks all - follow-up ticket created here [CMS-1223](https://officefornationalstatistics.atlassian.net/browse/CMS-1223)

Excellent, thanks @kevinhowbrook

PR #619, opened 2026-05-05
Update dataset and validate on bundle approval

MaciekBaron:

This PR relates to CMS-900 and introduces a check for dataset content-drift when approving a bundle. It expands on work done in #429.

When a user approves a bundle, the backend checks whether the datasets added to the bundle have changed since they were added to said bundle. If a change is detected (e.g. title or description), the dataset entries get updated and a validation error is returned. The user then can click "Approve" again to perform a successful approval.

### How to review

As this relies heavily on Dataset API behaviour, the recommended way to test this behaviour is to run the included functional test `"An approver is blocked when dataset metadata has changed in the source API"` and observing the behaviour in action by disabling headless mode and adding "slow-mo":

```bash
PLAYWRIGHT_HEADLESS=False PLAYWRIGHT_SLOW_MO=1000 poetry run behave functional_tests -n "An approver is blocked when dataset metadata has changed in the source API"
```

This will allow you to observe the process in a browser window.

Ensure that all test cases are correct.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

Consider adding UX improvements to highlight the changed datasets.


cms/bundles/forms.py
```diff
@@ -189,6 +191,67 @@ def _validate_bundled_datasets_status(self) -> None:
...
+        for form in self.formsets["bundled_datasets"].forms:
+            if not form.is_valid() or form.cleaned_data.get("DELETE"):
+                continue
+            dataset: Dataset | None = form.cleaned_data.get("dataset")
+            if not dataset:
+                continue
+
+            try:
+                item_from_api = queryset.get(pk=dataset.namespace)
+            except Exception:
```
kevinhowbrook: **Suggestion**

Is there not a more specific exception we can catch before this? Or perhaps add a comment as to why we need the bare catch?

cms/bundles/forms.py
```diff
@@ -189,6 +191,67 @@ def _validate_bundled_datasets_status(self) -> None:
...
             raise ValidationError(errors)
 
+    @datasets_bundle_api_enabled
+    def _validate_bundled_datasets_metadata(self) -> None:
+        """Compare bundled datasets' stored metadata against the dataset API.
+
+        On drift, update the local Dataset rows in place and block the approval
+        with a clear error so the approver must re-review the refreshed metadata
+        and reconfirm by submitting again.
+        """
```
kevinhowbrook, line 211: **nitpick, sort of**

It's a bit unusual to perform saves (e.g dataset.save) in a clean method IMO. But its obviously needed... I think it's worth extending this comment a bit with some more context as to _why_... something like:

```python

"""Compare bundled datasets' stored metadata against the dataset API and
  refresh local rows on drift.

  Note: this validator deliberately has a side effect, it writes refreshed
  title/description back to the local `Dataset` rows before raising. This is
  intentional so that when the approver clicks Approve a second time, the
  form re-renders with the up-to-date metadata they need to re-review.
  The save is safe to do here because Django caches `cleaned_data`, so
  `is_valid()` won't re-run this method within the same request.

  On drift, the approval is blocked with a `ValidationError` listing the
  changed fields, forcing the approver to reconfirm by submitting again.
  """

```

cms/bundles/forms.py
```diff
@@ -189,6 +191,67 @@ def _validate_bundled_datasets_status(self) -> None:
...
+        if not settings.BUNDLE_DATASET_STATUS_VALIDATION_ENABLED:
```
kevinhowbrook: **Nitpick**

This is documented in settings/base as `# Feature flag to enable/disable validation of bundled datasets status on bundle approval` - it explicitly says status. Reusing it here for metadata means that comment is now stale (at least from what I'm seeing here)

Potentially worth renaming to BUNDLE_DATASET_VALIDATION_ENABLED and extending the comment. Or adding a separate `BUNDLE_DATASET_METADATA_VALIDATION_ENABLED,`

cms/bundles/tests/test_forms_with_bundle_api.py
```diff
@@ -377,6 +402,142 @@ def test_dataset_validation_empty_contents_array(self):
...
+                    ),
+                    "teams": inline_formset([]),
+                }
+            ),
+            for_user=self.approver,
+        )
+
+        self.assertTrue(second.is_valid(), second.errors)
+
+
```
kevinhowbrook, line 594: **suggestion**

Is it worth adding a test for the  API fetch failing? (except Exception path) and asserting the error message

cms/bundles/forms.py
MaciekBaron: Fair point, I made them a bit more specific here: [1485b56](https://github.com/ONSdigital/dis-wagtail/pull/619/commits/1485b560a79b084ec2d784a5261b578e2c131431)

cms/bundles/tests/test_forms_with_bundle_api.py
MaciekBaron: Test added here: [1485b56](https://github.com/ONSdigital/dis-wagtail/pull/619/commits/1485b560a79b084ec2d784a5261b578e2c131431)

cms/bundles/forms.py
MaciekBaron: Turned it into two settings: [d2ce149](https://github.com/ONSdigital/dis-wagtail/pull/619/commits/d2ce149a6c20cd5844ab82602d3b7b6b78e8cbf9)

cms/bundles/forms.py
MaciekBaron: Fair point, worth highlighting the side-effect in case this causes unintended consequences later down the line: [d2ce149](https://github.com/ONSdigital/dis-wagtail/pull/619/commits/d2ce149a6c20cd5844ab82602d3b7b6b78e8cbf9)

kevinhowbrook (approved):

cms/bundles/tests/test_forms_with_bundle_api.py
```diff
@@ -377,6 +403,185 @@ def test_dataset_validation_empty_contents_array(self):
...
+        self.ons_patcher = patch("cms.bundles.forms.ONSDataset")
+        self.mock_ons = self.ons_patcher.start()
+        self.mock_ons.objects.get.side_effect = fake_get
+
+    def tearDown(self):
+        self.client_patcher.stop()
+        self.ons_patcher.stop()
+
+    def _approve_form(self, dataset: Dataset) -> Any:
```
BJacksonONS: Tiny nit: Approve is a verb, so at first read I thought this would also submit the form. Would `_approval_form` or `_get_approve_form` be a little bit more clear?

MaciekBaron: Really good point, I'm usually the one nit-picking about this so thanks for flagging! [fff83a9](https://github.com/ONSdigital/dis-wagtail/pull/619/commits/fff83a964acd07fd39994391efa31df15d457272)

cms/bundles/tests/test_forms_with_bundle_api.py
```diff
@@ -9,10 +10,33 @@
...
+    def fake_get(pk: Any = None) -> Any:
+        local = Dataset.objects.filter(namespace=pk).first()
+        item = MagicMock()
+        item.next = None
+        item.title = local.title if local else ""
+        item.description = local.description if local else ""
+        return item
+
+    patcher = patch("cms.bundles.forms.ONSDataset")
+    mock_ons = patcher.start()
```
BJacksonONS: Nit: `patcher`, `client_patcher`, `ons_patcher` and `mock_ons` are quite vague names, especially as the same names are used in both this and a different test class for different things.

cms/bundles/forms.py
```diff
@@ -189,6 +192,75 @@ def _validate_bundled_datasets_status(self) -> None:
...
+        """
+        if not settings.BUNDLE_DATASET_METADATA_VALIDATION_ENABLED:
+            return
+
+        if "bundled_datasets" not in self.formsets:
+            return
+
+        queryset = ONSDataset.objects  # pylint: disable=no-member
+        if self.datasets_bundle_api_user_access_token:
+            queryset = queryset.with_token(self.datasets_bundle_api_user_access_token)
```
MebinAbraham: If we didn't have it, it would be invalid to proceed, as in reality, we always need a token. So I suggest we raise instead.

cms/settings/base.py
```diff
@@ -993,6 +993,11 @@
     env.get("BUNDLE_DATASET_STATUS_VALIDATION_ENABLED", "false").lower() == "true"
 )
 
+# Feature flag to enable/disable validation of bundled datasets metadata on bundle approval
+BUNDLE_DATASET_METADATA_VALIDATION_ENABLED = (
+    env.get("BUNDLE_DATASET_METADATA_VALIDATION_ENABLED", "false").lower() == "true"
```
MebinAbraham: We should default to true.

cms/bundles/forms.py
MaciekBaron: Fair point – this technically is handled a few lines below with the more broader exception handling, but the result is a bit confusing to the user. I added a message asking to reauthenticate: [8cedaa0](https://github.com/ONSdigital/dis-wagtail/pull/619/commits/8cedaa0486fed6cacea694b7297c035e636e82cc)

(Technically, the session hasn't expired because this can only happen if the token is not present, but seems like a good enough way of presenting this to the user)

cms/settings/base.py
MaciekBaron: Currently the status validation defaults to `false`, should both be updated?

cms/bundles/tests/test_forms_with_bundle_api.py
MaciekBaron: Sorry, missed this. I improved the naming here: [ac5d659](https://github.com/ONSdigital/dis-wagtail/pull/619/commits/ac5d659ba83093f577334b83074b348cc2c61b14)

cms/bundles/forms.py
MebinAbraham: If they didn't have an access token, our middleware would kick them out when Cognito login is enabled so im not sure that message is the best approach as they would get redirect to Florence. I guess it is only an issue if they are logged in via Wagtail core login. Happy to leave the change as is given it is edge case that can't happen in prod as is.

cms/settings/base.py
MebinAbraham: I believe we set that to false by default due to a bug on their side. Let's leave the status one as is for now. I'll check with them and raise a PR to update the default if that bug is fixed on their side.

cms/bundles/forms.py
```diff
@@ -189,6 +192,78 @@ def _validate_bundled_datasets_status(self) -> None:
...
+
+            updated_fields = update_dataset_metadata(dataset, title=api_title, description=api_description)
+            if updated_fields:
+                dataset.save(update_fields=updated_fields)
+            drift_messages.append(f"'{old_title}': {'; '.join(changes)}")
+
+        if drift_messages:
+            errors = [
+                "Dataset metadata has changed in the source API since the bundle was last edited or submitted. "
+                "The local records have been refreshed - please re-review the changes and reconfirm to approve:",
```
MebinAbraham: How about:

```
Approval could not be completed because dataset metadata has changed since they were added. The latest metadata has been imported. Please review the changes below and approve the bundle again.
```

cms/bundles/forms.py
```diff
@@ -189,6 +192,78 @@ def _validate_bundled_datasets_status(self) -> None:
...
+                *drift_messages,
```
MebinAbraham, line 270: Can we better format drift messages via ValidationError, i.e can we show nested list items?. This message is not very clear, especially when the title and description change.
Image

If we are limited by the formatting, then I suggest we just list each one independently. Thoughts?
Image

cms/bundles/forms.py
```diff
@@ -189,6 +192,78 @@ def _validate_bundled_datasets_status(self) -> None:
...
+        drift_messages: list[str] = []
+        for form in self.formsets["bundled_datasets"].forms:
+            if not form.is_valid() or form.cleaned_data.get("DELETE"):
+                continue
+            dataset: Dataset | None = form.cleaned_data.get("dataset")
+            if not dataset:
+                continue
+
+            try:
+                item_from_api = queryset.get(pk=dataset.namespace)
```
MebinAbraham: Suggestion, and an inline cache map before the loop, as this would be the same duplicate call for datasets with different editions as we are only querying by namespace.
```py
api_items_by_namespace: dict[str, Any] = {}

for form in self.formsets["bundled_datasets"].forms:
    ...
    namespace = dataset.namespace

    if namespace not in api_items_by_namespace:
        api_items_by_namespace[namespace] = queryset.get(pk=namespace)

    item_from_api = api_items_by_namespace[namespace]
```

cms/bundles/forms.py
```diff
@@ -189,6 +192,78 @@ def _validate_bundled_datasets_status(self) -> None:
...
+            except (requests.exceptions.RequestException, ValueError):
+                logger.exception("Failed to fetch dataset %s for metadata validation on approval", dataset.namespace)
+                raise ValidationError(
+                    f"Could not verify the latest metadata for '{dataset.title}'. Please try again."
+                ) from None
```
MebinAbraham: Issue: The current loop saves each drifted Dataset as soon as it is detected, but a later API failure for another dataset will still fail approval. In that case, the user can see an error saying metadata could not be verified, while some earlier dataset records may already have been updated. That is surprising because the operation is reported as failed but has partially changed local state.

I suggest either validating all bundled datasets first and only then persisting all metadata updates, or wrapping the refresh in a transaction so the operation is atomic.

cms/settings/base.py
MaciekBaron: Changed to `true` here: [4bdf1aa](https://github.com/ONSdigital/dis-wagtail/pull/619/commits/4bdf1aaa7096b6bf842c45a9e12818e9ae6327a5)

Note that I defaulted to `False` in the functional tests settings so that this can be explicitly set to true when using the tag to enable the dataset API mock.

cms/bundles/forms.py
MaciekBaron: Addressed here: [11ba457](https://github.com/ONSdigital/dis-wagtail/pull/619/commits/11ba4577c1a0d97675ea6fdf8c1236a702499869)

cms/bundles/forms.py
MaciekBaron: Addressed here: [11ba457](https://github.com/ONSdigital/dis-wagtail/pull/619/commits/11ba4577c1a0d97675ea6fdf8c1236a702499869)

cms/bundles/forms.py
MaciekBaron: Changed here: [dcd2a23](https://github.com/ONSdigital/dis-wagtail/pull/619/commits/dcd2a23517d3d83663d093e4fb0ca2b549cad007)

cms/bundles/forms.py
MaciekBaron: Went with a transaction: [efde3a4](https://github.com/ONSdigital/dis-wagtail/pull/619/commits/efde3a4f8dd58319b7de2715246d370bc760355f)

MebinAbraham (approved):
Functionally all working as expected.

cms/bundles/forms.py
```diff
@@ -254,12 +255,13 @@ def _validate_bundled_datasets_metadata(self) -> None:
...
                 dataset.save(update_fields=updated_fields)
-            drift_messages.append(f"'{old_title}': {'; '.join(changes)}")
+            for change in changes:
+                drift_messages.append(f"'{old_title}': {change}")
 
         if drift_messages:
             errors = [
-                "Dataset metadata has changed in the source API since the bundle was last edited or submitted. "
-                "The local records have been refreshed - please re-review the changes and reconfirm to approve:",
+                "Approval could not be completed because dataset metadata has changed since they were added. "
```
BJacksonONS, line 268: Tests for this in `test_forms_with_bundle_api.py` need to be updated also.

BJacksonONS (changes requested):

cms/bundles/forms.py
MaciekBaron: Good spot, fixed here: [26cf982](https://github.com/ONSdigital/dis-wagtail/pull/619/commits/26cf982805ce7eba82347125dd2f5976c2fdad2c)

kevinhowbrook (approved):

BJacksonONS (approved):

sanjeevz3009 (commented):
LGTM!

PR #620, opened 2026-05-06
Clean up django tasks todos as we've updated to newer version of wagtail

BJacksonONS:

There were several places where we were overriding `django-tasks` default behaviour to enqueue tasks without a commit. As of our upgrade to Wagtail 7.3 we're using `django-tasks` version 0.11 which has this as the default behaviour.

### How to review

Ensure tests still run.
Double check we don't have any other places overriding the redundant setting.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

NA

### DIFF
--- a/cms/bundles/tests/test_viewsets.py
+++ b/cms/bundles/tests/test_viewsets.py
@@ -44,14 +44,6 @@ from cms.workflows.tests.utils import (
     progress_page_workflow,
 )
 
-# TODO: remove when Wagtail updates to django-tasks >= 0.11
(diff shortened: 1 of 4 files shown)


MebinAbraham (approved):

sanjeevz3009 (approved):
Thanks @BJacksonONS for this. LGTM!

MaciekBaron (approved):

RealOrangeOne (approved):

PR #621, opened 2026-05-06
Fix CSV data sanitisation for negative numbers

MaciekBaron:

This PR addresses DPD-3743 and fixes the sanitisation process used for values included in generated CSVs.

Previously, negative numbers such as `-123` would be prepended with a quote, resulting in `'-123`.

The minus/dash symbol is escaped to avoid formula injection. However, the symbol does not need to be escaped if we deal with a real number.

This PR modifies the logic to check if the value is in fact a valid number, and skips sanitisation when this is the case.

### How to review

1. Create or edit a page in which you can embed a table, such as a statistical article page
2. Populate the table with data, including some negative numbers
3. Save and publish the page
4. Go to the live page
5. Download the CSV for that table
6. Inspect the CSV - confirm that negative numbers are represented correctly

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

N/A

### DIFF
--- a/cms/data_downloads/utils.py
+++ b/cms/data_downloads/utils.py
@@ -95,15 +95,6 @@ def flatten_table_data(data: Mapping) -> list[list[str | int | float]]:
     return result
 
 
-def _is_number(val: str) -> bool:
-    """Check if a string represents a valid number."""
-    try:
-        float(val)
-        return True
-    except ValueError:
-        return False
-
-
 def sanitize_data_for_csv(data: list[list[str | int | float]]) -> list[list[str | int | float]]:
     """Sanitize data for CSV export by escaping formula triggers.
 
@@ -112,21 +103,22 @@ def sanitize_data_for_csv(data: list[list[str | int | float]]) -> list[list[str 
     """
     triggers = ("=", "+", "-", "@", "\t")
 
-    sanitized_row: list[list[str | int | float]] = []
+    sanitized_rows: list[list[str | int | float]] = []
     for row in data:
         new_row: list[str | int | float] = []
         for value in row:
             if isinstance(value, str) and value.startswith(triggers):
                 # If it's a number string (e.g. "-5"), do not prepend the quote
-                if _is_number(value):
+                try:
+                    float(value)  # Check if it can be converted to a number
                     new_row.append(value)
-                else:
+                except ValueError:
                     new_row.append(f"'{value}")
             else:
                 new_row.append(value)
-        sanitized_row.append(new_row)
+        sanitized_rows.append(new_row)
 
-    return sanitized_row
+    return sanitized_rows
 
 
 def create_data_csv_download_response_from_data(data: list[list[str | int | float]], *, title: str) -> HttpResponse:
--- a/cms/data_downloads/utils.py
+++ b/cms/data_downloads/utils.py
@@ -95,6 +95,15 @@ def flatten_table_data(data: Mapping) -> list[list[str | int | float]]:
     return result
 
 
+def _is_number(val: str) -> bool:
+    """Check if a string represents a valid number."""
+    try:
+        float(val)
+        return True
+    except ValueError:
+        return False
+
+
 def sanitize_data_for_csv(data: list[list[str | int | float]]) -> list[list[str | int | float]]:
     """Sanitize data for CSV export by escaping formula triggers.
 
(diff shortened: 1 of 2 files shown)


cms/data_downloads/utils.py
```diff
@@ -95,6 +95,15 @@ def fill_rowspan_columns(processed_row: list[str | int | float], col_index: int)
     return result
 
 
+def _is_number(val: str) -> bool:
+    """Check if a string represents a valid number."""
+    try:
+        float(val)
```
MebinAbraham: `float()` is probably too permissive as the allow-list. For example, `float("\t1")` returns 1.0, so a value starting with the explicit tab trigger would now bypass escaping:
```py
sanitize_data_for_csv([["\t1"]])
# currently returns [["\t1"]]
```

I’d suggest using a stricter plain number regex instead, so we only skip escaping for simple numeric strings and continue escaping anything else that starts with a trigger.

`float` also allows things like `+inf`, `-inf`, `-1e3` etc. Scientific notation a case can be made for, but inf, Infinity, and nan are definitely not plain CSV numbers and should not be treated as safe numeric strings.

RealOrangeOne: If the aim for sanitization is just for security, then leaving `Infinity` as-is should be fine. It will still be displayed correctly by any reader and won't lead to a security issue. Although then arguably `_is_number` is not the best name - inlining this might be better then.

MaciekBaron: @MebinAbraham I'd argue that while it's true these values would be allowed, none of them introduce security concerns as any real injection would fail this check (i.e. an actual injection cannot be expressed using a value which `float()` can convert). Happy to look into inlining as @RealOrangeOne suggested.

cms/data_downloads/utils.py
```diff
@@ -103,10 +112,21 @@ def sanitize_data_for_csv(data: list[list[str | int | float]]) -> list[list[str
     """
     triggers = ("=", "+", "-", "@", "\t")
 
-    return [
-        [f"'{value}" if isinstance(value, str) and value.startswith(triggers) else value for value in row]
-        for row in data
-    ]
+    sanitized_row: list[list[str | int | float]] = []
```
joepercival: Nit: `sanitized_data` or `sanitized_rows` may be better variable names for this

MaciekBaron: Good spot, updated here: [217f169](https://github.com/ONSdigital/dis-wagtail/pull/621/commits/217f169acbb16447d419581957ba4eea2d5f2e0c)

cms/data_downloads/utils.py
MaciekBaron: Inlined here: [217f169](https://github.com/ONSdigital/dis-wagtail/pull/621/commits/217f169acbb16447d419581957ba4eea2d5f2e0c)

joepercival (approved):
LGTM

BJacksonONS (approved):
LGTM

cms/data_downloads/utils.py
sanjeevz3009: Thanks @MaciekBaron, resolving on behalf.

cms/data_downloads/utils.py
sanjeevz3009: Resolving on behalf.

cms/data_downloads/utils.py
sanjeevz3009: On second thought, digging deeper.

I agree this is probably not directly exploitable from a CSV injection perspective, since an actual formula payload would fail the `float()` conversion anyway.

My concern (maybe @MebinAbraham ) is more around the sanitisation boundary itself. We currently treat “anything Python’s float parser accepts” as safe, which is broader than just plain numeric strings. For example:

```python
float("\t1")  # succeeds
```

So a value starting with one of our explicit trigger characters (`\t`) still bypasses escaping.

I don’t think `"\t1"` is itself dangerous, but it does make the rule harder to reason about:

> escape trigger-prefixed values unless Python happens to parse them as floats

rather than:

> only allow plain numeric literals through unescaped

Using `float()` here also permits values like `"Infinity"`, `"nan"`, leading whitespace/control chars, etc., which are probably outside the intended “safe numeric string” scope.

I think siding with Mebin, it would make sense to use a stricter regex instead, so the exception stays narrow and explicit? Something like:

```python
re.fullmatch(r"-?\d+(\.\d+)?", value)
```

That would still allow values like `-5` and `1.25`, while avoiding the more permissive parser behaviour.

MebinAbraham: @MaciekBaron given the requirement is specifically to prevent CSV injection, I'm happy with the move to inline so the function isn't misleading and keeping as is.

@sanjeevz3009 for the sake of simplicity, its worth keeping as is as the sanitation behaves different when trigger has number value Vs not but it would still ensure the csv itself is escaped correct. It was about boundary but not a blocker from me as there are no security issues.

MebinAbraham (approved):

cms/data_downloads/utils.py
MaciekBaron: @sanjeevz3009 thanks for your explanation of your reasoning.

The original bug was due to the escaping being too aggressive. The role of the function was to escape _potentially dangerous strings_. Our rule actually isn't to "only allow plain numeric literals through unescaped" because we also allow (safe) text. The more safe content we can allow through without escaping, the better, as the escaping is a "necessary evil" since we need to modify the cell value. My change essentially makes the escaping less aggressive and more permissive, while not affecting safety. I'd argue if we can find ways to be even more permissive in a safe way, the better, so that less data gets modified unnecessarily.

sanjeevz3009: Thanks, both @MaciekBaron @MebinAbraham

sanjeevz3009 (approved):
LGTM! Works as expected. Merge away!

PR #622, opened 2026-05-06
Order bundle listings by most recently updated first

joepercival:

Addresses [CMS-1215](https://officefornationalstatistics.atlassian.net/browse/CMS-1215?atlOrigin=eyJpIjoiNzc4MzEyMTM5YTllNDc5MGE1YjU3MTgwN2I2ZmFmYTAiLCJwIjoiaiJ9). 

We want the bundle listing page in Wagtail admin to open with the most recently changed bundles at the top rather than alphabetically, in the same way editors are used to seeing recently changed pages surfaced in Wagtail.

Additionally, this PR updates the ordering of the LatestBundlesPanel and BundlesInReviewPanel on the admin index page for consistency, as these were also previously ordered by name. 

### How to review

1. Create a bundle, e.g. "Bundle1". 
2. Create a second bundle, e.g. "Bundle2". 
3. Submit both bundles for preview.
4. Verify that the second bundle appears first in the Bundles list on the 'Latest Active Bundles' Panel.
5. ... and on the 'Bundles Ready for Preview' Panel.
6. ... and on the Bundles manager page (via the sidebar). 

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

N/A

### DIFF
--- a/cms/bundles/tests/test_wagtail_hooks.py
+++ b/cms/bundles/tests/test_wagtail_hooks.py
@@ -81,7 +81,7 @@ class WagtailHooksTestCase(WagtailTestUtils, TestCase):
     def test_latest_bundles_panel_ordering(self):
         """Checks that the latest bundles dashboard panel orders bundles by updated_at descending."""
         old_bundle = BundleFactory(
-            name="Old Bundle", created_by=self.superuser, updated_at=timezone.now() - timedelta(days=3)
+            name="3 Day Old Bundle", created_by=self.superuser, updated_at=timezone.now() - timedelta(days=3)
         )
         new_bundle = BundleFactory(name="New Bundle", created_by=self.superuser, updated_at=timezone.now())
 
@@ -142,7 +142,10 @@ class WagtailHooksTestCase(WagtailTestUtils, TestCase):
     def test_bundles_to_preview_panel_ordering(self):
         """Checks that the bundles preview dashboard panel orders bundles by updated_at descending."""
         old_bundle = BundleFactory(
-            in_review=True, name="Old Bundle", created_by=self.superuser, updated_at=timezone.now() - timedelta(days=3)
+            in_review=True,
+            name="3 Day Old Bundle",
+            created_by=self.superuser,
+            updated_at=timezone.now() - timedelta(days=3),
         )
         new_bundle = BundleFactory(
             in_review=True, name="New Bundle", created_by=self.superuser, updated_at=timezone.now()
--- a/cms/bundles/tests/test_wagtail_hooks.py
+++ b/cms/bundles/tests/test_wagtail_hooks.py
@@ -1,5 +1,8 @@
+from datetime import timedelta
+
 from django.test import TestCase
 from django.urls import reverse
+from django.utils import timezone
 from wagtail.test.utils.wagtail_tests import WagtailTestUtils
 
 from cms.articles.tests.factories import ArticlesIndexPageFactory, StatisticalArticlePageFactory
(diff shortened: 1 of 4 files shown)


cms/bundles/tests/test_wagtail_hooks.py
```diff
@@ -75,6 +78,22 @@ def test_latest_bundles_panel_content(self):
             self.assertContains(response, bundle.status.label)
         self.assertNotContains(response, self.published_bundle.status.label)
 
+    def test_latest_bundles_panel_ordering(self):
+        """Checks that the latest bundles dashboard panel orders bundles by updated_at descending."""
+        old_bundle = BundleFactory(
+            name="Old Bundle", created_by=self.superuser, updated_at=timezone.now() - timedelta(days=3)
+        )
+        new_bundle = BundleFactory(name="New Bundle", created_by=self.superuser, updated_at=timezone.now())
```
MaciekBaron, line 86: This is a huge nit-pick, but if you were to sort this alphabetically, _"New bundle"_ would come before _"Old bundle"_.

To confirm that this is not a false positive, how about we rename the old bundle to _"3 Day Old Bundle"_? This way if your code accidentally sorted it alphabetically (due to a bug or oversight), the test would fail (as the _"3 Day Old Bundle"_ would appear first)

MaciekBaron (changes requested):

cms/bundles/tests/test_wagtail_hooks.py
joepercival: Good catch, will update now.

joepercival: c7ef2d8

MaciekBaron (approved):

cms/bundles/wagtail_hooks.py
```diff
@@ -131,15 +131,15 @@ def get_latest_bundles(self) -> QuerySet[Bundle]:
...
+            queryset = Bundle.objects.active().order_by("-updated_at")[: self.num_bundles]
 
         return queryset
 
     def get_context_data(self, parent_context: RenderContext | None = None) -> RenderContext | None:
         """Adds the request, the latest bundles and whether the panel is shown to the panel context."""
         context = super().get_context_data(parent_context)
         context["request"] = self.request
-        context["bundles"] = sorted(self.get_latest_bundles(), key=lambda b: b.name)
+        context["bundles"] = sorted(self.get_latest_bundles(), key=lambda b: b.updated_at, reverse=True)
```
BJacksonONS: Looks like `sorted` isn't necessary any more given the ordering on the queryset itself.

cms/bundles/wagtail_hooks.py
```diff
@@ -186,7 +186,7 @@ def get_context_data(self, parent_context: RenderContext | None = None) -> Rende
         """Adds the request, the latest bundles and whether the panel is shown to the panel context."""
         context = super().get_context_data(parent_context)
         context["request"] = self.request
-        context["bundles"] = sorted(self.bundles[: self.num_bundles], key=lambda b: b.name)
+        context["bundles"] = sorted(self.bundles[: self.num_bundles], key=lambda b: b.updated_at, reverse=True)
```
BJacksonONS: Think `sorted` is no longer necessary here either as the ordering matches the `order_by` on the queryset now.

BJacksonONS (changes requested):

cms/bundles/wagtail_hooks.py
joepercival: Yes it's now redundant, thanks - 46d186f

cms/bundles/wagtail_hooks.py
joepercival: 46d186f

BJacksonONS (approved):

sanjeevz3009 (approved):
LGTM! Works well!

PR #628, opened 2026-05-11
Pin the Release Calendar index page to the top of the explorer page

charlotteromaniuk:

Addresses [CMS-1216
](https://officefornationalstatistics.atlassian.net/browse/CMS-1216)

This PR pins the Release Calendar index page to the top of the list of pages both on the Wagtail explorer page list. All other pages are ordered by recency apart from the home pages which are ordered by path to ensure the Home (English) and Home (Welsh) pages remain in fixed positions.

### How to review

- Create new child pages under the home page
- In both the explorer page check:
   - The Release Calendar index page remains at the top of the page list
   - The other child pages are ordered by recency
   - The home pages remain in the correct order (English home page first and Welsh home page second)

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

N/A


cms/release_calendar/tests/test_wagtail_hooks.py
```diff
@@ -0,0 +1,51 @@
...
+    @classmethod
+    def setUpTestData(cls):
+        cls.superuser = cls.create_superuser(username="admin")
+        cls.root = Page.get_first_root_node()
+        cls.homepages = HomePage.objects.specific()
+        cls.homepage = cls.homepages.first()
+        cls.release_calendar_index = ReleaseCalendarIndex.objects.first()
+
+    def setUp(self):
+        self.login()
```
MaciekBaron, line 30: `self.login()` is only necessary if you make `self.client` calls, and you only use it in one of your tests, so I would move the call there (`test_release_calendar_index_is_first_in_explorer_page`)

cms/release_calendar/tests/test_wagtail_hooks.py
```diff
@@ -0,0 +1,51 @@
...
+from cms.home.models import HomePage
+from cms.release_calendar.models import ReleaseCalendarIndex
+from cms.release_calendar.wagtail_hooks import pin_release_calendar_page
+from cms.standard_pages.tests.factories import IndexPageFactory
+
+
+class ReleaseCalendarHooksTestCase(WagtailTestUtils, TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.superuser = cls.create_superuser(username="admin")
```
MaciekBaron: You create a superuser but you don't use them. `self.login()` will automatically create a test user for you with enough privileges. This is only useful when we explicitly want to use a superuser.

cms/release_calendar/tests/test_wagtail_hooks.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def test_release_calendar_index_is_sorted_first(self):
+        """Checks that the Release Calendar index page is placed before all other pages in the returned ordering."""
+        IndexPageFactory(parent=self.homepage, title="New Index Page")
+        pages = self.homepage.get_children().specific()
+        result = pin_release_calendar_page(
+            self.homepage,
+            pages,
+            None,
+        )
```
MaciekBaron: Non-blocking nit: This should fit on one line, I don't think there's a benefit of splitting this into multiple lines.

cms/release_calendar/tests/test_wagtail_hooks.py
```diff
@@ -0,0 +1,51 @@
...
+        self.assertEqual(result[0], self.release_calendar_index)
```
MaciekBaron: Non-block nit: the third argument in `assertEqual` can be used to provide a message. Example:

https://github.com/ONSdigital/dis-wagtail/blob/9d9ac04cf762069e495989cab3ba27931a4a2fc6/cms/search/tests/test_publishers.py#L177

This is useful to make the test results more readable. In your particular case, it might be useful to add a message saying "Release calendar index page is not first" so that this error is displayed if the test fails - otherwise it will just say that one page does not equal another.

Bonus nit: `pin_release_calendar_page` returns a queryset, so you can do `result.first()` instead of `result[0]`, and even better, rename `result` to `query` (and have `query.first()`) for full correctness and clarity.

cms/release_calendar/tests/test_wagtail_hooks.py
```diff
@@ -0,0 +1,51 @@
...
+        cls.homepages = HomePage.objects.specific()
+        cls.homepage = cls.homepages.first()
+        cls.release_calendar_index = ReleaseCalendarIndex.objects.first()
+
+    def setUp(self):
+        self.login()
+
+    def test_release_calendar_index_is_sorted_first(self):
+        """Checks that the Release Calendar index page is placed before all other pages in the returned ordering."""
+        IndexPageFactory(parent=self.homepage, title="New Index Page")
```
MaciekBaron: I don't think `IndexPageFactory` is needed here? Tests work without it.

MaciekBaron (dismissed):

cms/release_calendar/wagtail_hooks.py
```diff
@@ -51,3 +54,18 @@ def register_chooser_viewset() -> FutureReleaseCalendarPageChooserViewSet:
 def hide_release_date_text_field_for_non_provisional_release_pages() -> str:
     """Hide the release date text field for non-provisional release pages."""
     return format_html('<script src="{}"></script>', static("js/hide-date-text-on-non-provisional-releases.js"))
+
+
+@hooks.register("construct_explorer_page_queryset")
+def pin_release_calendar_page(_parent_page: Page, pages: PageQuerySet, _request: HttpRequest) -> PageQuerySet:
+    """Pin the Release Calendar index to the top of the explorer page and explorer menu."""
+    if all(isinstance(page.specific, HomePage) for page in pages):
```
zerolab: suggestion: it would be great to do the ordering without evaluating the queryset.

Considering the release calendar index can only be created under HomePage, I would argue that the hook should something like (note: untested):

```python
@hooks.register("construct_explorer_page_queryset")
def pin_release_calendar_page(parent_page: Page, pages: PageQuerySet, _request: HttpRequest) -> PageQuerySet:
    if isinstance(parent_page.specific_deferred, HomePage):
        return pages.order_by(
            Case(
                When(content_type__model="releasecalendarindex", then=Value(0)),
                default=Value(1),
                output_field=IntegerField(),
            ),
            "-latest_revision_created_at",  # default ordering
        )

    return pages

```

Aside: it is important to test the default `-latest_revision_created_at` ordering set in the hook around the cases outlined in 
https://github.com/wagtail/wagtail/blob/f0b2c5bed43843efde1673c498290ed3d3a65fd4/wagtail/admin/views/pages/listing.py#L216-L242:
1. manual orderding view ("ord")
2. oldest or newest revision first (when they click on the "Updated" column header)

cms/release_calendar/tests/test_wagtail_hooks.py
charlotteromaniuk: Good point, I've moved `self.login()` [4c7519c](https://github.com/ONSdigital/dis-wagtail/pull/628/commits/4c7519cd2303c6e9b922d2ea543b8b29dba13c24)

cms/release_calendar/tests/test_wagtail_hooks.py
charlotteromaniuk: Makes sense thanks, I've removed it [4c7519c](https://github.com/ONSdigital/dis-wagtail/pull/628/commits/4c7519cd2303c6e9b922d2ea543b8b29dba13c24)

cms/release_calendar/tests/test_wagtail_hooks.py
charlotteromaniuk: I tried putting this on one line but the ruff formatter re-wrote it back to the multi-line version

cms/release_calendar/tests/test_wagtail_hooks.py
charlotteromaniuk: Thanks, updated both points here: [f25602a](https://github.com/ONSdigital/dis-wagtail/pull/628/commits/f25602ae524138f54626a8dd60dfa3a8fbb2ef4b)[f25602a](https://github.com/ONSdigital/dis-wagtail/pull/628/commits/f25602ae524138f54626a8dd60dfa3a8fbb2ef4b)

cms/release_calendar/wagtail_hooks.py
kevinhowbrook: **Issue**

Related to dans aside, when testing all the ordering options the pinning works but the ordering isn't respected:

Image

I think we need to add in a `secondary_sort` to respect all the ordering, e.g

```python
secondary_sort = "path" if ordering == "ord" else ordering
return pages.order_by(
    Case(
        When(content_type__model="releasecalendarindex", then=Value(0)),
        default=Value(1),
        output_field=IntegerField(),
    ),
    secondary_sort,
)
```

cms/release_calendar/wagtail_hooks.py
```diff
@@ -51,3 +54,18 @@ def register_chooser_viewset() -> FutureReleaseCalendarPageChooserViewSet:
...
+        return pages.order_by("path")
+    return pages.order_by(
+        Case(
+            When(content_type__model="releasecalendarindex", then=Value(0)),
+            default=Value(1),
+            output_field=IntegerField(),
+        ),
+        "-latest_revision_created_at",
+    )
```
kevinhowbrook, line 82: Praise

Using Case/When for ordering is great. Doing the sort at the database level keeps this as a queryset, which means Wagtail's pagination should still works correctly. A Python-level sort (converting to a list) would have broken that.

kevinhowbrook (changes requested):

cms/release_calendar/wagtail_hooks.py
MebinAbraham: 💯 Kudos

zerolab: I mirror this! Folks usually forget about hidden gems like this in the Django ORM

cms/release_calendar/tests/test_wagtail_hooks.py
MaciekBaron: Looks like you made it work, resolving

cms/release_calendar/tests/test_wagtail_hooks.py
charlotteromaniuk: I added this to make sure the test wouldn't pass falsely in the case where the release calendar index was the only page under home. It looks like the cookies page is always generated as well though so I'll remove the extra page.

charlotteromaniuk: [1a75382](https://github.com/ONSdigital/dis-wagtail/pull/628/commits/1a75382e91d9c3b9446df64161ce91a0dce7b789)

cms/release_calendar/wagtail_hooks.py
BJacksonONS: `secondary_sort = "path" if ordering == "ord" else ordering` would fail on `-ord` if that's a possible state?.

Not sure which of the following would be the best way to get the ordering in the hook.

`secondary_sort = request.GET.get("ordering", "-latest_revision_created_at")` (my preferred)
or
`secondary_sort = pages.query.order_by` (possibly not a stable django api)

charlotteromaniuk: Thanks all. I’ve added a fix [c3abec8](https://github.com/ONSdigital/dis-wagtail/pull/628/commits/c3abec8673f65153096bae1cebcff3d3eeca5013) which should now make it so that:
* In the default view, the release calendar index is pinned, and all other pages appear in their usual recency order.
* When a sort option is applied, the release calendar index is no longer pinned, and the pages are ordered according to the selected sort.


~~@MebinAbraham Does this logic sound okay?~~ - confirmed with Mebin

@zerolab This also removes the line that evaluates the queryset. For now, I’ve kept it so that all pages, except the home pages, are sorted by recency. When I applied the hook to only the children of the homepage, it caused some inconsistency in the sidebar as the rest of the pages were still ordered by path.

cms/release_calendar/wagtail_hooks.py
```diff
@@ -51,3 +53,26 @@ def register_chooser_viewset() -> FutureReleaseCalendarPageChooserViewSet:
...
+    """Pin the Release Calendar index to the top of the explorer page and explorer menu."""
+    ordering = request.GET.get("ordering")
+
+    # Respect any existing ordering
+    if ordering:
+        return pages.order_by(ordering)
+
+    # Order home pages by path
+    if parent_page.is_root():
+        return pages.order_by("path")
```
zerolab:
```suggestion
    # Respect any existing ordering
    if request.GET.get("ordering") or parent_page.is_root():
        return pages
```

Best to return the queryset as is and let Wagtail take care of the ordering on the root page, or when the user user changes the order

cms/release_calendar/wagtail_hooks.py
```diff
@@ -51,3 +53,20 @@ def register_chooser_viewset() -> FutureReleaseCalendarPageChooserViewSet:
...
+        return pages
+
+    return pages.order_by(
+        Case(
+            When(content_type__model="releasecalendarindex", then=Value(0)),
+            default=Value(1),
+            output_field=IntegerField(),
+        ),
+        "-latest_revision_created_at",
+    )
```
MebinAbraham: As discussed in the tech session, it's unfortunate that `construct_explorer_page_queryset` controls both the sidebar nav and the explorer page. However, I suggest we do not alter the default behaviour of the sidebar. The sidebar nav changing on recency may be confusing to users, and this is likely why the default Wagtail behaviour is that it is ordered by path for the sidebar and recency for the explorer.

Suggestion: You could do something like this to allow us to customise it for our needs while leaving sidebar behaviour unchanged. This also addresses [Dan's comment](https://github.com/ONSdigital/dis-wagtail/pull/628#discussion_r3220576774) about doing it only for `HomePage`, as RC will always be under home, therefore, we don't need custom logic if the parent isn't `HomePage`.

It could be condensed, but it might sacrifice readability.
Note: the `view_name` for the sidebar is `listing`.

```suggestion
    # Respect any existing user-selected ordering.
    if request.GET.get("ordering"):
        return pages

    # Only apply to the homepage Explorer view.
    resolver_match = getattr(request, "resolver_match", None)
    is_homepage_explorer = getattr(resolver_match, "view_name", "") == "wagtailadmin_explore" and isinstance(
        parent_page.specific_deferred, HomePage
    )

    if not is_homepage_explorer:
        return pages

    return pages.order_by(
        Case(
            When(content_type__model="releasecalendarindex", then=Value(0)),
            default=Value(1),
            output_field=IntegerField(),
        ),
        "-latest_revision_created_at",
    )
```

See if the Wagtail pros have a better alternative; otherwise, I suggest we go with this. 
@zerolab @MaciekBaron @kevinhowbrook

cms/release_calendar/tests/test_wagtail_hooks.py
```diff
@@ -0,0 +1,39 @@
...
+from cms.release_calendar.models import ReleaseCalendarIndex
+from cms.release_calendar.wagtail_hooks import pin_release_calendar_page
+from cms.standard_pages.tests.factories import IndexPageFactory
+
+
+class ReleaseCalendarHooksTestCase(WagtailTestUtils, TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.root = Page.get_first_root_node()
+        cls.homepages = HomePage.objects.specific()
```
sanjeevz3009: Nit
```suggestion
        cls.home_pages = HomePage.objects.specific()
```

cms/release_calendar/tests/test_wagtail_hooks.py
```diff
@@ -0,0 +1,39 @@
...
+        cls.homepage = cls.homepages.first()
```
sanjeevz3009: Nit
```suggestion
        cls.home_page = cls.homepages.first()
```

cms/release_calendar/tests/test_wagtail_hooks.py
zerolab: fwiw, `HomePage` is already specific. So `cls.home_pages = HomePage.objects.all()`

zerolab: at the same time, `homepages` is only used on the line below, so you probably only want `cls.home_page = HomePage.objects.first()`

cms/release_calendar/tests/test_wagtail_hooks.py
MaciekBaron: In Charlotte's defence, we have prior art:
https://github.com/ONSdigital/dis-wagtail/blob/e420a65fab507587d32282571ec0df588ca4fc33/cms/core/tests/test_views.py#L284-L287

sanjeevz3009 (commented):
Looking good. A couple of nits + making some of the test cases a bit more secure/ less brittle will be ideal as we spoke offline. Will re-review after it. It's nearly ready to go in.

cms/release_calendar/tests/test_wagtail_hooks.py
charlotteromaniuk: [5a06c07](https://github.com/ONSdigital/dis-wagtail/pull/628/commits/5a06c071ea944d71449fa440cfa55f89ee1c8717)

cms/release_calendar/tests/test_wagtail_hooks.py
charlotteromaniuk: [8314fb0](https://github.com/ONSdigital/dis-wagtail/pull/628/commits/8314fb07490bfbd0bab03c38d4e1380b6b93735a)

cms/release_calendar/tests/test_wagtail_hooks.py
charlotteromaniuk: @MaciekBaron I’ve added the creation of new pages back into the setup for this test here [9aef057](https://github.com/ONSdigital/dis-wagtail/pull/628/commits/9aef0573c263ea9471c583ae4a92e7c2ada392eb). The two topic pages are required for the other tests, but they also make this test less brittle. Since the Release Calendar page and cookies page already exist in the correct order, adding additional pages ensures that `pin_release_calendar_page()` is reordering the pages correctly and keeping the Release Calendar first as expected, instead of ordering all pages by recency.

BJacksonONS:
Given we've made a decision to preserve the original sort order in the sidebar, should we add a test to check that behaves correctly also?

charlotteromaniuk:
> Given we've made a decision to preserve the original sort order in the sidebar, should we add a test to check that behaves correctly also?

Good point, I've added a unit test to check this here: [9aef057](https://github.com/ONSdigital/dis-wagtail/pull/628/commits/9aef0573c263ea9471c583ae4a92e7c2ada392eb)

cms/release_calendar/tests/test_wagtail_hooks.py
sanjeevz3009: You forgot the snake casing @charlotteromaniuk

cms/release_calendar/tests/test_wagtail_hooks.py
sanjeevz3009: I would rather keep it consistent, as you can see above, there is:


```
cls.homepage = HomePage.objects.first() 
cls.index_page = IndexPageFactory(parent=cls.homepage) 
```

Yes, we don't want to go refactor everywhere, but she can follow snake casing for the code she writes.

cms/release_calendar/tests/test_wagtail_hooks.py
sanjeevz3009: Just saw the below commit, where you have made the change. Thanks =)

cms/release_calendar/wagtail_hooks.py
sanjeevz3009: Resolving!

cms/release_calendar/tests/test_wagtail_hooks.py
```diff
@@ -0,0 +1,67 @@
...
+    @classmethod
+    def setUpTestData(cls):
+        cls.root = Page.get_first_root_node()
+        cls.home_page = HomePage.objects.first()
+        cls.release_calendar_index = ReleaseCalendarIndex.objects.first()
+        cls.release_calendar_index.save_revision().publish()
+        cls.request = RequestFactory().get("/")
+
+        # Create two topic pages with initial revision timestamps
+        cls.older_topic = TopicPageFactory(parent=cls.home_page, title="Older Topic")
```
sanjeevz3009:
```suggestion
        cls.older_topic_page = TopicPageFactory(parent=cls.home_page, title="Older Topic")
```

cms/release_calendar/tests/test_wagtail_hooks.py
```diff
@@ -0,0 +1,67 @@
...
+        cls.older_topic.save_revision().publish()
+        cls.newer_topic = TopicPageFactory(parent=cls.home_page, title="Newer Topic")
```
sanjeevz3009:
```suggestion
        cls.newer_topic_page = TopicPageFactory(parent=cls.home_page, title="Newer Topic")
```

cms/release_calendar/tests/test_wagtail_hooks.py
```diff
@@ -0,0 +1,67 @@
...
+        cls.newer_topic.save_revision().publish()
+
+    def test_release_calendar_index_is_sorted_first(self):
+        """Checks that the Release Calendar index page is placed before all other pages in the returned ordering."""
```
sanjeevz3009: Make sure the docstring reflects that this test. Mention in short/ capture the following:
- Unit test directly calling the `pin_release_calendar_page` function. Direct function invocation.
- It's testing queryset ordering logic in isolation.
- Not an integration test. I.e., no HTTP requests or going through Wagtail hooks. Just database queries and Python logic here.


Write something short capturing the above in the doc string.

cms/release_calendar/tests/test_wagtail_hooks.py
```diff
@@ -0,0 +1,67 @@
...
+        pages = self.home_page.get_children().specific()
+        query = pin_release_calendar_page(self.home_page, pages, self.request)
+        self.assertEqual(query.first(), self.release_calendar_index, "Release calendar index page is not first")
+
+    def test_release_calendar_index_is_first_in_explorer_page(self):
+        """Checks that the Release Calendar index page is displayed at the top of the explorer page."""
```
sanjeevz3009: Related to https://github.com/ONSdigital/dis-wagtail/pull/628/changes#r3264645242.

Make sure the docstring mentions this is a integration test etc.

cms/release_calendar/tests/test_wagtail_hooks.py
```diff
@@ -0,0 +1,67 @@
...
+        response = self.client.get(reverse("wagtailadmin_explore", args=[self.home_page.id]))
+        pages = list(response.context["pages"])
+
+        self.assertEqual(pages[0], self.release_calendar_index, "Release calendar index page is not first in explorer")
+        self.assertEqual(pages[1], self.older_topic, "Updated topic page is not second in explorer")
+        self.assertEqual(pages[2], self.newer_topic, "Topic page is not ordered by recency as expected")
+
+    def test_sidebar_is_not_reordered(self):
+        """Checks that the sidebar pages keep their original path order."""
+        self.login()
```
sanjeevz3009: Nit: Now you have to log in 2 times; might as well put in a `setUp` class.

cms/release_calendar/wagtail_hooks.py
sanjeevz3009: Thanks @charlotteromaniuk Resolving!

cms/release_calendar/tests/test_wagtail_hooks.py
```diff
@@ -0,0 +1,67 @@
...
+
+        parent = self.home_page
+        children = parent.get_children()
+
+        # Simulate a sidebar request
+        sidebar_request = self.client.get(f"/admin/api/pages/?child_of={parent.id}").wsgi_request
+
+        sidebar_request.resolver_match = SimpleNamespace(
```
sanjeevz3009, line 67: Write a short comment on what's going on here please.

sanjeevz3009:
@charlotteromaniuk @MaciekBaron @BJacksonONS Something to consider, if you go to the Welsh home page -> the release calendar index doesn't appear at the top. This is expected, but not consistent with how, within the English home page, the release calendar index will always appear at the top of the sidebar.

I'll have a think about this.

sanjeevz3009 (commented):
@charlotteromaniuk left some comments. Expect one issue I found; everything else looks good.

cms/release_calendar/tests/test_wagtail_hooks.py
charlotteromaniuk: [da72bed](https://github.com/ONSdigital/dis-wagtail/pull/628/commits/da72bed430d01bb0b83d8924e37535a49c0b4da5)

cms/release_calendar/tests/test_wagtail_hooks.py
charlotteromaniuk: [da72bed](https://github.com/ONSdigital/dis-wagtail/pull/628/commits/da72bed430d01bb0b83d8924e37535a49c0b4da5)

BJacksonONS:
> @charlotteromaniuk @MaciekBaron @BJacksonONS Something to consider, if you go to the Welsh home page -> the release calendar index doesn't appear at the top. This is expected, but not consistent with how, within the English home page, the release calendar index will always appear at the top of the sidebar.
> 
> I'll have a think about this.

Looks like we don't make the alias for the Release Calendar page immediately. I think it doesn't exist until a page is created under the english one, then both aliases are made at the same time. This means we can get paths that are higher in the search order than the welsh Release Calendar if we've done other stuff first.

Not sure if that affects prod or not though, we'll already have the release calendar created by now so it'll have whatever path it has already.

charlotteromaniuk:
> Looks like we don't make the alias for the Release Calendar page immediately. I think it doesn't exist until a page is created under the english one, then both aliases are made at the same time. This means we can get paths that are higher in the search order than the welsh Release Calendar if we've done other stuff first.
> 
> Not sure if that affects prod or not though, we'll already have the release calendar created by now so it'll have whatever path it has already.

I came to the same conclusion when I was looking into this

cms/release_calendar/tests/test_wagtail_hooks.py
charlotteromaniuk: [60e7e43](https://github.com/ONSdigital/dis-wagtail/pull/628/commits/60e7e43b265b630c28cc1b505efc152e633d38ff)

cms/release_calendar/tests/test_wagtail_hooks.py
charlotteromaniuk: [60e7e43](https://github.com/ONSdigital/dis-wagtail/pull/628/commits/60e7e43b265b630c28cc1b505efc152e633d38ff)

cms/release_calendar/tests/test_wagtail_hooks.py
charlotteromaniuk: [60e7e43](https://github.com/ONSdigital/dis-wagtail/pull/628/commits/60e7e43b265b630c28cc1b505efc152e633d38ff)

cms/release_calendar/tests/test_wagtail_hooks.py
charlotteromaniuk: [7e1b10d](https://github.com/ONSdigital/dis-wagtail/pull/628/commits/7e1b10d5fcd4e81246dafd84fab8efe36a02daca)

cms/release_calendar/tests/test_wagtail_hooks.py
sanjeevz3009: @charlotteromaniuk I am happy with the tests! Thanks!

cms/release_calendar/wagtail_hooks.py
sanjeevz3009: Resolving on behalf!

cms/release_calendar/wagtail_hooks.py
sanjeevz3009: Resolving on behalf!

cms/release_calendar/tests/test_wagtail_hooks.py
sanjeevz3009: Excellent!

sanjeevz3009:
@charlotteromaniuk @BJacksonONS You two verified this already. I flushed my DB and ran the migrations again, and yes, when we programmatically/ via migrations, the RC index doesn't show for Welsh unless we create a page underneath.

On the other hand, when we create a topic page, it automatically creates the articles and methodology index, and that's reflected in the Welsh site.

This is due to how we handled the migrations 18 months ago. Looking at https://github.com/ONSdigital/dis-wagtail/blob/main/cms/release_calendar/migrations/0002_create_releasecalendarindex.py we are not taking into account having another locale, of course.

Hence, the path order is messed up.

I don't want to tamper with migrations or paths on this PR. We'll tackle this issue on a separate card/ PR, as it will need a great deal of thought on how to go about fixing this now that all this exists in `prod`.

Charlotte also found another issue out of the scope of this work, around the sort menu order not functioning properly. @charlotteromaniuk, if you can put that issue in the PR as a comment, that would be helpful for future us. Thanks!

Let's get this PR in! I am happy with it.

Action points for @charlotteromaniuk 
- Document in this PR the other issue you found with the sort menu order
  - Validate if it's our customisations, or if it's a Wagtail core issue, using the bakery demo, please
- Make sure to mention this issue and link this comment or any relevant comments on the JIRA card comments section/ write it up there

sanjeevz3009 (approved):
LGTM! Thanks @charlotteromaniuk

kevinhowbrook (approved):
nice one @charlotteromaniuk

BJacksonONS (approved):
Good job!

cms/release_calendar/tests/test_wagtail_hooks.py
sanjeevz3009: Resolving on behalf! @MaciekBaron Feel free to re-open if anything is horribly wrong.

charlotteromaniuk:
> Charlotte also found another issue out of the scope of this work, around the sort menu order not functioning properly. @charlotteromaniuk, if you can put that issue in the PR as a comment, that would be helpful for future us. Thanks!

The issue mentioned above by @sanjeevz3009 is related to the “Sort menu order” option under the “... Actions” dropdown on the explorer page. When using “Sort menu order” in the Explorer, the custom order is saved to the page tree, but the Explorer view doesn’t display that order. Instead, it reverts to its own default sorting, so the order set by the user only appears in the sidebar, not in the Explorer.

This issue has been resolved in Wagtail 7.4 via [PR #14052](https://github.com/wagtail/wagtail/pull/14052). I’ve tested it locally, and the Explorer now correctly displays the custom ordering set through “Sort menu order”.

PR #633, opened 2026-05-14
Handle Pillow detecting a decompression bomb

RealOrangeOne:

When fetching the size of an image, if Pillow thinks this is a decompression bomb it will raise an exception. This happens before the pixel sizes are returned, so Wagtail can't run its own validation.

This PR catches the exception and returns a modified error message, rather than receiving an unhandled exception.

### How to review

Try uploading an image (like the one linked to the ticket) and remove the 

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

Also raised upstream as https://github.com/wagtail/wagtail/pull/14225. This can be reverted one upstream ships.

### DIFF
--- a/cms/images/forms.py
+++ b/cms/images/forms.py
@@ -1,7 +1,6 @@
 from typing import Any
 
 from django.forms import ValidationError
-from django.utils.translation import gettext_lazy as _
 from PIL.Image import DecompressionBombError
 from wagtail.images.forms import BaseImageForm
 
@@ -26,7 +25,7 @@ class CustomImageForm(BaseImageForm):
                 "file",
                 ValidationError(
                     self.fields["file"].error_messages["file_too_many_pixels"]
-                    % {"num_pixels": _("unknown number"), "max_pixels_count": self.fields["file"].max_image_pixels},
+                    % {"num_pixels": "unknown number", "max_pixels_count": self.fields["file"].max_image_pixels},
                     code="file_too_many_pixels",
                 ),
             )
--- a/cms/images/forms.py
+++ b/cms/images/forms.py
@@ -1,5 +1,8 @@
 from typing import Any
 
+from django.forms import ValidationError
+from django.utils.translation import gettext_lazy as _
+from PIL.Image import DecompressionBombError
 from wagtail.images.forms import BaseImageForm
 
 
@@ -12,3 +15,18 @@ class CustomImageForm(BaseImageForm):
             self.fields[
                 "title"
             ].help_text = "The title field will be used as the file name when the image is downloaded."
+
+    def full_clean(self) -> None:
+        # HACK: Catch the decompression bomb error during clean, without modifying the field.
+        # Remove when https://github.com/wagtail/wagtail/pull/14225 is merged.
+        try:
+            super().full_clean()
+        except DecompressionBombError:
+            self.add_error(
+                "file",
+                ValidationError(
+                    self.fields["file"].error_messages["file_too_many_pixels"]
+                    % {"num_pixels": _("unknown number"), "max_pixels_count": self.fields["file"].max_image_pixels},
+                    code="file_too_many_pixels",
+                ),
+            )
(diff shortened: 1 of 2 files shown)


cms/images/tests/test_forms.py
```diff
@@ -19,3 +24,18 @@ def test_description_field_is_relabelled_and_required(self):
...
+
+        form = form_class(
+            data={
+                "title": "Test image",
+                "description": "test",
+            },
+            files={"file": SimpleUploadedFile("test.png", get_test_image_file().file.getvalue())},
+        )
+
+        self.assertFormError(form, "file", "This file has too many pixels (unknown number). Maximum pixels 10000000.")
```
BJacksonONS, line 41: Shouldn't the value in the error message be tied to the `MAX_IMAGE_PIXELS` setting value?

cms/images/forms.py
```diff
@@ -12,3 +15,18 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        # HACK: Catch the decompression bomb error during clean, without modifying the field.
+        # Remove when https://github.com/wagtail/wagtail/pull/14225 is merged.
+        try:
+            super().full_clean()
+        except DecompressionBombError:
+            self.add_error(
+                "file",
+                ValidationError(
+                    self.fields["file"].error_messages["file_too_many_pixels"]
+                    % {"num_pixels": _("unknown number"), "max_pixels_count": self.fields["file"].max_image_pixels},
```
MaciekBaron: Nit: We don't need `getttext` here as the admin panel doesn't need translating

MaciekBaron (changes requested):

cms/images/tests/test_forms.py
```diff
@@ -19,3 +24,18 @@ def test_description_field_is_relabelled_and_required(self):
             form.fields["title"].help_text,
             "The title field will be used as the file name when the image is downloaded.",
         )
+
+    @patch.object(Image, "MAX_IMAGE_PIXELS", 1)
```
kevinhowbrook, line 28: **praise**

That's a neat way to force the bomb without needing a massive image

kevinhowbrook (approved):
Thanks Jake (Aside from what's raised)

cms/images/tests/test_forms.py
RealOrangeOne: I'd rather not - otherwise you get a different "maximum value" depending on which limit you hit.

cms/images/tests/test_forms.py
RealOrangeOne: I actually stole this trick from Pillow's test suite. It's much nicer than trying to construct a 250MB image!

cms/images/forms.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/633/commits/62c61c2c333566aac6f6ca55885b7e083297e801

cms/images/tests/test_forms.py
BJacksonONS: Hmm okay I'd have expected it to display the lower of `WAGTAILIMAGES_MAX_IMAGE_PIXELS` and the pillow image `MAX_IMAGE_PIXELS`. Does wagtail usually pass in it's own max pixels value when it hasn't been patched? Curious what would happen if the wagtail setting was changed at some point to be higher than the default pillow value?

RealOrangeOne: Wagtail has a default, yes. We set ours lower than that. If wagtail (or we) updated the value, it would just kick in first. You can disable the protection, in which case you'd get a nonsensitcal error message here. That's handled in the upstream fix, but not here for simplicity.

BJacksonONS (approved):

RealOrangeOne:
@BJacksonONS is there a reason this is still marked "do not merge"?

sanjeevz3009 (approved):
LGTM!

MaciekBaron (approved):

PR #634, opened 2026-05-14
fix: Include series and edition in related data page title and breadcrumb

kevinhowbrook:

Related to [CMS-4385](https://officefornationalstatistics.atlassian.net/browse/DPD-4385?atlOrigin=eyJpIjoiYWM5MjhkYmVjN2I3NDA3M2E0ZmZjNmZlY2ZlZjkyNWEiLCJwIjoiaiJ9).

The related data page heading and breadcrumb were showing only the edition title (e.g. "financial year ending 2018"), omitting the series title. Since the series page is excluded from breadcrumbs, this left the page without enough context to identify which article the data belongs to.

The fix adds a breadcrumb_title extension point to BasePage (defaulting to self.title) and overrides it on StatisticalArticlePage to return the full series: edition format, matching what the rest of the site uses for article titles.

#### Other option considered

Instead of adding breadcrumb_title to BasePage, StatisticalArticlePage could override get_breadcrumbs and call super():

```python
def get_breadcrumbs(self, request):
    breadcrumbs = super().get_breadcrumbs(request)
    if getattr(request, "is_for_subpage", False):
        breadcrumbs[-1]["text"] = self.get_full_display_title()
    return breadcrumbs
```

We wouldn't need to adjust the base model and the fix is entirely self-contained in StatisticalArticlePage.
However, that relies on the structure of what super() returns and if get_breadcrumbs ever changes, this override could break silently. My thoughts are the breadcrumb_title approach makes the extension point explicit and harder to misuse.

### How to review

1. Create a statistical article with at least one dataset in the "Related data" tab and publish it
2. Navigate to [series-url]/related-data/ or use the "Related Data Page" preview mode on the article
3. Confirm the heading reads "All data related to [Series]: [Edition]"
4. Confirm the breadcrumb shows the article link as "[Series]: [Edition]" rather than "[Edition]" alone

### Before

image

### After

image

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

None


BJacksonONS (approved):
LGTM

sanjeevz3009 (approved):
LGTM! Thanks @kevinhowbrook

PR #635, opened 2026-05-18
Update megalinter config and CI for zizmor

BJacksonONS:

Megalinter added zizmor in new version over the weekend. I've whitelisted an environment variable for it's use.

The primary change is to pin github actions to a release hash. Zizmor ensures that hashes match their version comment.

We also disable accessing the cache on release branches, given some of the recent cache poisoning attacks in the last week. This might be unecessary for this project as I don't believe we currently build any release artifacts in these actions. We can revert that change if we want.

Bump the version of `basic-ftp` for a security release.

Add a 7 day cooldown to dependabot config.

Specify only read permissions for workflows.

### How to review

Double check config against documentation, ensure actions finish successfully.
https://megalinter.io/latest/descriptors/action_zizmor/

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

Enable a tool like [pinact](https://github.com/suzuki-shunsuke/pinact) to automate action version updates.


joepercival (approved):
Confirmed Megalinter passes locally with new tools, looks good

sanjeevz3009 (approved):
LGTM! Thanks @BJacksonONS, this is a great improvement security-wise, especially due to the recent security issues left, right, and centre.

PR #636, opened 2026-05-18
Ensure that the downloads heading size matches the footnotes heading size in the table block

helenb:

See https://officefornationalstatistics.atlassian.net/browse/CMS-1204

This is a follow-on PR from https://github.com/ONSdigital/dis-wagtail/pull/605.

In the above PR we updated the footnotes heading to adjust according to whether a table heading was rendered, and avoid incorrectly nested headings. We missed that the downloads heading needed a similar update, which is covered by this PR.

### How to review

Create two table blocks, one with a heading, and one without. Make sure that neither have footnotes. Previously, for the table with no title, the 'downloads' heading would have been a heading 4, following straight on from a heading 2. With this PR, it is now correctly adjusted to a heading 3 if there is no title present.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


cms/core/blocks/markup.py
```diff
@@ -195,13 +195,15 @@ def get_context(self, value: dict, parent_context: dict | None = None) -> dict:
             options["download"] = self._get_download_config(parent_context=parent_context, block_id=block_id, data=data)
 
         footnotes_heading_level = 4 if value.get("title") else 3
+        downloads_heading_level = 4 if value.get("title") else 3
```
MaciekBaron: Both evaluate to the same - should we just combine them semantically, and use one variable called something like "`additional_sections_heading_level`" or something better?

BJacksonONS: Seconding this

sanjeevz3009: I agree

sanjeevz3009 (approved):
Apart from the comments we have already. LGTM. Thanks!

sanjeevz3009 (approved):

cms/core/blocks/markup.py
helenb: I've now addressed that if you would like to re-review.

helenb: @MaciekBaron @sanjeevz3009 @BJacksonONS

BJacksonONS (approved):
Looks good, thanks for the change

cms/core/blocks/markup.py
sanjeevz3009: Resolving on behalf! @MaciekBaron

PR #609, opened 2026-04-29
Add external links for methodologies on Topic pages

BJacksonONS:
### What is the context of this PR?

Addresses CMS-882

Methodology content can be published at different times to statistical article content. This means not all methodology pages will be available in Wagtail immediately, and they will need to link to previously published content on approved domains.

### How to review

Exact requirements are on ticket CMS-822.

- Ensure you can add external links, but only to approved domains.
- Make sure there is no regression when picking methodology pages. They should be limited to those under the same topic.
- Make sure updated text "Quality, methods and supporting information" is used now instead of "Methods and quality information".
- Ensure manually added links are displayed above any 'topped-up' links.

### Deployment Safety

Bleed and Sandbox deploy automatically on merge, so PRs should be safe to deploy immediately.

Please select one:

- [ ] Safe to auto-deploy
- [x] Not safe to auto-deploy

This PR adds a new migration, so this should be verified to be safe against the production database. Any rollback measures need to be clarified.

### Follow-up Actions

Update welsh translations for "Quality, methods and supporting information".

PR #637, opened 2026-05-18
Add missing fields to external links in "Related Articles" section on topic page

joepercival:

Addresses [DPD-2526](https://officefornationalstatistics.atlassian.net/browse/DPD-2526?atlOrigin=eyJpIjoiMmE0ZjFmMTNkY2FjNDJjMWFmYzU2ZmJkMDVhYzZkZDYiLCJwIjoiaiJ9).

On the topic page, when adding an external link/URL item into the “Highlighted articles” section, the “Description/summary” field and “release date” field are both missing. This means that when you view the topic page on the front-end, the link item does not meet the design system format and is inconsistent with other automated links in the section.

This PR adds description, release date, and content type fields for use with external URLs so that they can be displayed in the same way as internal pages. Description is a mandatory field when an external URL is set.

### How to review

- Create a topic page
- Create an article
- On the topic page add the article as an internal page under "Highlighted Articles"
- Add an additional external URL under "Highlighted Articles"
- Edit the description, release date, and content type fields for the external article
- Ensure that the internal and external article are rendered consistently on the topic page under "Related Articles"

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

N/A

[DPD-2526]: https://officefornationalstatistics.atlassian.net/browse/DPD-2526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cms/topics/services/related_content.py
```diff
@@ -71,12 +71,17 @@ def _process_related_article(self, related: TopicPageRelatedArticle) -> ArticleD
         # Handle external articles
         if not related.page:
             if related.external_url:
-                return {
+                external_article: dict = {
```
MaciekBaron: If we replace `dict` with `ExternalArticleDict`, then you don't need to use `type: ignore[return-value]` below (`ArticleDict` is a union of `ExternalArticleDict` and `InternalArticleDict`). Just remember to update your imports.

MaciekBaron (commented):
I know this is a draft but I thought I'd do a quick review before my annual leave :) Hope you don't mind!

cms/core/formatting_utils.py
```diff
@@ -39,13 +39,20 @@ def format_as_document_list_item(
 
 def _format_external_link(page_dict: ExternalArticleDict) -> DocumentListItem:
     """Format external link dictionary into DocumentListItem."""
-    return format_as_document_list_item(
+    content_type = page_dict.get("content_type") or _("Article")
```
BJacksonONS: I think this has an issue I encountered on my similar pr, where we can just end up returning the uppercase enum value rather than the translation string. [This is how I addressed it](https://github.com/ONSdigital/dis-wagtail/pull/609/changes#diff-3bcad11828ced8c0477df9700932af902f73fb74dc5dbe447d82cb0611d881baR11)

[PR link](https://github.com/ONSdigital/dis-wagtail/pull/609/changes#diff-41bd569115b0efa2842063c90703ec2df41de8c6f74ab4f3d41b3966aec00a3f)

cms/topics/models.py
```diff
@@ -69,7 +69,24 @@ class TopicPageRelatedArticle(Orderable):
...
+    description: models.TextField = models.TextField(
+        blank=True,
+        default="",
+        null=True,
+        help_text="Description or summary for the external link.",
+    )
+    content_type: models.CharField = models.CharField(
+        max_length=255,
+        blank=True,
+        default="",
```
BJacksonONS: Nit: We can probably leave the default empty if it's nullable, though it's not a huge issue.

BJacksonONS (changes requested):
Only one thing I can spot that needs addressing, otherwise looks good

cms/core/formatting_utils.py
joepercival: Thanks - 78bcc0f

cms/topics/models.py
joepercival: ede888f

cms/topics/services/related_content.py
joepercival: Good spot, 42bf67b

cms/core/formatting_utils.py
```diff
@@ -39,13 +40,20 @@ def format_as_document_list_item(
 
 def _format_external_link(page_dict: ExternalArticleDict) -> DocumentListItem:
     """Format external link dictionary into DocumentListItem."""
-    return format_as_document_list_item(
+    content_type = get_related_content_type_label(str(page_dict["content_type"] or _("Article")))
```
BJacksonONS: We don't need to call `get_related_content_type_label` with `_("Article")`, the function converts from the all caps string to the translation string label which we already have.

Maybe something like
```
content_type = get_related_content_type_label(page_dict.get("content_type") if page_dict.get("content_type") else _("Article")
```
or
```
if page_type := page_dict.get("content_type"):
    content_type = get_related_content_type_label(page_type)
else:
    content_type = _("Article")
```

joepercival: Latest version is here - fb7ae59
(I also switched to using RelatedContentType.choices for input validation 93afabc)

BJacksonONS (approved):
LGTM

cms/topics/models.py
```diff
@@ -69,7 +70,23 @@ class TopicPageRelatedArticle(Orderable):
...
+        choices=RelatedContentType.choices,
+        blank=True,
+        null=True,
+        help_text="Content type of the external content, e.g. Article (default), Dataset, Methodology.",
+    )
+    release_date: models.DateField = models.DateField(
+        blank=True,
+        null=True,
+        help_text="Release date of the external content.",
+    )
```
MebinAbraham, line 90: We should look to model this differently, as this can now be used in conjunction with "Page", which is not valid per se. We'll see what we can do within the limitations of Wagtail while trying to make it as user-friendly as possible. The minimum might be some validation to make it mutually exclusive with the Page option, which currently doesn't exist.

MebinAbraham: This is an example card that needed some analysis and business input as I think only the content team has proposed this.

BJacksonONS: This and https://github.com/ONSdigital/dis-wagtail/pull/609 mirror the current behaviour of the related links streamfield block.

Do we potentially want to make the additional metadata fields mutually exclusive with a selected page across all of them?

MebinAbraham: > Do we potentially want to make the additional metadata fields mutually exclusive with a selected page across all of them?

Yes, that is the minimum we should do, but I'm not a fan of how this is modelled and represented to the users (might be limited by Wagtail features).

functional_tests/steps/topic_page_manual_related_articles.py
```diff
@@ -54,6 +54,9 @@ def user_has_added_one_external_related_article(context: Context) -> None:
         "https://example.com/test-article"
     )
     context.page.locator(f"#id_related_articles-{context.manual_article_index}-title").fill("Test External Article")
+    context.page.locator(f"#id_related_articles-{context.manual_article_index}-description").fill(
```
MebinAbraham, line 46: We should test all the new fields. Release date + content type.

cms/topics/models.py
MebinAbraham: As discussed, given how far this has gone and given we will have the migration overhead either way, we will do the remodelling as follow-ups.

functional_tests/steps/topic_page_manual_related_articles.py
joepercival: d4f5628

functional_tests/steps/topic_page_manual_related_articles.py
```diff
@@ -126,6 +137,30 @@ def related_article_is_second_in_list(context: Context, title: str) -> None:
     expect(second_article.get_by_role("link", name=title)).to_be_visible()
 
 
+@then('the related article "{title}" displays the description "{description}"')
+def related_article_displays_description(context: Context, title: str, description: str) -> None:
+    """Assert that the given related article's description appears in the list item."""
+    related_articles_section = context.page.locator("section#related-articles")
+    li = related_articles_section.locator(f'li:has-text("{title}")').first
```
MaciekBaron: Nit: I think you should be able to use the standard `get_by_text` locator here, which is a cleaner solution

MaciekBaron (changes requested):

functional_tests/steps/topic_page_manual_related_articles.py
joepercival: Had some issues getting this working as the get_by_text locator just returns the anchor tag containing the title text rather than the entire li, but got it working here 82e7422 (not sure if this is what you had in mind)

charlotteromaniuk (approved):
LGTM

sanjeevz3009 (approved):
LGTM! Thanks @joepercival

functional_tests/steps/topic_page_manual_related_articles.py
sanjeevz3009: Resolving on behalf!

MaciekBaron (approved):

cms/topics/migrations/0010_topicpagerelatedarticle_content_type_and_more.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 6.0.5 on 2026-05-18 15:48
```
BJacksonONS, line 1: Can we rename this to 0010 so we don't have two 9's?

Not sure if django would also want the dependencies to be updated.

joepercival: f09fc6e

sanjeevz3009 (changes requested):
Did another round of code review + functionality test review.

There is an issue where if I don't enter the release date within the block -> then there is a `500` error thrown. But I can save the page fine. Release date is optional, so page saves fine, but then at the same time, if the release date is not there, a 500 error.

Take a look at the line: https://github.com/ONSdigital/dis-wagtail/pull/637/changes#diff-3bcad11828ced8c0477df9700932af902f73fb74dc5dbe447d82cb0611d881baR55

Need to be fixed before merge. Should be a minor fix.

joepercival:
> Did another round of code review + functionality test review.
> 
> There is an issue where if I don't enter the release date within the block -> then there is a `500` error thrown. But I can save the page fine. Release date is optional, so page saves fine, but then at the same time, if the release date is not there, a 500 error.
> 
> Take a look at the line: https://github.com/ONSdigital/dis-wagtail/pull/637/changes#diff-3bcad11828ced8c0477df9700932af902f73fb74dc5dbe447d82cb0611d881baR55
> 
> Need to be fixed before merge. Should be a minor fix.

0e5f307 & fixed a similar issue in 5cd23c8

cms/core/formatting_utils.py
```diff
@@ -43,14 +44,22 @@ def format_as_document_list_item(
...
+    content_type = get_related_content_type_label(page_type)
+
+    page_datum: DocumentListItem = format_as_document_list_item(
         title=page_dict["title"],
         url=page_dict["url"],
-        content_type=get_related_content_type_label(page_dict["content_type"]),
+        content_type=content_type,
         description=page_dict.get("description", ""),
-        release_date=page_dict["release_date"],
+        release_date=page_dict.get("release_date", None),
```
MebinAbraham: Nit: None is the default already when using `.get()`.
```suggestion
        release_date=page_dict.get("release_date"),
```

sanjeevz3009 (approved):
LGTM!

sanjeevz3009 (approved):
`PO` file regenerations looks good.

PR #638, opened 2026-05-18
update webpack dev server version

BJacksonONS:

Appease osv-scanner for the second time today.

### How to review

Ensure dev server runs correctly with `honcho start` in container.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

NA


joepercival (commented):
Had just spotted this as well. Do we need to update package.json too or just package-lock.json?

BJacksonONS:
> Had just spotted this as well. Do we need to update package.json too or just package-lock.json?

Ah yeah probably should, I used `npm audit fix`, not entirely sure why it doesn't do both.

joepercival (approved):
LGTM

kevinhowbrook (approved):

sanjeevz3009 (approved):
LGTM!

PR #639, opened 2026-05-19
fix: image block download failure

kevinhowbrook:

**Update:** This has been re-set and re-implemented due to large conflicts in the relevant files, a rebase wasn't really possible

Ticket: [CMS-1032](https://officefornationalstatistics.atlassian.net/browse/CMS-1032?atlOrigin=eyJpIjoiNjMwZDY5MTc0ODBlNDdiZTljM2ViYTUwYjNiMzVlMTIiLCJwIjoiaiJ9)

The image block download link pointed directly at the S3 file URL. Browsers silently ignore the HTML download attribute for cross-origin URLs, so clicking the link opened the image in a new tab instead of downloading it.

This adds ImageDownloadView at /images/download/<rendition_id> which proxies the file through Django with Content-Disposition: attachment, forcing a download regardless of where the file is stored. The image block template is updated to use this same-origin URL instead of the direct S3 URL.

### How to review

1. Log into Wagtail and open an existing statistical article page (or any page with an image block)
2. Add an image block, enable the download option, save and publish
3. Navigate to the live page and click the Download link
4. The file should download to your file system rather than opening in a new tab

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

- [ ] Create followup ticket for documents

[CMS-1032]: https://officefornationalstatistics.atlassian.net/browse/CMS-1032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
Ticket: [cms-1032](https://officefornationalstatistics.atlassian.net/browse/cms-1032)


cms/jinja2/templates/components/streamfield/image_block.html
```diff
@@ -43,10 +43,10 @@ <h4 class="ons-u-fs-r--b ons-u-mb-2xs">{{ _("Notes") }}</h4>
         {% set download_aria_label = _("Download image") %}
     {% endif -%}
 
-    {%- if value.download and large_src %}
+    {%- if value.download and download_url %}
```
kevinhowbrook: This should be covered by the existing [InformationPageImageBlockRenderingTests.test_renders_when_image_is_missing]( https://github.com/ONSdigital/dis-wagtail/blob/48d3fbe8f63ec471f3bde4191c6b5f2d52026393/cms/core/tests/test_blocks.py#L1212-L1237)... if the image is deleted after publishing, get_context returns early without setting download_url, so this condition safely evaluates to False and the download section is skipped.

cms/core/tests/test_blocks.py
```diff
@@ -1235,3 +1253,79 @@ def test_renders_when_image_is_missing(self):
...
+        image_model = get_image_model()
+        self.image = image_model.objects.create(
+            title="Test image title",
+            file=get_test_image_file(),
+            description="Meaningful alt text",
+        )
+        self.small = self.image.get_rendition("width-1024")
+        self.large = self.image.get_rendition("width-2048")
+
+    def _make_information_page(self, *, download: bool) -> InformationPage:
```
BJacksonONS: Minor nit, does download need to be a param here when we're only ever using true in this test case?

BJacksonONS (approved):
Looks good

cms/core/tests/test_blocks.py
kevinhowbrook: Good spot, thanks
73429cd4105baa3c6238c14c50ea6680fded46d8

cms/core/blocks/embeddable.py
```diff
@@ -40,6 +41,7 @@ def get_context(self, value: StreamValue, parent_context: dict | None = None) ->
 
         context["small_src"] = small.url
         context["large_src"] = large.url
+        context["download_url"] = reverse("image_download", args=[large.pk])
```
RealOrangeOne: Issue: Rather than a new view, could we add a `?force_download=true` query param and reuse the existing `ImageServeView`? That view already handles permissions so it would be great to avoid duplication.

`force_download` would need to bypass the redirect to S3, set the `Content-Disposition` header and probably only work for `not IS_EXTERNAL_ENV` so it's not extra caching load on the public site.

BJacksonONS: Would `Cache-Control: no-store` not remove the need to account for different environments?

RealOrangeOne: It might, but then it also becomes a DoS vector, being able to reliably hit the origin server and cause it to do a fair bit of work.

sanjeevz3009:
@kevinhowbrook Thank you for picking up this ticket. I need to dig into some conversations, investigations Mebin and I did into this. You are on the right track, and I saw on the JIRA card the options you explored. Let's have a chat offline, before we progress this further.

cms/core/blocks/embeddable.py
BJacksonONS: Think my brain stopped working by 5, misread your initial comment as disabling for external env. This makes much more sense.

kevinhowbrook: Thanks, yea that makes sense. Signature only covers image_id and filter_spec so the query param won't affect it.

kevinhowbrook:
Added `do not merge` label as @sanjeevz3009 needs to review some wider conversations here

cms/core/blocks/embeddable.py
kevinhowbrook: bd7c4b59f83b14cc655a73040bb508e07a89905f

cms/private_media/views.py
```diff
@@ -67,16 +69,25 @@ def get(  # pylint: disable=too-many-arguments,too-many-positional-arguments
...
             except NotImplementedError:
                 file_url = rendition.file.path
             return self.redirect_to_file(file_url)
 
         # Serve file contents
+        if force_download:
+            response = self._serve_rendition(rendition)
+            add_never_cache_headers(response)
+            _, ext = os.path.splitext(rendition.file.name)
+            response["Content-Disposition"] = f'attachment; filename="{image.title}{ext}"'
```
RealOrangeOne: Suggestion: Use `django.utils.http.content_disposition_header` to ensure it's escaped properly

cms/private_media/views.py
```diff
@@ -67,16 +69,25 @@ def get(  # pylint: disable=too-many-arguments,too-many-positional-arguments
...
             try:
                 file_url = rendition.file.url
             except NotImplementedError:
                 file_url = rendition.file.path
             return self.redirect_to_file(file_url)
 
         # Serve file contents
+        if force_download:
+            response = self._serve_rendition(rendition)
+            add_never_cache_headers(response)
```
RealOrangeOne: Thought: This is duplicated from `serve_private_rendition` - perhaps it can be reused? Pretending the file is private when forcing a download is probably fine.

cms/core/blocks/embeddable.py
```diff
@@ -40,6 +40,7 @@ def get_context(self, value: StreamValue, parent_context: dict | None = None) ->
 
         context["small_src"] = small.url
         context["large_src"] = large.url
+        context["download_url"] = large.serve_url + "?force_download=true"
```
RealOrangeOne: Question: There's a check in the template for `if large_src` - is that just in case it's not set or can the value be empty? If so this could end up as `"?force_download=true"` which isn't ideal. Or is it just a generic template?

kevinhowbrook: Looks just generic to me... pretty sure an empty value isn't possible though due to the early return above on L:35, or am i missing something?

```python
if not image:
    return context
```

cms/private_media/views.py
kevinhowbrook: Thanks 56cc9663d84d6eebfbebc3f45257607ff47ba6c9

cms/private_media/views.py
kevinhowbrook: cheers 230679fde5e44fdbe3c03eb7952a9ea5accaedad

cms/core/blocks/embeddable.py
RealOrangeOne: Yeah sounds like it's a generic template then. Let's leave as-is.

RealOrangeOne (approved):

cms/private_media/tests/test_serve_views.py
```diff
@@ -163,6 +166,99 @@ def test_serve_with_invalid_filter_spec(self):
...
+    def test_force_download_public_image(self):
+        """Public image with force_download=true serves the file directly as an attachment."""
+        rendition = self.public_image_renditions[0]
+        url = rendition.serve_url + "?force_download=true"
+
+        response = self.client.get(url)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn("attachment", response.get("Content-Disposition", ""))
+
```
MaciekBaron, line 178: I think we should also evaluate negative scenarios to ensure the `Content-Disposition` is not set when:
* `?force_download=false`
* `?force_download=1` (any truthy value)
* No query present

We could use a list and iterate through sub tests

cms/private_media/views.py
```diff
@@ -67,16 +70,26 @@ def get(  # pylint: disable=too-many-arguments,too-many-positional-arguments
...
             except NotImplementedError:
                 file_url = rendition.file.path
             return self.redirect_to_file(file_url)
 
         # Serve file contents
+        if force_download:
+            response = self.serve_private_rendition(rendition)
+            _, ext = os.path.splitext(rendition.file.name)
+            response["Content-Disposition"] = cast(
+                str, content_disposition_header(as_attachment=True, filename=f"{image.title}{ext}")
```
MaciekBaron: AFAIK `content_disposition_header` will [encode the string](https://github.com/django/django/blob/0e3eed56a2c6515a74fee9bc1a8ada68ae3a4b26/django/utils/http.py#L370) to follow RFC 6266 (to prevent header injections) but it will not take care of OS forbidden characters:

```python
>>> from django.utils.http import content_disposition_header
>>> content_disposition_header(as_attachment=True, filename="Hello/World:?")
'attachment; filename="Hello/World:?"'
```

I think different browsers will use different fallback characters, and things like `What does the future hold?` might be converted to `What does the future hold_`.

Maybe we should consider using `slugify` for this to guarantee consistent behaviour?

MaciekBaron (changes requested):

cms/private_media/tests/test_serve_views.py
kevinhowbrook: :+1: added in 9be48acd06a3ebd2a767509170ccb462c5b46a8b

cms/private_media/views.py
kevinhowbrook: Nice spot - thanks @MaciekBaron 
db08be2841e6019e85f420d0a77b573cfa7c3f4e

kevinhowbrook:
Update: Removed labels of 'discussion' and 'dnm' - @sanjeevz3009 said this is in the right direction so I think it's ok to review now (althoug there was perhaps something mentioned about filename titles that is relevant?)

cms/private_media/views.py
```diff
@@ -67,16 +71,26 @@ def get(  # pylint: disable=too-many-arguments,too-many-positional-arguments
                 status=400,
             )
 
+        force_download = not settings.IS_EXTERNAL_ENV and request.GET.get("force_download") == "true"
```
MebinAbraham, line 78: Will come back to this, need to have a think about this approach.

MaciekBaron (approved):

cms/private_media/views.py
MebinAbraham, line 78: ~~Let's document why it's not external; otherwise, it might look like a gap vs intended behaviour.~~

This may need to change based on my comment https://github.com/ONSdigital/dis-wagtail/pull/639/changes#r3655418543. Do we need to check if it is internal env?. If it was a private image and it was external, it wouldn't get this far as we would have raised a PermissionDenied?

cms/private_media/views.py
```diff
@@ -67,16 +71,26 @@ def get(  # pylint: disable=too-many-arguments,too-many-positional-arguments
...
+        if not force_download and image.is_public and not image.has_outdated_file_permissions():
             try:
                 file_url = rendition.file.url
             except NotImplementedError:
                 file_url = rendition.file.path
             return self.redirect_to_file(file_url)
 
         # Serve file contents
+        if force_download:
+            response = self.serve_private_rendition(rendition)
```
MebinAbraham, line 91: At this point, the image could be public?

cms/private_media/views.py
```diff
@@ -67,16 +71,26 @@ def get(  # pylint: disable=too-many-arguments,too-many-positional-arguments
...
+            _, ext = os.path.splitext(rendition.file.name)
+            response["Content-Disposition"] = cast(
+                str, content_disposition_header(as_attachment=True, filename=f"{slugify(image.title)}{ext}")
```
MebinAbraham: Let's avoid cast, and use type ignore if needed as per our style guide.

cms/core/blocks/embeddable.py
MebinAbraham: Let's use `large.url + "?force_download=true"` which will use `file.url` if the image is public and only uses `serve_url` if the image is private.

MebinAbraham (commented):
Once finalised, we also need to make the same set of changes for Documents (happy to be a separate PR).

cms/private_media/views.py
kevinhowbrook: Sorry the rebase for this was awful so I had to reset - I've unwrapped the cast during the rebase

cms/private_media/views.py
kevinhowbrook: I think Jakes comment earlier is relevant here (https://github.com/ONSdigital/dis-wagtail/pull/639#discussion_r3267830909) -  won't removing it mean anyone can append `?force_download=true` to any published image URL and get an uncached full-file stream out of the origin.

cms/private_media/views.py
kevinhowbrook: Yea, it can be public. It's a naming mixup rather than anything to do with privacy: `serve_private_rendition` is just `_serve_rendition` plus `add_never_cache_headers`, and never-cache is what we want for a forced download. I originally had this more specific but using this method came from @RealOrangeOne's earlier suggestion: https://github.com/ONSdigital/dis-wagtail/pull/639#discussion_r3279741909

Happy to pull it out into a `serve_rendition_as_download` method so the call site stops implying the image is private. Would you prefer that?

cms/core/blocks/embeddable.py
kevinhowbrook: I've had to re-do this basically, because of the changes in the rebase. So the chunk to review is https://github.com/ONSdigital/dis-wagtail/pull/639/changes#diff-9243ef7c310ec9bd45a1aa2b706a32b81f66a6dbd2acf707e8ec2c93a4e7ed66R110-R132

sanjeevz3009 (approved):
LGTM! There are existing comments and decisions that need to be made, and we seem to be waiting on them. I had the same queries/ similar as the ones @MebinAbraham already pointed out, and some of these decisions go back and were being discussed.

cms/private_media/views.py
kevinhowbrook: Have moved this into it's own method - although it's duping code I think being specific here is better (and to avoid a method name like private too) c7a187534f81e4904200c2fdaa7b0b83d84543f1

cms/core/blocks/embeddable.py
```diff
@@ -120,7 +120,16 @@ def _get_download_options(self, large_image: AbstractRendition, request: HttpReq
 
         return {
             "title": _("Download this image"),
-            "itemsList": [{"text": link_text, "url": large_image.url, "download": "file", "attributes": attributes}],
+            "itemsList": [
+                {
+                    "text": link_text,
+                    # Public images resolve to the media URL and private ones to the serve view.
+                    # Both honour force_download, so the response is always an attachment.
+                    "url": large_image.url + "?force_download=true",
```
BJacksonONS: Q: Is there any world where we have other query params on these urls? Just wondering if we should merge in with urllib rather than use string concat.

kevinhowbrook: Thanks - I dont think there are but doesn't hurt, I've added an 'add_query_params` util for it - tests to follow 540ee56c3572cc27d3b7f5a661cda7cea1ff606a

BJacksonONS (approved):

PR #643, opened 2026-05-20
Add workflow that checks for required approver when label applied

BJacksonONS:

Add a label that applies a check for review from a whitelist of approved reviewers.

We've been using the `do not merge` label for this purpose for a little while, but the name doesn't make it clear the reason why the pr is being blocked.

Using codeowners to add a hard requirement has been considered and deemed overly restrictive for small prs around typos or other minor changes.

This proposes a new label that makes the reasoning more obvious to contributors why their PR is blocked, and reduce manual label management.

### How to review

Decide if you think this is a good solution to the problem.

Check logic looks sound.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/.github/workflows/required-approvers-label.yml
+++ b/.github/workflows/required-approvers-label.yml
@@ -1,0 +1,50 @@
+name: Require ONS Approval
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, labeled, unlabeled, edited]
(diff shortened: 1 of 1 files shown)


BJacksonONS:
Not planned for now

PR #645, opened 2026-05-21
Hide non editable fields in bundle minimap

BJacksonONS:

Addresses CMS-1214

Currently we need to keep status, approved_by and approved_at fields as hidden fields in the wagtail admin bundle UI, but they are handled exclusively by the form. The existing logic was not sufficient to hide them from the editor minimap.

This PR adds a new HiddenFieldPanel class that force overrides the heading to an empty string, so header templating isn't displayed and it's not picked up by the minimap js. We need to keep the non editable fields present in the form to be submitted at the moment given the current logic.

### How to review

Check in the UI that the fields don't get shown in the form or minimap.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/cms/bundles/panels.py
+++ b/cms/bundles/panels.py
@@ -213,7 +213,7 @@ class ReleaseChooserWithDetailsPanel(BundleFieldPanel):
 class HiddenFieldPanel(FieldPanel):
     """A FieldPanel that renders no output.
 
-    Used for fields managed purely via form logic that need to be excluded from all UI including minimap.
+    Used for fields managed purely via form logic that need to be excluded from all UI, including minimap.
     """
 
     def __init__(self, field_name: str, **kwargs: Any) -> None:
--- a/cms/bundles/panels.py
+++ b/cms/bundles/panels.py
@@ -221,6 +221,6 @@ class HiddenFieldPanel(FieldPanel):
         super().__init__(field_name, **kwargs)
 
     class BoundPanel(FieldPanel.BoundPanel):
-        def is_shown(self) -> bool:
-            # Exclude from the minimap and panel tree
-            return False
+        def __init__(self, **kwargs: Any) -> None:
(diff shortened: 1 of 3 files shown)


BJacksonONS:
Adding label until sign-off from @sanjeevz3009

kevinhowbrook (approved):
lgtm, thanks @BJacksonONS 

Image

cms/bundles/panels.py
```diff
@@ -208,3 +208,19 @@ def format_value_for_display(self, value: ReleaseCalendarPage | None) -> str:
...
+    """
+
+    def __init__(self, field_name: str, **kwargs: Any) -> None:
+        kwargs.setdefault("classname", "hidden w-hidden")
+        super().__init__(field_name, **kwargs)
+
+    class BoundPanel(FieldPanel.BoundPanel):
+        def __init__(self, **kwargs: Any) -> None:
+            super().__init__(**kwargs)
+            self.heading = ""
```
MaciekBaron, line 228: This would benefit from a comment explaining what's happening
```suggestion
            # Set heading to an empty string to hide it from the minimap
            self.heading = ""
```

MaciekBaron (changes requested):

cms/bundles/panels.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/645/commits/f5acec73b0521683572b82b38d0c62203d6e2475

sanjeevz3009 (approved):
LGTM, thank you for the changes @BJacksonONS

charlotteromaniuk (approved):
LGTM!

MaciekBaron (approved):
LGTM!

PR #647, opened 2026-05-22
Use stream from openssl to generate randomness for feature split script

BJacksonONS:

On PR #609, functional tests were failing to run due to an error when splitting the features.
This uses a stream from openssl to generate randomness for the `shuf` function, then kills the stream at the end of execution.

https://github.com/ONSdigital/dis-wagtail/actions/runs/25908707290/job/76148374282

### How to review

Ensure splits are generated randomly.
Ensure seeds correctly regenerate the same splits.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/.github/split_functional_test_features.sh
+++ b/.github/split_functional_test_features.sh
@@ -31,11 +31,12 @@ shard_zero_based=$((SHARD_INDEX - 1))
 
 seedfile=""
 order_cmd=(sort)
-
 if [[ -n "$SHUFFLE_SEED" ]]; then
     echo "Using seed '$SHUFFLE_SEED' for shuffling" >&2
-    seedfile=$(mktemp)
-    printf '%s' "$SHUFFLE_SEED" >"$seedfile"
+    seedfile=$(mktemp -u)
+    mkfifo "$seedfile"
+    openssl enc -aes-256-ctr -nosalt -k "$SHUFFLE_SEED" -in /dev/zero 2>/dev/null >"$seedfile" &
+    openssl_pid=$!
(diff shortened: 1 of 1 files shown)


.github/split_functional_test_features.sh
```diff
@@ -31,16 +31,20 @@ shard_zero_based=$((SHARD_INDEX - 1))
...
 seedfile=""
 order_cmd=(sort)
-
 if [[ -n "$SHUFFLE_SEED" ]]; then
     echo "Using seed '$SHUFFLE_SEED' for shuffling" >&2
-    seedfile=$(mktemp)
-    printf '%s' "$SHUFFLE_SEED" >"$seedfile"
+    seedfile=$(mktemp -u)
+    mkfifo "$seedfile"
+    openssl enc -aes-256-ctr -nosalt -k "$SHUFFLE_SEED" -in /dev/zero 2>/dev/null >"$seedfile" &
```
RealOrangeOne, line 38: Question: Why is this better? This appears to just derive a larger value for `shuf` to use, which it will be doing in itself. Writing the seed multiple times to the file would achieve the same.

It adds a lot of complexity to the split script.

BJacksonONS: Will close this in favour of this https://github.com/ONSdigital/dis-wagtail/pull/648

I swear at the time I tried this and it wasn't working, so using the stream just guaranteed it would always be enough. Went back to the problematic commit and was able to repeat the error on the old script, but printing 5 or so times works now.

PR #648, opened 2026-05-22
Print seed multiple times to create extra random bytes

BJacksonONS:

On PR https://github.com/ONSdigital/dis-wagtail/pull/609, functional tests were failing to run due to an error when splitting the features.
This prints the value multiple times to the temporary file to ensure there's enough random bytes.

https://github.com/ONSdigital/dis-wagtail/actions/runs/25908707290/job/76148374282

### How to review

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


kevinhowbrook (approved):

joepercival (approved):

MaciekBaron (approved):

.github/split_functional_test_features.sh
```diff
@@ -35,7 +35,10 @@ order_cmd=(sort)
 if [[ -n "$SHUFFLE_SEED" ]]; then
     echo "Using seed '$SHUFFLE_SEED' for shuffling" >&2
     seedfile=$(mktemp)
-    printf '%s' "$SHUFFLE_SEED" >"$seedfile"
+    # Print the seed to the file 10 times to ensure enough randomness for shuf
+    for _ in {1..10}; do
+        printf '%s' "$SHUFFLE_SEED"
+    done >"$seedfile"
```
MebinAbraham, line 41: I don't think this will be random enough, as we are just repeating the bytes and may not be long enough for the long term.

How about something like: https://www.gnu.org/software/coreutils/manual/html_node/Random-sources.html#Sources-of-random-data

BJacksonONS: I'd looked at that but got some push back as being too complex

https://github.com/ONSdigital/dis-wagtail/pull/647

MebinAbraham: Given the seed is small and we're only repeating it 10 times, I don't think it will be the "same" per se. It's not a problem we will see now. I'm not opposed to your previous change, given it is not something we actively need to look at. I'm also happy to see how this plays out.

BJacksonONS: both changes worked to address the issue with that particular commit failing to run, so I think for now either will do.

There are some other possible changes to the splitting we might want to consider I tested on https://github.com/ONSdigital/dis-wagtail/pull/649 so maybe just go with the most straightforward for now and pick up later?

MebinAbraham: Sounds like a plan

MebinAbraham (approved):

sanjeevz3009 (commented):
LGTM!

PR #649, opened 2026-05-22
Use browser context to login for functional tests

BJacksonONS:

Currently for all tests outside of cognito test cases we're going through the standard login flow 

1. Go to login page
2. Enter username + password
3. Click Button

Outside of a test specifically to validate this works, we can skip this and login by inserting a cookie directly into the browser context. This should speed up the functional tests by a small amount as almost every test starts with an authenticated user.

### How to review

Ensure all functional tests continue to pass.
Ensure there is still a test to validate the login flow.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


.github/split_functional_test_features.py
```diff
@@ -0,0 +1,90 @@
+# pylint: skip-file
```
MebinAbraham: This was kept simple on purpose. Has this changed to fix an issue, or is it more about better randomness/deterministic randomness?

BJacksonONS: It more evenly splits the actual workload between the shards. Some feature files are much longer/shorter than others so bucketing equal file counts could lead to large variability in the runtime of shards.
e.g.

Image

MebinAbraham: Worth adding some module-level docs on what this is intending to achieve, as we likely don't need to touch it much going forward, so worth having a summary to avoid needing to scan and find out.

functional_tests/step_helpers/auth_utils.py
```diff
@@ -209,3 +213,31 @@ def _capture(route: Any, request: Any) -> None:
...
+    context.browser_context.add_cookies(
+        [
+            {
+                "name": "sessionid",
+                "value": session_key,
+                "path": "/",
+                "domain": "localhost",
+                "secure": False,
+                "httponly": True,
+                "samesite": "Lax",
```
MebinAbraham: Pull these from settings/config?

```py
{
    "name": settings.SESSION_COOKIE_NAME,
    "value": session_key,
    "url": context.base_url,
    "httpOnly": True,
    "secure": settings.SESSION_COOKIE_SECURE,
    "sameSite": settings.SESSION_COOKIE_SAMESITE,
}
```

.github/split_functional_test_features.py
MebinAbraham: What are the pylint complaints?

MebinAbraham (approved):

.github/workflows/ci.yml
```diff
@@ -309,7 +309,7 @@ jobs:
 
       - name: Select feature subset
         run: |
-          SHARD_INDEX=${{ matrix.shard }} ./.github/split_functional_test_features.sh > features.txt
+          SHARD_INDEX=${{ matrix.shard }} poetry run python .github/split_functional_test_features.py
```
kevinhowbrook, line 312: Does this mean we can remove the old [.sh](https://github.com/ONSdigital/dis-wagtail/blob/test/tidy-functional-test-login-flows/.github/split_functional_test_features.sh) file? I can't see any other references to it anymore

BJacksonONS: Yes we would. I need to tidy up the python script a little first, sort out the pylint issues.

.github/split_functional_test_features.py
BJacksonONS: issue was in pre-commit pylint fails to run against files outside of the `cms` directory as the `pylint_django` plugin can't load `cms.settings.production`. I've put in a workaround to add the rood dir to `sys.modules` which works to allow this to happen, but alternatively we could just disable for this file.

Running a separate pylint for scripts outside of `cms` would require a new `.pylintrc` as I couldn't get the cli argument to override the plugins loaded from `pylintrc`.

functional_tests/step_helpers/auth_utils.py
BJacksonONS: [f26393b](https://github.com/ONSdigital/dis-wagtail/pull/649/commits/f26393b8d60430f1ca91b73075f47940c8f904c8)

.github/split_functional_test_features.py
BJacksonONS: [2c0da79](https://github.com/ONSdigital/dis-wagtail/pull/649/commits/2c0da7979b2915ad41a7e3890115ff57384160e7)

.pre-commit-config.yaml
```diff
@@ -27,6 +27,7 @@ repos: # Python linting and formatting
           - '--reports=n' # only display messages
           - '--output-format=colorized'
           - '--rcfile=.pylintrc'
+          - '--init-hook=import sys; sys.path.insert(0, ".")'
```
BJacksonONS, line 31: this ensures the root dir is loaded into system so pylint-django doesn't error on loading settings files in the `.github` directory.

kevinhowbrook (approved):

.pre-commit-config.yaml
MebinAbraham: Worth adding that comment inline?

BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/649/commits/2ecfb59ced221b99b846fb24777f099be92ff002

MebinAbraham (approved):

sanjeevz3009 (commented):
LGTM!

PR #651, opened 2026-05-26
Table accessibility improvements

helenb:

See https://officefornationalstatistics.atlassian.net/browse/CMS-1207 and https://officefornationalstatistics.atlassian.net/browse/DPD-3448?focusedCommentId=770748

- Fixes missing block id
- Adds a subtitle field
- Migrates data from the caption field to the subtitle field
- Re-labels the caption field to 'accessible label'
- Rather than using the DS aria-label field, uses the `caption` field and visually hides it with `hideCaption`
- Updates the additional heading logic for the footnotes and downloads
- Ensures a title is present if a subtitle is used
- Adds and updates tests

### How to review

**_Important note: if you previously reviewed this pr you will need to backwards migrate core to 0017 before applying the changes in 19075db0ab4d79e3c4e9f3668337e1751b35f1e9_**

- First create a page with several tables. Ensure that some have captions, and that at least one has a caption and no downloads.
- Add some footnotes to one of the tables with a caption, and check the heading levels for the footnotes and downloads.
- Check the ids used for each table - they are currently all 'table-'
- Run the data migrations
- Edit your table page - you should see that the caption field is now labelled 'Accessible label', and the caption data has moved to the subtitle field.
- Try to save the page: you should be prompted to add some content in the 'accessible label' field, and to fix the table where there is now a subtitle but no title.
- Fix the validation errors, and save the page
- Check the heading levels have been updated for footnotes and downloads
- Check the ids for each table - they should now all be 'table-`[unique id]`'

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy - subject to discussion though, as this includes a data migration.
- [ ] Not safe to auto-deploy

### Follow-up Actions


cms/core/blocks/markup.py
```diff
@@ -259,3 +262,8 @@ def _build_preview_table_download_url(page: Any, block_id: str, request: Any = N
     class Meta:
         icon = "table"
         template = "templates/components/streamfield/table_block.html"
+        # Marker class on the StructBlock wrapper used by an admin CSS rule to
+        # hide the inherited `caption` child from editors. The field is kept so
+        # its data is preserved and remains available for future use.
+        # See cms/core/static/css/admin.css.
+        form_classname = "ons-table-block"
```
helenb: Not sure if there is a better way to hide the `caption` field for now - the field will still be present, just hidden with CSS.

zerolab: Perhaps `self.child_blocks["caption"].classname = "w-hidden"` in the block `__init__`?

kevinhowbrook: I don't see anything specifically better (by this I mean what you've done is ok), both would work I think. Arguably yea the wagtail way would mean no custom css is needed.

ref [FieldBlockAdapter line 107](https://github.com/wagtail/wagtail/blob/ac49f19db509d9e353269b5f563d362b3446359b/wagtail/blocks/field_block.py#L104-L118)

From looking at that I think we would use `form_classname` - looks like classname is legacy and still supported though, so:

```python
self.child_blocks["caption"].meta.form_classname =
  "w-hidden"
```

cms/core/migrations/0018_migrate_table_caption_to_subtitle.py
kevinhowbrook, line 1: Looks fine to me this @helenb, forwards and backwards look like they do what's needed/intended. The single `_migrate` path is **good**. Personally I like to name the fwds and bkwds ops a bit more verbose, like what it's doing, those are hidden in the `forwards` and `backwards` methods themselves. E.g

```python
migrations.RunPython(migrate_caption_to_sub_heading, migrate_sub_heading_to_caption),
```

Just makes it more readable, but that's only my pref though...


I noticed `caption` might still be populated/used in tests, assume you might be still working on this so I'll hhold off on more comments for the moment, nice work :)

cms/core/blocks/markup.py
helenb: The issue I had with `form_classname` was that it hid the input field but not the 'Caption' label.

cms/core/migrations/0018_migrate_table_caption_to_subtitle.py
helenb: Yes I've updated the tests now, thanks for spotting though!

cms/core/blocks/markup.py
```diff
@@ -130,15 +131,28 @@ def render(self, value: dict, context: dict | None = None) -> str | SafeString:
 class ONSTableBlock(TinyTableBlock):
     """The ONS table block."""
 
+    sub_heading = blocks.CharBlock(label="Sub-heading", required=False)
```
helenb: I'm aware that 'Sub-heading' is somewhat inconsistent with 'Title' - 'Subtitle' might be better. I went with 'Sub-heading' as that was what we documented on the ticket, but happy to change if 'Subtitle' is preferred.

cms/core/migrations/0018_migrate_table_caption_to_subtitle.py
helenb: I'll hold off on renaming the `forwards` and `backwards` migrations for now, to see what other people prefer.

cms/core/blocks/markup.py
```diff
@@ -184,7 +198,7 @@ def get_context(self, value: dict, parent_context: dict | None = None) -> dict:
             return context
 
         options = {
-            "caption": value.get("caption"),
+            "ariaLabel": value.get("accessible_label"),
```
kevinhowbrook: **General observation**

this will populate as `None` as the migration doesn't set anything for `accessible_label` - that's probably expected but it's worth raising that anything existing will need populating in the CMS.

With None, won't it render aria-label=". " Is an empty (or whitespace) as an aria-label worse for screen readers rather than omitting it if not set?

helenb: Good point - I'm not sure how best to handle ensuring it is set everywhere that tables are used once this change goes through. Obviously as soon as a page is saved, then the editor will be forced to add an aria label.

helenb:
Related DS PR: https://github.com/ONSdigital/design-system/pull/3843

cms/core/blocks/markup.py
kevinhowbrook: We discussed this in our tech meeting today, we don't really consider it stopping the PR going forward. It was suggested that we could migrate the `caption` to the `accessible_label` so that there is data added that's better than `None`. But realistically there aren't too many tables at the moment - so it's worth raising to publishers that the `accessible_label` will need populating.

helenb:
I've marked this as 'do not merge' while I re-assess the approach.

cms/core/blocks/markup.py
helenb: This will no longer be relevant with the revised implementation that is coming shortly.

cms/core/blocks/markup.py
```diff
@@ -138,7 +140,24 @@ def __init__(
     ) -> None:
         super().__init__(local_blocks=local_blocks, search_index=search_index, **kwargs)
         # relabeled to match the publishing team's terminology
-        self.child_blocks["caption"].label = "Sub-heading"
+        self.child_blocks["caption"].label = "Accessible label"
```
MebinAbraham: Q: Any reason in charts we call it an "Audio description" and here "Accessible label". Are they not the same?. If so, we should align so we don't have inconsistency.

cms/core/blocks/markup.py
```diff
@@ -138,7 +140,24 @@ def __init__(
...
+        self.child_blocks["caption"].field.required = True
+        self.child_blocks["caption"].help_text = """
```
MebinAbraham: Any reason for triple quotes?.

cms/core/blocks/markup.py
MebinAbraham: Have asked content team to confirm.

cms/core/blocks/markup.py
MebinAbraham: Have asked content team to confirm if they want use to default to empty/"scrollable table" for now until they republish. Waiting on response.

helenb: Bear in mind that since this conversation was opened the implementation has change, and the accessible label now populates the caption element. However, the principle still applies. I don't think a default is a good idea - it will mean it is less likely to be updated with something meaningful when the page is edited, as the validation won't fail.

cms/core/blocks/markup.py
helenb: Yes, the audio description for charts is a longer piece of text describing the contents of the chart, whereas this has more of a naming function. I am pretty sure I discussed using this label with Zoe in refinement.

cms/core/blocks/markup.py
MebinAbraham: The default would be in the config, not the editor field value, so they still need to populate it and republish.

helenb: Ah makes sense.

MebinAbraham: Looking at the DS change you made, we don't need a default anymore. As it will simply omit caption when not provided. And aria won't be added since we don't configure that as we use caption.

helenb: Ok yes I see what you mean - we won't end up with `<caption></caption>` - so assume you'll update the content team @MebinAbraham ?

MebinAbraham: Ya publishing/content teams are already aware they need to republish the page once we get this into prod.

MebinAbraham: Marking this thread as resolved as the initial issue spotted is no longer a problem.

cms/core/blocks/markup.py
helenb: I have confirmation that we are going to use `Table title` and `Table subtitle`.

Question on how best to proceed: I can update the migration, but that means anyone who's applied this would need to backwards migrate. Is it better to add a new migration? Or we could update the label only, but that might lead to confusion in the future.

helenb: For context this is in relation to making these kinds of fields consistent for images, charts, iframes and tables. We are going to be using e.g. "Image title" and "Visualisation title" etc. I imagine for most of these we can simply update the label rather than making lots of migrations, but one to maybe discuss.

MebinAbraham: As discussed, keep the db model as title etc, but rename the admin label to "Table title" etc.

helenb: Further to this, given the conversation in refinement this morning, I will go with 'Title' and 'Subtitle' and leave the label prefixes for now.

helenb: I have now renamed sub-heading to subtitle.

cms/core/migrations/0018_migrate_table_caption_to_subtitle.py
```diff
@@ -0,0 +1,119 @@
+# Data migration to move ONSTableBlock `caption` data into the new `subtitle` field, and back again on reverse.
```
MebinAbraham, line 1: Do we actually need this?. You still have a caption now "Accessible label", is it really wrong that they become the "Accessible label". A caption becoming "sub title" is probably not always right anyway?

sanjeevz3009 (approved):
Apart from the comments already left. Looks good to me!

cms/core/migrations/0018_migrate_table_caption_to_subtitle.py
helenb: Yes I believe we do. The caption field was previously being labelled 'Sub-heading', so this is what it was being used for. So we need to migrate that to the new subtitle field. The caption is now used as the accessible label - because it is being passed to the hidden caption in the DS.

MebinAbraham: Ah right yes forgot it was relabelled as sub-heading. All good then.

cms/core/blocks/markup.py
```diff
@@ -139,7 +141,25 @@ def __init__(
...
         # relabeled to match the publishing team's terminology
-        self.child_blocks["caption"].label = "Sub-heading"
+        self.child_blocks["caption"].label = "Accessible label"
+        self.child_blocks["caption"].field.required = True
+        self.child_blocks["caption"].help_text = (
+            "A short label to explain what this table is about for those "
+            "using a screen reader. This will not be visible on the page "
+            "but is important for accessibility. "
+            "Note that this is rendered as a hidden caption element."
+        )
```
MebinAbraham: The help text is currently not working. It should be on the field. Let's add a test to cover the gap.

```py
        self.child_blocks["caption"].field.help_text = (
            "A short label to explain what this table is about for those "
            "using a screen reader. This will not be visible on the page "
            "but is important for accessibility. "
            "Note that this is rendered as a hidden caption element."
        )
```

However, since we are basically overriding more than just the label, I would suggest we redeclare the field instead of mutating the inherited field. Can achieve same thing by doing:

```py
class ONSTableBlock(TinyTableBlock):
    """The ONS table block."""

    caption = blocks.CharBlock(
        required=True,
        label="Accessible label",
        help_text=(
            "A short label to explain what this table is about for those "
            "using a screen reader. This will not be visible on the page "
            "but is important for accessibility. "
            "Note that this is rendered as a hidden caption element."
        ),
    )

    ...
```

helenb: @MebinAbraham I've now updated this redeclaring the field as you suggest. I haven't added a test as we'd now just be testing core wagtail functionality.

MebinAbraham (approved):

PR #652, opened 2026-05-26
Silently deduplicate internal related articles on Topic pages

kevinhowbrook:

:warning: Based on https://github.com/ONSdigital/dis-wagtail/pull/609

Ticket [CMS-1229](https://officefornationalstatistics.atlassian.net/browse/CMS-1229)

Fixes a bug where the same internal article page could be added multiple times to the "Highlighted articles" inline panel on a Topic page, with all duplicates appearing in both preview and on the live site.

As raised in recent discussions, the same silent deduplication has also been applied to the "Related methodologies" inline panel, which had the same issue.

  ### How to review

  1. On a Topic page, add the same article to "Highlighted articles" more than once and save.
  Verify only one instance appears in the preview and on the live page.
  2. Verify that external URL entries in the same panel are unaffected.
  3. On the same Topic page, add the same methodology to "Related methodologies" more than once
  and save. Verify only one instance appears.
  4. It's also worth checking the behaviour of 'save draft'/workflows.

  ### Deployment Safety

  - [x] Safe to auto-deploy

  ### Follow-up Actions

I haven't added dedicated unit tests directly against `DeduplicateInlinePanelAdminForm` for the new `deduplicate_nullable_page_formset` method, the behaviour is covered indirectly through the `TopicPageAdminForm` tests. Happy to add them in a new test_forms if preferred though.


BJacksonONS:
This PR is removing duplicate related articles at the form level, whereas the current behaviour with methodologies is to leave them both in but deduplicate in the output.

Not sure which is the correct behaviour, as far as the business is concerned. I'd personally expect a form submission error telling the user there's a duplicate, though appreciate that wasn't in the ticket.

image

zerolab:
@BJacksonONS we deduplicate bundle items (https://github.com/ONSdigital/dis-wagtail/blob/b35a70249ba0f51cd9efeb191ea067d61f6a77a8/cms/bundles/forms.py#L29) and topics (https://github.com/ONSdigital/dis-wagtail/blob/b35a70249ba0f51cd9efeb191ea067d61f6a77a8/cms/core/forms.py#L47-L53)

~@kevinhowbrook https://github.com/ONSdigital/dis-wagtail/blob/main/cms/core/forms.py#L19-L44 may come in handy~ edit: I saee the comment about nullable values

BJacksonONS:
@zerolab Thanks, looks like methodologies are the outlier then.

@sanjeevz3009 Do we want to align them in this PR or raise another ticket?

kevinhowbrook:
> @zerolab Thanks, looks like methodologies are the outlier then.
> 
> @sanjeevz3009 Do we want to align them in this PR or raise another ticket?

I don't really have a POV on what the aligned behaviour should be, but I  just chose to remove them from the form due to what's on the ticket:

> **expected behaviour**
>     The system should prevent duplicate entries OR
>     Automatically deduplicate repeated articles
>     Each related article should appear only once in both preview and live environments.

So I chose `or Automatically deduplicate repeated articles` - happy to follow what we think is best here though

@BJacksonONS - I looked at the methodologies too for similarities, but adding the same value twice threw an integrity error for me rather than any validation error - perhaps my DB is behind

kevinhowbrook:
UPDATE: As discussed in the tech sessions - I'm going to fixup methodologies here to de-duplicate in the same way.

cms/topics/forms.py
```diff
@@ -36,4 +35,8 @@ def clean(self) -> dict[str, Any] | None:
                 else:
                     seen_pages.add(internal_page)
 
+        # `page` is non-nullable on related methodologies, so
```
BJacksonONS: FYI my PR here https://github.com/ONSdigital/dis-wagtail/pull/609 is making page nullable for external links on methodologies, same as articles.

kevinhowbrook: Ah thanks - I think I'll re-base on top of #609 in that case

cms/core/forms.py
```diff
@@ -43,6 +43,27 @@ def deduplicate_formset(self, *, formset: str, target_field: str) -> None:
 
         return None
 
+    def deduplicate_nullable_page_formset(self, formset_name: str) -> None:
```
kevinhowbrook, line 46: As noted in the PR desc - I haven't added dedicated unit tests directly against DeduplicateInlinePanelAdminForm happy to though if people agree.

BJacksonONS (approved):

MaciekBaron (approved):

sanjeevz3009 (commented):
LGTM!

PR #653, opened 2026-05-27
Fix rich text rendering for notice and main_points_summary fields

kevinhowbrook:

Ticket: [CMS 1141](https://officefornationalstatistics.atlassian.net/browse/CMS-1141)

Two `RichTextField` values were being output without passing through Wagtail's rich text renderer, causing embedded links and other rich text features to break on the frontend. 

 ### How to review

1. Open a `ReleaseCalendarPage` in the admin, set the status to Cancelled, add rich text content to the **notice** field including an anchor link, save and publish. View the cancelled page,  the link should render correctly.
2. Open a `StatisticalArticlePage`, add rich text with an anchor link to the **main points summary** field, feature the article on a topic page, and view the topic page. The summary should render correctly in the featured article component.

### Deployment Safety

- [x] Safe to auto-deploy

### Follow-up Actions

None.


cms/release_calendar/tests/test_models.py
```diff
@@ -568,6 +568,16 @@ def test_rendered__changes_to_release_date(self):
...
 
+    def test_rendered_notice_renders_on_cancelled_page(self):
+        """Notice content should appear on the cancelled page."""
+        self.page.status = ReleaseStatus.CANCELLED
+        self.page.notice = "<p>Cancelled due to unforeseen circumstances</p>"
+        self.page.save_revision().publish()
+
+        response = self.client.get(self.page.url)
+
+        self.assertContains(response, "Cancelled due to unforeseen circumstances")
```
kevinhowbrook: fixing this

cms/articles/tests/test_article_models.py
```diff
@@ -1133,7 +1133,7 @@ def test_as_featured_article_macro_data(self):
         self.assertEqual(data["metadata"]["date"]["short"], "1 November 2024")
         self.assertEqual(data["metadata"]["date"]["iso"], "2024-11-01")
 
-        self.assertEqual(data["description"], "Test main points summary")
+        self.assertIn("Test main points summary", data["description"])
```
kevinhowbrook: fixing this

BJacksonONS (approved):
Looks good

cms/release_calendar/tests/test_models.py
```diff
@@ -568,6 +568,16 @@ def test_rendered__changes_to_release_date(self):
...
 
+    def test_rendered_notice_renders_on_cancelled_page(self):
+        """Notice content should appear on the cancelled page."""
+        self.page.status = ReleaseStatus.CANCELLED
+        self.page.notice = "<p>Cancelled due to unforeseen circumstances</p>"
+        self.page.save_revision().publish()
+
+        response = self.client.get(self.page.url)
+
+        self.assertContains(response, "<p>Cancelled due to unforeseen circumstances</p>", html=True)
```
MaciekBaron: Minor, non-blocking: Should we use some additional HTML tags internally, just in case the test returns a false positive if the raw text happens to be included inside a `p` tag in the template?

kevinhowbrook: Thanks - I pondered this so it's good to have your opinion. I thought adding html classes into the assertion might make the test a bit too brittle (html can change) but I think strengthening it is right so I've wrapped it in some of the markup in 11ab8d4395b21268e7c9e19b710825d587f322b0

MaciekBaron: Thanks, I like your approach better!

MaciekBaron (approved):

MebinAbraham (approved):

sanjeevz3009 (commented):
LGTM!

PR #654, opened 2026-05-27
Spike to investigate the switch to the wide container, and where we might need to restrict content width

helenb:

See https://officefornationalstatistics.atlassian.net/browse/CMS-1242

This pr experiments with a new class `ons-reading-width` which would be provided by the DS, to see how it could be used to restrict the width of content in the CMS where a maximum width applies to meet accessibility standards for reading width.

It is not intended to be merged - this PR is to share the spike code.

The comments add details of where we might add restrictions in the CMS, and where they might be needed in the DS. It isn't exhaustive of the places we'd need to update to ensure the reading width was restricted - more an exploration of the kinds of places it might apply.

Note that the `wide` option currently sets the overall container width at 1280px, but this is going to increase to 1304px.

### How to review

You can check out the branch and view in your browser, but the main thing is to review the code changes and comments.

### Deployment Safety

Please select one:

- [ ] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/cms/jinja2/templates/base.html
+++ b/cms/jinja2/templates/base.html
@@ -47,6 +47,7 @@
 {% set pageConfig = {
     "bodyClasses": body_class,
     "title": page_title,
+    "wide": true,
     "header": {
         "variants": "basic",
         "phase": {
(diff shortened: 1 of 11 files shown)


helenb:
Closing as this can still be used as a reference, but should not be merged.

PR #657, opened 2026-06-01
Normalise non breaking spaces in table content

BJacksonONS:

Addresses DPD-4630

Currently when copying tables from word it adds non breaking spaces to attempt to preserve formatting in other apps, but this causes undesirable behaviour when interacting with wagtail and CSV conversion.

This removes non breaking spaces and normalises them to regular ones when processing the html content of the block.

### How to review

1. Try pasting in content with non breaking spaces
2. Save and view the preview/published page
3. Inspect HTML and CSV download and ensure content is preserved as is

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

Possibly look at moving this logic up to the table block library to update the [paste_postprocess function](https://github.com/torchbox/wagtail-tinytableblock/blob/89e179b8a226ada2df3bfadbd8b71f91b59c233e/src/wagtail_tinytableblock/static/wagtail_tinytableblock/js/tiny-table-block.js#L118), or push for tinymce to address https://github.com/tinymce/tinymce/issues/9017.

### DIFF
--- a/cms/core/blocks/markup.py
+++ b/cms/core/blocks/markup.py
@@ -7,7 +7,7 @@ from django.utils.text import slugify
 from django.utils.translation import gettext_lazy as _
 from wagtail import blocks
 from wagtail.contrib.table_block.blocks import TableBlock as WagtailTableBlock
-from wagtail_tinytableblock.blocks import TinyTableBlock
+from wagtail_tinytableblock.blocks import TinyTableBlock, TinyTableFieldBlock
 
 from cms.core.utils import strip_unwanted_control_chars_from_json
 from cms.data_downloads.utils import flatten_table_data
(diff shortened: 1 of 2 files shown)


cms/core/blocks/markup.py
```diff
@@ -127,18 +127,50 @@ def render(self, value: dict, context: dict | None = None) -> str | SafeString:
         return rendered
 
 
+class NormalisingTinyTableFieldBlock(TinyTableFieldBlock):
+    """A TinyTableFieldBlock that normalises non-breaking spaces before parsing."""
+
+    @staticmethod
+    def _normalise_html(value: str) -> str:
```
zerolab, line 134: how about submitting a PR to https://github.com/torchbox/wagtail-tinytableblock instead?

BJacksonONS: Was planning on it as a followup. Wasn't sure if you'd want to handle in python or js, and have some method for allowing users to preserve non breaking spaces in some circumstances so thought I'd put a quick fix in here then look at upstream.

BJacksonONS:
There might also be an argument to move stripping the non-breaking spaces to `clean_cell_value` when exporting the csv itself, to preserve formatting in the web page.

MaciekBaron:
> There might also be an argument to move stripping the non-breaking spaces to `clean_cell_value` when exporting the csv itself, to preserve formatting in the web page.

Yeah I was about the make this point, I think sometimes editors might want to use non-breaking spaces to solve awkward formatting issues

BJacksonONS:
> * Try pasting in content with non breaking spaces
> * Save and view the preview/published page
> * Inspect HTML and CSV download and ensure content is preserved as is

Will close in favour of https://github.com/ONSdigital/dis-wagtail/pull/658 then

PR #658, opened 2026-06-02
Escape HTML entities and add BOM for correct display in excel

BJacksonONS:

Non breaking spaces are often inserted when copying tables out of Microsoft Word, or other text editors.

This strips them from CSV output to prevent malformed characters when viewing in excel.

### How to review

1. Try pasting in content with non breaking spaces
2. Save and view the preview/published page
3. Inspect HTML on output page and ensure formatting from non-breaking spaces is preserved
4. Download CSV and ensure it's using regular spaces and text hasn't been malformed

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

NA


MaciekBaron (approved):

joepercival (approved):
Ran tests locally, looks good

kevinhowbrook (approved):

cms/data_downloads/utils.py
```diff
@@ -19,6 +19,10 @@ def clean_cell_value(value: str | int | float) -> str | int | float:
     if not isinstance(value, str):
         return value
 
+    # Remove non-breaking spaces to prevent malformed characters in CSV output
+    # Do this before `strip` to ensure `&nbsp;` is also removed from the ends of strings
+    value = value.replace("\u00a0", " ").replace("&nbsp;", " ")
```
MebinAbraham: Nit pick on code comment, we aren't removing only, we are replacing/normalising.

MebinAbraham: Have you considered `html.unescape` (may have side effects, both good/bad, depending on how richtext block passes the value, i.e., it would convert `&amp;` to &, which is what we want) in CSV?

Also, nbsp can be represented in more ways than just this; it depends on where the text was copied from. https://developer.mozilla.org/en-US/docs/Glossary/Character_reference

A fuller implementation may be (unless we use unescape):
```py
value = (
    value
    .replace("\u00a0", " ")  # actual NBSP character already decoded into the string
    .replace("&nbsp;", " ")  # named HTML entity for NBSP
    .replace("&#160;", " ")  # decimal numeric HTML entity for NBSP
    .replace("&#xA0;", " ")  # hexadecimal numeric HTML entity for NBSP
    .replace("&#xa0;", " ")  # lowercase hexadecimal numeric HTML entity for NBSP
)
```

BJacksonONS: `html.unescape` by itself still leaves unicode characters that can make text malformed when viewing excel.

Good point on the other representations. It's possible if there's enough values to consider that regex might become more performant than chained replaces, will have a look.

BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/658/commits/da8d3f2b045308baa03532a9d40475e527c74430

BJacksonONS: `html.unescape` could work if we also add a BOM to indicate to excel the file is utf-8 encoded?

MaciekBaron: I think the solution would be to use both `unescape` and then `replace`:

```python
value = html.unescape(value).replace('\xa0', ' ')
```

`unescape` will convert `\u00a0`, `&nbsp;`, `&#160;` etc. into `\xa0`.

BJacksonONS: I think there may generally be an issue with unicode characters in general slipping in outside of just spaces (they're just the most common one we'd see). Explicitly marking the file as utf-8 encoded should solve most cases.

Happy to go with @MaciekBaron's suggestion if we're not concerned about that though.

MaciekBaron: I think that would be caused by malicious behaviour rather than accidental copy and paste (e.g. z̸̙͖̏̋́́̌̾̅̿͆̄̒͛̄͌̆͘͝a̶̧̝̣̙̹̯͙̫̩̬̔̿̑̽̎̈́̅̓͊͆̕l̷̢̧̛̹͈̫͐̿͂̾͛͊̆̇g̶̯͓̟̠̳̼̽͋̄̇ǫ̴̝̞̹̳͇̖̩̱̫̖̆̓̀͌͋͊ ̶̨̨̛̥͇͎̋̂̎̉̅̇̃́̓͆̇̚̚̕͝ṫ̵̨͉̪̼̯̝̟͇̜̭̱̠̓̋͆̔̉̐̓̀̀̄͘͠͠͝è̵̢̛̞̤̭̺̜̰̝͉̠͈͑͗̅͆̃͒̏̉̆̚͠x̴̢̨͚̬̗̲̱͇͚͈̩͔̤̜͖̔̽̎͗̀̔͌̍̚t̶̢͈̰͕̝̹̤͎͖̠̥̖̃̿͒̔̓͐́̚͜) so perhaps nothing to worry about in this particular case

BJacksonONS: Did a bit more testing and without the BOM we won't correctly render £, € or ¥ after `html.unescape`.

BJacksonONS: Actually let me double check, I was on the wrong branch.

BJacksonONS: Yes can replicate
image

Haven't specifically tried them yet but I also have a feeling there are some other characters like thin spaces that will have similar issues and would need to be specifically handled.

sanjeevz3009 (approved):
Apart from @MebinAbraham and @MaciekBaron comments, which are important, LGTM!

sanjeevz3009 (approved):
Apart from @MebinAbraham and @MaciekBaron comments, which are important, LGTM!

cms/data_downloads/utils.py
MebinAbraham: Let's discuss this offline. Some decisions here would need business consideration on what is expected vs what can be assumed.

BJacksonONS: Discussed with @MebinAbraham, going to check with business but the working solution for now will be to use `html.unescape` alongside a utf-8 BOM at the start of the file to ensure excel reads the file correctly. This should allow unicode characters to function correctly.

Have tested across a number of common apps, including Excel, Word, Numbers, LibreOffice and the pandas python lib that everything works either out of the box, or the app itself prompts the user by default to pick a suitable encoding.

Publish review process should hopefully catch any of @MaciekBaron's cursed text.

BJacksonONS:
Added "do-not-merge" until we receive confirmation from business.

MaciekBaron (approved):

cms/data_downloads/tests/test_utils.py
```diff
@@ -305,6 +306,13 @@ def test_strips_whitespace(self):
         self.assertEqual(clean_cell_value("  hello  "), "hello")
         self.assertEqual(clean_cell_value("\t\ntext\n\t"), "text")
 
+    def test_html_entities_unescaped(self):
+        """Test that html entites are unescaped to unicode characters."""
```
MebinAbraham:
```suggestion
        """Test that HTML entities are unescaped to unicode characters."""
```

cms/data_downloads/utils.py
```diff
@@ -19,7 +21,7 @@ def clean_cell_value(value: str | int | float) -> str | int | float:
     if not isinstance(value, str):
         return value
 
-    value = value.strip()
+    value = html.unescape(value).strip()
 
     # Replace <br> tags (all variations) with newlines
     value = re.sub(r"<br\s*/?>", "\n", value, flags=re.IGNORECASE)
```
MebinAbraham: The regex should likely be first but I can't see any real cases it would be an issue based on content we published (i.e. converting escaped br into newlines)

MebinAbraham (approved):

cms/data_downloads/utils.py
BJacksonONS: Yeah checked and there are a couple situations, like `&lt;br/&gt;` that would end up replacing what should be literal text with a newline, so have swapped. Not sure how likely that would be in what ONS publishes but swapped to be safe.

https://github.com/ONSdigital/dis-wagtail/pull/658/commits/a0f1a3908d914d74096c861878c01f2e3639678f

cms/data_downloads/tests/test_utils.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/658/commits/a0f1a3908d914d74096c861878c01f2e3639678f

MebinAbraham (approved):

PR #660, opened 2026-06-02
Last editor can request changes on their own work

kevinhowbrook:

Ticket [CMS-1231](https://officefornationalstatistics.atlassian.net/browse/CMS-1231). 

When a Publishing Officer submitted a page for release review and then immediately clicked "Request changes" on their own submission, two problems occurred: the option was available despite them being the last editor, and no email notification was triggered.

###  How to review

(locally, set EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend" in local.py for emails to log to console)

1. Log in as a Publishing Officer.
2. Create or edit a Topic/Methodology page and submit it for release review.
3. While still logged in as that PO, open the page, confirm no workflow actions appear in the action menu (previously "Request changes" was shown).
4. Log in as a different Publishing Admin/Officer, open the same page — confirm "Request changes", "Approve", and "Approve with comment" are all still available.
5. As the second user, click "Request changes", confirm the submitting PO receives a rejection email.

###  Deployment Safety

  - [x] Safe to auto-deploy

###  Follow-up Actions

  None.


MaciekBaron (approved):
Makes sense and works 👍

Non-blocking thought: Currently the message says "Cannot self-approve your changes". Now that this also handles self-rejection, should we reword this so that it says something like "Cannot review your own changes"?

cms/workflows/wagtail_hooks.py
```diff
@@ -123,7 +123,11 @@ def before_edit_page(request: HttpRequest, page: Page) -> HttpResponse | None:
...
         action_name = request.POST.get("workflow-action-name", "")
 
-        if action_name == "approve" and page.latest_revision and page.latest_revision.user_id == request.user.pk:
+        if (
+            action_name in ("approve", "reject")
+            and page.latest_revision
+            and page.latest_revision.user_id == request.user.pk
+        ):
             messages.error(
                 request, "Cannot self-approve your changes. Please ask another Publishing team member to do so."
```
MebinAbraham: Agree with Maciej, let's update this for clarity.
```suggestion
                request, "You cannot review your own changes. Please ask another Publishing team member to do so."
```

MebinAbraham (approved):

cms/workflows/wagtail_hooks.py
kevinhowbrook: applied be06258766bc22bcd43ecfe8f561af6beadf8fdd

MaciekBaron (approved):

cms/workflows/tests/test_workflow_tweaks.py
```diff
@@ -223,6 +224,25 @@ def test_before_edit_page(self):
...
+
+        workflow_state = mark_page_as_ready_for_review(self.page, self.publishing_admin)
+        latest_revision = self.page.latest_revision
+
+        data = self.get_simple_post_data(self.page)
+        data["workflow-action-name"] = "reject"
+        response = self.client.post(self.edit_url, data, follow=True)
+
+        self.assertContains(
+            response, "Cannot self-approve your changes. Please ask another Publishing team member to do so."
```
MebinAbraham: Will need updating.
```suggestion
            response, "You cannot review your own changes. Please ask another Publishing team member to do so."
```

MebinAbraham (approved):

BJacksonONS (approved):

sanjeevz3009 (approved):
LGTM!

@kevinhowbrook We can just use our `mailpit` server to test the emails, btw, https://github.com/ONSdigital/dis-wagtail#mailpit-email-testing.

Another thing to keep in mind now is that the last editor for a page cannot request any changes; they aren't able to approve, approve with comment or request changes on the pages they just worked on, which is what we want.

There is a misleading behaviour in one of the panels in the homepage where the page they just worked on, which they can't do anything with in terms of approving or requesting changes, is in the `Awaiting your review` panel, where truly the current user/ last editor is not meant to or responsible for reviewing that page. This might be a problem as there are many more pages getting built by the business, and it might add to the noise, depending on the volume of pages that get created at one time from different users' POs, PAs, etc.

Something to keep in mind, but not a blocker for this PR.

Image

MebinAbraham:
> LGTM!
> 
> @kevinhowbrook We can just use our `mailpit` server to test the emails, btw, [ONSdigital/dis-wagtail#mailpit-email-testing](https://github.com/ONSdigital/dis-wagtail#mailpit-email-testing).
> 
> Another thing to keep in mind now is that the last editor for a page cannot request any changes; they aren't able to approve, approve with comment or request changes on the pages they just worked on, which is what we want.
> 
> There is a misleading behaviour in one of the panels in the homepage where the page they just worked on, which they can't do anything with in terms of approving or requesting changes, is in the `Awaiting your review` panel, where truly the current user/ last editor is not meant to or responsible for reviewing that page. This might be a problem as there are many more pages getting built by the business, and it might add to the noise, depending on the volume of pages that get created at one time from different users' POs, PAs, etc.
> 
> Something to keep in mind, but not a blocker for this PR.
> 
> Image

Can you raise a ticket @sanjeevz3009. We can follow up.

PR #662, opened 2026-06-03
Add tick interval options for column charts and column confidence charts

helenb:

See https://officefornationalstatistics.atlassian.net/browse/DPD-2610

This PR introduces the option to configure the mobile and desktop tick interval for column charts, and column charts with confidence intervals.

Because column and bar charts share the same interface, validation is introduced to prevent this from being configured for bar charts. This follows the same pattern as other options that only apply to one or the other.

### How to review

- Create a column chart with some dummy data
- Try changing the tick interval for the category axis at both desktop and mobile, and observe the change in the way the chart is rendered in the preview
- Save the changes
- Switch the chart type to be a bar chart
- Observe the message that the preview is out of date and validation issues need to be resolved
- Try to save and see the validation error
- Repeat for a bar / column chart with confidence intervals

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


helenb:
Removed 'do not merge' label as my approach has been validated.

cms/datavis/tests/test_bar_column_chart_block.py
```diff
@@ -400,14 +400,27 @@ def test_value_axis_tick_interval(self):
...
+        self.raw_data["select_chart_type"] = BarColumnChartTypeChoices.COLUMN
         self.raw_data["x_axis"]["tick_interval_mobile"] = 5
         self.raw_data["x_axis"]["tick_interval_desktop"] = 6
-        self.block.clean(self.get_value())
-        x_axis_config = self.get_value().block.get_component_config(self.get_value())["xAxis"]
-        for key in ["tickIntervalMobile", "tickIntervalDesktop"]:
-            with self.subTest(key=key):
-                self.assertNotIn(key, x_axis_config)
+        try:
+            self.block.clean(self.get_value())
```
kevinhowbrook, line 416: **Suggestion**

After this clean() call, shouldn't we still call get_component_config() and assert both keys and values in xAxis? Mostly wondering about the removal of the get_component_config here

helenb: The previous test was ensuring that we didn't have tick intervals in the config for the x-axis - so the new test is just checking that now if they are included, there is no validation error. But yes perhaps we should also test that the config is correct - I'll amend.

helenb: I've updated the tick interval tests now so that the validation testing is separate from the config testing.

kevinhowbrook (approved):

BJacksonONS:
Should this be restricted to integers? Looks odd on certain types of chart at least.
image

helenb:
> Should this be restricted to integers? Looks odd on certain types of chart at least. 

I've passed this question on to Erin in the datavis team, as she raised the ticket. We allow decimals for tick intervals on the value access, but I can see that it might make sense to limit it to whole numbers for the category access.

helenb:
> > Should this be restricted to integers? Looks odd on certain types of chart at least.
> 
> I've passed this question on to Erin in the datavis team, as she raised the ticket. We allow decimals for tick intervals on the value access, but I can see that it might make sense to limit it to whole numbers for the category access.

Erin's confirmed they should be whole numbers on the category access so I'll amend that now.

cms/datavis/blocks/charts.py
```diff
@@ -145,6 +146,26 @@ class BarColumnChartBlock(BaseChartBlock):
...
                 ),
             ),
+            (
+                "tick_interval_mobile",
+                TextInputIntegerBlock(
+                    label="Tick interval (mobile)",
+                    required=False,
+                    help_text="""
+                        Category axis tick intervals are available for column charts only, and must use whole numbers.
+                    """,
```
MebinAbraham: Any reason these are all triple quotes?.

cms/datavis/tests/test_bar_column_chart_block.py
```diff
@@ -400,14 +400,47 @@ def test_value_axis_tick_interval(self):
...
+    def test_bar_chart_category_axis_has_no_tick_interval(self):
+        self.raw_data["select_chart_type"] = BarColumnChartTypeChoices.BAR
+        self.raw_data["x_axis"]["tick_interval_mobile"] = 5
+        self.raw_data["x_axis"]["tick_interval_desktop"] = 6
+        with self.assertRaises(blocks.StructBlockValidationError):
+            self.block.clean(self.get_value())
+            x_axis_config = self.get_value().block.get_component_config(self.get_value())["xAxis"]
+            for key in ["tickIntervalMobile", "tickIntervalDesktop"]:
+                with self.subTest(key=key):
+                    self.assertNotIn(key, x_axis_config)
```
MebinAbraham: These won't execute as `self.block.clean(self.get_value())` would raise before it got here. This test is a false positive per-se.

cms/datavis/blocks/charts.py
```diff
@@ -258,18 +279,28 @@ def validate_series_customisation(self, value: StructValue) -> StructValue:
...
-                    "x_axis": blocks.StructBlockValidationError(
-                        block_errors={
-                            "title": ValidationError(
-                                "Category axis title is not supported for horizontal bar charts.",
-                                code=self.ERROR_HORIZONTAL_BAR_NO_CATEGORY_TITLE,
-                            )
-                        }
-                    )
-                }
+                {"x_axis": blocks.StructBlockValidationError(block_errors=x_axis_errors)}
```
MebinAbraham: Nit: This is basically duplicate logic; it would be good to have a helper to dedupe so validation logic is in a single place

cms/datavis/tests/test_bar_column_confidence_interval_chart_block.py
```diff
@@ -298,6 +298,48 @@ def test_column_chart_blank_x_axis_title(self):
...
         self.assertNotIn("title", config["xAxis"])
 
+    def test_column_chart_category_axis_tick_interval_supported(self):
+        self.raw_data["select_chart_type"] = BarColumnConfidenceIntervalChartTypeChoices.COLUMN
+        self.raw_data["x_axis"]["tick_interval_mobile"] = 5
+        self.raw_data["x_axis"]["tick_interval_desktop"] = 6
+        x_axis_config = self.get_value().block.get_component_config(self.get_value())["xAxis"]
+        for key in ["tickIntervalMobile", "tickIntervalDesktop"]:
+            with self.subTest(key=key):
+                self.assertIn(key, x_axis_config)
```
MebinAbraham: Suggestion: A mapping bug could emit the keys with incorrect values, and these tests would still pass. I suggest that the values are also asserted.

cms/datavis/tests/test_bar_column_confidence_interval_chart_block.py
```diff
@@ -298,6 +298,48 @@ def test_column_chart_blank_x_axis_title(self):
...
+    def test_bar_chart_category_axis_has_no_tick_interval(self):
+        self.raw_data["select_chart_type"] = BarColumnConfidenceIntervalChartTypeChoices.BAR
+        self.raw_data["x_axis"]["tick_interval_mobile"] = 5
+        self.raw_data["x_axis"]["tick_interval_desktop"] = 6
+        with self.assertRaises(blocks.StructBlockValidationError):
+            self.block.clean(self.get_value())
+            x_axis_config = self.get_value().block.get_component_config(self.get_value())["xAxis"]
+            for key in ["tickIntervalMobile", "tickIntervalDesktop"]:
+                with self.subTest(key=key):
+                    self.assertNotIn(key, x_axis_config)
```
MaciekBaron: I think you perhaps intended to have this block not indented? This code is within the `with self.assertRaises` block, but you have an assertion at the end. I think Mebin raised something similar above.

cms/datavis/tests/test_bar_column_confidence_interval_chart_block.py
```diff
@@ -298,6 +298,48 @@ def test_column_chart_blank_x_axis_title(self):
...
+            self.fail("Expected no ValidationError for column chart category axis tick interval")
+
+    def test_bar_chart_category_axis_has_no_tick_interval(self):
+        self.raw_data["select_chart_type"] = BarColumnConfidenceIntervalChartTypeChoices.BAR
+        self.raw_data["x_axis"]["tick_interval_mobile"] = 5
+        self.raw_data["x_axis"]["tick_interval_desktop"] = 6
+        with self.assertRaises(blocks.StructBlockValidationError):
+            self.block.clean(self.get_value())
+            x_axis_config = self.get_value().block.get_component_config(self.get_value())["xAxis"]
+            for key in ["tickIntervalMobile", "tickIntervalDesktop"]:
```
MaciekBaron: Really minor but you use a list `[]` here and on line 336 you use a tuple `()`. Tuples are more idiomatic, would be good to keep things consistent.

MaciekBaron (changes requested):

cms/datavis/tests/test_bar_column_confidence_interval_chart_block.py
helenb: I was trying to test if the config is clean after the validation error is raised, but maybe I don't need to do that? If I move the block outwards it then fails because the config *does* have the `tickIntervalMobile` and `tickIntervalDesktop`. Any advice? Maybe I should just remove this test?

helenb: I'm going to remove this test after further investigation as it isn't meaningful.

cms/datavis/tests/test_bar_column_chart_block.py
helenb: It seems the tests I added were redundant - see comments below - @kevinhowbrook could you check again to make sure you're happy with the tests that remain?

cms/datavis/tests/test_bar_column_chart_block.py
helenb: I've now removed it.

cms/datavis/tests/test_bar_column_confidence_interval_chart_block.py
helenb: Added in fcac676fc6fee4edfd72e0f7b09fca63e2ce7d2d

cms/datavis/tests/test_bar_column_confidence_interval_chart_block.py
helenb: This code has changed in fcac676fc6fee4edfd72e0f7b09fca63e2ce7d2d so the list is no longer there.

cms/datavis/tests/test_bar_column_confidence_interval_chart_block.py
helenb: Removed the whole test in fcac676fc6fee4edfd72e0f7b09fca63e2ce7d2d

cms/datavis/blocks/charts.py
helenb: Updated in 36164ef5f383c1b99a5e999065b9f7e29644c671

cms/datavis/blocks/charts.py
helenb: Addressed in 83ec67c1a5961bc2b7c057c429a58dda4fd487b8

MaciekBaron (approved):

cms/datavis/blocks/charts.py
```diff
@@ -146,6 +175,26 @@ class BarColumnChartBlock(BaseChartBlock):
...
+                ),
+            ),
+            (
+                "tick_interval_desktop",
+                TextInputIntegerBlock(
+                    label="Tick interval (desktop)",
+                    required=False,
+                    help_text=(
+                        "Category axis tick intervals are available for column charts only, and must use whole numbers."
+                    ),
```
MebinAbraham: Nit: make the help text a constant and reuse since it is the same. Others do already.

cms/datavis/blocks/charts.py
```diff
@@ -400,7 +445,29 @@ class BarColumnConfidenceIntervalChartBlock(BaseChartBlock):
...
+                ),
+            ),
+            (
+                "tick_interval_desktop",
+                TextInputIntegerBlock(
+                    label="Tick interval (desktop)",
+                    required=False,
+                    help_text=(
+                        "Category axis tick intervals are available for column charts only, and must use whole numbers."
```
MebinAbraham: Same as above, use a constant?

MebinAbraham (approved):

BJacksonONS (approved):
+1 for Mebin's comments but otherwise LGTM

cms/datavis/blocks/charts.py
helenb: Done.

cms/datavis/blocks/charts.py
helenb: Done.

MebinAbraham (approved):

kevinhowbrook (approved):

sanjeevz3009 (commented):
Reviewed!

PR #668, opened 2026-06-08
fix(bundles): use API dataset URLs with topics

MebinAbraham:

Dataset URLs now include the topic path, for example, `/retail-industry/datasets/....`. The Data team fixed the API response so Wagtail receives the correct topic-aware `links.preview` URL for dataset preview. This means the "Preview" button URL now automatically includes the topic and no longer 404s in Sandbox. However, Wagtail was still manually constructing the published dataset Live URL, which dropped the topic path and caused published dataset links to 404. This PR updates Wagtail to use the Bundle API-provided dataset links instead, preserving the topic path.

This change also removes links from the content item payload Wagtail sends to the Bundle API, because those links are now [read-only](https://github.com/ONSdigital/dis-bundle-api/blob/develop/swagger.yaml#L668) and should be provided by the API response. (This should have always been the case, but we were waiting for the API team to fix this.

Tests and functional test mocks have been updated to use topic-prefixed relative dataset URLs.

### How to review

Hard to functionally test, as we don't have the Bundle API in our stack; therefore, ensure the existing tests have been corrected and updated so that:
- Wagtail no longer sends links in Bundle API content item payloads.
- Dataset preview iframe uses the Bundle API `links.preview` value. (Existing, no regression!)
- Published dataset View Live uses the Bundle API `links.preview` value, including the topic path.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


cms/bundles/tests/test_viewsets.py
```diff
@@ -773,8 +776,10 @@ def test_inspect_view__contains_datasets(self, mock_api_client):
                     },
                     "state": "APPROVED",
                     "links": {
-                        "edit": f"/data-admin/series/{dataset_id_b}/editions/{edition_b}/versions/{version_b}",
-                        "preview": f"/datasets/{dataset_id_b}/editions/{edition_b}/versions/{version_b}",
+                        "edit": expected_edit_url_b,
+                        "preview": (
+                            f"/retail-industry/datasets/{dataset_id_b}/editions/{edition_b}/versions/{version_b}"
```
MaciekBaron, line 781: Is there a reason why we didn't also create a `expected_view_url_b`?

MebinAbraham: There was only this single occurrence, and it wasn't reused in the assertion like `expected_view_url_a`

MaciekBaron (approved):

cms/bundles/viewsets/bundle.py
```diff
@@ -587,10 +587,11 @@ def _extract_api_fields(api_item: dict[str, Any] | None) -> tuple[str, str, str
     def _build_action_button(self, state: str, preview_url: str | None, dataset: Dataset) -> SafeString | str:
         """Build the action button HTML based on dataset state."""
         if state == BundleContentItemState.PUBLISHED:
-            view_url = get_data_admin_action_url("preview", dataset.namespace, dataset.edition, str(dataset.version))
+            if not preview_url:
+                return ""
```
kevinhowbrook, line 594: **nitpick, sort of**

Given this is new logic, do you think a test that passes preview_url=None with a PUBLISHED state and asserts no button is returned would be worthwhile?

kevinhowbrook (approved):

BJacksonONS (approved):

cms/bundles/viewsets/bundle.py
MebinAbraham: Good point https://github.com/ONSdigital/dis-wagtail/pull/668/commits/4ce4622053820ce0c93adaded401c60e1cd1d1f0

kevinhowbrook (approved):

sanjeevz3009 (commented):
LGTM!

PR #671, opened 2026-06-09
Upgrade DS version to 73.7.0

helenb:
See https://officefornationalstatistics.atlassian.net/browse/CMS-1241

Upgrades the DS version which includes some good wins in terms of accessibility and performance:

- The line height is now responsive if the font-size is changed in the browser or OS settings, which is a big win in terms of readability
- The charts JavaScript is now bundled separately to the main JavaScript, meaning we don't load it for any pages without charts

See the releases for full accessibility improvements and other enhancements

### How to review

- Try changing your browser font-size settings to the largest possible (chrome://settings/?search=size to change this in chrome). The line height on any page should now proportionally increase
- Visit a page with a chart - in the network tab you should see `chart.js` and `main.js` loading separately
- Visit a page without a chart - in the network tab you should only see `main.js`
- If you compare the size of `main.js` with another branch, you should see that it is reduced.
- If you run an automated accessibility checker, such as Axe, on any page, you should not see any issues about 'All page content should be contained by landmarks'.

Note: there is a `main.js` both from the CMS build and from the DS - it is the DS one that we are comparing.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


MaciekBaron (approved):

BJacksonONS (approved):

joepercival:
I think the latest release of the Design system is now 73.8.0, is it worth bumping the version here?

helenb:
> I think the latest release of the Design system is now 73.8.0, is it worth bumping the version here?

See my comment on the ticket - there's currently an issue with the static assets not loading for 73.8.0. @sveltifier advised updating to 73.7.0 instead.

joepercival (approved):

sanjeevz3009 (commented):
LGTM!

PR #673, opened 2026-06-10
chore(npm packages): update shell-quote and uuid package versions

BJacksonONS:

OSV-Scanner flagged a vulnerability in shell-quote yesterday. This updates to a fixed version.

We previously ignored an issue in the uuid package as a transitive dependency. This forces an override to use v11.

### How to review

- Give a quick skim through https://github.com/uuidjs/uuid/blob/main/CHANGELOG.md. Most of the breaking changes are around browser and node versions supported but I might have missed something.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


MebinAbraham (approved):

sanjeevz3009 (approved):
LGTM!

PR #675, opened 2026-06-10
Enable Wagtail autosave

BJacksonONS:

Addresses [CMS-1227](https://officefornationalstatistics.atlassian.net/browse/CMS-1227).

Enables autosave introduced in wagtail 7.3.

Uses the default wagtail value of 500ms delay after input, but allows for overriding or dissabling via environment variable.

### How to review

- Test under normal development environment
- Test under cognito auth scheme. Follow documentation in [auth.md](https://github.com/ONSdigital/dis-wagtail/blob/main/docs/integrations/auth.md) on how to set up. ACs are on the ticket but things to check for 
    - If tokens are missing/expired users won't see any missleading saved state
    - Autosave is correctly handled if a users token is refreshed mid-edit

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


docs/integrations/auth.md
```diff
@@ -46,7 +46,7 @@ TODO: add details about the middleware, auth.js, ONSLogoutView and the team sync
 4. Install node dependencies by running `npm install`, then `npm run build:prod `
```
MebinAbraham, line 46: I suggest we update this doc to say we have verified the compatibility of this with auto-save functionality and that everything was as expected. In case something changes in future, and we wonder whether if it ever was tested.

BJacksonONS: [18b79b5](https://github.com/ONSdigital/dis-wagtail/pull/675/commits/18b79b51ef32eb207b8ebe6d30295ade1c30d260)

MebinAbraham (changes requested):
Found an edge case that we should review, and it's worth checking whether there are any similar scenarios elsewhere. It's not an autosave functionality issue nor exclusively an issue with auto save itself, but auto save makes the issue more prominent.

We currently prevent self-approval by disallowing the last editor from approving a page. However, this logic does not behave as intended when autosave is enabled.

Example:

1. User A submits Page A for review.
2. User B opens the page and types something, triggering an autosave (without explicitly clicking Save).
3. User A can now approve the page because User B is technically recorded as the last editor, even if their change was accidental, temporary, or immediately removed.

This effectively allows User A to approve their own submission, which is what the restriction was intended to prevent.

The interesting part is that autosave is really exposing an issue that already exists. Even without autosave, if User B were to open the page and click Save without making any meaningful changes, User A would also become eligible to approve. Auto-save makes this more likely to happen unintentionally.

One possible solution would be to prevent the original submitter and last editor from approving, regardless of subsequent edits. This would be closer to how GitHub reviews work, where the author cannot approve their own changes. However, this would be a change from current behaviour and expectations, so it would need discussion with the business.

We could prevent both the original submitter and the last editor from approving, although this has its own implications. Another option is submitting for review should "lock" the page so this can't accidentally happen.

Have a think about any alternative approaches or edge cases we should consider. Happy to discuss.

MaciekBaron (approved):

MaciekBaron:
Question: Should we disable this for functional tests to avoid any flakiness?

BJacksonONS:
> Question: Should we disable this for functional tests to avoid any flakiness?

Was debating this myself, but I also don't think any of the tests would ever be on a given screen for more than 500ms, excepting some of the cognito auth timeout tests.

BJacksonONS:
Added "do-not-merge" until we have decision from the business around workflows.

kevinhowbrook (approved):
Aside from the points raised, this lgtm.

I was curious how this was going to act with [Expand activity logs](https://github.com/ONSdigital/dis-wagtail/pull/446) - so I merged them both locally. It's something to be mindful of because a save of a single field via autosave triggers a lot (rather than 1 page save) so there is a risk of 'overlogging' here perhaps. I'll bring it up in our next tech session for discussion.

MebinAbraham (approved):
Let's get this into Bleed/Sandbox. Let's disable in Staging/Prod for now. We need workflow changes before we enable this in Staging/Prod, or need explicit sign off on the gaps.

MebinAbraham:
> Aside from the points raised, this lgtm.
> 
> I was curious how this was going to act with [Expand activity logs](https://github.com/ONSdigital/dis-wagtail/pull/446) - so I merged them both locally. It's something to be mindful of because a save of a single field via autosave triggers a lot (rather than 1 page save) so there is a risk of 'overlogging' here perhaps. I'll bring it up in our next tech session for discussion.

As discussed, we'll explore with Legal whether they are happy for us to have a cooldown, given the revision history still provides the source of truth. If they are not happy, we will just have to accept it is noisy and storage needs.

BJacksonONS:
> > Aside from the points raised, this lgtm.
> > I was curious how this was going to act with [Expand activity logs](https://github.com/ONSdigital/dis-wagtail/pull/446) - so I merged them both locally. It's something to be mindful of because a save of a single field via autosave triggers a lot (rather than 1 page save) so there is a risk of 'overlogging' here perhaps. I'll bring it up in our next tech session for discussion.
> 
> As discussed, we'll explore with Legal whether they are happy for us to have a cooldown, given the revision history still provides the source of truth. If they are not happy, we will just have to accept it is noisy and storage needs.

I can add docs somewhere (in internal or part of the repo) but to summarise here, revision history keeps a compressed diff of all autosaves made during an editing session. This makes it impossible for someone to manipulate autosave to hide their activity.

We might consider introducing a cooldown to edits for the activity log changes mentioned by @kevinhowbrook, given it might get quite noisy. The revision history provides a source of truth so we aren't too concerned with people being able to make edits without an audit trail.

sanjeevz3009 (commented):
LGTM!

PR #504, opened 2026-02-17
Slack Notifications - Bundle Publishing flows - Part 1

MaciekBaron:
### What is the context of this PR?

Loom Demo Video: https://www.loom.com/share/8ea9acc7e03f4faeaae636253e400e11

This PR related to CMS-366 and expands our Slack notification system. It covers the bundles ACs from the ticket, which are listed below.

Please not that the "failed to under pre-publish state" AC seems to be redundant, as discussed with Mebin, and therefore has not been addressed.

This change also includes adding two new bundle statuses, `FAILED` and `PARTIALLY_PUBLISHED`. These statuses behave similarly to `PUBLISHED`, being immutable end states for the bundle, but represent failures in the bundle pubish.
Commit [fe89d8e](https://github.com/ONSdigital/dis-wagtail/pull/504/commits/fe89d8e9eb23ac63122e496a2825235e99b6f4e6) handles these new statuses in the admin UI, filtering them out from the default bundle listing and redirecting the edit URL (as we do for published bundles).
Commit [0a0f7f7](https://github.com/ONSdigital/dis-wagtail/pull/504/commits/0a0f7f78539bbc6258b1d8efd48e62b349e8d7d3) skips syncing these new bundle statuses with the dataset bundle API.

### AC Checklist

- [x] Scheduled Bundle Pre-Publish Alert via Slack
- [ ] ~Scheduled Page Pre-Publish Alert via Slack~
- [x] Slack Alert: Bundle Publishing Commencement
- [ ] ~Slack Alert: Page Publishing Commencement~
- [x] Slack Alert: Bundle Publishing Finished Successfully
- [ ] ~Slack Alert: Page Publishing Finished Successfully~
- [ ] Bundle failed to enter “Pre-publish state" Alarm **(may be redundant?)**
- [x] Bundle Content Publication Failure (unexpected issues) Alarm
- [ ] ~Page Publication Failure  (unexpected issues) Alarm~
- [ ] ~Slack Notification for Technical Publishing Errors Alarm~ 
- [x] Slack Alert: Dataset API Call Failed
- [x] Third party API integration fail (generic) - currently used for Bundle API

### How to review

1. Ensure that all changes cover the bundle ACs listed in the ticket (with the exception noted above)
2. Create a Slack workspace and/or provide a working Slack bot API
3. Perform actions which trigger the messages
4. Confirm the messages are sent as expected

### Follow-up Actions

- [ ] Complete ACs related to pages
[...]

PR #676, opened 2026-06-11
feat(bundles): add parallel post-publish actions for bundle publishing

RealOrangeOne:

Some actions as part of publishing need to be done after Wagtail has published the page, but don't impact the publish itself (Eg S3 ACLs, search index etc)..

This PR moves these actions into a thread pool to be run concurrently. The status of these actions is stored in the DB for future querying. An extra Slack message is added once all of these actions are done to report status.

A timeout is also introduced. If the actions take longer than this timeout, an error is reported. 

Because we're now using threads, I've added a pool to prevent consuming too many connections and reducing churn.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

A few things are intentionally out of scope:

- Improvements to the Slack message to include which actions failed / live actions status etc
- Run actions concurrently for non-bundle publishes

---

Ticket: [CMS-1226](https://officefornationalstatistics.atlassian.net/browse/CMS-1226)

[CMS-1226]: https://officefornationalstatistics.atlassian.net/browse/CMS-1226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RealOrangeOne:
("do not merge" label added to show this is for an architecture review rather than being fully ready to go)

cms/post_publish_actions/executor.py
```diff
@@ -0,0 +1,115 @@
...
+                "action_type": action_type.value,
+                "event": "post_publish_action_failed",
+                "duration": duration,
+            },
+        )
+        action.status = PostPublishActionStatus.FAILED
+        action.duration = timedelta(seconds=duration)
+        action.finished_at = timezone.now()
+        action.failed_reason = traceback.format_exception_only(e)[0].strip()
+        action.save(update_fields=["status", "finished_at", "fail_reason", "duration"])
```
BJacksonONS: Typo here, should be `failed_reason`

cms/bundles/viewsets/bundle.py
```diff
@@ -246,14 +247,49 @@ def save_instance(self) -> Bundle:
...
         """This method allows calling hooks or additional logic after an action has been executed.
 
         In our case, we want to replicate the scheduled publication (send Slack notification, publish pages, update RC).
         """
         if self.action == "publish" or (self.action == "edit" and self.object.status == BundleStatus.PUBLISHED):
+            start_time = time.monotonic()
+            start_date = timezone.now()
             publish_bundle(self.object, update_status=False)
 
+            while (time.monotonic() - start_time) <= settings.BUNDLE_POST_PUBLISH_TIMEOUT_SECONDS:
```
BJacksonONS: Can't this get hit by a user from a manual bundle publish?

Is there a risk that in the worst case scenario the user request is timed out by a proxy before we get to the 110s timeout limit?

I think one of the acceptance criteria is that UI bundle publishes shouldn't have to wait for side effects for a response, so seems like this should be handled differently.

BJacksonONS:
I've not fully followed this until now, and maybe I'm just being pedantic but "parallel bundle publishing" could potentially be interpreted as bundles can be published in parallel, with each bundle in it's own background task.

Would be great to document somewhere explicitly that this queues x side effects per y pages, for a total of x*y tasks. There's no batching of tasks for all the pages in a bundle into one background task.

cms/bundles/viewsets/bundle.py
RealOrangeOne: Ah yes good point. The only issue being we'd need something which runs after all the threads to send the final Slack update. That's not really possible with threads without something else in the background keeping track.

cms/bundles/notifications/slack.py
```diff
@@ -259,19 +258,132 @@ def notify_slack_of_publish_end(
     )
 
     if bundle_page_count > 0:
-        fields.append({"title": "Pages Published", "value": str(pages_published), "short": True})
+        fields.append(
+            {"title": "Pages Published", "value": str(bundle.get_bundled_pages().public().count()), "short": True}
+        )
```
MebinAbraham, line 283: Issue: We should keep passing the explicit `pages_published` count here rather than deriving it with `bundle.get_bundled_pages().public().count()`. 

`public()` tells us whether a bundled page currently has a live version, not whether this publish attempt successfully published the target revision. That can give the wrong result for retries or for bundles containing pages that are already live: the page may be counted as public even if the latest revision in this bundle failed to publish.

For the failure Slack message, the useful signal is “how many pages did this publish run successfully publish?".

cms/bundles/viewsets/bundle.py
MebinAbraham: How about running the entire thing in the executor?. And have 2 separate executors to avoid starvation?

cms/bundles/viewsets/bundle.py
```diff
@@ -246,14 +247,49 @@ def save_instance(self) -> Bundle:
...
+                        "outstanding_actions": outstanding_actions,
+                    },
+                )
+
+            run_in_executor(
+                notify_slack_of_post_publish_end,
+                bundle=self.object,
+                start_time=start_date,
+                end_time=timezone.now(),
+            )
```
MebinAbraham: One thing to consider is that we mark the action as failed (due to timeout), but the background tasks are not terminated and may still complete and update the DB even after the initial failure mark. So we end up with the DB saying Failed > Slack message saying failed > DB on investigation saying Success(or fail). This is not great for support teams.

It would be good to differentiate whether things completed after the timeout. I don't suggest we only update the record if it isn't already marked as finished; metadata is helpful for support. I wonder if we should have a previous_status and previous_finished_at? Something like that?.

Might be helped in future by status updates of each action in slack.

cms/post_publish_actions/models.py
```diff
@@ -0,0 +1,67 @@
...
+class PostPublishActionType(models.TextChoices):
+    S3_ACL = "S3_ACL", _("Private Media ACLs")
+    SEARCH_UPDATED = "SEARCH_UPDATED", _("Search updated")
+
+
+class PostPublishActionStatus(models.TextChoices):
+    READY = ("READY", _("Ready"))
+    RUNNING = ("RUNNING", _("Running"))
+    FAILED = ("FAILED", _("Failed"))
+    SUCCESSFUL = ("SUCCESSFUL", _("Successful"))
```
MebinAbraham: Suggestion: It would be clearer to differentiate a timeout explicitly with a `TIMED_OUT` status. Treat failed as a job has errors, and timeout as the job couldn't complete in time. Timeout is not the same as the underlying operation failing. S3, search, or CF may complete after the SLA has been breached.

Alternatively, keep using `FAILED` but introduce `timed_out_at` on `PostPublishAction`.

cms/post_publish_actions/handlers.py
```diff
@@ -0,0 +1,48 @@
...
+
+    PostPublishAction.objects.update_or_create(
+        page=page,
+        bundle=bundle,
+        action_type=action_type,
+        defaults={
+            "status": PostPublishActionStatus.READY,
+            "finished_at": None,
+        },
+    )
```
MebinAbraham: This would still keep old "failed_reason" and "duration". It's not fully reset. `enqueued_at` is also probalematic for reusing.

cms/settings/base.py
```diff
@@ -233,14 +234,8 @@
 
 AWS_REGION = env.get("AWS_REGION")
 
-
 # Database
 
-# None allows connections to be reused for longer, since opening them is expensive.
-# CONN_HEALTH_CHECK ensures they're still healthy before attempting to use them.
-db_conn_max_age = int(env["PG_CONN_MAX_AGE"]) if "PG_CONN_MAX_AGE" in env else None
-db_read_conn_max_age = int(env["PG_READ_CONN_MAX_AGE"]) if "PG_READ_CONN_MAX_AGE" in env else None
```
MebinAbraham, line 255: Q: how come we dropped these?

cms/settings/base.py
```diff
@@ -280,13 +274,21 @@
...
 
 if "read_replica" not in DATABASES:
     DATABASES["read_replica"] = deepcopy(DATABASES["default"])
 
+# Configure a connection pool to reduce connections when using threads.
+# The numbers are intentionally low, as contention on connections shouldn't be high.
+db_pool_options = {
+    "min_size": 1,
+    "max_size": 3,
+}
```
MebinAbraham, line 314: Is this best guess or?.

cms/settings/base.py
```diff
@@ -983,6 +985,10 @@
 # Feature flag for sending slack messages on bundle status changes (before pre-publish, e.g. entering review)
 SLACK_NOTIFY_ON_BUNDLE_STATUS_CHANGE = env.get("SLACK_NOTIFY_ON_BUNDLE_STATUS_CHANGE", "false").lower() == "true"
 
+BUNDLE_POST_PUBLISH_TIMEOUT_SECONDS = float(env.get("BUNDLE_POST_PUBLISH_TIMEOUT_SECONDS", 110))
```
MebinAbraham, line 972: Worth add a comment on why 110.

cms/settings/base.py
```diff
@@ -280,13 +274,21 @@
...
+        DATABASES["read_replica"] = dj_database_url.config(env="READ_REPLICA_DATABASE_URL", conn_max_age=0)
 
 if "read_replica" not in DATABASES:
     DATABASES["read_replica"] = deepcopy(DATABASES["default"])
 
+# Configure a connection pool to reduce connections when using threads.
+# The numbers are intentionally low, as contention on connections shouldn't be high.
+db_pool_options = {
+    "min_size": 1,
+    "max_size": 3,
```
MebinAbraham: Q: Do we want this to be atleast `BUNDLE_POST_PUBLISH_CONCURRENCY`, they all do db work?

MebinAbraham: Would be good if these were env var controlled as we can tune these per env as we scale envs / db.

cms/post_publish_actions/executor.py
```diff
@@ -0,0 +1,115 @@
...
+    action_type: PostPublishActionType,
+    page_id: int,
+    bundle_id: int | None,
+) -> None:
+    # TODO: Support page-only publishes
+    if bundle_id is None:
+        raise RuntimeError("Bundle id required")
+
+    action = PostPublishAction.objects.get(page_id=page_id, bundle_id=bundle_id, action_type=action_type)
+    action.status = PostPublishActionStatus.RUNNING
```
MebinAbraham, line 69: Even though in-process execution is accepted for now, we know stale rows are inevitable, process exits, deploys, network hangs etcs which would leave stale hanging actions. I wonder if we need a cleanup management command.

cms/private_media/signal_handlers.py
```diff
@@ -43,6 +41,23 @@ def publish_media_on_publish(instance: Model, **kwargs: Any) -> None:
...
+    """Signal handler to be connected to the 'page_published' and 'published'
+    signals for all publishable models. It is responsible for identifying any
+    privacy-controlled media used by the object, and ensuring that it is also
+    made public.
+    """
+    if isinstance(instance, Page):
+        run_post_publish_action(
+            PostPublishActionType.S3_ACL,
+            instance,
+            get_active_bundle_for_page(instance),
```
MebinAbraham: Not sure `get_active_bundle_for_page` would work as intended, we currently have a `FAILED` status for bundles, which we may choose to retry in future, this currently would return None as failed is not in the active states. Okay for now as it would immediately execute but needs some consideration.

MebinAbraham (commented):
On reflection, although we accepted in-process execution for this first pass, I think we should store enough information in `PostPublishAction` to retry later if the process dies.

Right now the DB gives us status visibility, but execution still depends on a process-local callable in the thread pool. If the worker dies after creating the row, we may leave `READY`/`RUNNING` actions that cannot be replayed cleanly.

Could we make the durable boundary `action_type + page_id + bundle_id`, and have the executor call `run_action(action_id)`? `run_action()` can then resolve the handler from `action_type`. This would also be helpful even if in future we decided to move to something like RQ.

We do not need to build a full queue system now, but this would allow a follow-up scheduled recovery job to find stale `READY`/`RUNNING` actions, increment retry metadata, and re-submit them. That keeps the happy path immediate while giving us a recovery story for process death/restarts.

Just thinking out loud, happy to discuss pros/cons.

cms/settings/base.py
RealOrangeOne: At that point the pool serves little value. Much of the time spent in each thread isn't really using the DB, so keeping this smaller reduces the number of connections.

cms/post_publish_actions/executor.py
RealOrangeOne: We will want something like that, yes. My suggestion is a future PR.

cms/private_media/signal_handlers.py
RealOrangeOne: I think it might be confusing if we leave it as implemented then. Is there a canonical way to determine "which bundle is this page in"? Given they can be in multiple, it sounds like we'd need to pass the bundle around, which gets a lot more complex.

cms/settings/base.py
RealOrangeOne: Because when using a pool, `CONN_MAX_AGE` must be 0.

cms/post_publish_actions/models.py
RealOrangeOne: Would the behaviour of a retry be the same for a "timed out" task vs a "failed" task? If so it's probably fine to keep them the same and have a flag on the model whether it timed out, as you say.

cms/bundles/viewsets/bundle.py
RealOrangeOne: We probably only need to track the current state in the DB. If we want historic data, the logs are better for that. I can't think of a reason we'd need to query the previous state.

cms/bundles/notifications/slack.py
RealOrangeOne: The reason I changed it is because that information isn't visible in the signals.

I think this and other reasons means we need to move these actions out of signals, and trigger them manually ourselves (which does add a fair bit of complexity).

cms/settings/base.py
RealOrangeOne: Yeah. It should avoid contention whilst not massively inflating connections.

cms/private_media/signal_handlers.py
BJacksonONS: @MebinAbraham Just had a look at this and right now I'm not sure we can retry the same bundle at present. The edit page redirects away if the bundle is in a published status, which currently includes partial and failed.

https://github.com/ONSdigital/dis-wagtail/blob/66a07442cc7ade08e1433ce91028c46c2ef72eb4/cms/bundles/viewsets/bundle.py#L146-L155
https://github.com/ONSdigital/dis-wagtail/blob/66a07442cc7ade08e1433ce91028c46c2ef72eb4/cms/bundles/enums.py#L36

It's possible to re add the page to a new bundle, but then we have restrictions on adding to multiple bundles.

This may be a regression introduced by the slack notification pr though as previously we only redirected away on status == published.

https://github.com/ONSdigital/dis-wagtail/pull/504/changes#diff-50ab22dc4dd51d8f8e12845953ce4b73af7d9c5e9097bb7eed3a77b476a39ba5L138

cms/bundles/viewsets/bundle.py
BJacksonONS: This is what I was thinking.

BJacksonONS:
Some other thoughts I've had that @MebinAbraham may have covered also

- Ensure the api boundary uses serialisable data structures. Right now we're passing model instances about which would require a fair bit of refactoring if we ever move to a different backend.
- Could we use futures or events to track completion rather than polling the database?

cms/post_publish_actions/handlers.py
RealOrangeOne: This assumes the reset has already happened - in `publish_bundle`.

RealOrangeOne:
Following some conversations with @MebinAbraham, I've made a few changes:

- Manual publishing runs the completion checks in a (separate) threadpool - https://github.com/ONSdigital/dis-wagtail/pull/676#discussion_r3395874203
- Messages are shown if the process terminates and the threadpool is still running, showing the number of blocking threads. Timing this out is deferred to the platform or gunicorn's `graceful_timeout`.
- A management command will retry actions which are stuck as running for too long, or have failed / timed out
- Consolidate signal handling. Actions are now registered as if they were signals, but wrapped centrally. This creates a registry where the actions can be listed
- Track whether an action timed out separately to its status. This deals with the race condition that an action may be marked as timed out, but then succeed shortly afterwards

cms/post_publish_actions/executor.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/3aa111c5be3f1a64f5fbc8dbc34c3e607127758b

cms/bundles/notifications/slack.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/744e4dcf39edce293dea20b43c7ba5bd9c09fb9e

cms/bundles/viewsets/bundle.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/a1cd76329b3ff9605465a8c4e97bdc536543fab9

cms/post_publish_actions/management/commands/retry_post_publish_actions.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def _get_actions_to_retry(self) -> PostPublishActionQuerySet:
+        """Find the actions to be retried.
+        An action should be retried if:
+            - It was enqueued 2x the timeout ago.
+            - It is marked as currently running (because of the above, it probably isn't).
+            - It is marked as failed.
+        """
+        return (
+            PostPublishAction.objects.unfinished()
```
BJacksonONS: Looks like we're setting a `finished_at` on exceptions and timeouts, so I feel like this query might not return some tasks that we'd like to retry.

cms/bundles/notifications/slack.py
```diff
@@ -263,19 +258,140 @@ def notify_slack_of_publish_end(
     )
 
     if bundle_page_count > 0:
-        fields.append({"title": "Pages Published", "value": str(pages_published), "short": True})
+        fields.append(
+            {
+                "title": "Pages Published",
+                "value": str(bundle.get_bundled_pages().filter(has_unpublished_changes=False).count()),
```
BJacksonONS: I think there is still a veeery small chance here that someone could save a draft in the time between publish and the task running.

Ideally we'd be tracking revision IDs on the bundle but that's probably a larger follow up task if we were to do that.

MebinAbraham: Yes, we accepted this as a gap for now. We can, in a follow-up, store the revision ID on bundle approval or page approval as that is when things become locked.

MebinAbraham: Also, we need to improve our bundle statuses. Currently, because we mark the bundle as "published" (before post publish finishes), it allows attached pages to be edited. If we had transition states, e.g., bundle post-publish, etc., then pages could stay locked until the bundle is fully published, including background tasks. But this needs some thorough discussion.

Let's capture this gap in the comments or the doc.

cms/post_publish_actions/management/commands/retry_post_publish_actions.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/130c6901f678d543575f731585d5ec75e8e46f78

cms/post_publish_actions/models.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/7911cb945c5d80fe430accc54a3fec8adbc5050f

cms/private_media/signal_handlers.py
RealOrangeOne: It doesn't work as expected - bundles are marked as "published" before actions start, so `get_active_bundle_for_page` regularly returns None.

I've called them manually instead https://github.com/ONSdigital/dis-wagtail/pull/676/commits/74fe33090aa9e7df5c8e20d506a6fb71a8375e39

cms/settings/base.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/b8b0e230821c423e2bb4f2f194696cd52380b3f5

cms/settings/base.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/f1ba26e398c3bcaa7371fe2151fe80e12f0fde01

cms/bundles/management/commands/publish_bundles.py
```diff
@@ -50,10 +56,55 @@ def _handle_bundle_action(self, bundle: Bundle) -> None:
                 logger.error("Bundle no longer approved", extra={"bundle_id": bundle.pk})
                 return
 
+            self.bundle_start_times[bundle] = timezone.now()
             publish_bundle(bundle)
         except Exception:  # pylint: disable=broad-exception-caught
             logger.exception("Publish failed", extra={"bundle_id": bundle.pk, "event": "publish_failed"})
 
+    def _await_bundle_post_publish_actions(self, bundles: list[Bundle]) -> None:
+        start_time = min(self.bundle_start_times.values())
```
BJacksonONS: We're starting our window for judging completion at the earliest bundle publish time. If there's multiple bundles being published it's possible that the 2nd or 3rd queued up might be marked as timed out because they didn't finish in 110 seconds of the first bundle publish starting, even if they actually run to completion fine.

If we don't want to give each bundle it's own timeout window then we might want to put a comment to explicitly address that.

cms/bundles/notifications/slack.py
```diff
@@ -263,19 +262,140 @@ def notify_slack_of_publish_end(
...
+
+    This is called after all pages are published, and all post-publish actions have finished.
+
+    Args:
+        bundle: The bundle that was published.
+        start_time: The time publishing started.
+        end_time: The time publishing ended.
+        url: The URL to link to the bundle (optional).
+    """
+    failed_post_publish_actions = (
```
MaciekBaron: Minor: Add a `_count` suffix for consistency

cms/post_publish_actions/models.py
```diff
@@ -0,0 +1,108 @@
...
```
MaciekBaron: I don't think we need to translate those, so we can drop the `gettext_lazy` calls.

cms/post_publish_actions/executor.py
```diff
@@ -0,0 +1,173 @@
...
+    if progress:
+        time.sleep(0.1)
+
+    executor_threads = _executor._threads | _support_executor._threads  # pylint: disable=protected-access
+
+    # Add 1 to force a print the first time around
+    last_running_threads = len(executor_threads) + 1
+
+    while running_threads := [t for t in executor_threads if t.is_alive()]:
+        if progress and last_running_threads > len(running_threads):
```
MaciekBaron, line 207: I think it might be good to add some sort of timeout here just in case.

Something like:
```python
timeout_seconds = 30  # Or a setting or parameter
start_time = time.monotonic()

while running_threads := [t for t in executor_threads if t.is_alive()]:

    if time.monotonic() - start_time > timeout_seconds:
        logger.error(
            "Shutdown timed out! %d threads still running.", 
            len(running_threads),
            extra={"thread_names": [t.name for t in running_threads]}
        )
        break
```

cms/post_publish_actions/executor.py
```diff
@@ -0,0 +1,173 @@
...
+                len(running_threads),
+                extra={
+                    "thread_ids": [t.native_id for t in running_threads],
+                    "thread_names": [t.name for t in running_threads],
+                },
+            )
+            last_running_threads = len(running_threads)
+
+        # 1 second is long enough for the other threads to do work, but not so long as to delay unnecessarily.
+        time.sleep(1)
```
MaciekBaron: Does this really need to be a second? Unless I misunderstand something, we could probably reduce this to 0.1 as each loop will just execute the while statement and if clauses, so not much CPU work. Let me know what you think.

cms/post_publish_actions/utils.py
```diff
@@ -0,0 +1,100 @@
...
+
+        # Only wait if there are bundles to check
+        if bundles_to_check:
+            time.sleep(settings.BUNDLE_POST_PUBLISH_POLL_FREQUENCY)
+
+    # Any remaining bundles will have timed out.
+    return bundles_to_check
+
+
+def post_publish_notify_slack(start_time: datetime, bundle: Bundle) -> None:
```
BJacksonONS: Should this also have `@force_write_db`?

Some of it's enqueuing functions [eg `run_after_hook`](https://github.com/ONSdigital/dis-wagtail/pull/676/changes#diff-50ab22dc4dd51d8f8e12845953ce4b73af7d9c5e9097bb7eed3a77b476a39ba5R252) have it but that would be thread local.

cms/bundles/management/commands/publish_bundles.py
RealOrangeOne: I think this is the correct behaviour. The timeout should be measured from when the bundle is due to publish, rather than when it actually starts. In reality, these are likely to be very similar.

cms/post_publish_actions/models.py
RealOrangeOne: In future, these are likely to be shown in the admin. Might as well start translating them now.

cms/post_publish_actions/executor.py
RealOrangeOne: We can, but unfortunately the process will still just wait for the threads anyway (this doesn't add the wait, it just adds some progress logging). Given that, we should just keep logging all the time there are threads running.

The external process manager (eg Kubernetes) will more forcefully terminate the process (without waiting for threads) if it takes too long.

cms/post_publish_actions/models.py
MaciekBaron: Sorry, this should have been documented somewhere: a decision has been made a while ago that no text in the admin panel needs translating (we even created a PR to remove all of those calls from existing strings)

RealOrangeOne: No worries - I'll get those dropped

cms/bundles/notifications/slack.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/f090e2b348e70de2ab0a8467ead6afb024c72023

cms/post_publish_actions/utils.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/63d7daee145e5a5283fb0d51fee02b3d0116912b

cms/post_publish_actions/executor.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/f8d1ac3dde036d131f73b500afb04480afb36b22

cms/post_publish_actions/models.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/259e789041262597dec25ea77f855807127c82b7

cms/bundles/notifications/slack.py
BJacksonONS: If two concurrent actions are running that send slack messages, it's possible that a stale bundle instance might result in the `slack_notification_ts` getting set incorrectly.

Think worst case is extra messages getting created though.

cms/post_publish_actions/executor.py
```diff
@@ -0,0 +1,172 @@
...
+            "Publish action successful",
+            extra={
+                "page_id": page_id,
+                "bundle_id": bundle_id,
+                "action_type": action_type.value,
+                "event": "post_publish_action_successful",
+                "duration": duration,
+            },
+        )
+        action.status = PostPublishActionStatus.SUCCESSFUL
```
BJacksonONS, line 123: Q: Should success clear any possible `timed_out_at` values?

RealOrangeOne: No, this behaviour is intentional. Because timing out doesn't actually kill the thread (it can't), we still want to keep track that an action took too long, even if it eventually succeeded.

cms/bundles/notifications/slack.py
RealOrangeOne: The notification TS should be set during the "start publish" part of the process, which is done synchronously

cms/post_publish_actions/management/commands/retry_post_publish_actions.py
```diff
@@ -0,0 +1,74 @@
...
+    def _get_actions_to_retry(self) -> PostPublishActionQuerySet:
+        """Find the actions to be retried.
+        An action should be retried if:
+            - It was enqueued 2x the timeout ago.
+            - It is marked as currently running (because of the above, it probably isn't).
+            - It is marked as failed.
+        """
+        return PostPublishAction.objects.active().filter(
+            enqueued_at__lte=timezone.now() - timedelta(seconds=settings.BUNDLE_POST_PUBLISH_TIMEOUT_SECONDS * 2),
+            status__in=[PostPublishActionStatus.RUNNING, PostPublishActionStatus.FAILED],
```
BJacksonONS: Should we get any stuck `READY` actions here too? We have the `enqueued_at__lte` filter so we shouldn't get any that are actually about to execute anyway.

Right now I think the only thing that can convert a `READY` action to `FAILED` is the polling, but there's no contingency for the process dying.

RealOrangeOne: Hmm, good point. Something needs to put ready items back in the queue if it dies before running.

cms/bundles/notifications/slack.py
BJacksonONS: `send_bundle_notification` mutates `slack_notification_ts` after sending the message, so we can still be using stale timestamps in different threads can't we?

cms/post_publish_actions/management/commands/retry_post_publish_actions.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/7a958df43c4b86d5efce8379e0ca2138d07e15bc

cms/locale/cy/LC_MESSAGES/django.po
```diff
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2026-06-26 10:50+0100\n"
+"POT-Creation-Date: 2026-06-30 14:48+0100\n"
```
MaciekBaron: We can probably revert the whole file now and remove it from the PR

MaciekBaron (approved):
Apart from the other comments raised (and one minor one I'm raising one) LGTM

cms/locale/cy/LC_MESSAGES/django.po
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/2d8edf2224dda3d5c59443acec8c16dcb3fcdf27

cms/bundles/management/commands/publish_bundles.py
BJacksonONS: `bundle_start_times` doesn't include filtered bundles that have left an `APPROVED` status. If this list is empty then I think `min` will throw an error.

cms/bundles/management/commands/publish_bundles.py
```diff
@@ -50,10 +56,55 @@ def _handle_bundle_action(self, bundle: Bundle) -> None:
...
+
+        concurrent.futures.wait(
+            bundle_complete_futures,
+            return_when=concurrent.futures.ALL_COMPLETED,
+            timeout=remaining_time + settings.BUNDLE_POST_PUBLISH_POLL_FREQUENCY,
+        )
+
+    @force_write_db()
+    def _handle_bundle_post_publish_complete(self, bundle: Bundle) -> None:
+        notify_slack_of_post_publish_end(bundle, self.bundle_start_times[bundle], timezone.now())
```
BJacksonONS: The bundles passed to `_await_bundle_post_publish_actions` are copied during the initial setup, but `bundle_start_times` filters out any that were moved out of an `APPROVED` status. This would key error for any bundle skipped when constructing the times dict.

cms/settings/base.py
```diff
@@ -280,13 +274,21 @@
...
     if "READ_REPLICA_DATABASE_URL" in env:
-        DATABASES["read_replica"] = dj_database_url.config(
-            env="READ_REPLICA_DATABASE_URL", conn_max_age=db_read_conn_max_age or 0
-        )
+        DATABASES["read_replica"] = dj_database_url.config(env="READ_REPLICA_DATABASE_URL", conn_max_age=0)
 
 if "read_replica" not in DATABASES:
     DATABASES["read_replica"] = deepcopy(DATABASES["default"])
 
+# Configure a connection pool to reduce connections when using threads.
```
BJacksonONS, line 309: While it's not likely and not a problem now, might be worth adding a comment about the issues with preloading apps that make DB connections in their init when pooling is combined with preloading.

Preloading is only configured with gunicorn so we'd be unlikely to catch that in local dev.

RealOrangeOne: Django raises a warning when you try and do DB queries before apps are loaded, so we're likely ok.

cms/bundles/management/commands/publish_bundles.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/494cd867e50276d8b50bc0628ad4c5b46092331d

cms/bundles/management/commands/publish_bundles.py
RealOrangeOne: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/494cd867e50276d8b50bc0628ad4c5b46092331d

cms/bundles/notifications/slack.py
```diff
@@ -263,19 +274,159 @@ def notify_slack_of_publish_end(
...
+    fields.extend(
+        [
+            {"title": "Dataset Count", "value": str(bundle_dataset_count), "short": False},
+            {
+                "title": "Post-Publish Actions Successful",
+                "value": str(
+                    PostPublishAction.objects.finished()
+                    .active()
+                    .filter(bundle=bundle, status=PostPublishActionStatus.SUCCESSFUL)
+                    .count()
```
MaciekBaron: **Suggestion**: We seem to be using this chain a lot. Maybe we could abstract it to return a count for a given bundle and status?

MaciekBaron (approved):

cms/bundles/notifications/slack.py
BJacksonONS: What do you think of these?

https://github.com/ONSdigital/dis-wagtail/pull/676/commits/e10c6cdeed94330ba8ceb61b105bf9ebdef2c4ec
https://github.com/ONSdigital/dis-wagtail/pull/676/commits/4e40124a5b7bbc6cc046473c6cc2172d0251848d

MebinAbraham:

cms/post_publish_actions/executor.py
```diff
@@ -0,0 +1,198 @@
...
+    action.status = PostPublishActionStatus.RUNNING
+    action.save(update_fields=["status"])
+
+    # use `specific` to ensure pages are cast to their subclass
+    # guarantees class attributes are available in post publish actions
+    page = Page.objects.get(id=page_id).specific
+    bundle = Bundle.objects.get(id=bundle_id)
+
+    logger.info(
+        "Starting publish action",
```
MaciekBaron: Question: Should this say "Starting post publish action"?

cms/post_publish_actions/executor.py
```diff
@@ -0,0 +1,198 @@
...
+    return ThreadPoolExecutor(
+        max_workers=settings.BUNDLE_POST_PUBLISH_CONCURRENCY, thread_name_prefix="post_publish_actions_support"
+    )
+
+
+# The main executor is reserved exclusively for post-publish actions
+_executor = _build_executor()
+
+# The "support" executor should be used for ancillary tasks
+_support_executor = _build_support_executor()
```
MebinAbraham, line 39: This is likely not fork-safe. We [preload the app](https://github.com/ONSdigital/dis-wagtail/blob/main/gunicorn.conf.py#L23) therefore, we would need to lazily initialise this or initialise it by hooking into the `post_fork` or `post_worker_init` Gunicorn hook. Leaving it as it may cause obscure issues in future. You could reuse the `flush_executor` func you have.

cms/post_publish_actions/executor.py
```diff
@@ -0,0 +1,198 @@
...
+        time.sleep(0.1)
+
+    executor_threads = _executor._threads | _support_executor._threads  # pylint: disable=protected-access
+
+    # Add 1 to force a print the first time around
+    last_running_threads = len(executor_threads) + 1
+
+    while running_threads := [t for t in executor_threads if t.is_alive()]:
+        if progress and last_running_threads > len(running_threads):
+            logger.debug(
```
MebinAbraham: Would be useful as `.info` while we observe behaviours and tune.

cms/post_publish_actions/executor.py
```diff
@@ -0,0 +1,198 @@
...
+                "Waiting for %d threads running post-publish actions",
+                len(running_threads),
+                extra={
+                    "thread_ids": [t.native_id for t in running_threads],
+                    "thread_names": [t.name for t in running_threads],
+                },
+            )
+            last_running_threads = len(running_threads)
+
+        time.sleep(0.1)
```
MebinAbraham, line 230: Non-blocker: I wonder if this should have a timeout; if this doesn't finish, it will be force-killed via `SIGKILL`. We have a graceful timeout of 110, so before that, if this was nearing that and had unfinished, we could log something nicer for support purposes.

gunicorn.conf.py
```diff
@@ -16,9 +17,13 @@
 # Log to stdout
 accesslog = "-"
 
-# Time out after 25 seconds (notably shorter than Heroku's)
+# Time out after 25 seconds
 timeout = int(os.environ.get("GUNICORN_TIMEOUT", "25"))
 
+# Allow the bundle publishing timeout to expire before considering a terminate "ungraceful".
+# gunicorn expects an int, but the timeout is parsed as a float elsewhere so convert to preserve behavior
+graceful_timeout = math.ceil(float(os.environ.get("BUNDLE_POST_PUBLISH_TIMEOUT_SECONDS", "110")))
```
MebinAbraham, line 25: We need to tune our `terminationGracePeriodSeconds` in k8s to match, the chart doesn't support it yet so we are using the default 30 atm. https://officefornationalstatistics.atlassian.net/browse/CMS-1351

cms/post_publish_actions/apps.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def ready(self) -> None:
+        from .executor import executor_stop_and_wait  # pylint: disable=import-outside-toplevel
+        from .signal_handlers import register_signal_handlers  # pylint: disable=import-outside-toplevel
+
+        # ThreadPoolExecutor threads are waited on when the process terminates.
+        # This adds some logging to make it clear the process didn't hang.
+        # This must be done using _register_atexit rather than atexit.register
+        # as atexit runs after threads have joined.
+        threading._register_atexit(executor_stop_and_wait)  # type: ignore[attr-defined] # pylint: disable=protected-access
```
MebinAbraham, line 17: Did you consider `on_exit` or `worker_exit` gunicorn hooks?.

cms/bundles/management/commands/publish_bundles.py
```diff
@@ -1,9 +1,11 @@
 import logging
```
MebinAbraham, line 1: Nit: Worth overriding the [scheduler.py shutdown ](https://github.com/ONSdigital/dis-wagtail/blob/main/cms/core/management/commands/scheduler.py#L36)method to wait to have prod like behaviour.
```py
    def shutdown(self, _signum: int | None = None, _frame: Any | None = None) -> None:
        """Shutdown handler."""
        if self.scheduler.running:
            self.scheduler.shutdown(wait=False) # wait=True
```

MebinAbraham (commented):
This feature needs to be well documented (follow up PR).

cms/post_publish_actions/management/commands/retry_post_publish_actions.py
```diff
@@ -0,0 +1,72 @@
...
+            - It is marked as currently running (because of the above, it probably isn't).
+            - It hasn't run yet (also unlikely).
+            - It is marked as failed.
+        """
+        return PostPublishAction.objects.active().filter(
+            enqueued_at__lte=timezone.now() - timedelta(seconds=settings.BUNDLE_POST_PUBLISH_TIMEOUT_SECONDS * 2),
+            status__in=[PostPublishActionStatus.RUNNING, PostPublishActionStatus.FAILED, PostPublishActionStatus.READY],
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
```
MebinAbraham, line 58: Use `force_write_db` here?. We have queries in here that must go to writer.

cms/post_publish_actions/management/commands/retry_post_publish_actions.py
```diff
@@ -0,0 +1,72 @@
...
+                # NB: These are enqueued on commit
+                action.enqueue()
+
+            # Update after enqueue so the iteration works
+            actions_to_retry.update(
+                status=PostPublishActionStatus.READY,
+                failed_reason="",
+                duration=None,
+                finished_at=None,
+                timed_out_at=None,
```
MebinAbraham, line 99: Should we not be resetting `enqueued_at`? Otherwise, we could end up retrying forever. Maybe we need a retry count / max retry limit.

cms/post_publish_actions/executor.py
```diff
@@ -0,0 +1,198 @@
...
+# The main executor is reserved exclusively for post-publish actions
+_executor = _build_executor()
+
+# The "support" executor should be used for ancillary tasks
+_support_executor = _build_support_executor()
+
+logger = logging.getLogger(__name__)
+
+
+def _executor_wrapper[**P](executor_fn: Callable[P, None], *args: P.args, **kwargs: P.kwargs) -> Any:
```
MebinAbraham:
```suggestion
def _executor_wrapper[**P](executor_fn: Callable[P, None], *args: P.args, **kwargs: P.kwargs) -> None:
```

cms/post_publish_actions/models.py
```diff
@@ -0,0 +1,125 @@
...
+        return self.active().finished()
+
+    def successful(self) -> Self:
+        return self.completed().filter(status=PostPublishActionStatus.SUCCESSFUL)
+
+    def failed(self) -> Self:
+        return self.completed().filter(status=PostPublishActionStatus.FAILED)
+
+    def count_for_bundle(self, bundle_id: int | Bundle) -> int:
+        return self.filter(bundle=bundle_id).count()
```
MebinAbraham: bundle or bundle_id?

cms/post_publish_actions/models.py
```diff
@@ -0,0 +1,125 @@
...
+
+    objects = PostPublishActionQuerySet.as_manager()
+
+    class Meta:
+        constraints: ClassVar[list[models.BaseConstraint]] = [
+            # Page and type must be unique, optionally in the scope of a bundle
+            models.UniqueConstraint(
+                fields=["bundle", "page", "action_type"],
+                condition=models.Q(bundle__isnull=False),
+                name="bundle_page_type",
```
MebinAbraham: Nit: maybe prefix these with `post_publish_actions_` as we may use this name elsewhere. Same for `page_type`

kevinhowbrook (approved):

sanjeevz3009 (approved):
Looks good. I know @MebinAbraham mentioned there are some things still to disucss with @BJacksonONS

cms/post_publish_actions/executor.py
BJacksonONS: Fixed across a few logs

https://github.com/ONSdigital/dis-wagtail/pull/676/commits/d7190b70151ec6aa6c17c0abbe9c6ee00f128497

cms/post_publish_actions/models.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/cc20b115a9c17a1daf4b1fd6a8025898cede67c9

cms/post_publish_actions/models.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/5033673b4536a73b066b1440df0f8d90dedc5b5f

cms/post_publish_actions/management/commands/retry_post_publish_actions.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/879ed38c9e7bb5226cc4da0fa37ab91b17818826

cms/bundles/management/commands/publish_bundles.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/c940f42d74fdd91c841b07d4fd2ec6ebcef69edd

cms/post_publish_actions/executor.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/d8932b9dd61cee08445c92d68122fb6a65f72fec

cms/post_publish_actions/executor.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/86eef7c6209ca51d9299c34bdff2b6b7b68238af

cms/post_publish_actions/executor.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/ee5cc0254e96d4e3aba7121a80f81bc8437a8b01

cms/post_publish_actions/apps.py
BJacksonONS: Think @RealOrangeOne made the right decision on this one.

`on_exit` only runs from the original process which will never submit any tasks so I don't think this is the right place to do this. As per one of your other comments we've added extra guarantees for thread safety by recreating executors in forked processes so it wouldn't hit them.

`worker_exit` could possibly work in future if we make the switch to running scheduled jobs via api calls, but for now wouldn't work in any cron jobs.

cms/post_publish_actions/management/commands/retry_post_publish_actions.py
BJacksonONS: Update enqueued at, and add a max retry limit

https://github.com/ONSdigital/dis-wagtail/pull/676/commits/27d8d321155eee920e420840f823c32e021a72e5

cms/post_publish_actions/management/commands/retry_post_publish_actions.py
```diff
@@ -0,0 +1,122 @@
...
+            dest="dry_run",
+            default=False,
+            help="Dry run -- don't change anything.",
+        )
+
+    def _get_failed_actions(self) -> PostPublishActionQuerySet:
+        return PostPublishAction.objects.active().filter(
+            enqueued_at__lte=timezone.now() - timedelta(seconds=settings.BUNDLE_POST_PUBLISH_TIMEOUT_SECONDS * 2),
+            status__in=[PostPublishActionStatus.RUNNING, PostPublishActionStatus.FAILED, PostPublishActionStatus.READY],
+        )
```
MebinAbraham, line 43: I suggest adding a comment more specifically around enqueued at logic, as it might not be immediately obvious why we do this.

MebinAbraham (approved):

MebinAbraham (approved):
Nice work @RealOrangeOne @BJacksonONS

cms/search/signal_handlers.py
```diff
@@ -7,19 +7,26 @@
...
+def update_index_post_publish_action(page: Page, _bundle: Bundle | None) -> None:
+    if is_indexable_page(page):
+        get_publisher().publish_created_or_updated(page)
+
+
 @receiver(page_published)
 def on_page_published(sender: type[Page], instance: Page, **kwargs: Any) -> None:  # pylint: disable=unused-argument
     """Called whenever a Wagtail Page is published (UI or code).
     instance is the published Page object.
     """
```
MebinAbraham: Should this be kept? It's no-op now?

BJacksonONS: Happy to remove it, assume anything else we do on page publish now will always be async.

MebinAbraham: Ya.

BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/676/commits/8de60e1a3f1d00af19db658777b3f5bd47309866

BJacksonONS:
@MebinAbraham 

Added an extra bit of handling for alias search indexing as bundle publishing now suppresses the signal that would have done this previously.

https://github.com/ONSdigital/dis-wagtail/pull/676/commits/821f458f210a6f4ee4dbb4f1d8e14059234102e2

BJacksonONS:
After discussion with @MebinAbraham and @sanjeevz3009, we're going to merge this as is, with a follow up to do a refactoring to avoid suppressing signals.

MebinAbraham (approved):

cms/post_publish_actions/executor.py
MebinAbraham: Resolving on behalf.

sanjeevz3009:
> After discussion with @MebinAbraham and @sanjeevz3009, we're going to merge this as is, with a follow up to do a refactoring to avoid suppressing signals.

@BJacksonONS Thanks Ben, great work =)

PR #677, opened 2026-06-11
Update times in slack notifications to use Europe/London timezone

charlotteromaniuk:

Addresses [CMS-1351](https://officefornationalstatistics.atlassian.net/browse/CMS-1251)

Updates the timestamps in the slack notifications to be displayed using the Europe/London timezone.

### How to review

1. Connect a slack workspace using environment variables
2. Create a bundle and update the status
3. Review the slack message to ensure the changed at time is correctly displayed using the Europe/London timezone
4. Publish the bundle
5. Review slack messages to ensure the publish start and end times are correctly displayed using the Europe/London timezone

### Deployment Safety

Please select one:

- [X] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/cms/bundles/notifications/slack.py
+++ b/cms/bundles/notifications/slack.py
@@ -6,6 +6,7 @@ from enum import Enum
 from typing import TYPE_CHECKING, Any
 
 from django.conf import settings
+from django.utils import timezone
 
 from cms.core.slack import send_or_update_slack_message
 
@@ -77,15 +78,18 @@ def _get_publish_type(bundle: Bundle) -> str:
 
 
 def _format_publish_datetime(dt: datetime) -> str:
-    """Format datetime as DD/MM/YYYY - HH:MM:SS.sss.
+    """Format datetime as DD/MM/YYYY - HH:MM:SS.sss in Europe/London local time.
 
     Args:
         dt: The datetime to format.
 
     Returns:
-        Formatted string in DD/MM/YYYY - HH:MM:SS.sss format.
+        Formatted string in DD/MM/YYYY - HH:MM:SS.sss format, representing the
+        datetime in Europe/London local time.
     """
-    return f"{dt:%d/%m/%Y - %H:%M:%S}.{dt.microsecond // 1000:03d}"
+    local_dt = dt.astimezone(timezone.get_default_timezone())
+
+    return f"{local_dt:%d/%m/%Y - %H:%M:%S}.{local_dt.microsecond // 1000:03d}"
 
 
 def _get_example_page_url(bundle: Bundle) -> str | None:
--- a/cms/bundles/notifications/slack.py
+++ b/cms/bundles/notifications/slack.py
@@ -341,7 +341,7 @@ def alert_slack_of_bundle_content_failure(
     """
     fields: list[dict[str, Any]] = [
         {"title": "Bundle Name", "value": f"<{bundle.full_inspect_url}|{bundle.name}>", "short": False},
-        {"title": "Timestamp", "value": _format_publish_datetime(datetime.now()), "short": True},
+        {"title": "Timestamp", "value": _format_publish_datetime(timezone.now()), "short": True},
         {"title": "Publish Type", "value": _get_publish_type(bundle), "short": True},
         {"title": "Alert Type", "value": alert_type, "short": True},
         {"title": "Exception", "value": exception_message, "short": False},
--- a/cms/bundles/notifications/slack.py
+++ b/cms/bundles/notifications/slack.py
@@ -87,7 +87,7 @@ def _format_publish_datetime(dt: datetime) -> str:
         Formatted string in DD/MM/YYYY - HH:MM:SS.sss format, representing the
         datetime in Europe/London local time.
     """
-    local_dt = dt.astimezone(timezone.get_default_timezone())
+    local_dt = timezone.localtime(dt)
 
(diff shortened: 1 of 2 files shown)


joepercival (approved):
Tested functionally with log outputs, LGTM

cms/bundles/notifications/slack.py
```diff
@@ -77,15 +78,18 @@ def _get_publish_type(bundle: Bundle) -> str:
...
     Args:
         dt: The datetime to format.
 
     Returns:
-        Formatted string in DD/MM/YYYY - HH:MM:SS.sss format.
+        Formatted string in DD/MM/YYYY - HH:MM:SS.sss format, representing the
+        datetime in Europe/London local time.
     """
-    return f"{dt:%d/%m/%Y - %H:%M:%S}.{dt.microsecond // 1000:03d}"
+    local_dt = dt.astimezone(timezone.get_default_timezone())
```
BJacksonONS: I think `local_dt = timezone.localtime(dt)` is a little bit tidier here and consistent with other utils in `custom_date_format.py`.

MebinAbraham: ☝️

charlotteromaniuk: [262a242](https://github.com/ONSdigital/dis-wagtail/pull/677/commits/262a2424f2a8bffc64945e3f1c0649eecee4a039)

kevinhowbrook (approved):

cms/bundles/notifications/slack.py
```diff
@@ -77,15 +78,18 @@ def _get_publish_type(bundle: Bundle) -> str:
...
+    """Format datetime as DD/MM/YYYY - HH:MM:SS.sss in Europe/London local time.
 
     Args:
         dt: The datetime to format.
 
     Returns:
-        Formatted string in DD/MM/YYYY - HH:MM:SS.sss format.
+        Formatted string in DD/MM/YYYY - HH:MM:SS.sss format, representing the
+        datetime in Europe/London local time.
     """
```
MaciekBaron: **Nit-pick**: Well _technically_, this function does not enforce the London time zone, but the current active time zone of the application or thread. We should update the docstring so it's not confusing.

```suggestion
    """Format datetime as DD/MM/YYYY - HH:MM:SS.sss in the current active local time.

    Args:
        dt: The datetime to format.

    Returns:
        Formatted string in DD/MM/YYYY - HH:MM:SS.sss format, representing the
        datetime in the current local time zone.
    """
````

**Review by MebinAbraham — APPROVED** [review]

File: cms/bundles/notifications/slack.py (hunk 1 above)
Comment (BJacksonONS) [reply cms/bundles/notifications/slack.py]: Think we might need to use local timezone instead of utc in some of the tests now.
`dt = datetime(2026, 2, 17, 14, 30, 45, 234000, tzinfo=timezone.get_default_timezone())`

**Review by MaciekBaron — CHANGES_REQUESTED** [review]

File: cms/bundles/tests/test_slack_notifications.py [hunk 3]
```diff
@@ -862,6 +862,12 @@ def test_format_publish_datetime(self):
         formatted = _format_publish_datetime(dt)
         self.assertEqual(formatted, "17/02/2026 - 14:30:45.234")
 
+    def test_format_publish_datetime_applies_bst_offset(self):
+        """Should return datetime with the BST offset."""
+        dt = datetime(2026, 6, 17, 14, 30, 45, 234000, tzinfo=UTC)
+        formatted = _format_publish_datetime(dt)
```
MaciekBaron: Following my previous comment, I think we should do this here:

```python
from django.utils import timezone
...
with timezone.override("Europe/London"):
    formatted = _format_publish_datetime(dt)
```

MaciekBaron (changes requested):

cms/bundles/notifications/slack.py
```diff
@@ -341,7 +341,7 @@ def alert_slack_of_bundle_content_failure(
...
     fields: list[dict[str, Any]] = [
         {"title": "Bundle Name", "value": f"<{bundle.full_inspect_url}|{bundle.name}>", "short": False},
-        {"title": "Timestamp", "value": _format_publish_datetime(datetime.now()), "short": True},
+        {"title": "Timestamp", "value": _format_publish_datetime(timezone.now()), "short": True},
```
charlotteromaniuk, line 344: I've updated this so that the timestamp is timezone aware and can be formatted to the correct timezone for the slack notifications. I could add something to handle naive datetimes in `_format_publish_datetime()` instead if that would be preferred e.g. 
```
if timezone.is_naive(dt):
    dt = timezone.make_aware(dt, timezone.UTC)
```

cms/bundles/notifications/slack.py
charlotteromaniuk: Good point, updated here [64e0e66](https://github.com/ONSdigital/dis-wagtail/pull/677/commits/64e0e668418650b8574b32b75cb92e6f312d1967)

cms/bundles/tests/test_slack_notifications.py
charlotteromaniuk: [6988b1a](https://github.com/ONSdigital/dis-wagtail/pull/677/commits/6988b1a84031b4674b13f85e50c7111c8102669f)

MaciekBaron (approved):

BJacksonONS (approved):

sanjeevz3009 (commented):
Reviewed!

PR #661, opened 2026-06-03
Conditionally display footnotes for iframe visualisations

helenb:
### What is the context of this PR?

This is a fix following on from https://github.com/ONSdigital/dis-wagtail/pull/586.

See this comment on https://officefornationalstatistics.atlassian.net/browse/DPD-2500?focusedCommentId=754707

The 'footnotes' accordion should only display if footnotes have been added.

### How to review

Edit a statistical article page, and add 2 visualisations. Add footnotes to one, but not to the other. Ensure that for the one that has no footnotes added, you don't see the accordion with the heading 'footnotes'.

### Deployment Safety

Bleed and Sandbox deploy automatically on merge, so PRs should be safe to deploy immediately.

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

List any follow-up actions (if applicable), like needed documentation updates or additional testing.

PR #688, opened 2026-06-17
Check for text content before displaying footnotes

joepercival:

A fix following on from #661 (and #586).
Addresses [CMS-1257](https://officefornationalstatistics.atlassian.net/browse/CMS-1257) and [CMS-1258](https://officefornationalstatistics.atlassian.net/browse/CMS-1258).

In certain cases on the Iframe Visualisation and Table content blocks, the 'Footnotes' info detail panel still displays even when there is no text in the field in Wagtail.

The issue only appears to apply when the pages are in draft and are not published. 
The issue is only visible from the in-page preview and disappears when the page is previewed in a separate tab. 
When opening the preview in a separate tab, the issue also disappears from the Wagtail in-page preview.
Additionally, when the Wagtail page is refreshed after opening the preview in an separate tab the issue re-appears.

Inspecting the in-page preview shows that the Footnotes ons-detail element can contain visually empty HTML when no text has been added, e.g.: 
`<div class="rich-text"><p></p></div>`
This PR modifies the existing content checks to strip HTML tags before checking if content exists, which should prevent empty Footnotes panels from being rendered.

### How to review

1.  Create a (draft) info page or statistical article page 
2. Under 'content', create both a Table and Iframe Visualisation block
3. Verify that the 'Footnotes' detail panel does not appear in the preview on either the table or the Iframe Visualisation until content is added to the respective Footnotes field
4. Check this behaviour across a range of scenarios (e.g. publishing/unpublishing/saving/refreshing a page) to ensure the behaviour is consistent

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


cms/datavis/tests/test_line_chart_block.py
```diff
@@ -214,3 +214,21 @@ def test_footnotes_configuration(self):
...
+        for html in ("<p></p>", "<p> </p>"):
+            with self.subTest(html=html):
+                self.raw_data["footnotes"] = html
+                config = self.get_component_config()
+                self.assertEqual(config["footnotes"], {})
+        html = "<p>Valid content</p>"
+        with self.subTest(html=html):
+            self.raw_data["footnotes"] = html
+            config = self.get_component_config()
+            self.assertEqual(
```
MaciekBaron: Question: Why do we perform the test slightly differently here compared to other test in `test_iframe_block.py`? The config looks the same I think

MaciekBaron (approved):

cms/datavis/tests/test_line_chart_block.py
joepercival: There were some slight differences between the way the footnotes logic was handled, for the `LineChartBlock` it inherits the logic from `BaseChartBlock` wheraas the `IFrameBlock` has its own separate logic, both implementations were slightly different, I've rewritten it now to try and keep the tests more consistent - 97fa1c0

BJacksonONS (approved):

PR #690, opened 2026-06-18
ignore vuln in `http-proxy-middleware` for now

BJacksonONS:

Ignore osv-scanner flagging issue with `http-proxy-middleware` as there currently is no valid upgrade path.

### How to review

Check ID is correct and reasoning is valid.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


kevinhowbrook (approved):

joepercival (approved):

sanjeevz3009 (commented):
Reviewed!

PR #692, opened 2026-06-19
Disable wagtail autosave by default in functional tests and add dedicated test

BJacksonONS:

Disable wagtail autosave by default for functional tests to improve stability.

Add a dedicated test to check autosave behaviour.

### How to review

Check settings, review functional test runs in CI.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


cms/settings/functional_test.py
```diff
@@ -8,5 +8,8 @@
 
 KAFKA_SERVERS = ["localhost:19094"]
 
+# Disable wagtail autosave for functional tests
```
MebinAbraham, line 11: Now that it is being disabled, I wondering if we should have at least a single test to prove this works. It's Wagtail core but we have a bunch of custom admin JS so we want to prove anything we have don't won't regress it. It could be a simple as make a change, after 1 second reload and assert the change is still there.

BJacksonONS: Yeah I did think this, I'm looking at how we can override settings in functional tests and clean up after. Thought I could get this in now and then put up a separate PR for the test, but happy to combine if you prefer.

BJacksonONS: Found some examples in cognito test cases so shouldn't take too long.

BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/692/commits/dc5d6625f1e0b763097cf89d9c48a00e9b3221c1

kevinhowbrook (approved):
Thanks @BJacksonONS - added this case to https://github.com/wagtail/wagtail/issues/14341 (requesting autosave as a user setting)

functional_tests/steps/autosave.py
```diff
@@ -0,0 +1,41 @@
...
+def user_types_autosave_content(context: Context) -> None:
+    context.autosave_title = "Autosave Test Page"
+    context.autosave_summary = "This content should be autosaved"
+
+    context.page.get_by_role("textbox", name="Title*").fill(context.autosave_title)
+    context.page.get_by_role("region", name="Summary*").get_by_role("textbox").fill(context.autosave_summary)
+
+
+@when("the user waits the autosave interval")
+def wait_for_autosave_interval(context: Context) -> None:
```
charlotteromaniuk: I've noticed this step is intermittently failing when running in slow motion with a 1s delay. It looks like the autosave POST request sometimes completes before this step listens for it and so it gets missed. I was wondering whether adding a UI check for the "saved" status here might make it more robust? I've only seen this issue when running it in slow motion though and seems to run fine normally so might be low risk that this would occur in practice.

BJacksonONS: Good catch. Trying some other methods but it's a bit tricky to ensure it catches multiple edits.

The "Saved" message at the top seems to persist after one autosave until it tries to save again, so if we edit the title then the summary it can think that we've saved but only one of the changes has been submitted.

Looking at some other methods of testing this that are resilient to slow mo, but might have to compromise on this one.

BJacksonONS: Have changed the steps to wrap each individual edit in an autosave check.
https://github.com/ONSdigital/dis-wagtail/pull/692/commits/959013bbb422518684133ba5996403e2af810080

charlotteromaniuk (approved):
LGTM!

joepercival (approved):

cms/settings/test.py
```diff
@@ -92,6 +92,7 @@
 # Don't send slack notifications from tests if env vars are set
 SLACK_NOTIFICATIONS_WEBHOOK_URL = None
 
+
```
MaciekBaron: NP: Change probably not needed and a leftover from a reverted edit

MaciekBaron (approved):
Tested, works well. LGTM, left one non-blocking comment about a whitespace change.

cms/settings/test.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/692/commits/de35c0e027445a5191028f9261d1c79ad1675eef

MebinAbraham (approved):

sanjeevz3009 (commented):
LGTM!

PR #696, opened 2026-06-22
feat(cache): generalise title change signal for cache purge

MaciekBaron:

This change generalises the title change signal created in PR #505.

It sends a signal whenever a page title changes so that all of its descendants can be purged. This is to update any mentioning of the titles in breadcrumbs etc..

This PR targets the branch of that PR.

### How to review

Ensure that tests pass.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

Approve and merge PR #505 

---

### DIFF
--- a/cms/core/models/base.py
+++ b/cms/core/models/base.py
@@ -1,6 +1,7 @@
-from typing import TYPE_CHECKING, ClassVar, Self, cast
+from typing import TYPE_CHECKING, Any, ClassVar, Self, cast
 
 from django.conf import settings
+from django.db import transaction
 from django.http import HttpRequest
 from django.urls import reverse
 from django.utils.decorators import method_decorator
@@ -17,6 +18,7 @@ from cms.core.cache import get_default_cache_control_decorator
 from cms.core.forms import DeduplicateTopicsAdminForm, ONSCopyForm
 from cms.core.permission_testers import BasePagePermissionTester
 from cms.core.query import order_by_pk_position
+from cms.core.signals import page_title_changed
 from cms.locale.utils import get_mapped_site_root_paths
 from cms.taxonomy.mixins import ExclusiveTaxonomyMixin
 
@@ -107,6 +109,23 @@ class BasePage(PageLDMixin, ListingFieldsMixin, SocialFieldsMixin, Page):  # typ
     def permissions_for_user(self, user: User) -> BasePagePermissionTester:
         """Override the permission tester class to use for our page models."""
         return BasePagePermissionTester(user, self)
+
+    @transaction.atomic
+    def save(  # type: ignore[override]
+        self, clean: bool = True, user: User | None = None, log_action: bool = False, **kwargs: Any
+    ) -> Self | None:
+        original_values = type(self).objects.values("title", "slug").filter(pk=self.pk).first() if self.pk else None
+
+        instance: Self | None = super().save(  # type: ignore[call-arg]
+            clean=clean, user=user, log_action=log_action, **kwargs
+        )
+
+        if original_values and self.title != original_values["title"] and self.slug == original_values["slug"]:
+            # The title has changed, but not the slug.
+            # The slug change is already handled by the `page_slug_changed` signal.
+            transaction.on_commit(lambda: page_title_changed.send(sender=type(self), instance=self))
+
+        return instance
 
     @cached_property
     def related_pages(self) -> PageQuerySet:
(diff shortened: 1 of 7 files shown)


cms/core/models/base.py
```diff
@@ -108,6 +110,23 @@ def permissions_for_user(self, user: User) -> BasePagePermissionTester:
...
+    def save(  # type: ignore[override]
+        self, clean: bool = True, user: User | None = None, log_action: bool = False, **kwargs: Any
+    ) -> Self | None:
+        original_values = type(self).objects.values("title", "slug").filter(pk=self.pk).first() if self.pk else None
+
+        instance: Self | None = super().save(  # type: ignore[call-arg]
+            clean=clean, user=user, log_action=log_action, **kwargs
+        )
+
+        if original_values and self.title != original_values["title"] and self.slug == original_values["slug"]:
```
kevinhowbrook, line 123: What happens (or is supposed to happen) when the title AND slug is changed?

cms/core/models/base.py
```diff
@@ -108,6 +110,23 @@ def permissions_for_user(self, user: User) -> BasePagePermissionTester:
         """Override the permission tester class to use for our page models."""
         return BasePagePermissionTester(user, self)
 
+    @transaction.atomic
+    def save(  # type: ignore[override]
```
kevinhowbrook, line 114: Mostly just speculating here, but wagtail will call self.save during a draft save, so if the editor has changed the title the condition on line 123 would be true and trigger a descendant purge even though nothing was published. Is that intentional?

MaciekBaron: Hmm I think this is a good point. My MR just generalises the pre-existing solution Dan wrote, but both of your comments raise a valid point. Perhaps this is something we should address in the main PR?
Not sure if it's worth just including this as part of the other PR as these discussion are relevant.

cms/core/models/base.py
MaciekBaron: I think in this scenario the slug related signal is triggerred - this prevents it from being purged twice.

cms/core/models/base.py
sanjeevz3009: @MebinAbraham

sanjeevz3009 (approved):
Apart from Kev's comments and needing @MebinAbraham input of whatever is happening with the front-end caching PR, looks good to me!

cms/core/models/base.py
MebinAbraham: Good spot; this isn't ideal. We don't want to be purging for draft changes; it needs to only be "on publish". We will need to think about some other options. Currently, the purge would be wasteful and not functional until the page has actually been published as it would recache with old title anyway.

Ideally, it behaves like `page_slug_changed`, which is fired on publish?

cms/frontend_cache/signal_handlers.py
```diff
@@ -0,0 +1,121 @@
...
+    return {model for model in get_page_models() if issubclass(model, BasePage) and model not in EXCLUDED_PAGE_TYPES}
+
+
+def disconnect_signal_handlers() -> None:
+    """Disconnect the core front-end cache signal handlers as we handle them with specific logic."""
+    for model in get_page_models():
+        page_published.disconnect(page_published_signal_handler, sender=model)
+        page_unpublished.disconnect(page_unpublished_signal_handler, sender=model)
+        post_page_move.disconnect(autocreate_redirects_on_page_move, sender=model)
+        page_slug_changed.disconnect(autocreate_redirects_on_slug_change, sender=model)
```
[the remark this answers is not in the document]
MaciekBaron: Addressed in #785

cms/core/models/base.py
MaciekBaron: Addressed in [5792af0](https://github.com/ONSdigital/dis-wagtail/pull/696/commits/5792af0fa26c20b079178031346a096cc29e5517) with test added

cms/core/models/base.py
```diff
@@ -115,6 +117,27 @@ def permissions_for_user(self, user: User) -> BasePagePermissionTester:
...
+        self, clean: bool = True, user: User | None = None, log_action: bool = False, **kwargs: Any
+    ) -> Self | None:
+        update_fields = kwargs.get("update_fields")
+        title_is_persisted = update_fields is None or "title" in update_fields
```
MebinAbraham, line 127: Let's add a comment to doc why we are doing this. It might not be clear when revisiting what exactly this is acheving.

MebinAbraham (approved):
🎉

cms/core/models/base.py
MaciekBaron: Added here: [4621bca](https://github.com/ONSdigital/dis-wagtail/pull/696/commits/4621bca277f911591ef630cf9a770f9f0afb07ed)

MebinAbraham (approved):

kevinhowbrook (approved):

PR #697, opened 2026-06-24
Prevent duplicate entries of page/topic combinations

MaciekBaron:

This PR addresses an issue raised CMS-1267 and allows Wagtail to handle gracefully the broken state.

It adds a check to avoid an attempt to add a topic/page relation which already exists.

### How to review

1. Ensure that tests run and pass
2. Remove the `save()` override from `taxonomy/models.py`
3. Rerun the tests
4. Ensure that an `IntegrityError` is raised (confirms that fix works)

**How to trigger the issue without tests:**
1. Disable the fix or checkout `main`
2. Create a new information/methodology page, no taxonomy
3. Save as draft (Tab **A**)
4. Open the same draft in a new tab (Tab **B**)
5. Now in Tab **A** add a topic, wait for autosave
6. In Tab **B**, add _the same topic_, try to save draft
7. This triggers the `500`

Confirm the fix addresses this.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

Determine the exact cause of the broken state.


kevinhowbrook:
Testing notes, (on main) the IntegrityError is thrown, but also there is an alert to show autosave is disabled which is good.

https://github.com/user-attachments/assets/e0b4dc39-932b-412d-9858-5a928e249901

On _this_ branch the fix is looking good

https://github.com/user-attachments/assets/c83d5978-efba-4eac-88d7-7e26f5ab7144

cms/taxonomy/models.py
```diff
@@ -129,3 +129,17 @@ class Meta:
         constraints: ClassVar[list[BaseConstraint]] = [
             UniqueConstraint(fields=["page", "topic"], name="unique_generic_taxonomy")
         ]
+
+    def save(self, **kwargs: Any) -> None:
+        """Override the save method to prevent duplicate entries for the same page and topic combination."""
```
kevinhowbrook:
```suggestion
        """Silently deduplicates when modelcluster tries to INSERT a (page, topic) pair that was already
        committed by a concurrent save/session."""
```

**Nitpick**

Docstring sort of says the same as the code, might be worth outlining the why here, and the concurrent save issue.

kevinhowbrook (commented):
Thanks @MaciekBaron, tests well. As for the follow up on "Determine the exact cause of the broken state." I think this may be a Wagtail/modelcluster limitation. Perhaps we could catch the IntegrityError and handle with get or create, but I doubt modelcluster is meant for session handling like this. There some slight related issues, eg [mc 194](https://github.com/wagtail/django-modelcluster/pull/194/changes) but to me they look more to do with the in-memory formset rather than the DB level checks 

**Other**

Looks like your latest commit (--amend) didn't amend correctly btw

BJacksonONS:
Adding context here now we know a bit more.

Steps to reproduce a broken state. I've reproduced on both Methodology and Information page types.

1. Publish an Information or Methodology page
2. Add a topic to the page and save as draft
3. Publish the page
4. Unpublish the page
5. Attempt to save

The root cause is that when saving a draft `commit=False` is used. This means that a row is never created in the through model for the many-many join between topics and pages. The revision saves with the page and topic but no PK to the through model.

When the page is published a real row is created.

When unpublished the page goes back to a revision that contains no PK to the through model, it attempts to save a new row, but one already exists due to the publish.

This is somewhat hidden earlier in the flow due to how the `DraftStateMixin` fetches real PKs via `get_latest_revision_as_object`.

```python
    def get_latest_revision_as_object(self):
        if not self.has_unpublished_changes:
            # Use the live database copy in preference to the revision record, as:
            # 1) this will pick up any changes that have been made directly to the model,
            #    such as automated data imports;
            # 2) it ensures that inline child objects pick up real database IDs even if
            #    those are absent from the revision data. (If this wasn't the case, the child
            #    objects would be recreated with new IDs on next publish - see #1853)
            return self

        latest_revision = self.get_latest_revision()

        if latest_revision:
            return latest_revision.as_object()
        else:
            return self
```

I think this fix here addresses this well, and I don't think we have any other many-many joins from pages in inline fields like this.

cms/methodology/tests/test_forms.py
```diff
@@ -0,0 +1,92 @@
...
+from django.urls import reverse
+from wagtail.test.utils import WagtailTestUtils
+from wagtail.test.utils.form_data import nested_form_data, rich_text, streamfield
+
+from cms.methodology.tests.factories import MethodologyPageFactory
+from cms.taxonomy.models import GenericPageToTaxonomyTopic
+from cms.taxonomy.tests.factories import TopicFactory
+
+
+class MethodologyTopicPublishUnpublishTestCase(WagtailTestUtils, TestCase):
```
BJacksonONS: The test itself is good, I just wonder if we should move it somewhere else and parametrise to run against a few different page types as I've been able to reproduce against Information and Article Series Index pages also.

I suppose given the fix is in the generic through model it probably shouldn't matter but would give extra peace of mind?

BJacksonONS (approved):

cms/methodology/tests/test_forms.py
kevinhowbrook: +1

MaciekBaron: I've moved the tests to the `taxonomy` app and added a couple of additional page types to test against

[3945cdd](https://github.com/ONSdigital/dis-wagtail/pull/697/commits/3945cdd6c85d2b6cf717ffe0876695618f13a986)

cms/taxonomy/models.py
```diff
@@ -129,3 +129,19 @@ class Meta:
         constraints: ClassVar[list[BaseConstraint]] = [
             UniqueConstraint(fields=["page", "topic"], name="unique_generic_taxonomy")
         ]
+
+    def save(self, **kwargs: Any) -> None:
+        """Silently deduplicates when modelcluster tries to INSERT a (page, topic) pair that was already
+        committed by a concurrent save/session.
```
kevinhowbrook, line 137:
```suggestion
        committed by a concurrent save/session.
        revisit when https://github.com/wagtail/wagtail/issues/14359 is addressed
```

cms/taxonomy/tests/test_mixins.py
```diff
@@ -205,3 +215,100 @@ def test_no_exclusivity_validation(self):
             info_page2.full_clean()
         except ValidationError:
             self.fail("Expected no ValidationError for shared topics in GenericTaxonomyMixin.")
+
+
+class TestGenericTaxonomyMixinIntegrityError(WagtailTestUtils, TestCase):
+    """Regression tests for the duplicate (page, topic) IntegrityError."""
```
kevinhowbrook:
```suggestion
    """Regression tests for the duplicate (page, topic) IntegrityError.
    ref https://github.com/wagtail/wagtail/issues/14359
    """
```

kevinhowbrook (approved):
Nice one @MaciekBaron - just a suggestion to add the issue link in some docstrings from me, but otherwise lgtm

BJacksonONS (approved):

cms/taxonomy/models.py
```diff
@@ -129,3 +129,20 @@ class Meta:
...
+        Revisit when https://github.com/wagtail/wagtail/issues/14359 is addressed.
+        """
+        if not kwargs.get("force_insert") and self._state.adding and self.page_id and self.topic_id:
+            existing_pk = (
+                GenericPageToTaxonomyTopic.objects.filter(page_id=self.page_id, topic_id=self.topic_id)
+                .values_list("pk", flat=True)
+                .first()
+            )
```
MebinAbraham, line 146: For this, I suggest we use `force_write_db_for`. A replication lag could cause this to return None resulting in the same issue.

However, I'm thinking whether it would still be possible for a concurrent request to cause this to error still, and I wonder whether we should actually handle the `IntegrityError`. That way the db is the source of truth and we should be able to avoid the race condition. It's not ideal to catch the error, but it avoids the race condition and is an edge case. Thoughts?

i.e.

```py
    try:
        return super().save(**kwargs)
    except IntegrityError:
        existing_pk = type(self).objects.only("pk").get(
            page_id=self.page_id,
            topic_id=self.topic_id,
        ).pk
        self.pk = existing_pk
        self._state.adding = False
```

sanjeevz3009 (approved):
Other than Mebin's comment, LGTM!

cms/taxonomy/models.py
MaciekBaron: I added `force_write_db_for` to address the potential replication lag issue here: [1e9d40f](https://github.com/ONSdigital/dis-wagtail/pull/697/commits/1e9d40f38862dc3adeefc7b21d9402391b6514ef)

However, I don't think the approach with `IntegrityError` works for two reasons:
* `IntegrityError` is too broad here - it could also be raised for a FK violation (page_id or topic_id doesn't exist). This could then raise a `DoesNotExist` when executing the `get()`
* When the `IntegrityError` is raised, the transaction is already "poisoned" and we wouldn't be able to resave. We would need to wrap the statement in `with transaction.atomic():` to create a save-point. See: https://docs.djangoproject.com/en/6.0/topics/db/transactions/#handling-exceptions-within-postgresql-transactions
> (...) when a call to a PostgreSQL cursor raises an exception (typically `IntegrityError`), all subsequent SQL in the same transaction will fail with the error “current transaction is aborted, queries ignored until end of transaction block”

I think the `force_write_db_for` approach is good for now 👍 Let me know what you think.

sanjeevz3009: Was going to resolve on behalf, but @MebinAbraham is back, so I will leave it up to him.

MebinAbraham: Nit: Ideally we should have a test or at least a comment so devs don't accidently remove it. @MaciekBaron

MebinAbraham: Good shout on the wider issues of catching `IntegrityError`. Happy with forcing write db 👍

MaciekBaron: Added test here: [f397b3c](https://github.com/ONSdigital/dis-wagtail/pull/697/commits/f397b3ca355600e7ab45bfebfa5e11feee42c440)

MebinAbraham (approved):

PR #702, opened 2026-06-29
feat: upgrade DS to the preview of 74.0.0, and update templates

helenb:

https://officefornationalstatistics.atlassian.net/browse/CMS-1266

Upgrading the DS with the breaking change version, 74.0.0

This PR currently uses the preview (un-released) version, 74.0.0-next.7. This is a deliberate decision to allow us to unify our templates and get this PR merged ahead of the official release. We'll still be able to do point upgrades when these are release against the DS `main` branch, as they will also be merged into `next`.

- Update base templates to remove overrides
- Updates all page templates to use the new blocks.
- Removes unused CSS overrides. Note that we may be able to also remove the grid overrides once the changes from https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-903 are included.

The following issues have been identified that need addressing on the DS side:

- Issue with fonts in the DS releases (the font was using the fallback of Helvetica not Open Sans) - **now addressed**
- Requirement for another block after the hero block, which is needed on the statistical article page, methodology page and release calendar page - **now addressed**
- Requirement for another block before the breadcrumb block, so that we can display the `show_localised_version_not_available_notice` in the correct position - **now addressed**
- A missing parameter in the base macro for header.search, which means the search icon was previously not visible on this PR.  - **now addressed**

### How to review

Have a look through the code, and test as many different page types as you can in order to spot any issues.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


cms/jinja2/templates/pages/release_calendar/release_calendar_page.html
```diff
@@ -64,10 +69,13 @@
         })
     }}
     {# fmt:on #}
+{% endblock %}
+
+{% block postHero %}
```
helenb, line 61: This is the new block which ~needs adding~ has now been added on the DS side.

cms/jinja2/templates/base.html
```diff
@@ -62,35 +62,6 @@
     {% endif %}
 {% endblock %}
 
-{% block header %}
```
helenb, line 68: We don't want to explicitly override the header call from the DS base template, as otherwise we can't then override the breadcrumb block separately on each page template. ~The only issue that needs resolving here is ensuring that the DS call to `onsHeader` correctly passes in `"search": pageConfig.header.search` - there's a fix for this coming shortly.~ - now resolved.

cms/jinja2/templates/base_page.html
```diff
@@ -70,57 +70,17 @@
     {%- set announcement_text -%}
         {% trans trimmed in_language=language_labels[request.LANGUAGE_CODE] %}This page is currently not available {{in_language}}. It is displayed in its original language.{% endtrans %}
     {%- endset -%}
-    {# Adds a full-width section for the header area before the main page content #}
-    {# Changes the location of the <main> tag from the base template - ensures it surrounds the header area #}
-    {# This change should be adopted by the design system soon #}
-    <main id="main-content">
```
helenb, line 76: We now no longer need to override the DS base page template as the a11y reasons for doing so have been resolved. The 'show_localised_version_not_available_notice' here moves to the `preBreadcrumbs` block, which is another new block ~needed~ now added on the DS side.

cms/jinja2/templates/components/preview-bar/preview-bar.html
```diff
@@ -2,41 +2,43 @@
 {% from "components/select/_macro.njk" import onsSelect %}
 {% from "components/button/_macro.njk" import onsButton %}
 
-{% call onsPanel({"classes": "ons-container ons-u-mb-xs preview-bar"}) %}
-    <div class="ons-grid">
+<section aria-label="Preview items">
```
helenb, line 5: As this is rendered outside the `<header>` and `<main>`, it needs to be rendered in a landmark level element in order to be easily discovered by screenreaders.

cms/jinja2/templates/base_page.html
```diff
@@ -70,57 +70,17 @@
...
-                    </h1>
-                {% endif %}
-            </div>
-        {% endblock %}
-
-        <div class="ons-page__container ons-container{{ containerClasses }}">
-            <div class="ons-grid">
-                <div class="ons-grid__col ons-col-{{ pageColNumber }}@m {{ pageColClasses }}">
-                    {# This is the main tag in the base template in the design system #}
-                    <div class="ons-page__main ons-u-mb-5xl {{ mainColClasses }}">
```
helenb, line 106: The custom bottom margin here moves to `mainColClasses`

cms/jinja2/templates/base_page.html
```diff
@@ -70,57 +71,41 @@
...
-                {% include "templates/components/navigation/breadcrumbs.html" %}
-                {% if not render_heading_in_content %}
-                    <h1 class="ons-u-fs-3xl common-header__heading">
-                        {{ page.title }}
-                    </h1>
-                {% endif %}
-            </div>
-        {% endblock %}
+{# This block should be overridden on individual page templates and include the breadcrumbs #}
+{% block breadcrumbs %}
```
helenb, line 87: I think this is overridden everywhere so I'm not sure we need it. For now I have fixed the untranslated aria-label in the breadcrumbs include (this was previously used on the cookies page so that is currently incorrect on live, but in this pr the cookies page now uses its own breadcrumb override).

cms/jinja2/templates/base_page.html
```diff
@@ -70,57 +71,41 @@
...
-                        {{ onsBackToTop({
-                            "description": back_to_top,
-                        }) }}
-                        {# fmt:on #}
-                    </div>
-                </div>
-            </div>
+{# Default heading if hero not used #}
+{% block hero %}
+    {% if not render_heading_in_content %}
```
helenb, line 93: I have kept the render_heading_in_content functionality but I don't think there is any need for it. I am not sure why it was needed previously, but the only place it is used is the cookies page - but we could just as easily override the hero block on the cookies template.

BJacksonONS: There's a comment thread [here](https://github.com/ONSdigital/dis-wagtail/pull/581#discussion_r3086593319) where that was added. IIRC headers were being rendered in the wrong element of the page?

helenb: Ok I think that now that the the `main` tag is rendered higher up the tree this should not be an issue. We could remove the default headings and breadcrumbs from the base template now (including the breadcrumbs include component), and simply use the `{% breadcrumbs %}` and `{% hero %}` block on the cookies page. But interested to hear if others agree.

cms/jinja2/templates/pages/wagtail/password_required.html
```diff
@@ -6,7 +6,7 @@
 
 {% block body_class %}template-password-required{% endblock %}
 
-{% block main %}
+{% block mainContent %}
     <h1>Password required</h1>
```
helenb: Might be good to move this h1 to the hero block for consistency with my proposed change for the cookie template.

MebinAbraham:
Marked as ready for review, as we want people to have a look, but marked as do not merge since it is prerelease.

cms/jinja2/templates/components/preview-bar/preview-bar.html
```diff
@@ -2,41 +2,43 @@
...
-            {# fmt:on #}
+                {{-
+                    onsButton({
+                        "text": "Back to bundle",
+                        "url": bundle_inspect_url,
+                        "variants": ["small"],
+                        "iconType": "exit",
+                    })
+                -}}
+                {# fmt:on #}
```
MaciekBaron: Minor: Let's match the indentation level of `fmt:off` above
```suggestion
            {# fmt:on #}
```

cms/jinja2/templates/components/navigation/main_menu_preview.html
```diff
@@ -2,5 +2,5 @@
 
 {% set main_menu = object %}
 
-{% block main %}{% endblock %}
+{% block mainContent %}{% endblock %}
 {% block footer %}{% endblock %}
```
MaciekBaron: Incredibly minor, but since we are touching this file we could add the missing new line at the end to match other files

MaciekBaron (approved):
Just a couple of very minor comments

sanjeevz3009 (approved):
Did a initial code review. LGTM! Will pull in @joepercival @charlotteromaniuk for a functional review. More the eyes the better.

helenb:
@MebinAbraham I notice this PR is failing based on the branch name, but I thought we were allowing older PRs that were created before the new naming standards to go through? I'd have to create a new pull request if I update the branch name.

cms/jinja2/templates/components/preview-bar/preview-bar.html
helenb: Addressed in ffa6a8515b17c2da34c4e10257b68cc8e786ef21

cms/jinja2/templates/components/navigation/main_menu_preview.html
helenb: Addressed in ffa6a8515b17c2da34c4e10257b68cc8e786ef21

cms/jinja2/templates/base_page.html
helenb: @MebinAbraham any input on this?

MebinAbraham:
> @MebinAbraham I notice this PR is failing based on the branch name, but I thought we were allowing older PRs that were created before the new naming standards to go through? I'd have to create a new pull request if I update the branch name.

You just need to add the `no-jira` label. Have added it.

cms/jinja2/templates/base_page.html
helenb, line 18: As a future piece of work we could investigate whether this override is now necessary - there is a ticket in progress on the DS side to make the og:image configurable.

MebinAbraham: Ya, it is not needed. The DS card was created of the back of: https://officefornationalstatistics.atlassian.net/browse/CMS-579. So we have a card on our side to track also 👍

cms/jinja2/templates/base_page.html
MebinAbraham: Will have a look and see if it is redundant now.

cms/jinja2/templates/base_page.html
helenb: @MebinAbraham are you happy for me to remove [templates/components/navigation/breadcrumbs.html](https://github.com/ONSdigital/dis-wagtail/blob/main/cms/jinja2/templates/components/navigation/breadcrumbs.html)?

cms/jinja2/templates/pages/release_index.html
```diff
@@ -1,8 +1,25 @@
 {% extends "templates/base_page.html" %}
-{% block header_area %}
```
helenb, line 2: Agreed with @sanjeevz3009 to keep this template as a placeholder, even though the release index page is not served from Wagtail. I have updated it to override the hero and breadcrumbs blocks rather than override the old `header_area` block.

cms/jinja2/templates/base.html
MebinAbraham: Does the page config we set still match the default in the design system?

Secondly, now we are not explicitly passing this config; you would need to update the `get_page_config` unit test to ensure all the "keys" we care about are tested. If someone changes the key names, then this would break the DS rendering as they no longer align. This previous component call worked as a mapping layer, which we no longer have; therefore, we need to have confidence in the tests. Worth adding a code comment to `get_page_config` that the config **MUST** align like-for-like with the DS config. Ideally, `get_page_config` should return a TypedDict.

cms/jinja2/templates/base_page.html
MebinAbraham: @helenb if you set `{% set render_heading_in_content = false %}` in cookies.html then view `/cookies` you can see the UX issue. If they can be fixed, then we could drop this. The h1 seems to be in a diff containers and that doesn't seem to have enough padding/margin and the margin bottom also looks too large.

cms/jinja2/templates/base_page.html
MebinAbraham: It looks like every page is duplicating this. The only difference in some of them seems to be the "variant", which should be part of the page config. So what we should do is drop the override from everything and set the variant at the python page config with a default value. Then the breadcrumb only needs to be defined in 1 place and only needs to be overridden if it is completely custom, which we should never really for breadcrumbs.

To be pragmatic, I suggest deduplicating it and using the base only as a separate follow up work.

cms/jinja2/templates/base_page.html
helenb: If you set `{% set render_heading_in_content = false %}` then it defaults to the heading rendered in the base page template on line 94, which doesn't have the same top margin set as when it is rendered at line 105 because in that scenario it is wrapped in the `ons-page__main` class. So we could just render the heading there. But is there a reason we can't use the hero on the cookies page?

Screenshot 2026-07-27 at 14 07 30

cms/jinja2/templates/pages/wagtail/password_required.html
helenb: @MebinAbraham again I would suggest we use the hero throughout so use it on the password required page too.

MaciekBaron (approved):

cms/jinja2/templates/base_page.html
MebinAbraham: Not our decision to make. It would need to be run past Amelia and Paula / DP as it's a decision for them to make.
Personally, I think it looks odd due to the amount of empty space. See what they say, if they are happy then technically no reason for us to not change. I would frame it from a lens of "should every page have a breadcrumb then?", if not, we still need to solve the wider issue.

cms/jinja2/templates/base_page.html
helenb: It would be helpful if you could give an example of how you would set that on an individual template - or do you mean it would be part of the `get_breadcrumbs` config? The `variant` is set at the breadcrumbs and hero component level, rather than being something passed down from the page config. In the mean time, in order to de-dupe, can I clarify that you mean to remove the included template for now?

cms/jinja2/templates/pages/wagtail/password_required.html
MebinAbraham: This html page is basically not used yet (feature flagged off), plus, it's a feature we won't have on anytime soon, so I'm happy for you update as needed.

cms/jinja2/templates/base_page.html
helenb: OK rather than hold this up by seeking approval for a change, we need to solve the issue that rendering a heading inside the hero block *without using the hero* doesn't work well, as the margins are incorrect and there is too much space below the title.

We either need a default in the base template that renders the heading in a hero, unless `render_heading_in_content` is defined (Option 1), *or* we only implement the positive case in the base template in which case we trust child templates to either override the hero block or set `render_heading_in_content` to be `true` (Option 2).

**Option 1**
```
{# Default heading if hero not overridden #}
{% block hero %}
    {% if not render_heading_in_content %}
        {# fmt:off #}
        {{
            onsHero({
                "title": page_heading or page.title,
                "variants": 'grey'
            })
        }}
        {# fmt:on #}
    {% endif %}
{% endblock %}

{% block mainContent %}
    {% if render_heading_in_content %}
        <h1 class="ons-u-fs-3xl ons-u-mb-xl common-header__heading">
            {{ page_heading or page.title }}
        </h1>
    {% endif %}
{% endblock %}
```

**Option 2**
```
# Hero block in example above is removed
{% block mainContent %}
    {% if render_heading_in_content %}
        <h1 class="ons-u-fs-3xl ons-u-mb-xl common-header__heading">
            {{ page_heading or page.title }}
        </h1>
    {% endif %}
{% endblock %}
```

helenb: Option 1 covers belt and braces so I'll implement that for now.

helenb: Implemented in bc7884ecdde73dba8faa896d04d494b004d8a5cd

cms/jinja2/templates/pages/wagtail/password_required.html
helenb: It looked a bit odd with a hero so I've used the same technique as the cookie page here - see ce45c7309f47070d4734d4db1005c73508b149bb

cms/jinja2/templates/base_page.html
```diff
@@ -70,57 +72,44 @@
...
-                {% include "templates/components/navigation/breadcrumbs.html" %}
-                {% if not render_heading_in_content %}
-                    <h1 class="ons-u-fs-3xl common-header__heading">
-                        {{ page.title }}
-                    </h1>
-                {% endif %}
-            </div>
-        {% endblock %}
+{# Default heading if hero not overridden #}
+{% block hero %}
```
BJacksonONS, line 92: The way this is implemented now means error pages and previews (snippets at least) will render a blank hero with an empty h1 tag.

Image

cms/jinja2/templates/pages/methodology_page.html
```diff
@@ -11,7 +11,19 @@
...
+{% block breadcrumbs %}
+    {# fmt:off #}
+    {{
+        onsBreadcrumbs({
+            "ariaLabel": breadcrumbs_aria_label,
+            "itemsList": page.get_breadcrumbs(request=request),
+            "variant": 'grey'
+        })
+    }}
+    {# fnt:on #}
```
BJacksonONS: Typo here

cms/jinja2/templates/pages/statistical_article_page.html
```diff
@@ -136,46 +142,50 @@
...
-                "openText": openText,
-                "closeText": closeText,
-                "detailsItems": corrections_and_notices
-            })
-        }}
-        {# fmt:on #}
+{% block postHero %}
 
-    {% endif %}
+    <section>
```
BJacksonONS, line 135: Should this also have an aria-label?

cms/jinja2/templates/base.html
helenb: In answer to the first point, the `get_page_config` tag is now just overriding the elements of the `pageConfig` that we care about, and everything else is defaults. This call to `onsHeader` was overriding the one in the Design System base page template, and in fact was missing a couple of keys because it had fallen out of sync over time. So I am not sure it was acting a mapping layer - in fact it was allowing us to miss potential improvements that came with DS upgrades.

I will look into adding a more rigorous test for the `get_page_config` to ensure that the keys match the DS

cms/jinja2/templates/base_page.html
```diff
@@ -70,57 +72,44 @@
     {%- set announcement_text -%}
         {% trans trimmed in_language=language_labels[request.LANGUAGE_CODE] %}This page is currently not available {{in_language}}. It is displayed in its original language.{% endtrans %}
```
BJacksonONS, line 71: This is in the header not the main body now, which means any screen readers using skip to main content won't pick it up.

cms/jinja2/templates/base_page.html
helenb: Ah... I will take another look at those.

cms/jinja2/templates/base_page.html
```diff
@@ -70,57 +72,44 @@
...
-                    <div class="ons-page__main ons-u-mb-5xl {{ mainColClasses }}">
-                        {% if render_heading_in_content %}
-                            <h1 class="ons-u-fs-3xl ons-u-mb-xl common-header__heading">
-                                {{ page_heading or page.title }}
-                            </h1>
-                        {% endif %}
-                        {% block main %}{% endblock %}
-                        {# used in small viewports #}
-                        {# fmt:off #}
-                        {{ onsBackToTop({
```
BJacksonONS, line 115: Should this be re-added somewhere? Think this was the only use of this component but I thought we wanted to keep this functionality.

cms/jinja2/templates/base.html
helenb: Added more tests in 0e039395c7726efaf63e63d9d8276a946b49b051. I have tried a `TypedDict` for the `get_page_config` return value in 8d757abb95638d10de36bfa06a9808d70a324d91 - let me know if this is what you had in mind.

cms/jinja2/templates/base_page.html
helenb: Yes you are right this is an oversight. Ideally I would add another block on the DS side to house this - maybe as `postMain` but as that can't happen for a little while while I may need to add this on each page template.

cms/jinja2/templates/base_page.html
MebinAbraham: > It would be helpful if you could give an example of how you would set that on an individual template - or do you mean it would be part of the get_breadcrumbs config? The variant is set at the breadcrumbs and hero component level, rather than being something passed down from the page config.

You can basically follow the same pattern as the analytics content type property. You can set a value on the `BasePage` class (default to None), then override the variant on a page-by-page basis.
- https://github.com/ONSdigital/dis-wagtail/blob/main/cms/core/models/base.py#L85
- https://github.com/ONSdigital/dis-wagtail/blob/main/cms/articles/models.py#L94

1. Python:

```py
class BasePage(PageLDMixin, ListingFieldsMixin, SocialFieldsMixin, Page):
      """Base page class with listing and social fields additions as well as cache decorators."""
  
      ...
  
      # Breadcrumb variant for the frontend component. Override in subclasses.
      breadcrumbs_variant: ClassVar[str | None] = None

class InformationPage(BasePage):
      """A generic information page model."""
  
      breadcrumbs_variant = "grey"
  
      ...
```

2. Then in the base HTML you can do:

```
{{
    onsBreadcrumbs({
        "ariaLabel": breadcrumbs_aria_label,
        "itemsList": page.get_breadcrumbs(request=request),
        "variant": page.breadcrumbs_variant
    })
}}
```

3. Now you can remove `{% block breadcrumbs %}` from `information_page.html` and repeat for all others.

---

> In the mean time, in order to de-dupe, can I clarify that you mean to remove the included template for now?

You want to keep the base_template one and remove all other override unless there is a reason to keep.

cms/jinja2/templates/base_page.html
helenb: This is now addressed in 5a20f12ac0cb4c1f4bc12ea0fd59f602c8d2981e

cms/jinja2/templates/pages/statistical_article_page.html
helenb: I've just tested the statistical article page in a screen reader (Voiceover). Without an aria-label here it moves straight to read out "Corrections show details" after the release date information. I think that is OK and the only improvement might be to see if we could improve what it reads out so "Corrections" sounds like a heading on its own - but that would be a change to the `onsDetails` component.

cms/jinja2/templates/base_page.html
helenb: I think that is unavoidable with the breadcrumbs moving to the header, as this has to appear before the breadcrumbs in the design.

cms/jinja2/templates/base_page.html
helenb: For now I've done the de-duping - default to the grey variant, and only override the breadcrumbs component when needed. See b38cd2d07c97f76144ecb80dcfecbcbd3db31824

cms/jinja2/templates/pages/methodology_page.html
helenb: This has now gone with the breadcrumbs de-duping

cms/jinja2/templates/base.html
MebinAbraham: > So I am not sure it was acting a mapping layer - in fact it was allowing us to miss potential improvements that came with DS upgrades.

Yeah, that doesn't change the fact that we were using it as a mapping layer (it mapped our keys to DS keys). The consequence is exactly that; we can miss changes if we don't keep them aligned for new keys.

> Added more tests in https://github.com/ONSdigital/dis-wagtail/commit/0e039395c7726efaf63e63d9d8276a946b49b051. I have tried a TypedDict for the get_page_config return value in https://github.com/ONSdigital/dis-wagtail/commit/8d757abb95638d10de36bfa06a9808d70a324d91 - let me know if this is what you had in mind.

👍

cms/jinja2/templates/base_page.html
helenb: Added to every template in 5c1424de8897da32ac854f0cd5d2d9a49d004974. I'll raise a ticket to tidy this up once we have the new block in the DS.

cms/jinja2/templates/base_page.html
MebinAbraham: Moving to Python would have been ideal to avoid the component overrides and to be consistent with existing but given it's significantly better than main, not a blocker 👍

cms/jinja2/templates/pages/errors/403.html
```diff
@@ -1,10 +1,10 @@
-{% extends "templates/base_page.html" %}
+{% extends "templates/pages/errors/base_error.html" %}
 
 {% block title %}Access Denied (403){% endblock %}
 
-{% block mainContent %}
-    <h1>Access Denied</h1>
+{% set title = _("Access Denied") %}
 
+{% block error_content %}
```
MebinAbraham, line 5: 👍

cms/jinja2/templates/base_page.html
MebinAbraham: Must it be in the mainContent, or is anywhere on the page fine, cause it could be added to `preFooter`, which is after `mainContent`, so we get what we need and don't need to duplicate in all templates and potentially don't even need a DS change. https://github.com/ONSdigital/design-system/blob/main/src/layout/_template.njk#L221

MebinAbraham:

cms/jinja2/templates/base_page.html
helenb: I did try in preFooter but it doesn't work so well from a spacing point of view, and semantically it really ought to be inside the `<main>` tag.

cms/jinja2/templates/pages/errors/404.html
```diff
@@ -1,5 +1,5 @@
 {% extends "templates/pages/errors/base_error.html" %}
-{% set page_title = _("Page not found") %}
+{% set title = _("Page not found") %}
```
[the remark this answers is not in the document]
helenb: Addressed in 21b8988c0a3fd48383c0d0e9bca0c2f8c1af2181

cms/jinja2/templates/pages/errors/500.html
```diff
@@ -1,5 +1,5 @@
 {% extends "templates/pages/errors/base_error.html" %}
-{% set page_title = _("Sorry, there’s a problem with the service") %}
+{% set title = _("Sorry, there’s a problem with the service") %}
```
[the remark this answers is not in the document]
helenb: Addressed in 21b8988c0a3fd48383c0d0e9bca0c2f8c1af2181

cms/jinja2/templates/pages/errors/403.html
```diff
@@ -1,10 +1,10 @@
-{% extends "templates/base_page.html" %}
+{% extends "templates/pages/errors/base_error.html" %}
 
 {% block title %}Access Denied (403){% endblock %}
 
-{% block main %}
-    <h1>Access Denied</h1>
+{% set title = _("Access Denied") %}
```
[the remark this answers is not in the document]
helenb: Addressed in 21b8988c0a3fd48383c0d0e9bca0c2f8c1af2181

cms/jinja2/templates/base_page.html
```diff
@@ -55,7 +55,7 @@
     {{ block.super }}
 {%- endblock -%}
 
-{% block pageContent %}
+{% block preBreadcrumbs %}
     {#- "Welsh" in Welsh is Cymraeg, but when we say "in Welsh" the noun uses soft mutation and becomes Gymraeg and the
     resulting phrase reads "yn y Gymraeg". Therefore we can't just translate the name of the language, so we use the
```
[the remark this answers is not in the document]
helenb: `pageContent` is a valid block still. So this template just overrides more of the page than other templates - probably fine for the dataset preview.

cms/jinja2/templates/pages/wagtail/password_required.html
```diff
@@ -6,9 +6,12 @@
 
 {% block body_class %}template-password-required{% endblock %}
 
-{% block main %}
-    <h1>Password required</h1>
+{# Render the page heading in the main content area #}
+{% set render_heading_in_content = true %}
+{% set title = "Password required" %}
```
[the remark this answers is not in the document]
helenb: Addressed in 21b8988c0a3fd48383c0d0e9bca0c2f8c1af2181

cms/jinja2/templates/base_page.html
helenb: New ticket: https://officefornationalstatistics.atlassian.net/browse/CMS-1338

cms/jinja2/templates/pages/errors/403.html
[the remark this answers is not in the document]
BJacksonONS: The `{% block title %}Access Denied (403){% endblock %}` is still in the template at line 3, not sure it actually does anything though.

cms/jinja2/templates/base_page.html
BJacksonONS: Think `definition_preview.html` and `login_page.html` still inherit this in a way that would show a banner but the login should be unreachable and I think the preview is low enough priority I'm happy to leave for now.

cms/jinja2/templates/pages/errors/404.html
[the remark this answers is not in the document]
BJacksonONS: LGTM

cms/jinja2/templates/pages/errors/500.html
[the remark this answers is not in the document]
BJacksonONS: LGTM

cms/jinja2/templates/base_page.html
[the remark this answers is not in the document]
BJacksonONS: LGTM

cms/jinja2/templates/pages/wagtail/password_required.html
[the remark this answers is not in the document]
BJacksonONS: LGTM

cms/jinja2/templates/pages/errors/403.html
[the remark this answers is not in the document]
helenb: Addressed in da4fbdc2fefb8b37070d15256b5990a80680a659

BJacksonONS (approved):

charlotteromaniuk (approved):
Functionally tested and checked for visual regressions - LGTM

MebinAbraham (approved):

PR #712, opened 2026-07-06
fix(csp): align iframe frame-src with allowed domain validation

MebinAbraham:

Updates the iframe CSP configuration so `frame-src` matches the existing iframe URL validation rules.

Previously, `IFRAME_VISUALISATION_ALLOWED_DOMAINS` was used directly in CSP, which only allowed exact hosts in `frame-src` meaning domains such as `www.ons.gov.uk` was failings to render visualisaiton.
The Iframe field validation already accepts the apex domain and any subdomain via `is_hostname_in_domain()` so this ensure the CSP policy matches that logic.

This PR adds a derived `IFRAME_VISUALISATION_CSP_SOURCES` setting that expands each allowed domain to:
- `example.com`
- `*.example.com`

This keeps the existing validation behaviour and makes CSP consistent with it.

### How to review

Inspect the response for a Wagtail page and ensure the CSP header shows `frame-src 'self' ons.gov.uk *.ons.gov.uk`

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

- Ensure iframes are now rendering correctly in Sandbox post deploy.

### DIFF
--- a/cms/settings/base.py
+++ b/cms/settings/base.py
@@ -1004,6 +1004,14 @@ if "IFRAME_VISUALISATION_ALLOWED_DOMAINS" in env:
 else:  # Default to ONS allowed link domains if not set
     IFRAME_VISUALISATION_ALLOWED_DOMAINS = ONS_ALLOWED_LINK_DOMAINS
 
+# CSP wildcards do not match the apex domain, so include both `example.com`
+# and `*.example.com` to mirror iframe URL validation.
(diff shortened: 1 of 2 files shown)


BJacksonONS (approved):

sanjeevz3009 (approved):
LGTM

PR #713, opened 2026-07-06
chore(deps): upgrade Wagtail to 7.4.3

kevinhowbrook:

Ticket [1260, part 1](https://officefornationalstatistics.atlassian.net/browse/CMS-1260)

Upgrades Wagtail from 7.3.3 to 7.4.2 (LTS). Also bumps `django-redis` 6 to 7 to align with `redis-py` 8.x.

Four fixes were required alongside the upgrade:

- **modelsearch 1.3.1** (bundled with Wagtail 7.4): `FilterField` on a `@cached_property` is now rejected. Replaced `FilterField("team_id")` in `Bundle.search_fields` with `RelatedFields("teams", [FilterField("team_id")])` and removed the workaround property.
- **Wagtail 7.4 merges wagtail/wagtail#13948**: `CancelWorkflowMenuItem` now correctly appears in the action menu during the review stage. Updated two count assertions (4 to 5) and enabled the previously-commented-out assertions.
- **redis-py 8.x unix socket change**: `ping()` silently returns `True` against a non-existent unix socket path. Switched the health-check test `LOCATION` values to refused TCP ports so `ConnectionError` is raised correctly.
- **Orphaned `ignore_required_constraints` parameter**: Wagtail 7.0 added this parameter to `StreamBlock.clean()`; Wagtail 7.4 removed it in favour of `defer_required_validation()`. Two custom block overrides (`DatasetStoryBlock`, `TimeSeriesPageStoryBlock`) still had the old signature, causing a pylint CI failure.

### Deferred StreamField validation

Wagtail 7.4 introduces [deferred validation for StreamField blocks when saving drafts](https://docs.wagtail.org/en/stable/releases/7.4.html#deferred-validation-for-streamfield-blocks-when-saving-drafts). On a plain "Save draft", and on autosave which posts through the same path, `FieldBlock.defer_required_validation()` temporarily sets `field.required = False`, so required fields inside StreamField blocks are no longer enforced until publish or workflow submission. A publisher could add a block, leave a required field empty, save, and see no error.

This PR restores the previous behaviour by adding `required_on_save=True` to the required leaf fields. 

**What this does not restore.** `required_on_save` only exists on `FieldBlock`, so three container-level checks stay deferred to publish time no matter how many leaves are annotated:

1. A StreamField's own "This field is required." when zero blocks have been added.
2. A nested StreamBlock's emptiness check, for example a section saved with a title and an entirely empty content stream.
3. `block_counts` and `min_num` minimums, for example a documents block containing zero documents.

### How to review

- Functional and unit tests all pass
- Boot the app and confirm the Wagtail admin loads
- Bundle action menu buttons render and work (Save as draft, Approve, etc.)
- Add a block to a page's content, for example a section on a Methodology page, leave the section heading empty and click Save draft. A validation error should appear.
- Save a page whose content StreamField is entirely empty. No error appears on save, and the error appears on publish. This is expected, see "What this does not restore" above.
- If reviewing the annotations: check each `required_on_save=True` sits on a leaf field and not on a `StructBlock`, `StreamBlock` or `ListBlock`, where it is silently ignored.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

- PT2: other Python security package updates (like wagtail polymath)

### Other/Misc

Initial (and extensive) investigation notes archived [here](https://gist.github.com/kevinhowbrook/06665741032a90ef6b32e21a9cadd536)


cms/bundles/models.py
```diff
@@ -224,12 +224,6 @@ def scheduled_publication_date(self) -> datetime.datetime | None:
     def active_team_ids(self) -> list[int]:
         return list(self.teams.filter(team__is_active=True).values_list("team__pk", flat=True))
 
-    @cached_property
-    def team_id(self) -> list[int]:
```
kevinhowbrook, line 228: This was fixed for 7.4, follow https://github.com/wagtail/django-modelsearch/pull/83

cms/workflows/tests/test_workflow_tweaks.py
```diff
@@ -423,11 +423,10 @@ def test_group_review_task_actions(self):
         response = self.client.get(self.edit_url)
         menu_items = response.context["action_menu"].menu_items
 
-        self.assertEqual(len(menu_items), 4)
+        self.assertEqual(len(menu_items), 5)
 
         self.assertItemWithPropertyIn("name", "action-unpublish", menu_items)
-        # TODO: uncomment when https://github.com/wagtail/wagtail/pull/13948 is fixed
-        # self.assertItemWithPropertyIn("name", "action-cancel-workflow", menu_items)
```
kevinhowbrook, line 430: **praise to the commenter here**

BJacksonONS:
Going over the changes in wagtail to understand how this all works so just have questions more than anything right now. 

Did you try overriding clean rather than `defer_required_validation` at any point? Looks like `StreamBlock` clean decides whether to run based on

`self.required and not self.is_deferred_validation`

 I think we can partly close the gaps you mentioned with an override in just `SectionStoryBlock`, `SectionContentBlock`, `CoreStoryBlock` and `CoreSectionContentBlock`. Could be put in a mixin possibly. Maybe would also need to be used else where to re-enforce min counts etc.

```python
def clean(self, value: StreamValue) -> StreamValue:
    deferred = self.is_deferred_validation
    self.is_deferred_validation = False
    try:
        return super().clean(value)
    finally:
        self.is_deferred_validation = deferred
```

This keeps the behaviour that a block is required in the content block of Information and Methodology pages, as well as `RichTextBlock`s.

Not sure if the business had a preference either way though so maybe it's just not necessary.

image

image

kevinhowbrook:
> Going over the changes in wagtail to understand how this all works so just have questions more than anything right now.
> 
> Did you try overriding clean rather than `defer_required_validation` at any point? Looks like `StreamBlock` clean decides whether to run based on
> 
> `self.required and not self.is_deferred_validation`
> 
> I think we can partly close the gaps you mentioned with an override in just `SectionStoryBlock`, `SectionContentBlock`, `CoreStoryBlock` and `CoreSectionContentBlock`. Could be put in a mixin possibly. Maybe would also need to be used else where to re-enforce min counts etc.
> 
> ```python
> def clean(self, value: StreamValue) -> StreamValue:
>     deferred = self.is_deferred_validation
>     self.is_deferred_validation = False
>     try:
>         return super().clean(value)
>     finally:
>         self.is_deferred_validation = deferred
> ```
> 
> This keeps the behaviour that a block is required in the content block of Information and Methodology pages, as well as `RichTextBlock`s.
> 
> Not sure if the business had a preference either way though so maybe it's just not necessary.
> image image

Thanks @BJacksonONS - I hadn't tried that, I'd assumed it was equivalent to overriding `defer_required_validation` on the StreamBlock.

> Not sure if the business had a preference either way though so maybe it's just not necessary.

Just to respond to this first, chatting earlier about this with @MebinAbraham , the consensus landed on us being ok with forgoing the StreamField 'emptiness' validation, but some more musings below...

I've tested both on my branch against a methodology page. Your `clean()` suggestion does  restore the container-level checks: an empty content StreamField, an empty nested section content stream, and `block_counts` minimums. But it doesn't reach the individual fields: with the `required_on_save=True` annotations removed, I can add a section with an empty heading, or an empty Rich text block, and it saves with no error. (Not sure if you were suggesting an either/or here, just flagging for clarity we would need both)

The one case `clean()` can't reach is things like `FooterMenu.columns`. Declared inline as `StreamField([("column", LinksColumn())])`, the outer block is a plain `wagtail.blocks.StreamBlock` with no class to attach a mixin to, similar situation with `LinksColumn` being a `StructBlock`. That's the `navigation_menus.feature` failure  this PR is fixing, so we'd need `required_on_save` there regardless.

While looking into this initially I also tested a third option: a **no-op** `defer_required_validation`, i.e. literally `pass`. That stops the cascade entirely, so everything below is enforced and no annotations are needed anywhere in that subtree. I tried it on the methodology page with all annotations removed and got full pre-7.4 behaviour from one mixin on one class. The trade-off is it needs applying to each top-level StreamBlock, plus the child `StructBlock` for the seven inline-declared StreamFields.

So there are three states we could proceed with:

1. `required_on_save` annotations only (current branch): field checks, no container checks
2. annotations + your mixin: full parity, nothing already done gets reverted
3. no-op mixin only: full parity, replaces the annotations, ~20 classes, and new required
   fields are enforced automatically in future without anyone remembering a kwarg

Happy with any of the three... but I'm keen to hear your view and everyone else's.

BJacksonONS:
> (Not sure if you were suggesting an either/or here, just flagging for clarity we would need both)

I was imagining it as an additional change.

> no-op mixin only: full parity, replaces the annotations, ~20 classes, and new required
fields are enforced automatically in future without anyone remembering a kwarg

I like the idea of not needing to remember the kwargs, but the current approach is good as it gives us the ability to incrementally adopt the new behaviour where we might want it.

joepercival (approved):
Tested functionally, looks good from my side

BJacksonONS:
Should probably bump to 7.4.3 for security fixes

pyproject.toml
```diff
@@ -13,12 +13,12 @@ requires-poetry = "2.4.1"
 [tool.poetry.dependencies]
 python = "^3.14"
 django = "~6.0.7"
-wagtail = "~7.3.3"
+wagtail = "~7.4.2"
```
kevinhowbrook: TODO - bump to .3 for merge

kevinhowbrook: 35798548603624ca4116469a8da443c941fd34c7

cms/core/tests/test_views.py
```diff
@@ -235,12 +235,12 @@ def test_unexpected_database_error(self):
...
             "default": {
                 "BACKEND": "django_redis.cache.RedisCache",
-                "LOCATION": "redis:///does-not-exist",
+                "LOCATION": "redis://127.0.0.1:1",
                 "OPTIONS": {},
             },
             "other": {
                 "BACKEND": "django_redis.cache.RedisCache",
-                "LOCATION": "redis:///also-does-not-exist",
+                "LOCATION": "redis://127.0.0.1:2",
```
BJacksonONS, line 243: A comment in the code as to why we changed this would be helpful

cms/core/elasticache.py
```diff
@@ -32,6 +32,8 @@ def __init__(self, user: str, cluster_name: str, region: str):
 
         self._session = botocore.session.get_session()
         credentials = self._session.get_credentials()
+        if credentials is None:
+            raise RuntimeError("Unable to resolve AWS credentials for ElastiCache IAM authentication")
```
BJacksonONS, line 36: Would `ImproperlyConfigured` be more consistent with the rest of our config?

BJacksonONS (changes requested):
Annoyingly github won't let me comment on files that aren't modified in the PR.

Would only be a couple non-blocking nits, but just realised we introduced a new `IframeBlock` in `cms/datavis/blocks/iframe.py` while this PR has been open that probably needs `required_on_save` set.

charlotteromaniuk (approved):
Functionally reviewed, apart from Ben's comments LGTM

kevinhowbrook:
> Annoyingly github won't let me comment on files that aren't modified in the PR.
> 
> Would only be a couple non-blocking nits, but just realised we introduced a new `IframeBlock` in `cms/datavis/blocks/iframe.py` while this PR has been open that probably needs `required_on_save` set.

Thanks @BJacksonONS - added in 73fb0ebcf1827abeac8589a104bf3c06b65d2dc3

cms/core/tests/test_views.py
kevinhowbrook: I've added a comment above this to explain it 2e26557a1bc13f88c9a3f138254025315de98d3f

BJacksonONS (approved):

cms/core/elasticache.py
kevinhowbrook: Yep, thanks 800d9eb7af372388871e8dba5c36cb6febfdc199

cms/core/tests/test_views.py
```diff
@@ -235,12 +235,12 @@ def test_unexpected_database_error(self):
...
                 "OPTIONS": {},
             },
```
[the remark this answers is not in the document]
BJacksonONS: Happy with the change and reasoning, have a comment in there also to explain so will resolve on behalf of the team

sanjeevz3009 (approved):
LGMT! Thanks @kevinhowbrook Thanks for the reviews @BJacksonONS, @charlotteromaniuk and @joepercival

PR #717, opened 2026-07-07
ci(github): validate PR branch names and semantic titles

MebinAbraham:

1. Enforces that branch names contain JIRA issue key unless opted out using a label (`no-jira`) at PR stage or by including `no-jira` in branch name. Dependabot PR's are skipped.
2. Ensure the PR title follows [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) standards.
~~Currently the "subject" in the title can be lower or sentence case, we could choose to enforce one. Dependabot currently uses lowercase.~~ (Lower case enforced)

### How to review

Ensure the changes make sense and work as expected.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/.github/workflows/pr-standards.yml
+++ b/.github/workflows/pr-standards.yml
@@ -1,0 +1,53 @@
+name: PR standards
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, edited, labeled, unlabeled]
+
+permissions:
+  contents: read
+
+jobs:
+  branch-name:
+    name: Validate branch name
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          script: |
+            const branchName = context.payload.pull_request?.head?.ref ?? "";
+            const prAuthor = context.actor;
+            const bypassLabel = "no-jira";
+            const labels = (context.payload.pull_request?.labels ?? [])
+              .map((label) => label.name?.toLowerCase())
+              .filter(Boolean);
+
+            const hasBypassLabel = labels.includes(bypassLabel);
+            const isBotBypass = prAuthor === "dependabot[bot]";
+            const jiraIssueKeyPattern = /(CMS|DPD)-\d+/;
+
+            if (hasBypassLabel || isBotBypass) {
+              core.info("Skipping Jira issue key validation.");
+              return;
+            }
+
+            if (!jiraIssueKeyPattern.test(branchName)) {
+              core.setFailed(
+                `Branch name must contain a Jira issue key in the form CMS-<number> or DPD-<number>, or the PR must have the '${bypassLabel}' label. Got '${branchName}'.`,
+              );
+              return;
+            }
+
+            core.info("Branch name contains a valid Jira issue key.");
+
+  pr-title:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    steps:
+      - uses: amannn/action-semantic-pull-request@48f256284bd46cdaab1048c3721360e808335d50 # v6.1.1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/pr-standards.yml
+++ b/.github/workflows/pr-standards.yml
@@ -18,33 +18,30 @@ jobs:
     permissions:
       contents: read
     steps:
-      - uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+      - uses: actions/checkout@v4
+      - name: Check if branch validation should be skipped
+        id: branch-validation-bypass
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/pr-standards.yml
```diff
@@ -0,0 +1,57 @@
...
+            const branchName = context.payload.pull_request?.head?.ref ?? "";
+            const prAuthor = context.actor;
+            const bypassLabel = "no-jira";
+            const labels = (context.payload.pull_request?.labels ?? [])
+              .map((label) => label.name?.toLowerCase())
+              .filter(Boolean);
+
+            const hasBypassLabel = labels.includes(bypassLabel);
+            const isBotBypass = prAuthor === "dependabot[bot]";
+            const jiraIssueKeyPattern = /(CMS|DPD)-\d+/;
```
MaciekBaron: I'd add a word boundary here:

```suggestion
            const jiraIssueKeyPattern = /\b(CMS|DPD)-\d+/;
```
That way this matches `feature/CMS-123` and `CMS-123-feature` but not `foobarCMS-123` (unless we want it to match that)

.github/workflows/pr-standards.yml
```diff
@@ -0,0 +1,57 @@
+name: PR standards
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, edited, labeled, unlabeled]
+
+permissions:
+  contents: read
+
```
MaciekBaron, line 9: Should we add a concurrency block?

```
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```
Ref: https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency

MebinAbraham: [`a92bce6` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/717/commits/a92bce6264a4ca5295899ad0dd892d39a4a67305) - While at it, decided to add to the pr-blocked-label job also.

.github/workflows/pr-standards.yml
MebinAbraham: Is it more etiquette or whether Jira supports it, you are worried about?. Jira should pick up either, but I'm happy to add; we probably want a boundary after also to prevent `CMS-123foobar`. I want to extend this workflow to improve PR etiquette in future, so it might be good to add.

MaciekBaron: Just etiquette really to enforce consistency - just a suggestion though.

MaciekBaron (approved):

sanjeevz3009 (approved):
LGTM!

.github/workflows/pr-standards.yml
MebinAbraham: [`b558a8a` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/717/commits/b558a8aefd8aea1470510e31518d2a796e3c4c18)

Made a couple of changes.
- Added boundary check
- Moved the branch check script to a shell script and add a pre-commit hook.
- Added extra bypass for dependabot branches + branches that include "no-jira" in the name.

When this merges, existings PRs will fail on pre-commit so people will need to skip them for now.

.github/workflows/pr-standards.yml
```diff
@@ -0,0 +1,60 @@
...
+  branch-name:
+    name: Validate branch name
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: 'false'
+      - name: Check if branch validation should be skipped
```
MaciekBaron: Minor: From what I understand, the validation can be skipped either because of a label (handled here), or a branch name (handled in the shell script). Might be good to specify that:

```suggestion
      - name: Check if branch validation should be skipped based on label
```

.github/workflows/pr-standards.yml
```diff
@@ -0,0 +1,60 @@
...
+permissions:
+  contents: read
+
+concurrency:
+  group: '${{ github.head_ref || github.ref }}-${{ github.workflow }}'
+  cancel-in-progress: true
+
+jobs:
+  branch-name:
+    name: Validate branch name
```
MaciekBaron, line 16: What if we add an env here:
```yaml
env:
  BYPASS_LABEL: "no-jira"
```

And then reference it in the JS code:
```js
const bypassLabel = process.env.BYPASS_LABEL;
```

_and_ in the shell script?

MebinAbraham: Why in the shell script?. I didn't add it there since that is also used for pre-commit which has no context of the PR and labels?

Let me know if I have misunderstood.

MebinAbraham: Updated description so now passing in the label: [`d67b5ab` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/717/commits/d67b5abc9690094debd69c67cf7ee982ceb96739)

sanjeevz3009 (approved):

MaciekBaron (approved):

MaciekBaron (approved):

PR #718, opened 2026-07-08
fix(tables): report all table block validation errors

helenb:
Addresses this comment from testing: https://officefornationalstatistics.atlassian.net/browse/CMS-1207?focusedCommentId=777873

This change ensures that the validation for the table block reports field validation errors and the custom 'subtitle without title' validation error at the same time, rather than separately.

### How to review

- Add a table to an information page, with a subtitle, but no title. Don't add an accessible label
- Try to save the page
- You should see 2 validation issues reported at the same time - the missing accessible label, and the fact that you can't have a subtitle without a title.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/cms/core/blocks/markup.py
+++ b/cms/core/blocks/markup.py
@@ -149,12 +149,19 @@ class ONSTableBlock(TinyTableBlock):
 
     def clean(self, value: dict) -> dict:
         """Validate that a subtitle is only present when a title is also provided."""
-        cleaned_value: dict = super().clean(value)
+        block_errors: dict = {}
+
+        try:
+            cleaned_value: dict = super().clean(value)
+        except StructBlockValidationError as e:
+            block_errors = dict(e.block_errors)
+            cleaned_value = value
 
         if cleaned_value.get("subtitle") and not cleaned_value.get("title"):
-            raise StructBlockValidationError(
-                block_errors={"subtitle": ValidationError("Please add a title if you want to add a subtitle.")}
-            )
+            block_errors["subtitle"] = ValidationError("Please add a title if you want to add a subtitle.")
+
+        if block_errors:
+            raise StructBlockValidationError(block_errors=block_errors)
 
         return cleaned_value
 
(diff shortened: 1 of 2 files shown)


sanjeevz3009 (approved):
Works as expected. Code looks code and testing it locally, it fixed the issue.

Not going to block this PR here but, there is a recurring theme with our blocks in general around field validations etc. There was a recent one too, which @joepercival merged.

It would be really good to have tests, around this. Maybe it's a wider tech-session discussion to be had.

@MebinAbraham @helenb

cms/core/blocks/markup.py
```diff
@@ -149,12 +149,19 @@ class ONSTableBlock(TinyTableBlock):
...
+            cleaned_value: dict = super().clean(value)
+        except StructBlockValidationError as e:
+            block_errors = dict(e.block_errors)
+            cleaned_value = value
 
         if cleaned_value.get("subtitle") and not cleaned_value.get("title"):
-            raise StructBlockValidationError(
-                block_errors={"subtitle": ValidationError("Please add a title if you want to add a subtitle.")}
-            )
+            block_errors["subtitle"] = ValidationError("Please add a title if you want to add a subtitle.")
```
MebinAbraham, line 161: This, in theory, would swallow other errors against the subtitle. Currently not a problem unless someone changes the subtitle model.

You could do the following to preserve existing errors first, but not a blocker.

```py
block_errors.setdefault(
    "subtitle",
    ValidationError("Please add a title if you want to add a subtitle."),
)
```

MebinAbraham (approved):

cms/core/blocks/markup.py
helenb: @MebinAbraham If you're happy to merge without this last tweak please go ahead, otherwise I'll look at this when I'm back in next week.

kevinhowbrook (approved):

PR #719, opened 2026-07-08
fix: ensure teardown on migration tests goes to latest project state

BJacksonONS:

Guarantees that we tear down to latest project state on migration testcase teardown.

I've only noticed this issue on one PR so far, but occasionally tests run after the migration test cases could potentially error due to the database state not matching the model state. By fetching the leaf nodes from the migration tree we can ensure we always leave the database in a clean state.

Also defaults the `next_migration` var to latest project state if not specified to avoid manually updating the string when new migrations are created.

### How to review

- Ensure CI is green.
- Sense check approach

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/cms/core/tests/__init__.py
+++ b/cms/core/tests/__init__.py
@@ -96,31 +96,32 @@ class MigrationTestCase(_TransactionTestCase):
     def setUpClass(cls):
         super().setUpClass()
         cls.executor = MigrationExecutor(connections[DEFAULT_DB_ALIAS])
+        # register class cleanup to ensure it runs even if setUpClass fails
+        cls.addClassCleanup(cls._restore_latest_state)
         if cls.next_migration is None:
             cls.next_migration = cls.executor.loader.graph.leaf_nodes()
         cls.executor.migrate(cls.next_migration)
         cls.apps = cls.executor.loader.project_state(cls.next_migration).apps
 
-    def tearDown(self):
-        """Restore the migration state to the next_migration after each test."""
-        self.migrate_to(self.next_migration)
-        super().tearDown()
+    def setUp(self):
+        """Restore the migration state to the next_migration before each test."""
+        super().setUp()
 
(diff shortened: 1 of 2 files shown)


sanjeevz3009 (approved):
I know I am aware of why this PR exists truly. But a bit of context on the PR comment itself for the rest of the dev team will be helpful =)

Thanks @BJacksonONS for this improvement as otherwise, we get into akward states.

cms/core/tests/__init__.py
```diff
@@ -96,9 +96,17 @@ class MigrationTestCase(_TransactionTestCase):
     def setUpClass(cls):
         super().setUpClass()
         cls.executor = MigrationExecutor(connections[DEFAULT_DB_ALIAS])
+        if cls.next_migration is None:
+            cls.next_migration = cls.executor.loader.graph.leaf_nodes()
         cls.executor.migrate(cls.next_migration)
         cls.apps = cls.executor.loader.project_state(cls.next_migration).apps
 
+    def tearDown(self):
```
MebinAbraham: We probably want `tearDownClass` as if `setUpClass` failed, and no tests ran, then it would be in a broken state.

MebinAbraham (approved):

cms/core/tests/__init__.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/719/commits/07d64224eed55d05e5125ad00ae0ec65dbcc6d86

and updated comment

https://github.com/ONSdigital/dis-wagtail/pull/719/commits/34fa665bbf297124213051b0ba45e77d4609b5fe

MebinAbraham (approved):

MebinAbraham (approved):

PR #723, opened 2026-07-09
fix(pr-standards): check for Jira link before update PR body

MebinAbraham:

This swaps the PR standards workflow logic from checking for the Jira issue key to checking for the full Jira issue link before appending a ticket link to the PR body.

This ensures the workflow only skips updates when the canonical Jira link is already present, rather than when the issue key appears as plain text.

### How to review

Ensure the change makes sense

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/.github/workflows/pr-standards.yml
+++ b/.github/workflows/pr-standards.yml
@@ -65,17 +65,16 @@ jobs:
             }
 
             const issueKey = match[0];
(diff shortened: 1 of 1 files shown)


BJacksonONS (approved):

joepercival (approved):
LGTM

sanjeevz3009 (commented):
Reviewed!

PR #725, opened 2026-07-09
fix: correct dataset duplicate detection in `DatasetStoryBlock`

MaciekBaron:

This PR addresses CMS-1255 and fixes the way duplicate datasets are detected in the `DatasetStoryBlock`.

Currently, the URL of a dataset just contains the namespace of the Dataset:

https://github.com/ONSdigital/dis-wagtail/blob/197fcf9906c562cf1f4463aad20182c076bfe1a8/cms/datasets/models.py#L247-L251

Meanwhile, the `clean()` method for the `DatasetStoryBlock` would only use the URLs to detect duplicates, creating false positives.

This change adds a more thorough checks for looked up datasets to consider namespace, edition and version of the dataset before flagging it as a duplicate.

### How to review

Follow the steps from the ticket and ensure no validation error is shown:
1. Create a new or select an existing Statistical Article page ensuring mandatory fields have been populated with valid data
2. Go to ‘Related data’ tab (Datasets) and click the + button and select 'Lookup Dataset option
3. ‘Choose a dataset' , select ‘Unpublished’ from the 'Published status’ dropdown and select at least two datasets. In this case it is two different editions for the same dataset.
4. As the user is currently unable to select more than one dataset at a time, steps 2 and 3 need to be performed twice
5. User selects to save the changes

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

N/A

---
Ticket: [CMS-1255](https://officefornationalstatistics.atlassian.net/browse/CMS-1255)

[CMS-1255]: https://officefornationalstatistics.atlassian.net/browse/CMS-1255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cms/datasets/tests/test_blocks.py
+++ b/cms/datasets/tests/test_blocks.py
@@ -78,6 +78,48 @@ class TestDatasetStoryBlock(TestCase):
             [
                 ("dataset_lookup", self.lookup_dataset.id),
                 (
+                    "dataset_lookup",
+                    Dataset.objects.create(
+                        # Same namespace as the first dataset, but different edition and version
+                        namespace=self.lookup_dataset.namespace,
+                        edition="test2_edition",
+                        version=2,
+                        title="test_title",
+                        description="test_description",
+                    ).id,
+                ),
+            ],
+            [
+                ("dataset_lookup", self.lookup_dataset.id),
+                (
+                    "dataset_lookup",
+                    Dataset.objects.create(
+                        # Same namespace and version as the first dataset, but different edition
+                        namespace=self.lookup_dataset.namespace,
+                        edition="test2_edition",
+                        version=1,
+                        title="test_title",
+                        description="test_description",
+                    ).id,
+                ),
+            ],
+            [
+                ("dataset_lookup", self.lookup_dataset.id),
+                (
+                    "dataset_lookup",
+                    Dataset.objects.create(
+                        # Same namespace and edition as the first dataset, but different version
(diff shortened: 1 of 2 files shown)


cms/datasets/tests/test_blocks.py
```diff
@@ -112,6 +154,8 @@ def test_successful_validation(self):
                     block,
                     stream_data=stream_data,
                 )
-
-                # Expect clean to not raise any errors
-                block.clean(value)
+                try:
```
MaciekBaron, line 157: I think it's best practice to fail a test, rather than rely on exceptions (for better readability and test result clarity)

kevinhowbrook (approved):
Code lgtm (can't test manually/locally due to dataset API access)

BJacksonONS (approved):

MebinAbraham:
I need to do some analysis as this may actually be expected behaviour (in some places).

MebinAbraham:
I'm going to close this, as what we currently have is mostly correct, but we need to make some changes to the URLs we need to link out to from different pages, which will essentially change the dedupe rules. We need to make datasets validate duplicates using different rules based on the pages they are linked from.

Placeholder: https://officefornationalstatistics.atlassian.net/browse/CMS-1334

PR #726, opened 2026-07-09
fix: add GTM data attributes to breadcrumbs component

charlotteromaniuk:

Adds the following GTM data attributes to the breadcrumbs links:

- `data-ga-event`
- `data-ga-navigation-type`
- `data-ga-link-text`
- `data-ga-click-content-group`
- `data-ga-click-content-type`
- `data-ga-click-path`
- `data-ga-click-content-theme`
- `data-ga-click-position`

See the ticket and specification attached to it for the expected values for each attribute.

### How to review

1. Using browser dev tools, inspect the breadcrumb links on a published page and ensure the data attributes listed above are added to each breadcrumb link
2. Ensure the expected values match the specification
3. For pages that don't have a content type, content theme or content group, the associated data attribute should not be added to the breadcrumb link
4. Verify the following specific cases:
   -  The Home page doesn't have a content group or content theme so `data-ga-click-content-group` and `data-ga-click-content-theme` should not be present
   - The Release Index page doesn't have a content type so `data-ga-click-content-type` should not be present

### Deployment Safety

Please select one:

- [X] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

N/A

---
Ticket: [DPD-4319](https://officefornationalstatistics.atlassian.net/browse/DPD-4319)

[DPD-4319]: https://officefornationalstatistics.atlassian.net/browse/DPD-4319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cms/core/models/base.py
+++ b/cms/core/models/base.py
@@ -158,10 +158,12 @@ class BasePage(PageLDMixin, ListingFieldsMixin, SocialFieldsMixin, Page):  # typ
             "data-ga-event": "navigation-click",
             "data-ga-navigation-type": "breadcrumb",
             "data-ga-link-text": link_text,
-            "data-ga-click-content-type": page.analytics_content_type,
             "data-ga-click-path": page.get_url(request=request),
             "data-ga-click-position": position,
         }
+
+        if content_type := page.analytics_content_type:
+            attributes["data-ga-click-content-type"] = content_type
 
         if content_group := page.analytics_content_group:
             attributes["data-ga-click-content-group"] = content_group
--- a/cms/core/models/base.py
+++ b/cms/core/models/base.py
@@ -340,6 +340,7 @@ class BasePage(PageLDMixin, ListingFieldsMixin, SocialFieldsMixin, Page):  # typ
             return cast(str, parent_topic_or_theme.slug)
         return None
 
+    # TODO: Revisit once the theme/page relationship is defined and the attribute's requirements are finalised.
     @cached_property
     def analytics_content_theme(self) -> str | None:
         """Returns the title of the top ancestor taxonomic topic for the pages parent topic or theme page,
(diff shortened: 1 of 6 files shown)


kevinhowbrook (approved):
Tests  well - all attributes present locally 👍

MaciekBaron (approved):

BJacksonONS (approved):

cms/core/models/base.py
```diff
@@ -150,24 +150,75 @@ def breadcrumb_title(self) -> str:
...
+        }
+
+        if content_type := page.analytics_content_type:
+            attributes["data-ga-click-content-type"] = content_type
+
+        if content_group := page.analytics_content_group:
+            attributes["data-ga-click-content-group"] = content_group
+
+        if content_theme := page.analytics_content_theme:
+            attributes["data-ga-click-content-theme"] = content_theme
```
MebinAbraham, line 172: For this, we need to check with Awen whether she expects the title or the slug. Currently, the topic title is being passed. In the spec, she uses the title in some places and the slug in others. For example, the breadcrumb spec seems to use a slug which this wouldn't match. Let's confirm with Awen Jones whether this should be the topic slug or the topic title. It should be standardised, so it is not conditionally different and always means the same.

charlotteromaniuk: Awen has suggested leaving this attribute out and made it optional for now as the requirement will need revisiting once it’s been established how to link pages to themes. I’ve left the attribute here so that it matches the other navigation links and left a note that it will need to be revisited - [79c08d9](https://github.com/ONSdigital/dis-wagtail/pull/726/commits/79c08d9ba5a85ca74ad612890ad41bde1e418e06)

MebinAbraham (approved):

sanjeevz3009 (commented):
Reviewed!

PR #731, opened 2026-07-15
feat: enforce specific version of `poetry`

MaciekBaron:

This PR enforces the usage of a specific version of `poetry`, to avoid any inconsistencies.

This involves:
* Pinning the exact version of `poetry` in `pyproject.toml` (previously set to `>=2.2`)
* Adding a simple linter checking the message at the top of the `poetry.lock` file
* Adding the inter to the `pre-commit` setup and the GitHub `ci.yml` configuration as a new step

Note: In order to install a specific version of poetry, you can use the following commands:

If you used the official installer:
```bash
poetry self update 2.4.1
```

If you use `pipx`:
```bash
pipx install poetry==2.4.1 --force
```
If you installed it using a different method, consult the [documentation](https://python-poetry.org/docs/#installation).

### How to review

1. Install a version of poetry different from `2.4.1`
2. Trigger `pre-commit`
3. Ensure that it fails, complaining about the version
4. Re-generate the lock file using a different version, or manually change it to something else
5. Repeat steps 2 and 3

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,8 @@ jobs:
         with:
           python-version-file: .python-version
           cache: ${{ steps.cache-flag.outputs.poetry-cache }} # disabled on release branches via setup-cache-flag action
+      - name: Check poetry lock file version
+        run: make check-poetry-version
 
       - name: Install dependencies
         run: poetry install --no-root
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,6 +50,15 @@ repos:
         # Only run on changes to pyproject.toml
         files: ^pyproject.toml$
 
+  - repo: local
+    hooks:
+      - id: check-poetry-version
+        name: check-poetry-version
+        language: system
+        entry: ./scripts/check-poetry-version.sh
+        pass_filenames: false
+        files: ^(pyproject\.toml|poetry\.lock)$
+
   # Template linting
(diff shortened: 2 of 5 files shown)


BJacksonONS (approved):

BJacksonONS:
Just occurred to me, is it possible this could break dependabot if it's using a different version of poetry to us?

MebinAbraham:
> Just occurred to me, is it possible this could break dependabot if it's using a different version of poetry to us?

It definitely [used to be a problem](https://github.com/dependabot/dependabot-core/issues/1556), but the recent ones seem to adhere to the version in pyproject.toml.
- https://github.com/ONSdigital/dis-wagtail/pull/642/changes
- https://github.com/ONSdigital/dis-wagtail/pull/629/changes
- https://github.com/ONSdigital/dis-wagtail/pull/572/changes

So in theory it should be fine, but let's see. If it becomes a problem, we can make it less strict, so it only enforces minor versions.

MebinAbraham (approved):

sanjeevz3009 (commented):
Reviewed!

sanjeevz3009:
@MaciekBaron Was on my backlog of things to do; I think I even had a branch locally but never got around to creating a PR in the chaos. Thanks @MaciekBaron

PR #732, opened 2026-07-16
feat: support purging wagtail responses by Cache-Tag header

joepercival:

This PR adds a Cloudflare Cache-Tag header to Wagtail server route responses so those routes can be purged from Cloudflare as a group.

### How to review

1. Run the app with `IS_EXTERNAL_ENV` as `True`
2. Use the Network tab in developer tools and confirm that the `Cache-Tag` header is set  to match the configured `CMS_CLOUDFLARE_CACHE_TAG`

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

n/a

---
Ticket: [CMS-1290](https://officefornationalstatistics.atlassian.net/browse/CMS-1290)

[CMS-1290]: https://officefornationalstatistics.atlassian.net/browse/CMS-1290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cms/core/middleware.py
```diff
@@ -43,3 +45,61 @@ def is_request_path_allowed(self, path: str) -> bool:
...
+
+    def _parse_cache_tags(self, header: str) -> list[str]:
+        """Parse the Cache-Tag header into a list of tags."""
+        return [tag.strip() for tag in header.split(",") if tag.strip()] if header else []
+
+    def process_response(self, request: HttpRequest, response: HttpResponse) -> HttpResponse:
+        """Add the Cloudflare cache tag header to the response."""
+        # Only tag successful responses, not client or server errors
+        if response.status_code >= HTTPStatus.BAD_REQUEST:
+            return response
```
MebinAbraham: For now we can set this header for all responses. We would want it for things like 404s, etc, but right now it is not worth having an allowlist of status codes; this header is graceful.

cms/core/middleware.py
```diff
@@ -43,3 +45,61 @@ def is_request_path_allowed(self, path: str) -> bool:
         """Check if the request path is allowed to be redirected."""
         stripped_path = path.lstrip("/")
         return any(stripped_path.startswith(allowed_path.lstrip("/")) for allowed_path in ALLOWED_REQUEST_PATHS)
+
+
+class CloudflareWagtailCacheTagMiddleware(MiddlewareMixin):
+    """Adds a Cloudflare cache tag header to Wagtail server route responses."""
+
+    def _is_wagtail_route(self, request: HttpRequest) -> bool:
```
MebinAbraham: I can see how this might have been confused from ticket description but you don't actually need this as all "requests" reaching Wagtail can have this tag as we are the catch all. We don't route or proxy to other services.

cms/settings/base.py
```diff
@@ -1189,3 +1192,7 @@
     "payment": [],
     "usb": [],
 }
+
+# Cloudflare cache tag for Wagtail responses
+# Used to purge Wagtail server route responses as a group
+WAGTAIL_CLOUDFLARE_CACHE_TAG = env.get("WAGTAIL_CLOUDFLARE_CACHE_TAG")
```
MebinAbraham: You can set the default to `wagtail`

cms/settings/base.py
```diff
@@ -170,6 +170,9 @@
 if CSP_ENABLED:
     MIDDLEWARE.append("django.middleware.csp.ContentSecurityPolicyMiddleware")
 
+if IS_EXTERNAL_ENV and "WAGTAIL_CLOUDFLARE_CACHE_TAG" in env:
```
MebinAbraham: You either need to define `WAGTAIL_CLOUDFLARE_CACHE_TAG = env.get("WAGTAIL_CLOUDFLARE_CACHE_TAG")` before this line, or alternatively, don't check the env value as we would set a default anyway.

docs/custom-features/deployment-environments.md
```diff
@@ -23,3 +23,5 @@ The URLs and apps required for the Wagtail admin are disabled. This includes the
 Only a subset of media stored in S3 are accessible to the external env. Notably the `images/*` prefix needed for Wagtail's rendition generation.
 
 Certain static files (for example those related to the Wagtail admin) are inaccessible in the external environment. This is configured using `cms.core.whitenoise.CMSWhiteNoiseMiddleware`.
+
+External responses carry a Cloudflare Cache-Tag header (default wagtail, configurable via the `WAGTAIL_CLOUDFLARE_CACHE_TAG` env var) so they can be purged from Cloudflare as a group; purge invocation itself is manual/operational, not application-driven.
```
MebinAbraham:
```suggestion
External responses carry a Cloudflare `Cache-Tag` header (default `wagtail`, configurable via the `WAGTAIL_CLOUDFLARE_CACHE_TAG` env var) so they can be purged from Cloudflare as a group via the purge prefix option; purge invocation itself is manual/operational, not application-driven.
```

cms/core/middleware.py
```diff
@@ -43,3 +45,61 @@ def is_request_path_allowed(self, path: str) -> bool:
...
+        if not resolver_match:
+            return False
+
+        excluded_paths = {
+            "/api/",
+            "/v1/",
+            "/wagtail/",
+            "/admin/",
+            "/__debug__/",
+        }
```
MaciekBaron: I think this should live outside of the method, and potentially be derived from some settings values, e.g. `WAGTAILADMIN_HOME_PATH`

cms/settings/base.py
joepercival: c51c715

cms/core/middleware.py
joepercival: removed this method here bd355d8

cms/core/middleware.py
joepercival: removed this method here bd355d8 as per Mebin's comment

cms/settings/base.py
joepercival: addressed here c51c715

cms/core/middleware.py
joepercival: bd355d8

MaciekBaron (approved):

cms/core/middleware.py
```diff
@@ -43,3 +44,30 @@ def is_request_path_allowed(self, path: str) -> bool:
         """Check if the request path is allowed to be redirected."""
         stripped_path = path.lstrip("/")
         return any(stripped_path.startswith(allowed_path.lstrip("/")) for allowed_path in ALLOWED_REQUEST_PATHS)
+
+
+class CloudflareWagtailCacheTagMiddleware(MiddlewareMixin):
+    """Adds a Cloudflare cache tag header to Wagtail server route responses."""
```
MebinAbraham:
```suggestion
    """Add a Cloudflare `Cache-Tag` header to responses from Wagtail routes."""
```

cms/core/middleware.py
```diff
@@ -43,3 +44,30 @@ def is_request_path_allowed(self, path: str) -> bool:
...
+        """Parse the Cache-Tag header into a list of tags."""
+        return [tag.strip() for tag in header.split(",") if tag.strip()] if header else []
+
+    def process_response(self, request: HttpRequest, response: HttpResponse) -> HttpResponse:
+        """Add the Cloudflare cache tag header to the response."""
+        cache_tag = getattr(settings, "WAGTAIL_CLOUDFLARE_CACHE_TAG", None)
+        if cache_tag is None:
+            return response
+
+        cache_tag = str(cache_tag).strip()
```
MebinAbraham: Nit: We should do this when parsing the setting in base.py

cms/core/middleware.py
```diff
@@ -43,3 +44,30 @@ def is_request_path_allowed(self, path: str) -> bool:
...
+        if not cache_tag:
+            return response
+
+        # Add tag if header already exists and tag is not present
+        existing_header = response.get(CLOUDFLARE_CACHE_TAG_HEADER, "")
+        existing_tags = self._parse_cache_tags(existing_header)
+        if cache_tag not in existing_tags:
+            updated_tags = [*existing_tags, cache_tag]
+            response[CLOUDFLARE_CACHE_TAG_HEADER] = ",".join(updated_tags)
```
MebinAbraham: ~~We need to add the header if it doesn't exist. This currently adds the header only if it's already present. It should add it if it doesn't exist; otherwise, update it.~~ Behaviour is correct, but the comment is misleading.

MebinAbraham: [`e1d25bf` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/732/commits/e1d25bf98ff44678ea58543268b97e2e0a866138)

cms/core/middleware.py
MebinAbraham: [`e1d25bf` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/732/commits/e1d25bf98ff44678ea58543268b97e2e0a866138)

cms/core/middleware.py
MebinAbraham: [`e1d25bf` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/732/commits/e1d25bf98ff44678ea58543268b97e2e0a866138)

cms/core/middleware.py
```diff
@@ -43,3 +44,27 @@ def is_request_path_allowed(self, path: str) -> bool:
...
+    def process_response(self, request: HttpRequest, response: HttpResponse) -> HttpResponse:
+        """Add the Cloudflare cache tag header to the response."""
+        cache_tag = settings.WAGTAIL_CLOUDFLARE_CACHE_TAG
+        if not cache_tag:
+            return response
+
+        # Append the configured tag when missing, or create the header if absent.
+        existing_header = response.get(CLOUDFLARE_CACHE_TAG_HEADER, "")
+        existing_tags = self._parse_cache_tags(existing_header)
+        if cache_tag not in existing_tags:
```
kevinhowbrook, line 66: **praise**

nice touch, should avoid a messy Cache-Tag header if this middleware ever runs twice or another layer already sets the tag.

kevinhowbrook (approved):

MaciekBaron (approved):

cms/settings/base.py
```diff
@@ -1189,3 +1192,7 @@
     "payment": [],
     "usb": [],
 }
+
+# Cloudflare cache tag applied to CMS responses.
+# Used to purge Wagtail CMS responses as a group.
+CMS_CLOUDFLARE_CACHE_TAG = env.get("CMS_CLOUDFLARE_CACHE_TAG", "wagtail").strip()
```
BJacksonONS, line 1198: Should we add an extra check somewhere that the tag doesn't contain a space? The docs don't specify exactly what happens, but something like `wagtail another-tag` isn't a valid tag but we'd still try to use it anyway.

MebinAbraham: I agree; however, I suggest we revisit this point, as we don't validate a lot of our config, so there are similar gaps across the board. We currently trust that most things are configured correctly rather than validating them which isn't ideal but need to find the right balance.

MebinAbraham: Placeholder to revisit: https://officefornationalstatistics.atlassian.net/browse/CMS-1332

BJacksonONS: Sounds good

BJacksonONS (approved):

sanjeevz3009 (commented):
Reviewed!

PR #736, opened 2026-07-21
feat: upgrade design system to 73.10.0 and integrate with figure component

helenb:

This PR implements 2 tickets:

https://officefornationalstatistics.atlassian.net/browse/CMS-1308 - Upgrade DS to 73.10.0
https://officefornationalstatistics.atlassian.net/browse/CMS-1272 - Figure component implementation for charts, images and tables

- Updates the DS to 73.10.0
- Updates the table, image and chart components to fully utilise the figure component
- Updates the field ordering and labelling for table, image, chart and featured chart
- Adds a new field to the image block to allow overriding of the image alt text in the context of the page
- Adds a new checkbox field to the image block to mark the image as decorative allow the alt text to be empty
- Adds validation to ensure that the alt text override cannot be saved when the 'Image is decorative' option is checked.
- Updates the featured article to adapt to the new way the DS component works - arbitrary child content instead of mandating a chart or image
- Removes the option to add footnotes to a featured chart (previously you could add them in the wagtail admin but they would not display - confirmed with Zoe that they should be removed)

**Other DS changes included**

- Fix dynamic island bug
- Add Welsh translation for search label
- Table of contents ordered variant
- Improvements to table aria labels
- Font preloading

### How to review

- Check out the branch and run `make load-design-system-templates`

- Test add an image block and a table block to a standard page
- Check all fields display including the new 'Figure number'
- Ensure that the headings for title, subtitle, footnotes and downloads are as expected, regardless of whether a title or subtitle are included
- Check validation to ensure subtitle cannot be added without a subtitle still works

- Test adding a chart to an article page
- Check all fields display including the new 'Figure number'
- Check that the heading levels for title, subtitle, footnotes and downloads behave as expected, regardless of whether a title or subtitle are included
- Check all options ares still present

- Test adding an image to a featured article
- Check all fields display as expected

- Test adding a chart to a featured article
- Ensure that you cannot add subtitle, source text or footnotes
- Ensure all fields display as expected

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

- This needs to be merged before https://github.com/ONSdigital/dis-wagtail/pull/702 and is also needed for the iframe implementation ticket to start.

---
Ticket: [CMS-1308](https://officefornationalstatistics.atlassian.net/browse/CMS-1308)

[CMS-1308]: https://officefornationalstatistics.atlassian.net/browse/CMS-1308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cms/core/blocks/embeddable.py
```diff
@@ -37,8 +37,22 @@ class ImageBlock(blocks.StructBlock):
     figure_subtitle = blocks.CharBlock(required=False, label="Subtitle")
     supporting_text = blocks.TextBlock(required=False, label="Source text")
     notes_section = blocks.RichTextBlock(required=False, features=settings.RICH_TEXT_BASIC, label="Footnotes")
+    empty_alt_text = blocks.BooleanBlock(
```
zerolab: drive-by thought: should this be called `decorative_image` or `is_decorative` to better reflext its purpose to both editors and devs?

helenb: I'm actually waiting on confirmation for the naming as I've deviated from what was on the ACs - I'll settle on a final name once I've had that.

helenb: Now updated as my label change has been agreed to.

zerolab: Nice one! Feel free to resolve (as I dont have the ability :()

helenb: Resolved by me as @zerolab could not do so.

cms/core/blocks/markup.py
```diff
@@ -132,19 +132,17 @@ def render(self, value: dict, context: dict | None = None) -> str | SafeString:
...
+    table_number = blocks.CharBlock(
+        required=False, label="Table number", help_text="Include a label for the figure, for example Table 1."
+    )
     subtitle = blocks.CharBlock(required=False)
-    source = blocks.CharBlock(label="Source", required=False)
+    source = blocks.CharBlock(label="Source text", required=False)
     footnotes = blocks.RichTextBlock(label="Footnotes", features=settings.RICH_TEXT_BASIC, required=False)
     # Redeclare the inherited caption field
     caption = blocks.CharBlock(
         required=True,
```
charlotteromaniuk, line 144: AC3 has this as an optional field

helenb: That is a mistake in the ACs I'm pretty sure - the accessible text should be required. I'll double check.

helenb: I've now updated that AC after getting confirmation from Nicola - good spot to find the discrepancy :)

cms/core/blocks/embeddable.py
```diff
@@ -38,22 +66,52 @@ def get_context(self, value: StreamValue, parent_context: dict | None = None) ->
...
+        _base, ext = os.path.splitext(large.file.name)
         file_type = ext.lstrip(".").upper() or "IMG"
-        context["file_type"] = file_type
-
         size_bytes = getattr(large.file, "size", None)
-        context["file_size_human"] = filesizeformat(size_bytes) if size_bytes is not None else None
+        file_size_human = filesizeformat(size_bytes) if size_bytes is not None else None
+
+        options = {
+            "id": f"image-{context.get('block_id') or ''}",
```
kevinhowbrook: If block ID is missing here we would get `image-` is that intended? (the trailing -)

cms/core/blocks/markup.py
```diff
@@ -209,38 +207,34 @@ def get_context(self, value: dict, parent_context: dict | None = None) -> dict:
             return context
 
         options = {
+            # fallback is only when block_id is not available, which should not happen in normal usage
+            "id": f"table-{context.get('block_id') or ''}",
```
kevinhowbrook: again just checking that `table-` with the trailing `-` is ok here?

cms/core/blocks/embeddable.py
helenb: That's true, but I think the only time it would ever happen might be in tests. We could somehow generate a uuid as a fallback but I wasn't sure it was needed - would welcome your thoughts.

cms/core/blocks/markup.py
helenb: Same reply as above.

cms/core/blocks/embeddable.py
```diff
@@ -38,22 +66,52 @@ def get_context(self, value: StreamValue, parent_context: dict | None = None) ->
...
+        form_layout = [  # noqa
+            "figure_number",
+            "figure_title",
+            "figure_subtitle",
+            "alternative_text",
+            "decorative_image",
+            "image",
+            "supporting_text",
+            "notes_section",
+            "download",
```
charlotteromaniuk, line 114: If we're expecting the field order to match the ACs, I think the download checkbox should sit below the decorative image checkbox

cms/core/blocks/embeddable.py
```diff
@@ -19,11 +21,37 @@ class ImageBlock(blocks.StructBlock):
...
+    figure_number = blocks.CharBlock(
+        required=False, label="Figure number", help_text="Include a label for the figure, for example Figure 1."
+    )
+    figure_title = blocks.CharBlock(required=False, label="Title")
+    figure_subtitle = blocks.CharBlock(required=False, label="Subtitle")
+    supporting_text = blocks.TextBlock(required=False, label="Source text")
+    notes_section = blocks.RichTextBlock(required=False, features=settings.RICH_TEXT_BASIC, label="Footnotes")
+    decorative_image = blocks.BooleanBlock(
+        required=False,
+        label="Image is decorative, so alt text should be empty",
```
MaciekBaron, line 43: Hmm, I'm not sure about this label. Maybe it could say "Image is decorative and does not require alt text"? Just a personal opinion.

cms/core/tests/test_blocks.py
```diff
@@ -1276,14 +1226,56 @@ def test_renders_small_and_large_renditions_and_alt_text(self):
...
+        self.assertContains(response, 'alt="Custom alt text"')
+        self.assertNotContains(response, 'alt="Meaningful alt text"')
+
+    def test_decorative_image_renders_empty_alt_attribute(self):
+        page = self._make_information_page(download=False, decorative_image=True)
+
+        response = self.client.get(page.url)
+        self.assertEqual(response.status_code, HTTPStatus.OK)
+
+        self.assertContains(response, 'alt=""')
```
MaciekBaron: I'm worried that this test is too easy to satisfy and might prevent catching a regression in the future. It's not unlikely/impossible for an image to be introduced in the base layout which has `alt=""` (or somewhere else within the information page), and if that happens, this test would always pass regardless of the code.

A better approach might be to use BeautifulSoup to query the exact image and check its properties. Here's an existing example:

https://github.com/ONSdigital/dis-wagtail/blob/035d1c395c50ebb386895baf5e4be10822dcf62e/cms/articles/tests/test_pages.py#L83-L94

cms/core/tests/test_blocks.py
```diff
@@ -1296,12 +1288,14 @@ def test_renders_download_link_with_file_type_and_size_when_enabled(self):
         # HTML5 download attribute present once (avoid base-template noise)
         self.assertContains(response, " download", count=1)
 
-        # Large rendition used for download (assert exact URL appears twice in the response)
-        # Assert exact downloadable rendition file URL appears twice: once in href, once in onsImage srcset
+        # Large rendition used for download
+        # Appers twice in the response - once in the image srcset attribute and once in the download url
```
MaciekBaron: Typo:
```suggestion
        # Appears twice in the response - once in the image srcset attribute and once in the download url
```

MaciekBaron (changes requested):

cms/core/blocks/embeddable.py
helenb: The idea is that the order the fields appear in the admin always reflect the order that they are rendered on the front-end, so in this case it is correct that the download option comes last, as downloads display after footnotes ("notes_section").

cms/core/blocks/embeddable.py
helenb: I did get this re-worded version approved by Nathan and Zoe - empty is more accurate as it is rendered with `alt=""` rather than no `alt` attribute. Although maybe that detail isn't relevant to publishing officers. I think let's leave it for now and we can adjust later if needed.

cms/core/tests/test_blocks.py
helenb: Improved in 662950c2ad528a70e48e7171fb70cdf4550e724e

cms/core/blocks/embeddable.py
kevinhowbrook: It's very much a nitpick... but if there is an element of 'unsure' I think on the safe/cheap side, it might be worth chucking in a `uuid.uuid4().hex[:8]` - my main thought being that if two images on the same page both lack a block_id, we'd get duplicate id="image-" elements, which is invalid HTML, even if the "no block_id" case itself is rare.

charlotteromaniuk:
I think there’s a bug in the design system with the search label translation. The header macro seems to be checking the wrong language value, so it always returns the English label. 
image
@MebinAbraham should I raise a bug with the design system?

cms/core/blocks/embeddable.py
charlotteromaniuk: Makes sense, thanks :)

cms/core/blocks/embeddable.py
helenb: Addressed in 2421b39b40f9e0f9c25adc2105ab68c7bc09c5ac

cms/core/blocks/markup.py
helenb: Addressed in 2421b39b40f9e0f9c25adc2105ab68c7bc09c5ac

helenb:
> I think there’s a bug in the design system with the search label translation. The header macro seems to be checking the wrong language value, so it always returns the English label. image @MebinAbraham should I raise a bug with the design system?

Hey @charlotteromaniuk I think you're right. There *was* a [recent fix](https://github.com/ONSdigital/design-system/pull/3852) but it tests against `{{ params.lang }}` which I don't think exists/ It needs to use `{{ currentLanguageIsoCode }}`. If you raise a bug @MaciekBaron or I can look at fixing it.

(This is not a bug introduced by this PR - it is already an existing bug as far as I can see).

cms/core/blocks/embeddable.py
```diff
@@ -19,11 +22,37 @@ class ImageBlock(blocks.StructBlock):
...
+            "If this field is left blank, alt text will be sourced from the mandatory "
+            "'alternative text' field when the image was uploaded to Wagtail"
+        ),
+    )
+    figure_number = blocks.CharBlock(
+        required=False, label="Figure number", help_text="Include a label for the figure, for example Figure 1."
+    )
+    figure_title = blocks.CharBlock(required=False, label="Title")
+    figure_subtitle = blocks.CharBlock(required=False, label="Subtitle")
+    supporting_text = blocks.TextBlock(required=False, label="Source text")
```
charlotteromaniuk: Small thing I noticed - the source text here uses TextBlock but the table and chart source text field uses CharBlock. Just wondering if CharBlock should be used here as well for consistency?

cms/locale/cy/LC_MESSAGES/django.po
```diff
@@ -816,3 +807,9 @@ msgstr "Cymraeg"
 #: cms/workflows/templates/wagtailadmin/workflows/workflow_status.html:13
 msgid "Not started"
 msgstr "Heb ei ddechrau"
+
+#~ msgid "Notes"
+#~ msgstr "Nodiadau"
+
+#~ msgid "Download image"
+#~ msgstr "Lawrlwythwch y ddelwedd"
```
charlotteromaniuk: I think these can be removed

cms/core/blocks/embeddable.py
helenb: Addressed in e75b6eb571cb00dfe9a0ad55b00ddb7a66957fec

cms/locale/cy/LC_MESSAGES/django.po
helenb: Addressed in e75b6eb571cb00dfe9a0ad55b00ddb7a66957fec

charlotteromaniuk:
> Hey @charlotteromaniuk I think you're right. There _was_ a [recent fix](https://github.com/ONSdigital/design-system/pull/3852) but it tests against `{{ params.lang }}` which I don't think exists/ It needs to use `{{ currentLanguageIsoCode }}`. If you raise a bug @MaciekBaron or I can look at fixing it.

Thanks @helenb! I've raised a bug [here](https://officefornationalstatistics.atlassian.net/browse/ONSDESYS-985?search_id=f8e3f9b1-c86e-4f83-a1a5-8aa9676a4712)

> (This is not a bug introduced by this PR - it is already an existing bug as far as I can see).

Yeah, just mentioning it here since its one of the DS changes included in the upgrade

MaciekBaron (approved):

charlotteromaniuk (approved):
LGTM!

kevinhowbrook (approved):

sanjeevz3009 (commented):
Reviewed!

PR #737, opened 2026-07-21
chore(kics): remove redundant SHA pinning exclusion

MebinAbraham:

Remove the KICS exclusion for unpinned GitHub Actions.

Commit SHA pinning is already enforced by [zizmor](https://github.com/zizmorcore/zizmor), so this disable is misleading, as it is still checked (just by another linter atm), so this override is unnecessary and can coexist with zizmor, which also checks the same rule.

### How to review

Ensure changes make sense.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


BJacksonONS (approved):

MaciekBaron (approved):

sanjeevz3009 (commented):
Reviewed!

PR #744, opened 2026-07-28
feat: iframe implementation

helenb:

Brings the iframe implementation into the CMS using the figure component to wrap it.

### How to review

- Edit a statistical article page
- Add an iframe visualisation block
- Check the fields match the ACS
- Test the validation - e.g. no subtitle without a title. Note that the allowed domains is handled via an environment variable so may not exactly match the given help text.
- Test with JavaScript disabled
- Test at mobile

When reviewing the code it will be helpful to first look at b06f1746904f76c1e8839967f4ede18e220dd072, which is all the code changes required to bring the iframe into the CMS. Then look at cc3d43abf8b0c973e860e738bcc61a2dbb732e24 which moves the iframe block to its own file as `charts.py` was too long for pylint.

Useful iframe urls for testing:

https://www.ons.gov.uk/visualisations/dvc3294/fig01/index.html  - simple line chart
https://www.ons.gov.uk/visualisations/dvc3388/dashwrapper/index.html - a very tall visualisation which pym.js should adjust

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

---
Ticket: [CMS-1273](https://officefornationalstatistics.atlassian.net/browse/CMS-1273)

[CMS-1273]: https://officefornationalstatistics.atlassian.net/browse/CMS-1273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cms/jinja2/templates/partials/content_scripts.html
```diff
@@ -1,10 +1,10 @@
...
-This is a partial include to reuse on pages with equations/ONS embeds.
+This is a partial include to reuse on pages with equations/iframe visualisations.
 The conditionals should help with only adding the extra JavaScript when needed.
 -#}
 {% if page.has_equations %}
     <script async src="{{ mathjax() }}"></script>
 {% endif %}
-{% if page.has_ons_embed %}
+{% if page.has_iframe_visualisation %}
     <script src="{{ static('js/third-party/pym.min.js') }}"></script>
```
MebinAbraham: Now that this is back in the CMS, we should drop this from the CMS and point to the CDN. `https://cdn.ons.gov.uk/vendor/pym/1.3.2/pym.min.js`. The CDN base URL needs to be passed to the template, which can be done by adding it to https://github.com/ONSdigital/dis-wagtail/blob/main/cms/core/context_processors.py#L12

cms/jinja2/templates/components/streamfield/datavis/iframe_visualisation_block.html
```diff
@@ -0,0 +1,20 @@
...
+
+{# fmt:off #}
+{% call onsFigure(figure_config) %}
+    <div
+        data-chart-iframe
+        {# id needs prefix to be different from figure id #}
+        id="iframe--{{ block_id }}"
+        data-url="{{ iframe_config.iframeUrl }}"
+        data-title="{{ iframe_config.iframeTitle }}">
+        <iframe src="{{ iframe_config.iframeUrl }}"
```
MaciekBaron: Super minor but I think `src` should also be on it's own line to follow the convention above

cms/core/blocks/stream_blocks.py
```diff
@@ -19,3 +19,9 @@ class Meta:
     def has_equations(self, value: StreamValue) -> bool:
         """Checks if there are any equation blocks."""
         return any(block.value["content"].first_block_by_name(block_name="equation") is not None for block in value)
+
+    def has_iframe_visalisations(self, value: StreamValue) -> bool:
+        """Checks if there are any iframe visualisation blocks."""
+        return any(
+            block.value["content"].first_block_by_name(block_name="iframe_visualisation") is not None for block in value
+        )
```
MaciekBaron: Maybe now that we have two examples of this, we could abstract it like so:
```suggestion
    def has_block_with_name(self, name: str, value: StreamValue) -> bool:
        """Checks if there are any blocks with the given name."""
        return any(block.value["content"].first_block_by_name(block_name=name) is not None for block in value)
        
    def has_equations(self, value: StreamValue) -> bool:
        """Checks if there are any equation blocks."""
        return self.has_block_with_name("equation", value)

    def has_iframe_visalisations(self, value: StreamValue) -> bool:
        """Checks if there are any iframe visualisation blocks."""
        return self.has_block_with_name("iframe_visualisation", value)
```

Just a suggestion.

cms/jinja2/templates/partials/content_scripts.html
helenb: Makes sense to use the CDN. I assume you are suggesting this is stored as a setting in https://github.com/ONSdigital/dis-wagtail/blob/main/cms/settings/base.py then?

MebinAbraham: Yup, it is already a setting: https://github.com/ONSdigital/dis-wagtail/blob/main/cms/settings/base.py#L979C1-L979C12

helenb: Addressed in 5a53350e3422b214c1261026183213685ce94e2f

cms/jinja2/templates/components/streamfield/datavis/iframe_visualisation_block.html
helenb: Addressed in 5b11bd2af2d7a3bf11fa0cf4751ddcb51f0c7a65

cms/core/blocks/stream_blocks.py
helenb: Addressed in 5b11bd2af2d7a3bf11fa0cf4751ddcb51f0c7a65

MaciekBaron (approved):

cms/datavis/blocks/iframe.py
```diff
@@ -0,0 +1,191 @@
...
+
+        return config
+
+    def get_iframe_config(
+        self,
+        value: StructValue,
+        *,
+        # We don't call super() here, so these args are unused
+        parent_context: dict[str, Any] | None = None,  # pylint: disable=unused-argument
+        block_id: str | None = None,  # pylint: disable=unused-argument
```
MebinAbraham: Q: Why do these exist?. They should be removed if unused. Am I missing something?

cms/datavis/blocks/iframe.py
```diff
@@ -0,0 +1,191 @@
...
+            )
+        return errors
+
+    def get_figure_config(
+        self,
+        value: StructValue,
+        *,
+        # We don't call super() here, so these args are unused
+        parent_context: dict[str, Any] | None = None,  # pylint: disable=unused-argument
+        block_id: str | None = None,  # pylint: disable=unused-argument
```
MebinAbraham: Q: Why do these exist?. They should be removed if unused. Am I missing something?

cms/jinja2/templates/components/streamfield/datavis/iframe_visualisation_block.html
```diff
@@ -0,0 +1,21 @@
+{% from "components/figure/_macro.njk" import onsFigure %}
+
+{%- do figure_config.update(id=block_id) -%}
```
MebinAbraham: We should avoid template logic as much as possible; we should do this in Python and pass the template the final config.

cms/core/blocks/stream_blocks.py
```diff
@@ -16,6 +16,14 @@ class SectionStoryBlock(StreamBlock):
...
+    def has_block_with_name(self, name: str, value: StreamValue) -> bool:
+        """Checks if there are any blocks with the given name."""
+        return any(block.value["content"].first_block_by_name(block_name=name) is not None for block in value)
+
     def has_equations(self, value: StreamValue) -> bool:
         """Checks if there are any equation blocks."""
-        return any(block.value["content"].first_block_by_name(block_name="equation") is not None for block in value)
+        return self.has_block_with_name("equation", value)
+
+    def has_iframe_visalisations(self, value: StreamValue) -> bool:
```
MebinAbraham: Typo `visalisations` > `visualisations`

cms/core/models/base.py
```diff
@@ -139,6 +139,17 @@ def has_equations(self) -> bool:
 
         return False
 
+    @cached_property
+    def has_iframe_visualisation(self) -> bool:
+        """Checks if there are any iframe visualisation blocks."""
+        if (streamvalue := getattr(self, self.content_field_name)) and hasattr(
+            streamvalue.stream_block, "has_iframe_visalisations"
```
MebinAbraham: Typo `visalisations` > `visualisations`

cms/datavis/blocks/iframe.py
```diff
@@ -0,0 +1,191 @@
...
+        # Check for meaningful text before displaying footnotes
+        if (footnotes := value.get("footnotes")) and strip_tags(str(footnotes)).strip():
+            config["footnotes"] = {
+                "title": _("Footnotes"),
+                "content": str(footnotes),
+            }
+
+        return config
+
+    def get_iframe_config(
```
MebinAbraham: This doesn't seem to have any tests.

cms/jinja2/templates/components/streamfield/datavis/iframe_visualisation_block.html
```diff
@@ -0,0 +1,21 @@
...
+
+{# fmt:off #}
+{% call onsFigure(figure_config) %}
+    <div
+        data-chart-iframe
+        {# id needs prefix to be different from figure id #}
+        id="iframe--{{ block_id }}"
+        data-url="{{ iframe_config.iframeUrl }}"
+        data-title="{{ iframe_config.iframeTitle }}">
+        <iframe
```
MebinAbraham, line 15: Suggestion: I'm wondering if the iframe should be lazy-loaded via `loading="lazy"`?

cms/datavis/blocks/iframe.py
```diff
@@ -0,0 +1,191 @@
...
+        ),
+    )
+    # Overrides audio_description in BaseVisualisationBlock in order to update the help text
+    audio_description = blocks.TextBlock(
+        required=True,
+        help_text=(
+            "An overview of what the embed shows for screen reader users, for example"
+            "“GDP is the highest in London and lowest in the North East” or "
+            '"Inputs for users to describe what their household spends on different categories, which gives an estimate'
+            'of how much monthly spend has increased over the past year and compares to previous years"'
```
MebinAbraham: This needs to use smart quote, curly quotes, not straight quotes. The help text for accessible label uses curly quotes, which is the correct form. `“Bar chart of GDP per region” or “Interactive personal inflation calculator tool”"`

We talked about possibly adding a linter for this at some point. Maybe it's worth investing time in that.

Image

cms/datavis/blocks/iframe.py
```diff
@@ -0,0 +1,191 @@
...
+        ),
+        label="Accessible description",
+    )
```
MebinAbraham, line 52: Non-blocker: Can this not just be generalised and applied to the base instead of having variants?

cms/datavis/blocks/iframe.py
```diff
@@ -0,0 +1,191 @@
...
+            "A brief but descriptive label for the embed, for example "
+            "“Bar chart of GDP per region” or “Interactive personal inflation calculator tool”"
+        ),
+    )
+    # Overrides audio_description in BaseVisualisationBlock in order to update the help text
+    audio_description = blocks.TextBlock(
+        required=True,
+        help_text=(
+            "An overview of what the embed shows for screen reader users, for example"
+            "“GDP is the highest in London and lowest in the North East” or "
```
MebinAbraham: This is missing a space, renders as `for example“GDP is the`

cms/datavis/blocks/iframe.py
MebinAbraham: This is missing a space; it renders as `estimateof` > `estimate of`

MebinAbraham (commented):
Nothing major, functionally looks good. 👍

BJacksonONS:
This works fine on articles, but information pages can have iframe visualisations also.

image

Maybe we can refactor the changes to `SectionStoryBlock` into a mixin which is inherited by both `SectionStoryBlock` and `CoreStoryBlock`?

cms/jinja2/templates/components/streamfield/datavis/iframe_visualisation_block.html
BJacksonONS: I think it should be wrapped in `<noscript></noscript>`. Looks like pym creates an iframe element itself, so right now we're getting double fetches from the first initial load, then again after pym creates the element.

Image

BJacksonONS: > The Pym.js library and a small bit of javascript are injected onto the parent page. This code writes an iframe to the page in a container of your choice. Pym.js will actually replace the contents of the container with the newly created iframe, so it is better to use an empty container from the start to avoid surprises.

[From the docs.](https://blog.apps.npr.org/pym.js/)

cms/datavis/blocks/iframe.py
helenb: This was previously there - I just split it out from one `get_config` method to two separate methods. I assumed that the arguments were required and the comment was to explain the pylint disables. Happy to remove if they aren't needed.

cms/datavis/blocks/iframe.py
helenb: As above

cms/jinja2/templates/components/streamfield/datavis/iframe_visualisation_block.html
helenb: > Suggestion: I'm wondering if the iframe should be lazy-loaded via `loading="lazy"`?

Good idea. I think if there are no corrections and notices or headline figures, an iframed visualisation could be in the initial viewport. I might add the loading attribute if it isn't the first block on the page.

helenb: Actually I am not sure this is worth doing as it would only be for the non-js version. pym.js has a `trackscroll` option which allows custom lazy loading from the child visualisation, but nothing to set the `loading` attribute on the actual iframe that I can see. EDIT: I now realise @BJacksonONS's comment was replying to the lazy loading suggestion, so apologies - good idea to separate the iframe itself into a `noscript` tag.

helenb: @BJacksonONS I think the only benefit of leaving the `iframe` inside the the div that pym.js replaces is that if JS is unavailable due to a network issue, the iframe might still load. I believe `<noscript>` will only kick in if JS is explicitly disabled. However, I think if the pym.js docs suggest that the container div should be empty we should follow that recommendation.

cms/datavis/blocks/iframe.py
helenb: The help text here was an AC specifically for the iframe so I think it needs to be overridden here.

cms/datavis/blocks/iframe.py
helenb: Seems OK to remove them - see 05053a1c869314434d1f3491de479621a78ea8bd

cms/datavis/blocks/iframe.py
helenb: Seems ok to remove them - see 05053a1c869314434d1f3491de479621a78ea8bd

cms/jinja2/templates/components/streamfield/datavis/iframe_visualisation_block.html
helenb: Addressed in 65120556413622f0bdf2e9e50b0b73483d16b581

cms/core/blocks/stream_blocks.py
helenb: Addressed in dae01a2f7b3713abd2a7a439482c436327fe5cc5

cms/core/models/base.py
helenb: Addressed in dae01a2f7b3713abd2a7a439482c436327fe5cc5

cms/jinja2/templates/components/streamfield/datavis/iframe_visualisation_block.html
helenb: @BJacksonONS your point addressed in 6bdd0e19f8bf66f58639f30efaa787afbd2fd5f3

cms/datavis/blocks/iframe.py
helenb: Addressed in eda4af85c6e004ce0122a5b82d2ddb7b1631b3c8

cms/datavis/blocks/iframe.py
helenb: Addressed in eda4af85c6e004ce0122a5b82d2ddb7b1631b3c8

cms/datavis/blocks/iframe.py
helenb: Addressed in eda4af85c6e004ce0122a5b82d2ddb7b1631b3c8

cms/datavis/blocks/iframe.py
helenb: Looking at the diffs we didn't previously have any specific tests around the config - the existing tests focus on validation. Is it config or rendering tests you would like added here, or both?

helenb:
> This works fine on articles, but information pages can have iframe visualisations also.

> Maybe we can refactor the changes to `SectionStoryBlock` into a mixin which is inherited by both `SectionStoryBlock` and `CoreStoryBlock`?

@BJacksonONS addressed in 99cd696fb3e5310e1ef6cb90efa50ccc08d1977e

MebinAbraham:
> > This works fine on articles, but information pages can have iframe visualisations also.
> 
> > Maybe we can refactor the changes to `SectionStoryBlock` into a mixin which is inherited by both `SectionStoryBlock` and `CoreStoryBlock`?
> 
> @BJacksonONS addressed in [99cd696](https://github.com/ONSdigital/dis-wagtail/commit/99cd696fb3e5310e1ef6cb90efa50ccc08d1977e)

This will need a test given info page was broken prior to this change, to ensure we don't regress.

cms/datavis/blocks/iframe.py
MebinAbraham: I guess my point is does it need to be?. Non blocker can revisit in future.

cms/jinja2/templates/components/streamfield/datavis/iframe_visualisation_block.html
MebinAbraham: Resolving on behalf.

cms/datavis/blocks/iframe.py
MebinAbraham: We want to test the config `get_iframe_config` like we do the other methods and we also need tests to verify the iframe is loaded on the page correctly both iframe and nonscript tag variant, I don't think we have any tests that actually verify this gets rendered?

MaciekBaron:
@MebinAbraham - Tests added in [af34d1b](https://github.com/ONSdigital/dis-wagtail/pull/744/commits/af34d1b73f5c74b9b3d929c74b0216a094692d12)

cms/jinja2/templates/components/streamfield/datavis/iframe_visualisation_block.html
```diff
@@ -0,0 +1,23 @@
...
+        data-title="{{ iframe_config.iframeTitle }}">
+    </div>
+    <noscript>
+       {# Load the iframe directly if JS is disabled #}
+        <iframe
+            src="{{ iframe_config.iframeUrl }}"
+            title="{{ iframe_config.iframeTitle }}"
+            class="chart-embed"
+        >
+        </iframe>
```
MebinAbraham, line 20: We should set width to 100% and height to a sensible value maybe 500px etc?. Height 100% doesn't seem to work. Default value is 100x300 which doesn't look great.

MebinAbraham (approved):

cms/jinja2/templates/components/streamfield/datavis/iframe_visualisation_block.html
charlotteromaniuk: Not sure if this is the same point, but [this visualisation](https://www.ons.gov.uk/visualisations/dvc3388/dashwrapper/index.html) doesn't seem to be re-sizing correctly

Locally:
image

Current ONS website:
image

MebinAbraham: Not related to the above, as that is non-JS only. Good spot, though; that is likely because our "container" size is smaller than the current ons hence the issue. When we do the wider width, it should be resolved. Not sure if pymjs can handle this. If our container size was 69% it fits (currently it is 66%).

MaciekBaron: I think you may have been using outdated CSS, Helen set the aspect ratio to 16/9 so it shouldn't render as 100x300 (it does with the old CSS, I checked). Looks like this now:

Image

MebinAbraham: 👍 Had outdated css, working as expected.

charlotteromaniuk (approved):
Functionally tested, LGTM

cms/jinja2/templates/components/streamfield/datavis/iframe_visualisation_block.html
sanjeevz3009: Resolving on behalf @MebinAbraham

sanjeevz3009 (approved):
LGTM!

PR #745, opened 2026-07-28
fix: add GTM data attributes to file download links

charlotteromaniuk:

Adds the following GTM data attributes to the chart CSV, table and image download links (the DS adds these to the parent `li` element by default):
- `data-ga-event`
- `data-ga-file-extension`
- `data-ga-file-name`
- `data-ga-link-text`
- `data-ga-link-url`
- `data-ga-link-domain`
- `data-ga-file-size`

For the chart CSV download links, these data attributes are also added:
- `data-ga-chart-title`
- `data-ga-chart-type`

See the ticket and specification attached to it for the expected values for each attribute.

**Other changes:**
- Separated out the logic for determining the download file name for charts and tables to the helper functions `get_csv_download_filename` and `get_table_csv_download_title`, allowing the same file name to be used for the `data-ga-file-name` data attribute and ensure it stays aligned with the actual downloaded file name
- Fixed a small bug where the 'Download CSV' link text for charts wasn't being translated

### How to review

- Add an image block, table block and chart block to a page
- Save and publish the page
- On the published page, inspect the download links using browser dev tools. Ensure the correct data attributes are added to the link's parent `li` element for the chart and table CSV download links and the image download link (the chart image download hasn't been implement yet)
- Ensure the expected values match the specification
- Check there is no regression in the behaviour of downloading chart and table CSVs and images

### Deployment Safety

Please select one:

- [X] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

- [ ] GTM data attributes need to be added to the chart image download link once the functionality has been implemented

---
Ticket: [DPD-4318](https://officefornationalstatistics.atlassian.net/browse/DPD-4318)

[DPD-4318]: https://officefornationalstatistics.atlassian.net/browse/DPD-4318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cms/core/blocks/markup.py
```diff
@@ -286,6 +311,22 @@ def _build_preview_table_download_url(page: Any, block_id: str, request: Any = N
...
+            "data-ga-event": "file-download",
+            "data-ga-file-extension": "csv",
+            "data-ga-file-name": file_name,
+            "data-ga-link-text": text,
+            "data-ga-link-url": parsed_url.path,
+            "data-ga-file-size": file_size,
+        }
+        if parsed_url.hostname is not None:
+            attributes["data-ga-link-domain"] = parsed_url.hostname
+
```
MaciekBaron: Question: This is nearly identical to the image one but does not add the file size attribute. Is this intended?

cms/core/blocks/embeddable.py
```diff
@@ -90,15 +87,62 @@ def get_context(self, value: StreamValue, parent_context: dict | None = None) ->
...
+            file_size_bytes=size_bytes,
+        )
+
+        return {
+            "title": _("Download this image"),
+            "itemsList": [{"text": link_text, "url": large_image.url, "download": "file", "attributes": attributes}],
+        }
+
+    @staticmethod
+    def _get_gtm_attributes_image_download(
```
MaciekBaron: Since this is a static method to begin with, how about we move this to a utility function? This will avoid repetition, as you do nearly the exact same thing with the CSV download.

Might also be good to force keyword arguments with (`*`) for better readability

MaciekBaron (changes requested):
Nice work, just a couple of comments

cms/core/blocks/embeddable.py
charlotteromaniuk: Thanks that makes sense, I’ve added it here [f56d444](https://github.com/ONSdigital/dis-wagtail/pull/745/commits/f56d444017cb911571ab78ef3a92d6f80d9460cc)

I’ve also used the utility function for the existing document download link data attributes. I added a fix there as well - the download url is relative so the domain was always “None”. I’ve updated it now to build an absolute url and retrieve the domain from that instead.

cms/core/blocks/markup.py
charlotteromaniuk: They both add the file size attribute - its added directly to the table attributes but conditionally to the image attributes because there could be a case where the image doesn’t exist when the page is rendered, so I wanted to prevent a value of “None” being added to the data attributes. Both tables and images follow the same logic now though that I’ve created a utility function.

cms/data_downloads/views.py
```diff
@@ -10,7 +10,11 @@
 from wagtail.models import Page
 
 from cms.data_downloads.permissions import get_revision_page_for_request, user_can_access_csv_download
-from cms.data_downloads.utils import create_data_csv_download_response_from_data
+from cms.data_downloads.utils import (
+    create_data_csv_download_response_from_data,
+    get_csv_download_filename,
+    get_table_csv_download_title,
```
helenb, line 16: I don't think that this is introduced by your pr, but I noticed that if a table has no title set, the filename is `access.csv` and I was wondering why as it isn't a an intuitive filename.

charlotteromaniuk: Hmm it should use the accessible label if the title isn't set, is the accessible label for your table set to "access"?

helenb: Ah yes you are right - my lazy testing data confusing me.

cms/datavis/blocks/base.py
```diff
@@ -346,9 +351,29 @@ def _get_csv_download_item(
...
             "url": csv_url,
+            "attributes": self._get_gtm_attributes_csv_download(link_text, csv_url, file_size_kb, value),
+        }
+
+    def _get_gtm_attributes_csv_download(
+        self, text: str, url: str, file_size: str, value: StructValue
+    ) -> dict[str, str]:
+        parsed_url = urlparse(url)
+        filename = get_csv_download_filename(title=value.get("title"), fallback_stem="chart")
+        return {
```
BJacksonONS, line 369: I think we can simplify this a little bit with something like the following

```python
return {
    **get_gtm_attributes_file_download(
        text=text, url=url, file_extension="csv", file_name=filename, file_size_kb=file_size
    ),
    "data-ga-chart-title": value.get("title"),
    "data-ga-chart-type": self.get_highcharts_chart_type(value),
}
```

cms/core/blocks/embeddable.py
```diff
@@ -90,15 +89,39 @@ def get_context(self, value: StreamValue, parent_context: dict | None = None) ->
...
 
         context["options"] = options
         return context
 
+    def _get_download_options(self, large_image: AbstractRendition, request: HttpRequest | None) -> dict:
+        """Build the download payload for the image."""
+        _base, ext = os.path.splitext(large_image.file.name)
+        file_type = ext.lstrip(".").upper() or "IMG"
+        size_bytes = getattr(large_image.file, "size", None)
+        size_kb = str(size_bytes / 1000) if size_bytes is not None else None  # Convert from Bytes to KB
```
BJacksonONS: This divides by 1000 rather than 1024 and doesn't round up, so isn't consistent with the result of `get_approximate_file_size_in_kb`.

Not sure which is the correct one but I think we probably want to be consistent?

BJacksonONS (commented):
Just a couple things I noticed, otherwise looks good.

cms/datavis/blocks/base.py
charlotteromaniuk: Thanks, updated here [fdb421c](https://github.com/ONSdigital/dis-wagtail/pull/745/commits/fdb421c19495d4db0e9a8adf8dff7039e1b6e0b8)

cms/core/blocks/embeddable.py
charlotteromaniuk: Good point, I'll check which one we should be using

charlotteromaniuk: Checked with Awen and she's happy for the actual unrounded file size to be used - updated here [09c4a94](https://github.com/ONSdigital/dis-wagtail/pull/745/commits/09c4a948c63cb104e266520f1b53910c1d42f705)

MebinAbraham: @charlotteromaniuk I suggest we stick with 1024 to be consistent with `get_approximate_file_size_in_kb`; otherwise, we will cause confusion for ourselves.

charlotteromaniuk: @MebinAbraham thanks, I've updated this. Dividing by 1024 was causing a lot of long decimals to be returned so I've rounded the values to 3 d.p. which I think should give enough precision to compare file sizes. Also moved this into a new helper function `format_file_size_kb` to reduce duplication. 

[b5470c7](https://github.com/ONSdigital/dis-wagtail/pull/745/commits/b5470c7367766093964038ec4383df254de0c8e4)

cms/locale/cy/LC_MESSAGES/django.po
```diff
@@ -73,29 +73,30 @@ msgstr "Dangos popeth"
...
+#: cms/core/blocks/embeddable.py:121
 msgid "Download this image"
 msgstr "Lawrlwythwch y ddelwedd hon"
 
-#: cms/core/blocks/markup.py:262
+#: cms/core/blocks/markup.py:271 cms/datavis/blocks/base.py:354
+#, fuzzy, python-format
+#| msgid "Download CSV %(size)s"
+msgid "Download CSV (%(size)s)"
+msgstr "Lawrlwythwch CSV %(size)s"
```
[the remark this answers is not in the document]
charlotteromaniuk: [090b3b9](https://github.com/ONSdigital/dis-wagtail/pull/745/commits/090b3b9e4194cda2d0df09d67d94e42d98fbceff)

cms/core/blocks/markup.py
```diff
@@ -247,22 +266,33 @@ def _get_download_config(self, *, parent_context: dict, block_id: str, data: dic
...
+        file_name = get_csv_download_filename(title=csv_title, fallback_stem="table")
+
+        csv_url = self._get_table_download_url(parent_context=parent_context, page=page, block_id=block_id)
+        attributes = get_gtm_attributes_file_download(
+            text=link_text,
+            url=csv_url,
+            file_extension="csv",
+            file_name=file_name,
+            file_size_kb=file_size_kb,
         )
```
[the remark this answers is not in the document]
charlotteromaniuk: `csv_url` has been an absolute URL in practice, but this isn’t guaranteed so I’ve added Copilot’s recommendation [e85da94](https://github.com/ONSdigital/dis-wagtail/pull/745/commits/e85da942c9c4dd72b140e107bdf1ba824cf3bfb7)

cms/core/tests/test_blocks.py
```diff
@@ -1061,6 +1068,67 @@ def test_ons_table_block_builds_published_url_correctly(self):
...
+
+        expected_attributes = {
+            "data-ga-event": "file-download",
+            "data-ga-file-extension": "csv",
+            "data-ga-file-name": get_csv_download_filename(title=self.full_data.get("title"), fallback_stem="table"),
+            "data-ga-link-text": expected_link_text,
+            "data-ga-link-url": urlparse(download_items_list["url"]).path,
+            "data-ga-link-domain": urlparse(download_items_list["url"]).hostname,
+            "data-ga-file-size": expected_file_size,
+        }
```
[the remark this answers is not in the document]
charlotteromaniuk: Same as above [e85da94](https://github.com/ONSdigital/dis-wagtail/pull/745/commits/e85da942c9c4dd72b140e107bdf1ba824cf3bfb7)

cms/datavis/blocks/base.py
```diff
@@ -346,9 +352,24 @@ def _get_csv_download_item(
...
             csv_url = self._build_chart_download_url(page, block_id, superseded_version)
 
+        link_text = _("Download CSV (%(size)s)") % {"size": file_size_with_unit}
+
         return {
-            "text": f"Download CSV{suffix}",
+            "text": link_text,
             "url": csv_url,
+            "attributes": self._get_gtm_attributes_csv_download(link_text, csv_url, file_size_kb, value),
+        }
```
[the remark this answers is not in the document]
charlotteromaniuk: Same as above [e85da94](https://github.com/ONSdigital/dis-wagtail/pull/745/commits/e85da942c9c4dd72b140e107bdf1ba824cf3bfb7)

cms/datavis/tests/test_chart_blocks_base.py
```diff
@@ -262,3 +269,80 @@ def test_download_config_not_in_preview_mode_uses_real_url(self):
...
+            "data-ga-event": "file-download",
+            "data-ga-file-extension": "csv",
+            "data-ga-file-name": get_csv_download_filename(title=self.raw_data.get("title"), fallback_stem="chart"),
+            "data-ga-link-text": expected_link_text,
+            "data-ga-link-url": urlparse(expected_url).path,
+            "data-ga-link-domain": urlparse(expected_url).hostname,
+            "data-ga-chart-title": self.raw_data.get("title"),
+            "data-ga-chart-type": "line",
+            "data-ga-file-size": expected_file_size,
+        }
```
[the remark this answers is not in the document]
charlotteromaniuk: Same as above [e85da94](https://github.com/ONSdigital/dis-wagtail/pull/745/commits/e85da942c9c4dd72b140e107bdf1ba824cf3bfb7)

sanjeevz3009 (approved):
Apart from the comments already made. LGTM! Good work!

cms/data_downloads/utils.py
```diff
@@ -123,21 +123,33 @@ def sanitize_data_for_csv(data: list[list[str | int | float]]) -> list[list[str
...
     Returns:
         A Django HttpResponse object configured for CSV file download.
     """
-    filename = slugify(title) or "chart"
     response = HttpResponse(
         content_type="text/csv",
         headers={
-            "Content-Disposition": f'attachment; filename="{filename}.csv"',
+            "Content-Disposition": f'attachment; filename="{filename}"',
         },
```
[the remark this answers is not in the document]
charlotteromaniuk: Addressed here [5fe890d](https://github.com/ONSdigital/dis-wagtail/pull/745/commits/5fe890d34ff989b920f9cf49e7d42d1dfb3b9620)

cms/core/blocks/embeddable.py
```diff
@@ -90,15 +90,39 @@ def get_context(self, value: StreamValue, parent_context: dict | None = None) ->
             options["footnotes"] = {"title": _("Footnotes"), "content": notes}
 
         if value.get("download"):
-            size_text = f" ({file_size_human})" if file_size_human else ""
-            options["download"] = {
-                "title": _("Download this image"),
-                "itemsList": [{"text": f"{file_type}{size_text}", "url": large.url, "download": "file"}],
-            }
+            options["download"] = self._get_download_options(large, context.get("request"))
```
BJacksonONS, line 93: praise: tidier, and also avoids calculating the sizes when we won't even use them

kevinhowbrook (approved):

BJacksonONS:
Resolved some copilot comments, think this is good once @MaciekBaron gets a chance to re-review.

cms/core/blocks/embeddable.py
```diff
@@ -90,15 +90,39 @@ def get_context(self, value: StreamValue, parent_context: dict | None = None) ->
...
+        size_text = f" ({file_size_human})" if file_size_human else ""
+        link_text = f"{file_type}{size_text}"
+
+        # Build an absolute URL when possible so GTM attributes include the link domain
+        absolute_url = request.build_absolute_uri(large_image.url) if request else large_image.url
+        file_name = os.path.basename(urlparse(large_image.url).path)
+
+        attributes = get_gtm_attributes_file_download(
+            text=link_text,
+            url=absolute_url,
```
MebinAbraham, line 115: The spec example uses a path instead of url, we should check with Awen, if it needs to be path or if url is fine.

cms/core/tests/test_blocks.py
```diff
@@ -72,8 +77,8 @@ def test_document_block__block_value(self):
                     "data-ga-file-name": self.document.title,
                     "data-ga-link-text": "The block document",
                     "data-ga-link-url": self.document.url,
-                    "data-ga-link-domain": urlparse(self.document.url).hostname,
-                    "data-ga-file-size": "0.025",  # KB
+                    "data-ga-link-domain": urlparse(absolute_url).hostname,
```
MebinAbraham, line 80: This might not be as expected, as this will be pointing to the domain for the media bucket (it won't be `www.ons.gov.uk` as per the spec example. I'm not sure if there is any value in this attribute, especially if the `url` above doesn't need to be a path. We should check with Awen. This, on publish, will be pointing to the `media.ons.gov.uk` domain.

cms/core/blocks/embeddable.py
charlotteromaniuk: The url is passed into this function, but within the function the attribute itself is set to the path https://github.com/ONSdigital/dis-wagtail/pull/745/changes#diff-f7024ab0c58b477bbc012fd809e58ba6e31e789d3cae9ea91ce399d0b1db7742R42

cms/core/tests/test_blocks.py
charlotteromaniuk: I checked this with Awen and she's expecting the attribute to be the domain where the file is downloaded from, so she's happy for that to be `media.ons.gov.uk` in this case.

MebinAbraham (approved):

MaciekBaron (approved):

joepercival (approved):

PR #746, opened 2026-07-29
fix: serve an empty png instead of the external jspreadsheet logo

helenb:

Addresses an issue with the functional tests. The compiled js file for `jspreadsheet` (`static/wagtailtables/js/jspreadsheet.js`) references an external logo file: `img.src = '//bossanova.uk/jspreadsheet/logo.png';` which violates the CSP. This console error then blocks the page from loading fully and there is a timeout error.

This PR serves an empty png file instead of any external images from the bossanova.uk domain.

### How to review

See if functional tests now pass, check code changes.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/functional_tests/environment.py
+++ b/functional_tests/environment.py
@@ -129,6 +129,22 @@ def before_scenario(context: Context, scenario: Scenario) -> None:
 
     context.page = context.playwright_context.new_page()
 
+    # Intercept external image requests from jspreadsheet to avoid CSP violations in CI.
+    # jspreadsheet.js tries to load //bossanova.uk/jspreadsheet/logo.png at runtime.
+    _empty_png = (
+        b"\x89PNG\r\n\x1a\n\x00\x00\x00\rIHDR\x00\x00\x00\x01\x00\x00\x00\x01"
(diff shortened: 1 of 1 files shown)


BJacksonONS:
CSP looks to be a red herring in this case, the actual issue is the design system script progressive enhancement causing a race condition. The design system first [hides the content when it adds some of the classes](https://github.com/ONSdigital/design-system/blob/4813ae0cc4d90010cd46a55d7eccde10ddc31e63/src/components/details/_details.scss#L55), then changes the [role from button to link](https://github.com/ONSdigital/design-system/blob/4813ae0cc4d90010cd46a55d7eccde10ddc31e63/src/components/details/details.js#L17), so the locator in the test can sometimes pass.

I think the right fix is to wait until the design system has initialized, then expand content by text then find the link.

```python
download_details = context.page.locator('.ons-js-details[id^="figure-downloads--"]')
expect(download_details).to_have_class("ons-details--initialised")
download_details.get_by_text("Download: line chart").click()
csv_download_link = context.page.get_by_role("link", name="Download CSV")
expect(csv_download_link).to_be_visible()
```

PR #747, opened 2026-08-03
fix(articles): separate link analytics values into own property to prevent recursive page rendering

BJacksonONS:

When adding google analytics attributes to related links for other article pages, we were generating the entire `cached_analytics_values` for the related article.

This dictionary contained several fields unnecessary in the related link block, which included word count for the related page. Word count is currently generated by rendering the streamfield content of the related page, which would also trigger the same `cached_analytics_values` property for _its_ related pages. Due to the way the query ran, this would not take advantage of any property caching and also end up traversing the entire network of all connected related articles, causing a huge spike in the number of queries necessary to render a page. In the worst case, if there was a link cycle between pages, this would trigger a recursion error.

This PR separates the attributes relevant to related links into it's own property, to prevent unnecessary word count calculations.

### How to review

1. Create a cycle of related pages, a large network of related pages, and try to link a page to itself
2. Publish and visit all the pages
3. Should get no recursion errors, and query count in debug toolbar should show ~40 queries. (Varies based on auth status and page level cache warming, try refreshing a few times)

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

Confirm if pages should be able to pick themselves as related links (probably not?).

---
Ticket: [CMS-1270](https://officefornationalstatistics.atlassian.net/browse/CMS-1270)

[CMS-1270]: https://officefornationalstatistics.atlassian.net/browse/CMS-1270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BJacksonONS:
In my testing I made 6 pages, which link each other as follows

1 - 2 - 3 - 4 - 5 & 6

With fix:

image

Without fix:

image

Edit:

Once cache is warmed, this took page load time from ~2000ms to ~450ms.

cms/articles/models.py
```diff
@@ -847,16 +847,20 @@ def serve(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse
...
             "releaseDate": format_date_for_gtm(self.release_date),
+        }
+
+    @cached_property
+    def cached_analytics_values(self) -> dict[str, str | bool]:
+        values: dict[str, str | bool] = {
+            "pageTitle": self.display_title,
             "latestRelease": bool_to_yes_no(self.is_latest),
-            "wordCount": self.word_count,
+            "wordCount": str(self.word_count),
```
MaciekBaron, line 880: Question: Why do we convert this to a string, other than matching the existing return type?

BJacksonONS: That was my only reason for changing it, not entirely sure why mypy was happy with it before but it was throwing an error after as the func claims to return string attributes.

MebinAbraham: We should check whether this should be a str or an int in the GTM spec.

BJacksonONS: I'll revert and update the func signature instead

MebinAbraham: In the spec it, it is a string so this would be fine.

BJacksonONS: Put it back to a string https://github.com/ONSdigital/dis-wagtail/pull/747/commits/2cf834a439d6008046f2499808d4b45a183dac8b

This is what the docs specify it should be, and updating the type to accept int ends up propagating across all page types.

MebinAbraham (commented):
Will need new tests.

cms/articles/models.py
sanjeevz3009: Resolving on behalf!

cms/articles/tests/test_article_models.py
kevinhowbrook, line 1: **Suggestion**

I think we would benefit from a testing two articles whose content each contains a section and related_links block pointing at the other (plus a third page linking to itself), publish, then self.client.get() each page and asserting the query count stays under a (sensible) limit (like 60).

sanjeevz3009 (approved):
Apart from the comments, already on the PR. LGTM!

cms/articles/tests/test_article_models.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/747/commits/20d6f30aff62bb9086d23a749e50a5e1334f21a6

BJacksonONS:
@MebinAbraham 

Tests added here

https://github.com/ONSdigital/dis-wagtail/pull/747/commits/20d6f30aff62bb9086d23a749e50a5e1334f21a6

kevinhowbrook (approved):
awesome @BJacksonONS - thanks

cms/articles/tests/test_article_models.py
```diff
@@ -1356,3 +1382,78 @@ def test_known_type_in_chart_block_types(self):
 
     def test_known_non_chart_type_not_in_chart_block_types(self):
         self.assertNotIn("rich_text", CHART_BLOCK_TYPES)
+
+
+class StatisticalArticlePageRelatedLinksRecursionTestCase(WagtailTestUtils, TestCase):
+    """Regression tests for CMS-1270.
```
sanjeevz3009: Hmm, we would want to rephrase this to something else, other than `Regression tests for CMS-1270`. Don't really want the ticket number here. We can just remove this line. The docstring below does justice.

sanjeevz3009 (approved):
One minor comment but other than that LGTM!

MebinAbraham (approved):

sanjeevz3009 (approved):
Did some further testing locally, and all seems to be working well.

cms/articles/tests/test_article_models.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/747/commits/575533cce43b89f079d2600057d233691dca5f2a

sanjeevz3009: Thanks @BJacksonONS

sanjeevz3009 (approved):

PR #750, opened 2026-08-04
chore(deps): ignore cryptography warnings for now

BJacksonONS:

OSV-Scanner is giving warnings for cryptography now, but they [dropped support for intel macs](https://cryptography.io/en/latest/changelog/#v49-0-0) in 49 upwards. Add an ignore for now until the team has either fully upgraded to apple silicon, or we set up tooling for emulating arm on x64.

### How to review

Review `osv-scannerl.toml` changes.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


MebinAbraham (approved):

MebinAbraham (commented):
I suggest pin to <49

BJacksonONS:
> I suggest pin to <49

@MebinAbraham 

https://github.com/ONSdigital/dis-wagtail/pull/750/commits/73ed0ec25dcd1c1dcaea84af6e5135977869376f

pyproject.toml
```diff
@@ -44,7 +44,7 @@ django-ses = "^4.7.2"
 aws-msk-iam-sasl-signer-python = "^1.0.2"
 wagtailtables = "^0.2.2"
 pyjwt = "^2.13.0"
-cryptography = "^48.0.1"
```
MebinAbraham, line 47: Ah, forgot this already doesn't allow `49` anyway, but maybe the new rule is more explicit in our intentions vs default rule.

MebinAbraham (approved):

pyproject.toml
BJacksonONS: Yeah it wouldn't have auto updated major version but happy to have the explicit version given it'll just totally break our local dev.

MaciekBaron (approved):

sanjeevz3009 (commented):
Reviewed!

PR #751, opened 2026-08-04
fix(functional-tests): ensure design system is initialised, then look for csv download elements

BJacksonONS:

There is a race condition between a playwright test assertion and the design system loading and performing progressive enhancement on the site in a functional test. This ensures the design system is loaded and then does the necessary checks to improve reliability of the test.

### How to review

Ensure the functional test `Scenario: A chart on a statistical article page has a CSV download link` passes.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

---
Ticket: [CMS-1337](https://officefornationalstatistics.atlassian.net/browse/CMS-1337)

[CMS-1337]: https://officefornationalstatistics.atlassian.net/browse/CMS-1337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


functional_tests/steps/statistical_article_page.py
```diff
@@ -731,7 +731,9 @@ def a_statistical_article_page_with_configured_listing_image_exists(
 @then("the page has a CSV download link for the chart")
 def the_page_has_a_csv_download_link_for_the_chart(context: Context) -> None:
     """Check that the page has a CSV download link for the chart."""
-    context.page.get_by_role("button", name="Download: line chart").click()
+    download_details = context.page.locator('.ons-js-details[id^="figure-downloads--"]')
+    expect(download_details).to_have_class("ons-details--initialised")
+    download_details.get_by_text("Download: line chart").click()
```
MebinAbraham, line 738: Nit: Worth adding a comment to make this explicit.

MebinAbraham (approved):
👍

functional_tests/steps/statistical_article_page.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/751/commits/e98a05950c8b0564f913ead84f05d363b997eaa7

functional_tests/steps/statistical_article_page.py
```diff
@@ -731,7 +731,11 @@ def a_statistical_article_page_with_configured_listing_image_exists(
 @then("the page has a CSV download link for the chart")
 def the_page_has_a_csv_download_link_for_the_chart(context: Context) -> None:
     """Check that the page has a CSV download link for the chart."""
-    context.page.get_by_role("button", name="Download: line chart").click()
+    download_details = context.page.locator('.ons-js-details[id^="figure-downloads--"]')
+    # Wait for this class as it signifies the design system js has loaded, initialised and updated
+    # element roles and styles
+    expect(download_details).to_contain_class("ons-details--initialised")
```
MaciekBaron, line 737: Question: are we happy with the default timeout?

MaciekBaron (approved):
LGTM apart from the non-blocking question

functional_tests/steps/statistical_article_page.py
BJacksonONS: I think realistically it should only ever fail if there's a bug or major change in the design system, but we could shorted it a bit if we want to fail fast.

The failure condition for the race was that the DS had loaded quicker than our assertions so it definitely shouldn't need to be longer.

sanjeevz3009 (commented):
Reviewed!

PR #753, opened 2026-08-04
chore(deps): configure dependabot to ignore cryptography

BJacksonONS:
### What is the context of this PR?

Config dependabot so it won't try to bump cryptography package before we're able to run it (see https://github.com/ONSdigital/dis-wagtail/pull/750).

### How to review

Check config against schema

### Deployment Safety

Bleed and Sandbox deploy automatically on merge, so PRs should be safe to deploy immediately.

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

List any follow-up actions (if applicable), like needed documentation updates or additional testing.

PR #755, opened 2026-08-04
chore(deps): ignore all cryptography updates including security

BJacksonONS:

Previous PR (https://github.com/ONSdigital/dis-wagtail/pull/753) doesn't apply to security updates, change ignore config so dependabot doesn't apply any updates to cryptography.

See https://github.com/ONSdigital/dis-wagtail/pull/750 for why we're doing this.

### How to review

Check config.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -50,5 +50,7 @@ updates:
           - '*'
     ignore:
       - dependency-name: cryptography
+        # Currently 49 and up don't support intel mac
+        versions: ['>=49.0.0']
     cooldown:
       default-days: 7
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -50,7 +50,5 @@ updates:
           - '*'
     ignore:
       - dependency-name: cryptography
-        update-types:
-          - 'version-update:semver-major'
     cooldown:
       default-days: 7


.github/dependabot.yml
```diff
@@ -50,7 +50,5 @@ updates:
           - '*'
     ignore:
       - dependency-name: cryptography
-        update-types:
-          - 'version-update:semver-major'
```
MebinAbraham, line 54: Maybe try `versions: [ ">=49.0.0" ]` instead. 

Docs did say: `update-types only affects version updates, not security updates. Security updates will always be created regardless of the update-types setting.` but doesn't mention anything about `versions`.

BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/755/commits/eea6f09798cb862e450e402a9a1346fb492ada2c

sanjeevz3009 (approved):
Apart from @MebinAbraham comments on this PR and Slakc thread, LGTM!

BJacksonONS:
Given the conditional dependency solution will close this, remove today's dependabot config changes and see what it tries to do.

PR #763, opened 2026-08-11
refactor(bundles): avoid suppressing signals for post publish actions

BJacksonONS:

Refactor the signal handling for post-publish actions.

Previously we suppressed signals and executed actions manually, but this has issues with signals not running for aliases. As this may introduce large maintenance burden later if other signals are added to core, we'd like to avoid this.

This PR adds a context variable to track the currently being published bundle to pass into the signal as it enqueues action. Context variables were chosen as in future we may possibly be executing in non-os threads.

This ensures we also have dedicated actions tracked for aliases, and preserves the action registry pattern.

### How to review

Ensure tests run successfully
Check action enqueueing runs correctly

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

Add a cleanup card to convert existing usages of thread locals to context variables.

### DIFF
--- a/cms/bundles/utils.py
+++ b/cms/bundles/utils.py
@@ -439,11 +439,12 @@ def publish_bundle(bundle: Bundle, *, update_status: bool = True) -> bool:  # py
     failed_pages: list[int] = []
     total_pages = bundle.get_bundled_pages().count()
 
+    PostPublishAction.objects.filter(bundle=bundle).delete()
+
     for page in bundle.get_bundled_pages().specific(defer=True).select_related("latest_revision"):
         try:
             # Durable ensures no other savepoint will roll back the publish
             with transaction.atomic(durable=True):
-                PostPublishAction.objects.filter(bundle=bundle, page=page).delete()
                 if workflow_state := page.current_workflow_state:
                     with enqueue_post_publish_actions_for_bundle(bundle):
                         # finish the workflow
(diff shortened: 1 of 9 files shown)


cms/post_publish_actions/signal_handlers.py
```diff
@@ -1,41 +1,38 @@
 from collections.abc import Generator
 from contextlib import contextmanager
-from typing import Any
+from contextvars import ContextVar
```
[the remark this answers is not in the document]
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/763/changes/3c990b2c444898341e3778dd134477f65cfa9f4b

cms/bundles/utils.py
```diff
@@ -446,26 +445,20 @@ def publish_bundle(bundle: Bundle, *, update_status: bool = True) -> bool:  # py
             with transaction.atomic(durable=True):
                 PostPublishAction.objects.filter(bundle=bundle, page=page).delete()
```
MebinAbraham: We need to do this for aliases?. And likely handle actions created by aliases elsewhere, our filtering etc doesn't account for actions created by aliases.

BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/763/commits/58d44675abd91f006627af2f47de65a71f1b6fe3

MebinAbraham (approved):
~~Search already excludes alias published signed, but we need to update s3 update to exclude alias.~~ You already did it 🙌  Only cache purge would be valid to trigger post publish for aliases.

MebinAbraham (approved):

sanjeevz3009 (approved):
LGTM!

sanjeevz3009 (approved):
Smashing approval to fix GitHub possible stacked PR bug 1

PR #764, opened 2026-08-11
fix: require linked release calendar approval before a bundle can be approved

charlotteromaniuk:

Adds validation when approving a bundle to ensure its linked release calendar is approved and ready to publish.

### How to review

1. Create a calendar entry bundle
2. Add an article and send this for review
3. Make changes to the release calendar page (e.g. add a contact detail snippet), and send for review
4. As a second user, view the bundle and approve the article - do not look at or approve the calendar entry changes
5. As the original user, check that you cannot approve the bundle for publication
6. Approve the calendar entry
7. Check that the bundle can now be approved and published

### Deployment Safety

Please select one:

- [X] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions
N/A

---
Ticket: [DPD-5228](https://officefornationalstatistics.atlassian.net/browse/DPD-5228)

[DPD-5228]: https://officefornationalstatistics.atlassian.net/browse/DPD-5228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cms/bundles/tests/test_forms.py
```diff
@@ -201,6 +201,23 @@ def test_clean__validates_page_must_be_ready_for_review(self):
         self.assertIsNone(form.cleaned_data["approved_by"])
         self.assertIsNone(form.cleaned_data["approved_at"])
 
+    def test_clean__validates_release_calendar_must_be_ready_for_review(self):
```
sanjeevz3009: You are checking if the bundle is ready to publish. Update the test name please.

Something like `test_clean__validates_release_calendar_must_be_ready_to_publish`

sanjeevz3009: Good test case for negative test case here, but we need to have a positive test case/ happy path.

Currently, I don't see any form-level test case for a passing case, so `RC page ready -> form valid`

cms/bundles/models.py
```diff
@@ -242,12 +242,17 @@ def has_unpublished_changes(self) -> bool:
     def can_be_approved(self) -> bool:
```
MebinAbraham: Where do we currently use this?

sanjeevz3009: I was digging into this and the answer is nowhere.

There have been some changes here, but we don't use it anywhere.

Actually, it's used in tests. Within `bundles/tests/test_models.py`

But then, this shouldn't be in here anyway. It's not a test util.

_**Edit: I edited my comments to collapse it all into one to keep it tidy**_

MebinAbraham (commented):
While we are here, I suggest we update the `ReleaseChooserWithDetailsPanel` to include the page workflow status, this will make it consistent with how we show it for bundled pages. This was raised by the business before, so now would be a good time to add this. Can be a follow up PR, but let's do it as part of the same ticket.

Image

cms/bundles/models.py
sanjeevz3009: @charlotteromaniuk

sanjeevz3009:
> While we are here, I suggest we update the `ReleaseChooserWithDetailsPanel` to include the page workflow status, this will make it consistent with how we show it for bundled pages. This was raised by the business before, so now would be a good time to add this. Can be a follow up PR, but let's do it as part of the same ticket.
> 
> Image

@charlotteromaniuk @MebinAbraham Let's do this as part of this PR, please; I don't want that PR count going up and breaking my heart.

This PR is fairly simple, so it should be fine to do it here.

sanjeevz3009:
@charlotteromaniuk 

We also have really good UI tests around bundles, so please explore the feature files to see if you can add or update anything. It may be that we don't need anything there, but worth a look.

MebinAbraham:
> > While we are here, I suggest we update the `ReleaseChooserWithDetailsPanel` to include the page workflow status, this will make it consistent with how we show it for bundled pages. This was raised by the business before, so now would be a good time to add this. Can be a follow up PR, but let's do it as part of the same ticket.
> > Image
> 
> @charlotteromaniuk @MebinAbraham Let's do this as part of this PR, please; I don't want that PR count going up and breaking my heart.
> 
> This PR is fairly simple, so it should be fine to do it here.

@sanjeevz3009 I suggest, let's merge this as is, and do follow up, actually. It will be easier to review and easier to distinguish in release notes. Then this PR can merge quick and changes are easily isolated as the changes don't depend on each other.
@charlotteromaniuk tag that pr branch as the same ticket number.

cms/bundles/models.py
charlotteromaniuk: Ah thanks for flagging, I've checked and looks like it was added in the initial implementation but isn't being used - I'll remove it

cms/bundles/tests/test_forms.py
charlotteromaniuk: Updated test name - [7d4e378](https://github.com/ONSdigital/dis-wagtail/pull/764/commits/7d4e378260b3215d339cbd5ad9edd01c43383b3b)

charlotteromaniuk: Added positive test case - [7d4e378](https://github.com/ONSdigital/dis-wagtail/pull/764/commits/7d4e378260b3215d339cbd5ad9edd01c43383b3b)

BJacksonONS:
I think there's still a bug here not caused by these changes, but is now highlighted by them.

The picker for the RC page for a bundle doesn't allow picking a page that was published as part of a bundle, but does allow for picking pages that were published independently.

If an RC page has been published independently of a bundle previously, has changes made in a new draft that then goes through the workflow and gets approved, the changes don't actually get published with the bundle.

When updating the RC page we do the following (comments mine)

```python
def update_bundle_linked_release_calendar_page(bundle: Bundle) -> None:
    """Updates the release calendar page related to the bundle with the pages in the bundle."""
    page = bundle.release_calendar_page # this is the live revision of the page
    if page:
        language_code = get_language_code_from_page(page)
        content = serialize_bundle_content_for_published_release_calendar_page(bundle, language_code)
        datasets = serialize_datasets_for_release_calendar_page(bundle)

        page.content = cast(StreamField, content)
        page.datasets = cast(StreamField, datasets)
        page.status = ReleaseStatus.PUBLISHED
        revision = page.save_revision(log_action=True) # this creates a new revision off the live version, not the workflow approved one
        revision.publish()
        logger.info(
            "Updated release calendar page linked to bundle",
            extra={"bundle_id": bundle.pk, "page_id": page.pk, "event": "release_calendar_page_updated"},
        )
```

Because we create a revision off the live page we lose all of the changes that were in the workflow.

With the changes in this PR we can't add a page that _isn't_ in a workflow, so we can probably just change to something like the following, like we do when publishing pages in a bundle.

```
# finish the workflow
page.current_workflow_state.current_task_state.approve()
```

If we also want to allow a live page without any edited revision to be the bundle's scheduled page then we would want some branching logic like in the bundle publishing.

It's also possible the solution is to not allow RC pages to be standalone published, or to widen the exclusions in the bundle picker and prevent them being added to a bundle.

cms/bundles/models.py
charlotteromaniuk: [3df6bc1](https://github.com/ONSdigital/dis-wagtail/pull/764/commits/3df6bc189d8f10e292412a09d791f6323d7d1290)

BJacksonONS:
> I think there's still a bug here not caused by these changes, but is now highlighted by them.

Looks like there is an existing ticket for this that I didn't spot previously 👍.

cms/bundles/forms.py
```diff
@@ -319,6 +315,15 @@ def _validate_bundled_pages_status(self) -> None:
                 f"page{pluralize(num_pages_not_ready)} not ready to be published."
             )
 
+    def _validate_release_calendar_page_status(self) -> None:
+        release_calendar_page = self.cleaned_data["release_calendar_page"]
+        if not release_calendar_page:
+            return
+
+        if not is_page_ready_to_publish(release_calendar_page):
```
BJacksonONS: Right now this doesn't allow live pages with no draft revisions. @MebinAbraham Worth verifying whether we want to allow those through also.

MebinAbraham: Have checked wth business and yes we want to change this so a bundle can be approved if the RC is approved or if the RC is live and has no unpublished changes (i.e. no dirty revisions). 

@charlotteromaniuk, can you add some extra AC's to the card to cover this please.

Good spot 👍

charlotteromaniuk: Thanks both, I've added the condition to allow live pages without unpublished changes here [1fe4f54](https://github.com/ONSdigital/dis-wagtail/pull/764/commits/1fe4f543cccf37d6cb9f5a0d89201bbe31e164f1) and added ACs to the ticket to cover both cases mentioned above

cms/bundles/tests/test_forms.py
charlotteromaniuk: Since updated again to align with newly added tests [5ac3205](https://github.com/ONSdigital/dis-wagtail/pull/764/commits/5ac32050cb5eed0ba7cc8569747a9e8dbfa84d93)

kevinhowbrook (approved):
Tested locally, all lgtm - thanks

cms/bundles/forms.py
BJacksonONS: Thanks @charlotteromaniuk, looks good

BJacksonONS (approved):

charlotteromaniuk:
> We also have really good UI tests around bundles, so please explore the feature files to see if you can add or update anything. It may be that we don't need anything there, but worth a look.

Thanks @sanjeevz3009, I had a look through the feature files and I don’t think anything needs adding here as I’ve covered the validation logic with unit tests and I’m not adding any new frontend behaviour.

sanjeevz3009:
> > We also have really good UI tests around bundles, so please explore the feature files to see if you can add or update anything. It may be that we don't need anything there, but worth a look.
> 
> Thanks @sanjeevz3009, I had a look through the feature files and I don’t think anything needs adding here as I’ve covered the validation logic with unit tests and I’m not adding any new frontend behaviour. 

@charlotteromaniuk Bundles is very custom app we have + we do override the frontend behaviour a lot. What you see is not out of the box Wagtail. Not on this PR but below...

There should be one UI test failing on your follow up PR that you will need to update. This is around the release calendar UI tests. A scenario should fail. So something to keep in mind.

MebinAbraham (approved):

cms/bundles/forms.py
```diff
@@ -319,6 +315,19 @@ def _validate_bundled_pages_status(self) -> None:
...
+    def _validate_release_calendar_page_status(self) -> None:
+        release_calendar_page = self.cleaned_data["release_calendar_page"]
+        if not release_calendar_page:
+            return
+
+        is_live_with_no_unpublished_changes = (
+            release_calendar_page.live and not release_calendar_page.has_unpublished_changes
+        )
+
+        if not (is_page_ready_to_publish(release_calendar_page) or is_live_with_no_unpublished_changes):
```
MebinAbraham: Suggestion: Check `is_live_with_no_unpublished_changes` first and short-circuit, and you can avoid an extra query. `is_live_with_no_unpublished_changes` doesn't make a DB request, but `is_page_ready_to_publish` does, so we only need to check `is_page_ready_to_publish`  if `is_live_with_no_unpublished_changes` is False, hence avoiding an extra query round trip.
```suggestion
        if not (is_live_with_no_unpublished_changes or is_page_ready_to_publish(release_calendar_page)):
```

charlotteromaniuk: Makes sense, thanks [6c864be](https://github.com/ONSdigital/dis-wagtail/pull/764/commits/6c864bef98ad055b7181688e38739422d5d89d76)

MebinAbraham (approved):
Good work @charlotteromaniuk 🎉

cms/bundles/tests/test_forms.py
sanjeevz3009: Thanks @charlotteromaniuk

sanjeevz3009:
> I think there's still a bug here not caused by these changes, but is now highlighted by them.
> 
> The picker for the RC page for a bundle doesn't allow picking a page that was published as part of a bundle, but does allow for picking pages that were published independently.
> 
> If an RC page has been published independently of a bundle previously, has changes made in a new draft that then goes through the workflow and gets approved, the changes don't actually get published with the bundle.
> 
> When updating the RC page we do the following (comments mine)
> 
> ```python
> def update_bundle_linked_release_calendar_page(bundle: Bundle) -> None:
>     """Updates the release calendar page related to the bundle with the pages in the bundle."""
>     page = bundle.release_calendar_page # this is the live revision of the page
>     if page:
>         language_code = get_language_code_from_page(page)
>         content = serialize_bundle_content_for_published_release_calendar_page(bundle, language_code)
>         datasets = serialize_datasets_for_release_calendar_page(bundle)
> 
>         page.content = cast(StreamField, content)
>         page.datasets = cast(StreamField, datasets)
>         page.status = ReleaseStatus.PUBLISHED
>         revision = page.save_revision(log_action=True) # this creates a new revision off the live version, not the workflow approved one
>         revision.publish()
>         logger.info(
>             "Updated release calendar page linked to bundle",
>             extra={"bundle_id": bundle.pk, "page_id": page.pk, "event": "release_calendar_page_updated"},
>         )
> ```
> 
> Because we create a revision off the live page we lose all of the changes that were in the workflow.
> 
> With the changes in this PR we can't add a page that _isn't_ in a workflow, so we can probably just change to something like the following, like we do when publishing pages in a bundle.
> 
> ```
> # finish the workflow
> page.current_workflow_state.current_task_state.approve()
> ```
> 
> If we also want to allow a live page without any edited revision to be the bundle's scheduled page then we would want some branching logic like in the bundle publishing.
> 
> It's also possible the solution is to not allow RC pages to be standalone published, or to widen the exclusions in the bundle picker and prevent them being added to a bundle.

@BJacksonONS An existing issue. There should be a card for it; if not, you can create one. I forgot where we stood with this one or what we wanted.

I remember coming across this issue a long time ago, I think, and there are many more weird, odd and strange behaviours in bundles/ there is, but we haven't captured everything on a card, etc. We prioritise the features and any bugs the business finds currently.

BJacksonONS:
> I remember coming across this issue a long time ago, I think, and there are many more weird, odd and strange behaviours in bundles/ there is, but we haven't captured everything on a card, etc. We prioritise the features and any bugs the business finds currently.

@sanjeevz3009 

It's in [DPD-2847](https://officefornationalstatistics.atlassian.net/browse/DPD-2847)

[DPD-2847]: https://officefornationalstatistics.atlassian.net/browse/DPD-2847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

sanjeevz3009:
> > I remember coming across this issue a long time ago, I think, and there are many more weird, odd and strange behaviours in bundles/ there is, but we haven't captured everything on a card, etc. We prioritise the features and any bugs the business finds currently.
> 
> @sanjeevz3009
> 
> It's in [DPD-2847](https://officefornationalstatistics.atlassian.net/browse/DPD-2847)

Oh yes I remember this now @BJacksonONS Thanks Ben!

[DPD-2847]: https://officefornationalstatistics.atlassian.net/browse/DPD-2847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

sanjeevz3009 (approved):
LGTM! Thanks @charlotteromaniuk

PR #765, opened 2026-08-11
fix(post-publish-actions): reduce connection pool usage

BJacksonONS:

Background actions can hold onto database connections for their entire lifetime, even if they are not actively using them. This is especially impactful in the support thread pool where we sleep and poll the database for post-publish actions to be complete.

This PR adds a transaction safe util to shut down unused connections, and adds calls in select places where we either sleep, or wait for external api calls (kafka, slack) so we don't hold connections if these services are slow or exhaust our pool polling for actions.

This also allows for more granular control over thread and database connection pool sizing to allow for optimisation.

### How to review

Check that all relevant post publish actions are updated correctly.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


cms/settings/base.py
```diff
@@ -971,6 +973,7 @@ def _iam_database_config(host: str) -> dj_database_url.DBConfig:
 # 110 seconds is comfortably under the 2 minute cron interval
 BUNDLE_POST_PUBLISH_TIMEOUT_SECONDS = float(env.get("BUNDLE_POST_PUBLISH_TIMEOUT_SECONDS", 110))
 BUNDLE_POST_PUBLISH_CONCURRENCY = int(env.get("BUNDLE_POST_PUBLISH_CONCURRENCY", 4))
+BUNDLE_POST_PUBLISH_SUPPORT_CONCURRENCY = int(env.get("BUNDLE_POST_PUBLISH_SUPPORT_CONCURRENCY", 4))
```
MebinAbraham: Since we release the connections before the api calls, I think we could safely set this to 6 (actions) and 3 (support doesn't need much). Thoughts?.

BJacksonONS: 6 action, 2 or 3 for support was roughly what I thought after we talked yesterday, then we might want todo some testing before updating the default pool size but maybe up to 5 for that.

BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/765/commits/6698e774bb150b2cf4f2a91ab4958191bd8d8e24

cms/core/utils.py
```diff
@@ -165,3 +166,13 @@ def __iter__(self) -> Generator[T, None, R]:
     def consume(self) -> None:
         for _ in iter(self):
             pass
+
+
+def release_db_connections() -> None:
```
MebinAbraham: Non-blocker, might be nice to have as a decorator also.

cms/core/utils.py
```diff
@@ -165,3 +166,13 @@ def __iter__(self) -> Generator[T, None, R]:
...
+    """Releases current thread's database connections to the pool.
+
+    Unline Django's `close_old_connections`, this is safe to use inside a transaction block.
+    """
+    for conn in connections.all(initialized_only=True):
+        if not conn.in_atomic_block:
```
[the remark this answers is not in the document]
BJacksonONS: @MebinAbraham let me know what you think on this. `close_if_unusable_or_obsolete` would close a connection if called inside a manual transaction block. Think we only use that in one test so not a likely edge case.

Happy to swap it or just update the doc string.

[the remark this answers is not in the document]
MebinAbraham: Let's swap.

MebinAbraham (approved):
🎉

cms/core/utils.py
BJacksonONS: Dropping before, after or both?

MebinAbraham: Configurable via a flag on the decorator?. Defaulted to our most used need.

cms/core/utils.py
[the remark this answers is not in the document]
BJacksonONS: Think we want both to ensure we keep it transaction safe.

https://github.com/ONSdigital/dis-wagtail/pull/765/commits/6371ce23e0cd3e00c5cdee5c12ba5db9bf76d2d4

cms/core/utils.py
sanjeevz3009: I second this. I have a thing for decorators. They are awesome.

sanjeevz3009 (approved):
LGTM!

cms/core/utils.py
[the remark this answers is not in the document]
MebinAbraham: Doesn't hurt, but do we?  `get_autocommit()` is False when we are inside a transaction? https://github.com/django/django/blob/main/django/db/transaction.py#L195-L235

cms/core/utils.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/765/commits/92797e11a3108545a04ee153296b3823eaeda7b2

cms/core/utils.py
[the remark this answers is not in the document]
BJacksonONS: There might be edge cases where a connection fails mid block. We might also want to add a check that `conn.connection is not None` also possibly?

cms/core/utils.py
MebinAbraham: Given we have it now, how about using it for: `bulk_set_file_permissions`, `wait_for_bundle_notifications`, kafka `self._publish`?

cms/core/utils.py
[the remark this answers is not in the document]
BJacksonONS: Yes that fixes the test errors, it was an unexpected side effect in some of the tests for functions where we were using this.

https://github.com/ONSdigital/dis-wagtail/pull/765/commits/2a4de8e60f8c5b07e766837a66f8d33c66e5f6c1

cms/core/utils.py
BJacksonONS: Added that and flipped the default to before because it makes more sense.

https://github.com/ONSdigital/dis-wagtail/pull/765/commits/3b952a6a449e2f59af3910be6c743f3a7a125e32

cms/core/utils.py
```diff
@@ -165,3 +167,63 @@ def __iter__(self) -> Generator[T, None, R]:
...
+        @release_db_connections  # release after call
+        @release_db_connections(before=True)  # release before and after each decorated function call
+        @release_db_connections(before=True, after=False)  # release before each decorated function call
+    """
+    if func is None and before is None and after is None:
+        _release_db_connections()
+        return None
+
+    release_before = True if before is None else before
+    release_after = False if after is None else after
```
MebinAbraham, line 211: Instead of accepting None, just make type bool and default? 

```py
def release_db_connections(
    func: Callable[..., Any] | None = None, *, before: bool = True, after: bool = False
```

Doc string needs updating.

BJacksonONS: It's `None` there to check if the kwarg has been passed. The following two calls are different.

```python
release_db_connections() # release
release_db_connections(after=True) # return a decorator
```

Have updated the docstring here.
https://github.com/ONSdigital/dis-wagtail/pull/765/commits/881be4cb14198ad559b8410b62d3702a4af111c3

MebinAbraham (dismissed):

sanjeevz3009 (approved):
LGTM!

sanjeevz3009 (dismissed):
Smashing approval to fix GitHub possible stacked PR bug

sanjeevz3009 (approved):

charlotteromaniuk (approved):
Approving to fix GitHub Bug

MebinAbraham (approved):

sanjeevz3009 (approved):

PR #767, opened 2026-08-12
fix: update dataset link destinations

kevinhowbrook:

[CMS-1334](https://officefornationalstatistics.atlassian.net/browse/CMS-1334)

Dataset links now point to a destination chosen by the page doing the linking, rather than always to the dataset series page.

- **Release calendar entries** link to the edition assigned to the release: `/datasets/<dataset-id>/editions/<edition-id>/versions/`. The version is deliberately omitted, so a correction published to that edition is followed automatically without anyone editing the link.
- **Topic and related data pages** are unchanged and continue to link to the series page, `/datasets/<dataset-id>`, so readers reach the most recent data.

### How to review

(This was pretty tricky to work on given the data api)

Seed a `Dataset` row the way `DatasetChosenView` does.

**Destinations**

1. Put a looked-up dataset on a release calendar entry, publish, and check the link under the Data heading is `/datasets/<id>/editions/<edition>/versions/`.
2. Put the same dataset on a topic page and a related data page, and check both are `/datasets/<id>`.

**Validation**, all on a release calendar page:

3. Two versions of the same edition, save, expect an error quoting the shared URL path and explaining that the version does not affect the destination.
4. Two different editions of the same dataset, save, expect it to succeed.
5. A looked-up dataset plus a hand-typed URL resolving to the same place, for example `https://www.ons.gov.uk/datasets/<Id>/editions/<Edition>/versions` against a lookup for the same edition. Expect an error, despite differing host, case and trailing slash.
6. Two manual URLs differing only by a trailing slash, expect an error.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

[CMS-1334]: https://officefornationalstatistics.atlassian.net/browse/CMS-1334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cms/datasets/models.py
+++ b/cms/datasets/models.py
@@ -251,6 +251,29 @@ class Dataset(models.Model):  # type: ignore[django-manager-missing]
         return f"/datasets/{self.namespace}"
 
     @property
+    def latest_version_url_path(self) -> str:
+        """The path to the latest published version of this dataset's edition.
+
+        The path deliberately omits the version. The website resolves it to the latest published
+        version of the edition, so a link built from it follows any correction published for that
+        edition rather than pinning the version that was current when the link was created.
+        """
+        return f"/datasets/{self.namespace}/editions/{self.edition}/versions/"
+
+    def get_url_path(self, *, link_to_latest_version: bool = False) -> str:
+        """Return the public path this dataset should be linked to.
+
+        Which destination is correct depends on the page doing the linking rather than on the
+        dataset. Pages tied to a specific edition, such as release calendar entries, pass True.
+        Pages that are not tied to an edition, such as topic and related data pages, link to the
+        series page so that readers reach the most recent data.
+
+        Both rendering and duplicate validation resolve destinations through here, so that a page
+        cannot validate against one destination and then render another.
+        """
+        return self.latest_version_url_path if link_to_latest_version else self.url_path
+
+    @property
     def compound_id(self) -> str:
         """Return the compound ID for this local Dataset instance.
 
--- a/cms/datasets/models.py
+++ b/cms/datasets/models.py
@@ -256,11 +256,11 @@ class Dataset(models.Model):  # type: ignore[django-manager-missing]
         Which destination is correct depends on the page doing the linking rather than on the
         dataset. Callers read link_to_latest_version off the DatasetStoryBlock the linking page's
         field was built with, so pages tied to a specific edition, such as release calendar
-        entries, get the edition path. That path stops at "versions/", so the destination is not
+        entries, get the edition path. That path stops at "versions", so the destination is not
         pinned to the version recorded here.
         """
         if link_to_latest_version:
-            return f"/datasets/{self.namespace}/editions/{self.edition}/versions/"
+            return f"{self.url_path}/editions/{self.edition}/versions"
         return self.url_path
 
     @property
--- a/cms/datasets/models.py
+++ b/cms/datasets/models.py
@@ -254,9 +254,10 @@ class Dataset(models.Model):  # type: ignore[django-manager-missing]
         """Return the public path this dataset should be linked to.
 
         Which destination is correct depends on the page doing the linking rather than on the
-        dataset: pages tied to a specific edition, such as release calendar entries, pass True.
-        The edition path stops at "versions/", so the destination is not pinned to the version
-        recorded here.
+        dataset. Callers read link_to_latest_version off the DatasetStoryBlock the linking page's
+        field was built with, so pages tied to a specific edition, such as release calendar
+        entries, get the edition path. That path stops at "versions/", so the destination is not
+        pinned to the version recorded here.
         """
         if link_to_latest_version:
             return f"/datasets/{self.namespace}/editions/{self.edition}/versions/"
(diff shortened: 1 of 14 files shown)


cms/datasets/tests/test_blocks.py
```diff
@@ -115,3 +156,161 @@ def test_successful_validation(self):
...
+            self.assertIn('"/datasets/LOOKUP/editions/Lookup_Edition/versions/"', error.message)
+
+    @override_settings(ONS_ALLOWED_LINK_DOMAINS=["example.com"])
+    def test_validation_fails_for_a_manual_link_to_the_resolved_url(self):
+        """Each case is the March lookup's destination, written a different way."""
+        block = DatasetStoryBlock(link_to_latest_version=True)
+        manual_url_cases = [
+            "/datasets/ds/editions/march/versions/",
+            "/datasets/ds/editions/march/versions",
+            "https://example.com/datasets/ds/editions/march/versions/",
```
BJacksonONS, line 264: Should we also cover "/datasets/ds/editions/march/versions/1" as a possible variant?

BJacksonONS (commented):
Think this all looks good to me, except for one test case.

cms/datasets/tests/test_blocks.py
kevinhowbrook: you mean by adding it to `manual_url_cases` or elsewhere?

I think adding it here will fail... those three are all the same destination written differently, and they normalise to `/datasets/ds/editions/march/versions`.

I've added it as a passing case in 66704c30b176a59462feef5b59d15b14be2d8595 instead.

BJacksonONS: Yeah I'm not 100% what the correct behaviour should be if it's manually liked via the long path to the latest dataset rather than `/dataset/<id>` which is very explicit. The ACs don't entirely cover this.

That test case looks good to me as long as there's no requirement that we don't resolve those for the user to the newest one.

cms/datasets/models.py
```diff
@@ -250,6 +250,19 @@ def url_path(self) -> str:
...
+        """Return the public path this dataset should be linked to.
+
+        Which destination is correct depends on the page doing the linking rather than on the
+        dataset. Callers read link_to_latest_version off the DatasetStoryBlock the linking page's
+        field was built with, so pages tied to a specific edition, such as release calendar
+        entries, get the edition path. That path stops at "versions/", so the destination is not
+        pinned to the version recorded here.
+        """
+        if link_to_latest_version:
+            return f"/datasets/{self.namespace}/editions/{self.edition}/versions/"
```
MaciekBaron: Two minor points:

1. Wondering if we should do:
```python
return f"{self.url_path}/editions/{self.edition}/versions/"
```

2. Do we need a trailing slash at the end? We get inconsistent behaviour depending on `link_to_latest_version` - if that is `True`, we get a slash, otherwise we don't. Not sure if this is expected.

MaciekBaron (changes requested):

cms/datasets/models.py
kevinhowbrook: Yea good shout - amended with ff51667039158f9c2f80a9889a9b0e7efbc93d62

MaciekBaron (approved):

cms/datasets/tests/test_blocks.py
MebinAbraham: I'll come back to this, as there are a few more edge cases that needs some thought, and it also looks like this isn't working as expected on the Dataset side. I need to speak to their TL, as I think there is a bug on their side that needs fixing for our logic to work. If they can't fix it, our approach would not work. In short, currently the redirect is taking the user to the latest version of the **latest** edition, not the requested edition.

kevinhowbrook: That makes sense, thanks Mebin

MebinAbraham: They have confirmed the bug and will get it fixed. 

Re: deduplication logic, one option is to enforce a structure for manual entry instead of allowing any url. We now have a strict structure i.e ends `/versions` when `link_to_latest_version` True, else `/datasets/<namespace>`. The other option is to take this as best efforts and deduplicate this that look like a duplicate vs all possible variants. @kevinhowbrook let's have a quick chat about this.

MebinAbraham: Given this should almost always be using the lookup, let's keep manual as an override functionality. So for dedupe logic, if manual url matches look up then error otherwise, keep manual as is. For manual vs manual, treat as dupe if they are the same url, i.e if versions differ, then allow it.

We can revisit this in future, and that likely will involve making the user input restricted to follow our structure so that manually entered url must be `/datasets/<namespace>` not other url parts etc.

Maybe make a note that the destination normalisation only happens for lookups.

This does contridct one of the AC's, I'll update that as a future consideration.

kevinhowbrook: Thanks, I've added a comment in 9e74a6d0dda5ce345444d282287f4f1978a5a78b to explain normalisation only happens for lookups

sanjeevz3009 (approved):
LGTM! Apart from the existing comment or decision @MebinAbraham needs to get back on.

MebinAbraham (approved):

sanjeevz3009 (approved):

PR #768, opened 2026-08-13
chore: remove C90 suppression and set maximum complexity

MaciekBaron:

~This PR removes `C90` from rules, because the only rule that provides is `C901`, which we ignore explicitly.~

This PR removes the suppression of rule `C901`, sets `max-complexity` to 15 and splits `register_bundle_log_actions` to reduce the complexity.

Reference: https://docs.astral.sh/ruff/rules/#mccabe-c90

### How to review

1. Review the reference link above
2. Confirm that "complex functions" are not flagged by ruff

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

N/A

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -123,13 +123,14 @@ ignore = [
     "W191",
     # Prefer explicit type rather than union (|) shorthand
     "UP007",
-    # Allow complex functions
-    "C901",
     # Allow non-top-level imports
     "PLC0415",
     # contextlib.suppress is slower
     "SIM105"
 ]
+
+[tool.ruff.lint.mccabe]
+max-complexity = 15
 
 [tool.ruff.lint.pydocstyle]
 convention = "google"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -104,7 +104,6 @@ select = [
     "C4",     # flake8-comprehensions
     "S",      # flake8-bandit
     "D",      # pydocstyle - Enforce existing docstrings only
-    "C90",    # mccabe
     "RUF",    # Ruff specific rules
 
     # PL - Pylint is only partially supported, we also use the pylint tool to catch all the rules.
@@ -123,8 +122,6 @@ ignore = [
     "W191",
     # Prefer explicit type rather than union (|) shorthand
     "UP007",
-    # Allow complex functions
-    "C901",
     # Allow non-top-level imports
     "PLC0415",
     # contextlib.suppress is slower
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -104,6 +104,7 @@ select = [
     "C4",     # flake8-comprehensions
     "S",      # flake8-bandit
     "D",      # pydocstyle - Enforce existing docstrings only
+    "C90",    # mccabe
     "RUF",    # Ruff specific rules
 
     # PL - Pylint is only partially supported, we also use the pylint tool to catch all the rules.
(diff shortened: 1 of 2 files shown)


BJacksonONS (approved):

AWitcherONS (approved):
Looks good to me lint and format do not cause any further errors 
_(i'm getting warnings for PLW0108 in both main and branch)_

pyproject.toml
```diff
@@ -123,8 +122,6 @@ ignore = [
     "W191",
     # Prefer explicit type rather than union (|) shorthand
     "UP007",
-    # Allow complex functions
-    "C901",
```
MebinAbraham, line 127: I wonder how many fail when this is enabled?. Maybe we need to revisit and tweak max complexity to a reasonable level for this project, this was prob disabled on adoption to not be a blocker. Is there any other rules/linters currently checking complexity?

sanjeevz3009 (approved):
LGTM

pyproject.toml
MaciekBaron: Currently five instances:

```
C901 `_validate_bundled_datasets_metadata` is too complex (13 > 10)
   --> cms/bundles/forms.py:197:9
    |
196 |     @datasets_bundle_api_enabled
197 |     def _validate_bundled_datasets_metadata(self) -> None:
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
198 |         """Compare bundled datasets' stored metadata against the dataset API and
199 |         refresh local rows on drift.
    |

C901 `register_bundle_log_actions` is too complex (16 > 10)
   --> cms/bundles/wagtail_hooks.py:317:5
    |
316 | @hooks.register("register_log_actions")
317 | def register_bundle_log_actions(actions: LogActionRegistry) -> None:
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
318 |     """Registers custom logging actions.
    |

C901 `clean_corrections` is too complex (13 > 10)
  --> cms/core/forms.py:78:9
   |
77 | class PageWithCorrectionsAdminForm(DeduplicateTopicsAdminForm):
78 |     def clean_corrections(self) -> StreamValue:
   |         ^^^^^^^^^^^^^^^^^
79 |         corrections: StreamValue = self.cleaned_data["corrections"]
   |

C901 `_process_blocks` is too complex (11 > 10)
  --> cms/core/migrations/0011_migrate_imageblock_caption.py:86:5
   |
86 | def _process_blocks(blocks, direction):
   |     ^^^^^^^^^^^^^^^
87 |     """Recursively process StreamField blocks to migrate ImageBlock data."""
88 |     modified = False
   |

C901 `fetch_api_response` is too complex (13 > 10)
  --> cms/datasets/models.py:52:9
   |
50 |         return results
51 |
52 |     def fetch_api_response(self, url: str | None = None, params: Mapping | None = None) -> dict:  # noqa: PLR0912
   |         ^^^^^^^^^^^^^^^^^^
53 |         # Import here to avoid circular import
54 |         from cms.bundles.notifications.api_failures import (  # pylint: disable=import-outside-toplevel
   |
   ```
   
How about we set it to 15 and address `register_bundle_log_actions`?

MebinAbraham: Let's do that 👍 We don't want unbounded complexity.

MaciekBaron: Updated here [5737622](https://github.com/ONSdigital/dis-wagtail/pull/768/commits/5737622fcbed472f90c63e9f1b0cd321f4d6a391)

BJacksonONS (approved):

kevinhowbrook (approved):

MebinAbraham (approved):

PR #769, opened 2026-08-14
test(functional-tests): functional test reliability improvements

BJacksonONS:

This PR tries to avoid a few common functional test failure scenarios that have been seen commonly in CI.

1. [Recently some have failed due to attempting to connect to the database before it's ready](https://github.com/ONSdigital/dis-wagtail/actions/runs/31609727216/job/94158504392). This PR adds healthchecks to the docker compose file and waits for the supporting services (postgres, kafka) to be healthy before starting test runs
2. [Occasionally after populating a datetime field the popup lingers and blocks clicks to other elements](https://github.com/ONSdigital/dis-wagtail/actions/runs/30983727280/job/92233970400). This adds a utility to ensure it's closed before continuing with the test.

### How to review

- Check healthchecks work
- Check all datetime fills have been updated

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/functional_tests/docker-compose-dev.yml
+++ b/functional_tests/docker-compose-dev.yml
@@ -28,11 +28,6 @@ services:
       service: db
     ports: !override
       - '15432:5432'
-    healthcheck:
-      test: ['CMD-SHELL', 'pg_isready -q']
-      interval: 2s
-      timeout: 5s
-      retries: 30
 
   redis:
     extends:
@@ -40,11 +35,6 @@ services:
       service: redis
     ports: !override
       - '16379:6379'
-    healthcheck:
-      test: ['CMD-SHELL', 'redis-cli ping']
-      interval: 2s
-      timeout: 5s
-      retries: 30
 
   broker:
     extends:
@@ -54,15 +44,6 @@ services:
       KAFKA_CFG_ADVERTISED_LISTENERS: 'PLAINTEXT://broker:9092,EXTERNAL://localhost:19094'
     ports: !override
       - '19094:9094'
-    healthcheck:
-      test:
-        [
-          'CMD-SHELL',
-          '/opt/bitnami/kafka/bin/kafka-broker-api-versions.sh --bootstrap-server localhost:9094',
-        ]
-      interval: 5s
-      timeout: 20s
-      retries: 20
 
 volumes:
   node_modules:
--- a/functional_tests/docker-compose-dev.yml
+++ b/functional_tests/docker-compose-dev.yml
@@ -28,6 +28,11 @@ services:
       service: db
     ports: !override
       - '15432:5432'
+    healthcheck:
+      test: ['CMD-SHELL', 'pg_isready -q']
+      interval: 2s
+      timeout: 5s
+      retries: 30
 
   redis:
     extends:
@@ -35,6 +40,11 @@ services:
       service: redis
     ports: !override
       - '16379:6379'
+    healthcheck:
+      test: ['CMD-SHELL', 'redis-cli ping']
+      interval: 2s
+      timeout: 5s
+      retries: 30
 
   broker:
     extends:
@@ -44,6 +54,15 @@ services:
       KAFKA_CFG_ADVERTISED_LISTENERS: 'PLAINTEXT://broker:9092,EXTERNAL://localhost:19094'
     ports: !override
       - '19094:9094'
+    healthcheck:
+      test:
+        [
+          'CMD-SHELL',
+          '/opt/bitnami/kafka/bin/kafka-broker-api-versions.sh --bootstrap-server localhost:9094',
+        ]
+      interval: 5s
+      timeout: 20s
+      retries: 20
 
 volumes:
   node_modules:
--- a/functional_tests/docker-compose.yml
+++ b/functional_tests/docker-compose.yml
@@ -5,6 +5,11 @@ services:
       service: db
     ports: !override
       - '15432:5432'
+    healthcheck:
+      test: ['CMD-SHELL', 'pg_isready -q']
+      interval: 2s
+      timeout: 5s
+      retries: 30
 
   redis:
     extends:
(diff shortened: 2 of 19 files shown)


functional_tests/step_helpers/utils.py
kevinhowbrook, line 1: I think there are a few missed ones that use datetime in `step_helpers/release_page_helpers` e.g in `handle_changes_to_release_date_feature` ? https://github.com/ONSdigital/dis-wagtail/blob/d5c53e1b7ca63b7e9a22f8bcaf4daded6464753d/functional_tests/step_helpers/release_page_helpers.py#L243

functional_tests/docker-compose-dev.yml
kevinhowbrook, line 1: It looks like the healthchecks aren't in the stack CI runs?

I've been trying to wrap my head around the docker inheritance here... I could be wrong but `functional_tests/docker-compose.yml` inherits `../docker-compose-dev.yml`, and that ../ points up to the root one, not the one next to it.

BJacksonONS: Ah yeah good point, I use the dev one for local but missed that.

https://github.com/ONSdigital/dis-wagtail/pull/769/commits/a07b8c247966f0ff9e2c88da3b3b211e29d1940a

kevinhowbrook: nice one, thanks

functional_tests/step_helpers/utils.py
BJacksonONS: Updated that file here

https://github.com/ONSdigital/dis-wagtail/pull/769/commits/1cbbda7f323da9bebe681527c4d7e055c1a79a2f

kevinhowbrook (approved):

functional_tests/docker-compose-dev.yml
```diff
@@ -44,6 +54,15 @@ services:
       KAFKA_CFG_ADVERTISED_LISTENERS: 'PLAINTEXT://broker:9092,EXTERNAL://localhost:19094'
     ports: !override
       - '19094:9094'
+    healthcheck:
```
sanjeevz3009: I initially wanted to add health checks for kafka, forgot to do it. Nice!

functional_tests/step_helpers/utils.py
```diff
@@ -16,6 +18,24 @@
...
+def fill_datetime_field(field: Locator, value: str) -> None:
+    """Fill a datetime field, and wait for the datetime picker to close."""
+    field.fill(value)
+
+    picker = field.page.locator(DATETIME_PICKER_SELECTOR)
+
+    try:
+        picker.first.wait_for(timeout=DATETIME_PICKER_OPEN_TIMEOUT)
+    except PlaywrightTimeoutError:
+        return
```
MebinAbraham, line 35: `field.blur()` before we return as a slow picker might still appear after it returns?.

functional_tests/step_helpers/utils.py
```diff
@@ -16,6 +18,24 @@
...
+
+
+def fill_datetime_field(field: Locator, value: str) -> None:
+    """Fill a datetime field, and wait for the datetime picker to close."""
+    field.fill(value)
+
+    picker = field.page.locator(DATETIME_PICKER_SELECTOR)
+
+    try:
+        picker.first.wait_for(timeout=DATETIME_PICKER_OPEN_TIMEOUT)
```
MebinAbraham, line 32: This will force a 3-second wait every time, right?. Does it need to be that high?.

functional_tests/docker-compose.yml
```diff
@@ -5,13 +5,23 @@ services:
       service: db
     ports: !override
       - '15432:5432'
+    healthcheck:
+      test: ['CMD-SHELL', 'pg_isready -q']
+      interval: 2s
+      timeout: 5s
+      retries: 30
```
MebinAbraham: Q: Why do we need these here if they’ve already been added to `docker-compose-dev.yml`? Wouldn’t they be inherited from there, or am I missing something?

functional_tests/step_helpers/utils.py
BJacksonONS: It will wait up to 3 seconds for the first element matching the picker locator to appear. Most likely it will be a few ms. I think the default is 30s so this will fail slightly faster if the picker never appears.

functional_tests/docker-compose.yml
BJacksonONS: This file inherits from the root not the functional test compose file which I'd also edited. Moved all the healthchecks to the root one here so everything can inherit from it.

[12cb910](https://github.com/ONSdigital/dis-wagtail/pull/769/commits/12cb910562a4ebcfc84907df3aef3aab3565b423)

functional_tests/step_helpers/utils.py
BJacksonONS: https://github.com/ONSdigital/dis-wagtail/pull/769/commits/0fb45d40b0e1ed7cf2d9af614bc0f054e3ce19fc

functional_tests/docker-compose.yml
MebinAbraham: Right didn't clock it was the func test one but yes this is what I was getting it. Good stuff 👍

MebinAbraham (approved):

functional_tests/docker-compose.yml
BJacksonONS: One point worth highlighting here is that right now I've only added the `--wait` to the functional test up command. Can add that elsewhere but didn't want to go too far out of scope.

functional_tests/docker-compose-dev.yml
MebinAbraham: Shoddy job @sanjeevz3009. Always getting someone else to clean up your mess. 😄 jk ofc

sanjeevz3009 (approved):
LGTM!

functional_tests/docker-compose-dev.yml
sanjeevz3009: Thanks, @MebinAbraham, for your kind words =)

PR #775, opened 2026-08-18
fix: add workflow/page status to release calendar chooser

charlotteromaniuk:

Adds the Release Calendar workflow/page status to the Release Calendar details on the chooser panel. This should match the way that the workflow/page status is displayed for the bundled pages chooser panel.
image

### How to review

1. Create/edit a bundle in the admin
2. Using the release calendar chooser panel, select a release calendar
3. Ensure that the status of the release calendar page is displayed correctly on the release calendar chooser panel
4. Update the status of the release calendar and ensure the status displayed on the chooser panel in the bundle is updated

### Deployment Safety

Please select one:

- [X] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions
N/A

---
Ticket: [DPD-5228](https://officefornationalstatistics.atlassian.net/browse/DPD-5228) (relates to the last AC)

[DPD-5228]: https://officefornationalstatistics.atlassian.net/browse/DPD-5228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cms/core/tests/test_formatting_utils.py
+++ b/cms/core/tests/test_formatting_utils.py
@@ -1,6 +1,8 @@
+from dataclasses import dataclass
 from datetime import date, datetime, timedelta
 from datetime import timezone as tz
 from types import SimpleNamespace
+from typing import Any
 
 from django.test import SimpleTestCase, TestCase
 from django.utils import timezone
@@ -45,11 +47,11 @@ class DummyPageWithNoReleaseDate(DummyPage):
         abstract = True
 
 
+@dataclass
 class DummyStatusPage:
-    def __init__(self, *, current_workflow_state=None, live=False, has_unpublished_changes=True):
-        self.current_workflow_state = current_workflow_state
-        self.live = live
-        self.has_unpublished_changes = has_unpublished_changes
+    current_workflow_state: Any | None = None
+    live: bool = False
+    has_unpublished_changes: bool = True
 
 
 class GetFormattedPagesListTests(TestCase):
--- a/cms/core/tests/test_formatting_utils.py
+++ b/cms/core/tests/test_formatting_utils.py
@@ -1,5 +1,6 @@
 from datetime import date, datetime, timedelta
 from datetime import timezone as tz
+from types import SimpleNamespace
 
 from django.test import SimpleTestCase, TestCase
 from django.utils import timezone
@@ -10,6 +11,7 @@ from cms.core.formatting_utils import (
     format_as_document_list_item,
     get_document_metadata,
     get_formatted_pages_list,
+    get_page_display_status,
     to_rfc3339_datetime,
 )
 from cms.core.models.base import BasePage
@@ -41,6 +43,13 @@ class DummyPageWithNoReleaseDate(DummyPage):
 
     class Meta:
         abstract = True
+
+
+class DummyStatusPage:
+    def __init__(self, *, current_workflow_state=None, live=False, has_unpublished_changes=True):
+        self.current_workflow_state = current_workflow_state
+        self.live = live
+        self.has_unpublished_changes = has_unpublished_changes
 
 
 class GetFormattedPagesListTests(TestCase):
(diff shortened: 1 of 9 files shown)


MebinAbraham:
I wouldn't class this as a bug tbh. A "small delighter" 🙌

cms/release_calendar/utils.py
```diff
@@ -96,12 +96,28 @@ def get_translated_string(string_to_translate: str, language_code: str) -> str:
...
+    """Returns the page status shown for a release calendar page in the chooser panel.
+
+    When the page is in workflow, this returns the current workflow task name.
+    Otherwise it returns the page status based on whether the page is live without unpublished changes or still a draft.
+    """
+    if workflow_state := release_calendar_page.current_workflow_state:
+        return workflow_state.current_task_state.task.name  # type: ignore[no-any-return]
+    if release_calendar_page.live and not release_calendar_page.has_unpublished_changes:
+        return "Live"
+    return "Draft"
```
MebinAbraham: I have asked the business what their preference is; this may be acceptable, but they might prefer "Live + Draft" if page is live but has changes while not in a workflow. Will let you know once I hear back.

MebinAbraham (commented):
Looks good, nice work. Just waiting on the 1 query before approving.

cms/release_calendar/utils.py
MebinAbraham: Decision: to be consident with other pages, let's follow Wagtail behaviour.

```py
if workflow:
	return "<Workflow Status>" # i.e. In Preview

if live:
	return "Live + Draft" if has_unpublished_changes else "Live"

return "Draft"
```

charlotteromaniuk: Thanks @MebinAbraham, just to note this doesn't match how the status is displayed when bundled pages are selected (a page that is "Live + Draft" is displayed as "Draft"). Is that something business are okay with or should this also be updated in future?

MebinAbraham: Let's make it consistent, and sort here. I suggest you have a single method for this, so collapse your new method and `get_page_title_with_workflow_status` into a single function. The other data for the RC page (release date etc) can be an optional parameter that is appended if exist, then we have one way of doing it.

cms/core/tests/test_formatting_utils.py
```diff
@@ -43,6 +45,13 @@ class Meta:
         abstract = True
 
 
+class DummyStatusPage:
```
MaciekBaron, line 51: **Minor**: in modern python we can do this instead:

```python
@dataclass
class DummyStatusPage:
    current_workflow_state: Any | None = None
    live: bool = False
    has_unpublished_changes: bool = True
```

https://docs.python.org/3/library/dataclasses.html

MaciekBaron (commented):
Great work, just left one comment

MebinAbraham (approved):
🎉

cms/core/tests/test_formatting_utils.py
charlotteromaniuk: Thanks! [2bd2547](https://github.com/ONSdigital/dis-wagtail/pull/775/commits/2bd254794db59bd571ab1f9120175145eb9decf3)

cms/release_calendar/utils.py
charlotteromaniuk: [018dfcb](https://github.com/ONSdigital/dis-wagtail/pull/775/commits/018dfcbbd6c52efb61f92927448311c12be7d648)

MaciekBaron (approved):
LGTM (once the checks go green)

sanjeevz3009 (commented):
LGTM! Reviewed. Thanks @charlotteromaniuk

sanjeevz3009 (commented):
Thank you for the rapid follow-up @charlotteromaniuk

PR #777, opened 2026-08-19
chore(functional-tests): increase the number of test shards from 6 to 8

sanjeevz3009:

### How to review

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


joepercival (approved):
Looks good

MebinAbraham (commented):
As discussed this doesn't provide enough value given a single shard is still longer than rest. We also have to account for total jobs across the org and concurrency issues.

If runtime is < slowest other test (unit test), keep it unchanged, if runtime is more and it is because of test length, then either split the test or increase to max 8.

Don't forget this is also only using a single core, what we should be doing is parallelising the actual test to run on all core (ticket in backlog). But we have to be mindful to increasing shards.

sanjeevz3009:
> As discussed this doesn't provide enough value given a single shard is still longer than rest. We also have to account for total jobs across the org and concurrency issues.
> 
> If runtime is < slowest other test (unit test), keep it unchanged, if runtime is more and it is because of test length, then either split the test or increase to max 8.
> 
> Don't forget this is also only using a single core, what we should be doing is parallelising the actual test to run on all core (ticket in backlog). But we have to be mindful to increasing shards.

Thanks @MebinAbraham Yes, will go with 8 for now, which with some of Ben's changes should bring functional test in parity with the unit tests we have.

BJacksonONS (approved):

BJacksonONS (approved):

PR #780, opened 2026-08-19
feat: add cache overpurge for paginated URLs

MaciekBaron:

This PR build on the work done in #505 and adds cache over-purge functionality for paginated URLs.

The reason for the over purge is to handle trailing pages so pages orphaned by shrinking pagination are invalidated too.

### How to review

Ensure that the code changes make sense.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

Merge #505 into main once everything is done.

---

### DIFF
--- a/cms/articles/models.py
+++ b/cms/articles/models.py
@@ -928,10 +928,9 @@ class StatisticalArticlePage(  # type: ignore[django-manager-missing]
     def get_cached_paths(self) -> Generator[str]:
         yield "/"
         yield "/related-data"  # always include, should a correction remove related datasets
-        if self.datasets:
-            pages = ceil(len(self.datasets) / settings.RELATED_DATASETS_PER_PAGE)
-            # over-purge trailing pages so pages orphaned by shrinking pagination are invalidated too
-            for page_number in range(1, pages + settings.CMS_PAGINATION_OVER_PURGE + 1):
-                yield f"/related-data?page={page_number}"
+        pages = ceil(len(self.datasets or []) / settings.RELATED_DATASETS_PER_PAGE)
+        # over-purge trailing pages so pages orphaned by shrinking pagination are invalidated too
+        for page_number in range(1, pages + settings.CMS_PAGINATION_OVER_PURGE + 1):
+            yield f"/related-data?page={page_number}"
 
         yield from self.get_downloadable_block_paths()
--- a/cms/articles/models.py
+++ b/cms/articles/models.py
@@ -257,7 +257,8 @@ class ArticleSeriesPage(  # type: ignore[django-manager-missing]
         yield "/editions"
         # ensure we always account for ?page=1
         pages = max(1, ceil(self.get_children().live().public().count() / settings.PREVIOUS_RELEASES_PER_PAGE))
-        for page_number in range(1, pages + 1):
+        # over-purge trailing pages so pages orphaned by shrinking pagination are invalidated too
+        for page_number in range(1, pages + settings.CMS_PAGINATION_OVER_PURGE + 1):
             yield f"/editions?page={page_number}"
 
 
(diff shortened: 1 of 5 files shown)


cms/articles/models.py
```diff
@@ -928,7 +929,9 @@ def get_cached_paths(self) -> Generator[str]:
         yield "/"
         yield "/related-data"  # always include, should a correction remove related datasets
         if self.datasets:
```
MebinAbraham: We would also need to purge the URL even if there are no datasets as the latest version may have removed them, so all previously published URL needs purging.

Ideally we would prefix purge, this is best efforts and we should document it that way vs it being the preferred choice. Think over page is fine for shrinking but on removal it is hard to know how many previous pages may have been removed.

MebinAbraham: The other over purge approach is to set over purge based on max datasets per article NFR. I.e if the max is 100 and it is 10 per page, our rule would be is we always purge 10 pages worth.

I think once Wagtail support prefix purge, we would change this approach. I think we should document that as a TODO:

sanjeevz3009 (approved):
Apart from @MebinAbraham comments, LGTM!

cms/articles/models.py
MaciekBaron: Updated the logic so it always over-purges pages even when none left: [bbbd70a](https://github.com/ONSdigital/dis-wagtail/pull/780/commits/bbbd70a957ee295d99a13d9f7b9abafa8ef57cd7)

kevinhowbrook (approved):

BJacksonONS (changes requested):
My only issue here is that a lot of the tests implicitly rely on the default `CMS_PAGINATION_OVER_PURGE` env value being 2. If we ever change it then we get ~22 failures.

It'd be good to either explicitly `@override_settings` in each test, or in `cms.settings.test` add an override so modifying the base settings won't break them.

BJacksonONS (approved):

MaciekBaron:
@BJacksonONS good point, addressed your comment here: [9078026](https://github.com/ONSdigital/dis-wagtail/pull/780/commits/9078026cb4035ad6ed9f3ab3ee02a421c3b2ed68)

joepercival (approved):

cms/settings/base.py
```diff
@@ -892,6 +892,10 @@
 CMS_RELEASES_INDEX_REDIRECT_ENABLED = env.get("CMS_RELEASES_INDEX_REDIRECT_ENABLED", "true").lower() == "true"
 RELATED_DATASETS_PER_PAGE = int(env.get("RELATED_DATASETS_PER_PAGE", DEFAULT_PER_PAGE))
 
+# Number of extra trailing pages to purge from the front-end cache beyond the current page count,
+# so that pages orphaned by shrinking pagination (e.g. after a correction removes items) get invalidated too.
+CMS_PAGINATION_OVER_PURGE = int(env.get("CMS_PAGINATION_OVER_PURGE", 2))
```
MebinAbraham, line 897: I suggest we add a comment like this.
```suggestion
# Number of extra trailing pages to purge from the front-end cache beyond the current page count,
# so that pages orphaned by shrinking pagination (e.g. after a correction removes items) get invalidated too.
# This is a workaround until Wagtail supports prefix-based cache purging:
# https://github.com/wagtail/wagtail/pull/13773
CMS_PAGINATION_OVER_PURGE = int(env.get("CMS_PAGINATION_OVER_PURGE", 2))
```

MebinAbraham (approved):

MebinAbraham (commented):
Now that we updated the base branch the `ci.yml` needs updating.

MaciekBaron:
@MebinAbraham `ci.yml` updated here: #794

BJacksonONS (approved):

PR #789, opened 2026-08-24
chore: upgrade the DS to 74.0.0-next.8 and make the back to top link code more DRY

helenb:

Follows on from https://github.com/ONSdigital/dis-wagtail/pull/702

- Updates the DS to 74.0.0-next.8
- The only change in 74.0.0-next.8 is to add a postMainContent block
- This allows us to add the back to top link in the base template as it was previously
- Remove the back-top-top include and the call to the include on child templates

### How to review

Check that the back to top link appears at mobile on all page templates

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

---
Ticket: [CMS-1338](https://officefornationalstatistics.atlassian.net/browse/CMS-1338)

[CMS-1338]: https://officefornationalstatistics.atlassian.net/browse/CMS-1338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cms/jinja2/templates/base_page.html
```diff
@@ -110,6 +111,15 @@ <h1 class="ons-u-fs-3xl ons-u-mb-xl common-header__heading">
     {% endif %}
 {% endblock %}
 
+{% block postMainContent %}
+    {# used in small viewports #}
+    {# fmt:off #}
+    {{ onsBackToTop({
+        "description": back_to_top,
+    }) }}
```
MaciekBaron: Super minor but I think we tend to format this like this:
```suggestion
    {{
        onsBackToTop({
            "description": back_to_top,
        })
    }}
```

MaciekBaron (approved):
Just one minor comment

benedictjackson (dismissed):

BJacksonONS (approved):

charlotteromaniuk:
@helenb I think this is an existing issue introduced in https://github.com/ONSdigital/dis-wagtail/pull/702 rather than this PR, but the Back to Top JS seems to be throwing an error on some pages. I’ve reproduced it on Index pages and the Release Calendar Index page when viewing on small viewports. Looking into it a bit, it seems like the Back to Top JS expects a previous sibling which is missing on these pages. 
image

Error in the console: Uncaught TypeError: Cannot read properties of null (reading 'getBoundingClientRect')

helenb:
> @helenb I think this is an existing issue introduced in #702 rather than this PR, but the Back to Top JS seems to be throwing an error on some pages. I’ve reproduced it on Index pages and the Release Calendar Index page when viewing on small viewports. Looking into it a bit, it seems like the Back to Top JS expects a previous sibling which is missing on these pages. 
> 
> Error in the console: Uncaught TypeError: Cannot read properties of null (reading 'getBoundingClientRect')

Thanks @charlotteromaniuk I've been able to repeat this on the release calendar index page (in preview) and on an index page with no children.

As you say, the back-to-top code is expecting it to follow on from a content sibling, but for the release calendar index page this is empty.

I think this could do with a fix in the design system to be more robust in the way that it calculates the position of the end of the previous content. But in the mean time we could get around it by outputting an empty div before the back to top code - not very clean but I haven't thought of a better way as an interim measure.

FWIW from a quick look at the code history I think this issue would have existed prior to #702.

cms/jinja2/templates/base_page.html
helenb: This suggestion was applied but I rebased as I forgot to pull before making a further change, so it isn't showing here.

sanjeevz3009:
> > @helenb I think this is an existing issue introduced in #702 rather than this PR, but the Back to Top JS seems to be throwing an error on some pages. I’ve reproduced it on Index pages and the Release Calendar Index page when viewing on small viewports. Looking into it a bit, it seems like the Back to Top JS expects a previous sibling which is missing on these pages.
> > Error in the console: Uncaught TypeError: Cannot read properties of null (reading 'getBoundingClientRect')
> 
> Thanks @charlotteromaniuk I've been able to repeat this on the release calendar index page (in preview) and on an index page with no children.
> 
> As you say, the back-to-top code is expecting it to follow on from a content sibling, but for the release calendar index page this is empty.
> 
> I think this could do with a fix in the design system to be more robust in the way that it calculates the position of the end of the previous content. But in the mean time we could get around it by outputting an empty div before the back to top code - not very clean but I haven't thought of a better way as an interim measure.
> 
> FWIW from a quick look at the code history I think this issue would have existed prior to #702.

@charlotteromaniuk  @helenb Correct, I saw this issue way before #702, for some reason, I thought it was already fixed.

sanjeevz3009 (approved):
LGTM apart from the existing comments!

helenb:
> > @helenb I think this is an existing issue introduced in #702 rather than this PR, but the Back to Top JS seems to be throwing an error on some pages. I’ve reproduced it on Index pages and the Release Calendar Index page when viewing on small viewports. Looking into it a bit, it seems like the Back to Top JS expects a previous sibling which is missing on these pages.
> > Error in the console: Uncaught TypeError: Cannot read properties of null (reading 'getBoundingClientRect')
> 
> Thanks @charlotteromaniuk I've been able to repeat this on the release calendar index page (in preview) and on an index page with no children.
> 
> As you say, the back-to-top code is expecting it to follow on from a content sibling, but for the release calendar index page this is empty.
> 
> I think this could do with a fix in the design system to be more robust in the way that it calculates the position of the end of the previous content. But in the mean time we could get around it by outputting an empty div before the back to top code - not very clean but I haven't thought of a better way as an interim measure.
> 
> FWIW from a quick look at the code history I think this issue would have existed prior to #702.

Empty div added in e5f7d25ff1f9f92023e8876170ae6b60cc190531 and I'll raise a ticket in the DS to make the selector for the previous content less likely to fail

sanjeevz3009 (approved):

charlotteromaniuk:
LGTM, thanks Helen

PR #793, opened 2026-08-25
chore(taxonomy): add make target to delete topics

MebinAbraham:

Adds a `delete-topics` Make target for removing all taxonomy topics from the database.

### How to review

Does it make sense, is it useful?. Try run `make delete-topics` then try load topics/sync topics to ensure no regression.

Will be used more often when being able to sync topics from diff envs.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,6 +115,10 @@ load-design-system-templates:  ## Load the design system templates
 load-topics:  ## Load our fixture of taxonomy topics
 	poetry run python ./manage.py loaddata cms/taxonomy/fixtures/topics.json
 
+.PHONY: delete-topics
+delete-topics: ## Delete all topics from the database (root is preserved via the model manager)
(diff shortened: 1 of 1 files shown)


BJacksonONS:
Not sure I've come across a situation where this would have helped yet, but I don't have anything against leaving handy one-liners in the Makefile if they're going to be useful once we have a local dataset API.

BJacksonONS (approved):

MebinAbraham:
> Not sure I've come across a situation where this would have helped yet, but I don't have anything against leaving handy one-liners in the Makefile if they're going to be useful once we have a local dataset API.

Same, only recently with some of the testing with topics and pointing at diff envs. It should be more used with local dataset stack. Not a big deal, can have an alias but think it will become useful.

MaciekBaron (approved):

sanjeevz3009 (approved):
LGTM!

PR #799, opened 2026-08-28
chore(headers): set appropriate origin cache headers

MaciekBaron:

This PR implements the caching rules defined in CMS-1388 (which combines the requirements from CMS-1315, CMS-1363 and CMS-1374 respectively).

### How to review

It's important that you consult the tickets to ensure that the correct headers are being set, based on the requirements.

Make requests to the server and check the headers set. For example, you can use `curl`:

```
❯ curl -I localhost:8000
HTTP/1.1 200 OK
Date: Fri, 28 Aug 2026 15:27:18 GMT
Server: WSGIServer/0.2 CPython/3.14.7
Content-Type: text/html; charset=utf-8
Cache-Control: public, max-age=60, stale-while-revalidate=0, stale-if-error=300
Cloudflare-CDN-Cache-Control: max-age=31536000, stale-while-revalidate=86400, stale-if-error=432000
Content-Security-Policy: default-src 'self'; frame-src 'self' ons.gov.uk *.ons.gov.uk; style-src 'self' https://cdn.ons.gov.uk cdnjs.cloudflare.com 'unsafe-inline' *.hotjar.com fonts.googleapis.com; img-src 'self' https://cdn.ons.gov.uk www.googletagmanager.com *.hotjar.com; script-src 'self' https://cdn.ons.gov.uk cdnjs.cloudflare.com *.hotjar.com www.googletagmanager.com 'unsafe-inline'; font-src 'self' https://cdn.ons.gov.uk cdnjs.cloudflare.com *.hotjar.com fonts.gstatic.com; connect-src 'self' www.googletagmanager.com www.google.com *.hotjar.com *.hotjar.io wss://*.hotjar.com ws://localhost:3000; manifest-src 'self' https://cdn.ons.gov.uk; frame-ancestors 'self'
Content-Length: 57065
Vary: Cookie, Accept-Language
Content-Language: en-gb
Permissions-Policy: accelerometer=(), autoplay=(), camera=(), display-capture=(), encrypted-media=(), fullscreen=(), geolocation=(), gyroscope=(), interest-cohort=(), microphone=(), midi=(), payment=(), usb=()
X-Content-Type-Options: nosniff
Referrer-Policy: no-referrer-when-downgrade
Cross-Origin-Opener-Policy: same-origin
```

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

N/A

---
Ticket: [CMS-1388](https://officefornationalstatistics.atlassian.net/browse/CMS-1388)

[CMS-1388]: https://officefornationalstatistics.atlassian.net/browse/CMS-1388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cms/core/models/base.py
```diff
@@ -428,6 +430,10 @@ def serve_preview(self, request: HttpRequest, mode_name: str) -> TemplateRespons
         response: TemplateResponse = super().serve_preview(request, mode_name)
         return response
 
+    def serve(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
```
BJacksonONS: Just started reviewing and things mostly look good, but looking at the override of serve on the base page reminded me of this pr https://github.com/ONSdigital/dis-wagtail/pull/446 (comment here if linking to resolved ones works properly https://github.com/ONSdigital/dis-wagtail/pull/446#discussion_r3172833183).

The pages that inherit from `RoutablePageMixin` will call that serve method and this won't get called. We either need to move some of this logic into the mixin or override the specific models.

Is there value in a test somewhere to assert the correct value for each page type?

MaciekBaron: Good point, that's a foot gun for sure.

Re-implemented this using hooks [7f1080e](https://github.com/ONSdigital/dis-wagtail/pull/799/commits/7f1080eb2beb7f2bfb11d3cc138414f37dc7b8fa)

Seems to be the recommended approach anyway:

https://docs.wagtail.org/en/latest/reference/hooks.html#on-serve-page

BJacksonONS: Interesting, if we can detect we're serving a preview from the hook maybe we can make a dev ticket to refactor the logging from that PR I linked to this hook also?

This is a fairly big change so I'll resolve this comment, re-review and raise anything else separately.

BJacksonONS (commented):
One case I've found here that I think need addressing, and a question

1. The localised homepages are added after the decorator in `urls.py`, so the welsh homepage under path based localisation (e.g. http://localhost:8000/cy ) doesn't get the headers.

2. I need to review the cache purging code again, but that does cover download links right? Just making sure that this is the appropriate behaviour for non-page urls.

Image

cms/core/wagtail_hooks.py
```diff
@@ -14,18 +15,30 @@
...
-    from collections.abc import Sequence
+    from collections.abc import Callable, Sequence
 
     from django.http import HttpRequest, HttpResponsePermanentRedirect, HttpResponseRedirect
     from wagtail.admin.views.bulk_action import BulkAction, ModelLogEntry
     from wagtail.log_actions import LogActionRegistry
 
 
+@hooks.register("on_serve_page")
+def apply_cache_headers_on_serve(next_serve_page: Callable) -> Callable:
```
kevinhowbrook, line 32: Do we need to check view restrictions at all? e.g https://github.com/wagtail/wagtail/blob/bddc5eaf2e77d6777b11d0e8cd4a3b62cf707417/wagtail/wagtail_hooks.py#L36

This might be a moot point, but if someone adds a password protected page (core feature in wagtail) how will that be handled?

cms/core/wagtail_hooks.py
```diff
@@ -14,18 +15,30 @@
...
+    """Applies Cache-Control headers to page responses."""
+
+    def wrapper(page: Page, request: HttpRequest, args: tuple, kwargs: dict) -> HttpResponse:
+        response: HttpResponse = next_serve_page(page, request, args, kwargs)
+        return apply_page_cache_headers(page, response)
```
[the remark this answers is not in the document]
MaciekBaron: Addressed in [15b973d](https://github.com/ONSdigital/dis-wagtail/pull/799/commits/15b973d92786fff996b7f156bc3d41ccb5754c7a)

MaciekBaron:
@BJacksonONS Welsh homepage issue addressed here: [15b973d](https://github.com/ONSdigital/dis-wagtail/pull/799/commits/15b973d92786fff996b7f156bc3d41ccb5754c7a)

cms/core/wagtail_hooks.py
MaciekBaron: That's a good question - I don't think there will ever be password protected pages. If there will be, we should explicitly define those rules in a ticket.

cms/core/cache.py
```diff
@@ -56,26 +58,65 @@ def purge_cache_on_all_sites(path: str) -> None:
...
+    routed page, since they can differ (e.g. an article series delegating to an edition).
+    """
+    rendered_page = getattr(response, "context_data", {}).get("self", page)
+    cache_control_kwargs = (
+        get_publishing_rule_cache_control_kwargs()
+        if getattr(rendered_page, "is_publishing_rule_page", False)
+        else get_default_cache_control_kwargs()
+    )
+    patch_cache_control(response, **cache_control_kwargs)
+    response["Cloudflare-CDN-Cache-Control"] = get_cdn_cache_control_header_value()
```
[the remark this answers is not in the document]
MaciekBaron: Cache purge already introduced, comment left before merge with `main` and now redundant.

BJacksonONS (approved):
LGTM, thanks @MaciekBaron 

Just so there's a record on this PR, it was decided by the business we want the headers on download links as queried here https://github.com/ONSdigital/dis-wagtail/pull/799#pullrequestreview-5088217794.

kevinhowbrook (approved):

MaciekBaron:
@MebinAbraham headers added to 404 pages [02b935c](https://github.com/ONSdigital/dis-wagtail/pull/799/commits/02b935c6e72a5b253c6d1ca67c6b9cc4c7c1aefd)

cms/core/whitenoise.py
```diff
@@ -25,6 +25,10 @@ class CMSWhiteNoiseMiddleware(WhiteNoiseMiddleware):
     The manifest file is hidden in the external environment to prevent fingerprinting. It is not filtered.
     """
 
+    # Immutable (hashed) static files are cached for 1 year, rather than whitenoise's
+    # 10 year default, per the static asset caching policy.
+    FOREVER = 31536000
```
MebinAbraham, line 30: Nit: Worth adding a small test for this. A future white noise change could regress the functionality without us noticing.

MebinAbraham (approved):

charlotteromaniuk (approved):

cms/core/whitenoise.py
MaciekBaron: Added explicit tests here: [9fb16ca](https://github.com/ONSdigital/dis-wagtail/pull/799/commits/9fb16cae84d4f737d0164828fd2873821fb7ea49)

MebinAbraham (approved):

sanjeevz3009 (approved):
LGTM!

PR #802, opened 2026-09-02
fix: missing ignore_required_constraints in StreamBlock.clean override

MebinAbraham:

Editing any page that contains a `DatasetStoryBlock` or `TimeSeriesPageStoryBlock` raised a `TypeError` on save/validation: TypeError: DatasetStoryBlock.clean() got an unexpected keyword argument 'ignore_required_constraints' Wagtail's form machinery (`FieldBlock._clean_bound_field` → `block.clean(...)`) calls `StreamBlock.clean()` with an `ignore_required_constraints` keyword argument. 

Our custom `clean()` overrides only accepted a single positional `value` argument, so the call blew up. This PR updates both overrides to accept `ignore_required_constraints` (defaulting to `False`) and forward it to `super().clean()`, preserving the parent's required-constraint handling and our existing duplicate-detection logic. 

These were the only two `StreamBlock` subclasses in the codebase overriding `clean()`; the other `clean()` overrides are on `StructBlock`/`FieldBlock`, whose signature is unaffected. Not caught by CI because mypy is configured with `ignore_missing_imports` and Wagtail ships no stubs, so the parent signature resolves to `Any` and the override-mismatch check can't fire. Pylint picked it up incorrectly as `unexpected-keyword-arg`. Regression tests are added to close that gap.

### How to review

- Ensure changes make sense.
- On main, manually test both a dataset story block and a time series story block, and confirm they trigger a server error.
- Repeat the same tests on this branch and confirm the server error no longer occurs.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

### DIFF
--- a/cms/datasets/blocks.py
+++ b/cms/datasets/blocks.py
@@ -68,8 +68,8 @@ class DatasetStoryBlock(StreamBlock):
         # version of the chosen edition rather than to the dataset series page.
         link_to_latest_version = False
 
-    def clean(self, value: StreamValue) -> StreamValue:
-        cleaned_value = super().clean(value)
+    def clean(self, value: StreamValue, ignore_required_constraints: bool = False) -> StreamValue:
+        cleaned_value = super().clean(value, ignore_required_constraints=ignore_required_constraints)
 
         # Validate there are no duplicate datasets, including between manual and looked up datasets
         # that point at the same place, however each was written. Destination resolution only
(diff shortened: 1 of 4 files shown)


BJacksonONS (approved):
LGTM

kevinhowbrook (approved):
Thanks @MebinAbraham - that's my mistake

sanjeevz3009 (approved):
LGTM

MebinAbraham:
@kevinhowbrook Nope, it's correct. User error on my part I wasn't on 7.4... 

PR not needed, behaviour is correct as of Wagtail 7.4. I'll take the approvals though 😄 ✅

PR #806, opened 2026-09-04
build(deps): upgrade django-treebeard to 5.3.1

kevinhowbrook:

Ticket [CMS-1356](https://officefornationalstatistics.atlassian.net/browse/CMS-1356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)

Upgrades `django-treebeard` from 4.8.0 to 5.3.1.

The version bump itself is two lines. The rest of the PR exists because a straight bump breaks 456 tests  with `Topic.DoesNotExist`.

We previously chose to hide a dummy root topic by filtering `depth__gt=1` in `TopicManager.get_queryset()`. Every real topic is a child of that. This PR doesn't alter that (although there is an argument to say we can and should now). The dummy root has to be visible to treebeard and hidden from the app, and the default manager can't do both.

Without the fixes we end up with  `_Root Topic → Topic → Subtopic` in the topic chooser.

#### WHY?!

From looking into it, the updates in Treebeard means that our custom manager is getting in the way and we cannot filter the depth out with it... because treebeard needs the full tree when doing things like add and change etc. See below:

V5 now does: 

```
        node = self.node_cls.objects.select_for_update().get(pk=self.node.pk)
```

The above will use our custom manager (that previously filtered depth and we need the full tree with the root node)

Previously v4 did this:

```
        if self.node_cls.node_order_by and not self.node.is_leaf():
```

So, more operations now use the ORM, which will use the custom manager we have defined.

### How to review

```bash
poetry install  # confirm treebeard 5.3.1
```

Check adding a topic to a page in the admin, the chooser is the only place topics are visible to an editor

- `_Root Topic` should **not** appear in a selectable topic row (in the chooser popup table)
- the Parent Topics column reads `About us → Transparency and governance`, not
  `_Root Topic → About us → Transparency and governance`

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

- **Remove the dummy root** (follow-up doc in confluence). `Topic.objects.all()` now includes the dummy root, so anything listing or counting topics must use  `Topic.objects.topics()`

[CMS-1356]: https://officefornationalstatistics.atlassian.net/browse/CMS-1356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cms/taxonomy/management/commands/sync_topics.py
+++ b/cms/taxonomy/management/commands/sync_topics.py
@@ -220,7 +220,9 @@ def _check_for_removed_topics(existing_topic_ids: set[str]) -> None:
 
 
 def _get_all_existing_topic_ids() -> set[str]:
-    return set(Topic.objects.values_list("id", flat=True))
+    # topics() rather than all(): the dummy root is never returned by the API, so including it here would
+    # make every sync decide it had been removed and flag it.
+    return set(Topic.objects.topics().values_list("id", flat=True))
 
 
 def _set_topic_as_removed(removed_topic_id: str) -> None:
--- a/cms/taxonomy/management/commands/sync_topics.py
+++ b/cms/taxonomy/management/commands/sync_topics.py
@@ -84,16 +84,15 @@ def _request_topics(url: str) -> list[dict[str, Any]]:
 
 
 def _extract_subtopic_links(raw_topics: Iterable[Mapping[str, Any]]) -> list[tuple[str, str | None]]:
-    """Return a list of tuples of any subtopic links and their parent topic IDs found in raw_topics."""
+    """Return a list of tuples of the subtopics URL and parent topic ID for each topic that has subtopics.
+
+    The subtopics URL is built from `TOPIC_API_BASE_URL` and the topic ID rather than following the
+    `links.subtopics.href` value from the response, so we always call our configured API host.
+    """
     return [
-        (subtopic_link, raw_topic["id"])
+        (f"{settings.TOPIC_API_BASE_URL}/{raw_topic['id']}/subtopics", raw_topic["id"])
         for raw_topic in raw_topics
-        # NOTE: This assumes we can simply call the links in the response as they are provided,
-        # perhaps switch to building the URL ourselves
-        if (
-            (subtopic_link := raw_topic.get("links", {}).get("subtopics", {}).get("href"))
-            and raw_topic.get("subtopics_ids")
-        )
+        if raw_topic.get("subtopics_ids")
     ]
 
 
--- a/cms/taxonomy/models.py
+++ b/cms/taxonomy/models.py
@@ -12,6 +12,9 @@ from wagtail.search import index
 
 from cms.core.db_router import force_write_db_for
 
+# The dummy root sits at depth 1 and real topics start below it, so every depth comparison in this module
+# is offset by one level that users never see.
+DUMMY_ROOT_DEPTH = 1
 BASE_TOPIC_DEPTH = 2
 
 if TYPE_CHECKING:
@@ -20,14 +23,25 @@ if TYPE_CHECKING:
 
 class TopicManager(MP_NodeManager):
     def get_queryset(self) -> TreeQuerySet:
-        """Filter out the dummy root topic from all querysets."""
+        """Return every row, dummy root included.
+
+        This must not filter. Treebeard resolves the tree through `cls.objects` by name, in roughly twenty
+        places, and offers no hook to point it at a different manager. Hiding the dummy root here makes it
+        invisible to add_child, move, get_root_nodes and the rest, which then fail with Topic.DoesNotExist.
+
+        Anything that shows or enumerates topics wants `topics()` below instead.
+        """
         # Reuse Wagtail's custom tree QuerySet for helpful utils
-        return TreeQuerySet(self.model, using=self._db, hints=self._hints).order_by("path").filter(depth__gt=1)
+        return TreeQuerySet(self.model, using=self._db, hints=self._hints).order_by("path")
+
+    def topics(self) -> TreeQuerySet:
+        """Return the real topics, excluding the dummy root."""
+        return self.get_queryset().filter(depth__gt=DUMMY_ROOT_DEPTH)
 
     def root_topic(self) -> Topic:
         """Return the dummy root topic."""
         # We create the dummy root in a migration so we know it will exist, so cast to "Topic" for mypy
-        return typing.cast(Topic, super().get_queryset().filter(depth=1).get())
+        return typing.cast(Topic, self.get_queryset().filter(depth=DUMMY_ROOT_DEPTH).get())
 
 
 # This is the main 'node' model, it inherits mp_node
@@ -95,7 +109,8 @@ class Topic(index.Indexed, MP_Node):
         """
         if self.depth == BASE_TOPIC_DEPTH:
             return self
-        return typing.cast("Topic", self.get_ancestors().first())
+        # Ancestors are ordered root to leaf, so without dropping the dummy root this would return it.
+        return typing.cast("Topic", self.get_topic_ancestors().first())
 
     def move(self, target: Topic | None = None, pos: str = "sorted-child") -> None:
         """Move the topic to underneath the target parent. If no target is passed, move it underneath our root."""
@@ -105,9 +120,18 @@ class Topic(index.Indexed, MP_Node):
     def __str__(self) -> str:
         return str(self.title)
 
+    def get_topic_ancestors(self) -> models.QuerySet[Topic]:
+        """Return the ancestors a user would recognise, without the dummy root.
+
+        get_ancestors goes through cls.objects, which cannot filter the dummy root out, so anything walking
+        up the tree for display has to drop it here instead.
+        """
+        # treebeard is untyped, so get_ancestors() comes back as Any
+        return typing.cast("models.QuerySet[Topic]", self.get_ancestors().filter(depth__gt=DUMMY_ROOT_DEPTH))
+
     @property
     def display_parent_topics(self) -> str:
-        if ancestors := [topic.title for topic in self.get_ancestors()]:
+        if ancestors := [topic.title for topic in self.get_topic_ancestors()]:
             return " → ".join(ancestors)
         return ""
 
@@ -117,7 +141,7 @@ class Topic(index.Indexed, MP_Node):
         Used for linking to search listing pages.
         """
         # Ancestors are ordered root to leaf.
-        ancestor_slugs = list(self.get_ancestors().values_list("slug", flat=True))
+        ancestor_slugs = list(self.get_topic_ancestors().values_list("slug", flat=True))
         return "/".join([*ancestor_slugs, self.slug])
 
 
--- a/cms/taxonomy/tests/test_management_commands.py
+++ b/cms/taxonomy/tests/test_management_commands.py
@@ -31,7 +31,7 @@ class SyncTopicsTests(TestCase):
 
         # Then
         self.mock_requests.get.assert_called_once()
-        self.assertEqual(Topic.objects.all().count(), 0, "Expect no topics to be saved")
+        self.assertEqual(Topic.objects.topics().count(), 0, "Expect no topics to be saved")
 
     def test_sync_empty_response(self):
         # Given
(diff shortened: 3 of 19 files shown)


cms/taxonomy/models.py
```diff
@@ -20,14 +23,25 @@
...
+
+        Anything that shows or enumerates topics wants `topics()` below instead.
+        """
         # Reuse Wagtail's custom tree QuerySet for helpful utils
-        return TreeQuerySet(self.model, using=self._db, hints=self._hints).order_by("path").filter(depth__gt=1)
+        return TreeQuerySet(self.model, using=self._db, hints=self._hints).order_by("path")
+
+    def topics(self) -> TreeQuerySet:
+        """Return the real topics, excluding the dummy root."""
+        return self.get_queryset().filter(depth__gt=DUMMY_ROOT_DEPTH)
```
kevinhowbrook, line 39: This re-work is basically how the tests now pass. the depth filter is removed from `.all()` calls, and instead moved to a new `.topics()` method. Upside is that `Topic.objects.all()` now reflects the table, dummy root included, so the root node system is visible when you use the ORM (rather than filtered).

Treebeard 5 resolves the tree through `self.node_cls.objects` [ref](https://github.com/django-treebeard/django-treebeard/blob/988f8551c349569fe92b1e7dbddcfab329297416/treebeard/mp_tree.py#L265) by name, so the depth filter can't live in`get_queryset()` any more: it would hide the dummy root from `add_child` and `move`.

cms/taxonomy/management/commands/sync_topics.py
```diff
@@ -220,7 +220,9 @@ def _check_for_removed_topics(existing_topic_ids: set[str]) -> None:
 
 
 def _get_all_existing_topic_ids() -> set[str]:
-    return set(Topic.objects.values_list("id", flat=True))
+    # topics() rather than all(): the dummy root is never returned by the API, so including it here would
+    # make every sync decide it had been removed and flag it.
+    return set(Topic.objects.topics().values_list("id", flat=True))
```
MebinAbraham, line 224: Q: Doesn't `all()` omit the root topic, given we have a custom manager which only returns depth > 1?

MebinAbraham: Right just seen you manager changes 👀

cms/taxonomy/models.py
MebinAbraham: > We previously chose to hide a dummy root topic by filtering depth__gt=1 in TopicManager.get_queryset(). Every real topic is a child of that. This PR doesn't alter that (although there is an argument to say we can and should now).

What are the options? Can you now have multiple "roots"?. What exactly changed between the previous and the new version that allows us to change, or was the existing change not necessary?

cms/taxonomy/management/commands/sync_topics.py
kevinhowbrook: Sorry, I'm still working on this (and understanding it)- It's just easier to run all the test in CI rather than locally (hence the draft) But FWIW Treebeard 5 resolves the tree through Topic.objects by name (here https://github.com/django-treebeard/django-treebeard/blob/988f8551c349569fe92b1e7dbddcfab329297416/treebeard/mp_tree.py#L265) - previously I think it used raw SQL, I'm still digging into it

cms/taxonomy/models.py
kevinhowbrook: I don't think the dummy root was strictly necessary no - as it stand the important changes down to things shifting to queryset methods e.g: 

v5 does this: https://github.com/django-treebeard/django-treebeard/blob/stable/5.x/treebeard/mp_tree.py#L265

v4 did this  https://github.com/django-treebeard/django-treebeard/blob/718ac15e4951b28b934b3c6fed4d607b786796a1/treebeard/mp_tree.py#L325 (4.8.x)

From what I looked at yesterday and the way we have topics built, all under a root node, it was to make moving and adding easier (smart call really). It makes sense in way as it mirrors how wagtail page tree is structured too. Looks to me like this update just breaks that setup. I spiked removing the root node for topics, and it's possible - but it will be a larger change (fixtures, tests, live topic paths etc) so I'm trying to keep this small as possible for the update itself

kevinhowbrook: (added a why section to the PR description to explain this)

cms/taxonomy/models.py
```diff
@@ -20,14 +23,25 @@
...
 
     def root_topic(self) -> Topic:
```
kevinhowbrook, line 41: TODO - consider using [get_first_root_node()](https://django-treebeard.readthedocs.io/en/stable/api.html#treebeard.models.NodeManager.get_first_root_node) instead, depends if it returns `DoesNotExist` or `None` though

kevinhowbrook: I don't think we should do this, it looks like this would fail with None and DoesNotExist is a better failure

BJacksonONS:
I might add a follow up card to see if we can clean up the `fix_tree` calls in some of the data loading scripts, as I think (but need to check) part of the problem was due to the dummy root being hidden from `objects`.

AWitcherONS:
`poetry install `does show Updating django-treebeard (4.8.0 -> 5.3.1)
also `poetry show | grep django-treebeard`
django-treebeard               5.3.1        ....
ran compose build without errors 
created new topic page I could select the Taxonomy directly but the search failed(I'm not sure if this is a problem or not)

joepercival (approved):
Tested functionally, looks good

cms/taxonomy/models.py
```diff
@@ -12,6 +12,9 @@
 
 from cms.core.db_router import force_write_db_for
 
+# The dummy root sits at depth 1 and real topics start below it, so every depth comparison in this module
+# is offset by one level that users never see.
+DUMMY_ROOT_DEPTH = 1
```
BJacksonONS, line 17: Suggestion: Given the comment maybe we can define `BASE_TOPIC_DEPTH = DUMMY_ROOT_DEPTH + 1`.

cms/taxonomy/models.py
BJacksonONS: This bit of docstring needs updating now.

cms/taxonomy/models.py
```diff
@@ -20,14 +23,25 @@
...
+        places, and offers no hook to point it at a different manager. Hiding the dummy root here makes it
+        invisible to add_child, move, get_root_nodes and the rest, which then fail with Topic.DoesNotExist.
+
+        Anything that shows or enumerates topics wants `topics()` below instead.
+        """
         # Reuse Wagtail's custom tree QuerySet for helpful utils
-        return TreeQuerySet(self.model, using=self._db, hints=self._hints).order_by("path").filter(depth__gt=1)
+        return TreeQuerySet(self.model, using=self._db, hints=self._hints).order_by("path")
+
+    def topics(self) -> TreeQuerySet:
```
BJacksonONS, line 46: Suggestion: Add this as a queryset method also, so it doesn't always need to be called first off the manager.

BJacksonONS (commented):
A couple of non-blocking suggestions + a docstring needs updating.

Also I confirmed the topic search not working reported by @AWitcherONS exists on main, so is not related to this PR. Possibly something related to `ENFORCE_EXCLUSIVE_TAXONOMY` as it works in sandbox.

cms/taxonomy/models.py
kevinhowbrook: 673dcfae3653b317da5b7ed66e6fdcd911390dc5

cms/taxonomy/models.py
```diff
@@ -21,6 +21,17 @@
     from django.db.models import BaseConstraint
 
 
+# Reuse Wagtail's custom tree QuerySet for helpful utils
```
kevinhowbrook: to remove

cms/taxonomy/models.py
kevinhowbrook: 7547939ee1e0963eeb192c59c0c7a5891b2f0092

cms/taxonomy/models.py
kevinhowbrook: f1d439174c1a49f829e1f787f7106b6c020fb3e4

cms/taxonomy/models.py
```diff
@@ -20,14 +23,25 @@
...
 class TopicManager(MP_NodeManager):
     def get_queryset(self) -> TreeQuerySet:
-        """Filter out the dummy root topic from all querysets."""
+        """Return every row, dummy root included.
+
+        This must not filter. Treebeard resolves the tree through `cls.objects` by name, in roughly twenty
+        places, and offers no hook to point it at a different manager. Hiding the dummy root here makes it
+        invisible to add_child, move, get_root_nodes and the rest, which then fail with Topic.DoesNotExist.
+
+        Anything that shows or enumerates topics wants `topics()` below instead.
```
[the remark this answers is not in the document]
BJacksonONS: Was addressed already in https://github.com/ONSdigital/dis-wagtail/commit/f1d439174c1a49f829e1f787f7106b6c020fb3e4

cms/taxonomy/tests/test_management_commands.py
```diff
@@ -391,7 +391,7 @@ def test_moved_subtopics(self):
 
         # Then
         self.assertEqual(self.mock_requests.get.call_count, 3, "Expect 2 calls to retrieve topics")
-        self.assertEqual(Topic.objects.all().count(), 4, "Expect 3 topics to be saved")
+        self.assertEqual(Topic.objects.topics().count(), 4, "Expect 3 topics to be saved")
```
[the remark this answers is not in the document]
kevinhowbrook: bf6176587e70d7e06f5c8eeeda8030ba8d6574ae

cms/taxonomy/viewsets.py
```diff
@@ -11,7 +11,9 @@ class TopicChooseViewMixin:
     model_class = Topic
 
     def get_object_list(self) -> QuerySet[Topic]:
-        return self.model_class.objects.all().order_by("path")  # type: ignore[no-any-return]
+        # topics() rather than all(): the dummy root is a real row and would otherwise be offered as a
+        # selectable "_Root Topic" in the chooser.
+        return self.model_class.objects.topics().order_by("path")  # type: ignore[no-any-return]
```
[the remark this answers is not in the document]
kevinhowbrook: ebc53861be86b674c3b54f620c206e48aad6f8ea

cms/taxonomy/models.py
```diff
@@ -20,14 +23,25 @@
...
+        """
         # Reuse Wagtail's custom tree QuerySet for helpful utils
-        return TreeQuerySet(self.model, using=self._db, hints=self._hints).order_by("path").filter(depth__gt=1)
+        return TreeQuerySet(self.model, using=self._db, hints=self._hints).order_by("path")
```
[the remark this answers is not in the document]
kevinhowbrook: I've looked into this, I don't think this is 'high' or even a security risk... it's probably worth addressing though as this PR does introduce it. Basically we don't guard against the root taxonomy being selected as an FK - the chooser just filters it out. We _could_ do this in the form validation layer but that does leave the ORM layer - and considering we sync taxononmies, it's a potential hole. It can be mitigated by the fact we may remove the root node in the future - but really I think we should handle this here. It will involve limiting FK choices at the DB level though and therefore; migrations.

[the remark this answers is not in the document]
kevinhowbrook: (for reviewers that pass by, here is what I've done to address this)

- test to prove the issue at the POST and ORM layers: af662f32cdc0b054c1db58839caff4538389d5e6, 
- fix by limiting choices, plus three AlterField migrations: f2c0dbe6b5b7e303ef709ace0c0838d48f604dc0,
- guard to ensure something fails if/when the dummy root gets removed  dbca26d95d41de6524a66c6a65b4e47c89768464

cms/taxonomy/management/commands/sync_topics.py
```diff
@@ -260,7 +260,9 @@ def _check_for_removed_topics(fetched_topic_ids: set[str]) -> None:
 
 
 def _get_active_topic_ids() -> set[str]:
-    return set(Topic.objects.filter(removed=False).values_list("id", flat=True))
+    # topics() rather than all(): the dummy root is never returned by the API, so including it here would
+    # make every sync decide it had been removed and flag it.
+    return set(Topic.objects.topics().filter(removed=False).values_list("id", flat=True))
```
kevinhowbrook, line 265: Added comment for clarity because this was a merge conflict

BJacksonONS (approved):
LGTM now, will make a follow up ticket around whether the `fix_tree` calls can be removed.

AWitcherONS (approved):
Unit tests, lint and format all pass
was able to create/review topic page (able to select a topic using manual find but search for taxonomy didn't work but this was expected) was able to view and approve
 Lgtm

PR #807, opened 2026-09-04
fix(table): render first column as header when option enabled

AWitcherONS:

Improvements to table rendering for release calendar page
Support for both row and column headers
While search shows only one explicit usage in release calendar, `BasicTableBlock` is exported and available for use throughout the application 

### How to review
create a release calendar page (Home ->  Release Calendar -> `release calendar page`)
In the pre-release section of the `release calendar page`
add a table  
then in the preview panel select the preview mode should be `Published`
then the headers can be changed by clicking the different options on table headers

### Deployment Safety

Please select one:

- [ ] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

---
Ticket: [DPD-3534](https://officefornationalstatistics.atlassian.net/browse/DPD-3534)

[DPD-3534]: https://officefornationalstatistics.atlassian.net/browse/DPD-3534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cms/core/blocks/markup.py
```diff
@@ -83,6 +83,14 @@ class Meta:
     def value_from_form(self, value: str) -> dict:
         return super().value_from_form(strip_unwanted_control_chars_from_json(value))  # type: ignore[no-any-return]
 
+    def _first_row_is_table_header(self, value: dict) -> bool:
+        """Return whether the first row should be rendered as table headers."""
+        return value.get("first_row_is_table_header", False) or value.get("table_header_choice") in {"row", "both"}
+
+    def _first_col_is_header(self, value: dict) -> bool:
+        """Return whether the first column should be rendered as row headers."""
+        return value.get("first_col_is_header", False) or value.get("table_header_choice") in {"column", "both"}
```
MaciekBaron: I think you intended to use this helper function below? At the moment both are unused.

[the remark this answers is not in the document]
MaciekBaron: Duplicate comment, resolving

https://github.com/ONSdigital/dis-wagtail/pull/807#discussion_r3933187308

cms/core/tests/test_blocks.py
```diff
@@ -606,6 +606,161 @@ def test_basictableblock__strip_control_characters(self):
...
+        ]
+        self.assertEqual(result, expected)
+
+    def test_get_rows_with_first_row_as_header(self):
+        """Test _get_rows when first_row_is_table_header is True."""
+        value = {
+            "data": [
+                ["R1C1", "R1C2", "R1C3"],
+                #   ["R2C1", "R2C2", "R2C3"],
+                #   ["R3C1", "R3C2", "R3C3"],
```
MaciekBaron: Code is commented out causing the test to fail

MaciekBaron (changes requested):
Left one comment.

Could you please also update your PR description:

- The description doesn't really reflect what the ticket asked for, it was mainly ensuring that the column renders as a header when necessary
- There is a stray "unit" there
- The "How to review" section is supposed to explain how to review your PR, not what you checked or tested
- The "follow up" section needs cleaning up (removing the template text, I don't understand what "ensure that this has not other pages" means); you mention "unit and functional tests for release calendar page" there, would this be covered by a separate ticket and PR?

cms/core/tests/test_blocks.py
```diff
@@ -606,6 +606,161 @@ def test_basictableblock__strip_control_characters(self):
...
+            "first_row_is_table_header": False,
+            "first_col_is_header": False,
+        }
+        result = self.block._get_rows(value)  # pylint: disable=W0212
+        expected = [
+            {
+                "tds": [
+                    {"value": "R1C1", "heading": False},
+                    {"value": "R1C2", "heading": False},
+                    {"value": "R1C3", "heading": False},
```
MaciekBaron: Sorry, realised after my initial comment that the commented out code wasn't the only reason the tests were failing: the result does not contain `"heading": False`. Your code only appends `"heading": True`.

`_get_rows()` code also does _not_ set `"heading": True` when `first_row_is_table_header` is true. It instead removes the first row, and the header is returned by `_get_header()`.

So you need to update all of your tests, so that the expected lists are correct.

cms/core/tests/test_blocks.py
```diff
@@ -606,6 +606,114 @@ def test_basictableblock__strip_control_characters(self):
         self.assertEqual(block.clean(value)["data"], [["unwanted", ""]] * len(UNWANTED_CONTROL_CHARACTERS))
 
 
+class BasicTableBlockGetRowsTestCase(TestCase):
+    """Test for BasicTableBlock._get_rows method."""
```
MaciekBaron: You updated the tests to check the generated context, which is great, I think that works better. However we should now update the comments as we are not just testing `_get_rows` now, but the context (`get_context` calls `_get_header` and `_get_rows`). So I would simply rename this to:
```suggestion
    """Test for BasicTableBlock.get_context method."""
```

and also update the comments in other methods.

MaciekBaron (approved):
LGTM but just left a comment regarding comments

MebinAbraham:
@AWitcherONS, should this still be marked as draft?

cms/core/tests/test_blocks.py
AWitcherONS: addressed in d52eb9c29787924ac065eb99b2da6f729b16c9f4

cms/core/blocks/markup.py
AWitcherONS: The helpers have been removed in commit e8beffc5c1249b064094f706e35af73a1d543b6b
neither were needed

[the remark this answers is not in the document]
AWitcherONS: addressed in e8beffc5c1249b064094f706e35af73a1d543b6b

cms/core/tests/test_blocks.py
AWitcherONS: the unit tests have been corrected in both 4c411e32d0e299a071cdb250653e320577b13084 and d52eb9c29787924ac065eb99b2da6f729b16c9f4

BJacksonONS (approved):

sanjeevz3009 (approved):

helenb (approved):
Functionally tested and double checked the DS options - LGTM

kevinhowbrook (approved):
lgtm, thanks 

Only thing I noticed when functionally testing this was that no headers still produce markup, perhaps not great for accessibility:

Image

But I don't think these changes introduced this :)

cms/core/tests/test_blocks.py
```diff
@@ -606,6 +606,111 @@ def test_basictableblock__strip_control_characters(self):
         self.assertEqual(block.clean(value)["data"], [["unwanted", ""]] * len(UNWANTED_CONTROL_CHARACTERS))
 
 
+class BasicTableBlockGetRowsTestCase(TestCase):
```
charlotteromaniuk: I think something like `BasicTableBlockContextTestCase` would be a clearer name here, since these tests mainly cover `get_context()` rather than just `_get_rows()`

AWitcherONS: Good spot 
resolved recommendation bd9eca599d48a6b905cdc6d4ad9f0682bd620a02

charlotteromaniuk (approved):
LGTM

MebinAbraham:
> lgtm, thanks
> 
> Only thing I noticed when functionally testing this was that no headers still produce markup, perhaps not great for accessibility:
> 
> Image
> But I don't think these changes introduced this :)

@kevinhowbrook
Could you capture this on a card please.

PR #808, opened 2026-09-04
feat(charts-api): add initial chart exporter API integration

MaciekBaron:

This PR is part of a stack, and addresses CMS-1286. It adds the initial Chart Exporter API integration by consuming the API. It also creates a data model which is used when rendering the images (see #812).

### How to review

* Consult [design-system-chart-exporter](https://github.com/ONSdigital/design-system-chart-exporter) and ensure that this implementation matches what the chart exporter API looks like
* Functional tests can be performed higher up the stack

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

Merge full stack

---
Ticket: [CMS-1286](https://officefornationalstatistics.atlassian.net/browse/CMS-1286)

[CMS-1286]: https://officefornationalstatistics.atlassian.net/browse/CMS-1286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cms/datavis/clients/chart_exporter.py
```diff
@@ -0,0 +1,145 @@
...
+                if attempt > self.max_retries:
+                    raise
+                backoff_seconds = 0.5 * (2 ** (attempt - 1))
+                logger.warning(
+                    "Retrying chart exporter API request (attempt %d/%d) after %.1fs",
+                    attempt,
+                    self.max_retries,
+                    backoff_seconds,
+                )
+                time.sleep(backoff_seconds)
```
MebinAbraham: Suggestion: Let's implement this natively. We also have a card in the backlog to create a reusable retryable session, so introducing it now seems like a good idea. We can then reuse it in other parts of the app in future work.

Im thinking something like:

```py
from collections.abc import Collection

import requests
from requests.adapters import HTTPAdapter
from urllib3.util.retry import Retry


def create_session(
    *,
    retries: int = 3,
    backoff_factor: float = 0.5,
    status_forcelist: Collection[int] = (429, 500, 502, 503, 504),
    allowed_methods: Collection[str] = ("GET", "HEAD", "OPTIONS"),
    respect_retry_after_header: bool = True,
) -> requests.Session:
    retry = Retry(
        total=retries,
        connect=retries,
        read=retries,
        status=retries,
        backoff_factor=backoff_factor,
        status_forcelist=status_forcelist,
        allowed_methods=allowed_methods,
        respect_retry_after_header=respect_retry_after_header,
    )

    adapter = HTTPAdapter(max_retries=retry)

    session = requests.Session()
    session.mount("https://", adapter)
    session.mount("http://", adapter)

    return session
```

zerolab: have you considered something like https://python-backoff.github.io/backoff/ ?

MebinAbraham: Using a package is the other option if we deem it appropriate; the native one historically was enough for our needs. That one seems quite new (the fork). If we were to use a package, I would suggest https://tenacity.readthedocs.io/en/latest/ for longevity.

MebinAbraham: Or https://github.com/hynek/stamina which is a wrapper on top of tenacity.

cms/datavis/clients/chart_exporter.py
```diff
@@ -0,0 +1,145 @@
...
+class ChartExporterClient:
+    """Client for the ONS Chart Exporter API."""
+
+    _MALFORMED_REQUEST_STATUSES = frozenset(
+        {HTTPStatus.BAD_REQUEST, HTTPStatus.REQUEST_ENTITY_TOO_LARGE, HTTPStatus.UNSUPPORTED_MEDIA_TYPE}
+    )
+    _RETRYABLE_STATUSES = frozenset({HTTPStatus.INTERNAL_SERVER_ERROR, HTTPStatus.SERVICE_UNAVAILABLE})
+
+    def __init__(self, *, base_url: str | None = None):
+        self.base_url = base_url or settings.CMS_CHART_EXPORTER_API_BASE_URL
```
BJacksonONS, line 53: Right now there's nothing that verifies this is a real url.

`CMS_CHART_EXPORTER_API_BASE_URL` defaults to an empty string if not in the environment so validation could happen there, but given we also want to potentially override that with `base_url` maybe here also makes sense?

MaciekBaron: I don't think we do checks like this for other URLs. The highlighted code is very similar to `BundleAPIClient` where we do the same thing. What did you have in mind to check?

BJacksonONS: If we're constructing the client then it seems like a good idea to check at least one of the URL vars is provided.

Right now if we don't specify the environment variable requests will throw a type of `RequestsException` which we treat as retryable even though it can never succeed. We could just short circuit all of the chart rendering logic at least if we don't have a valid API to target?

Given we don't know exactly when the chart exporter and this PR will be released we could potentially treat it's presence as part of the enabled feature flag maybe?

cms/datavis/clients/chart_exporter.py
MaciekBaron: Used stamina here: [3813f7e](https://github.com/ONSdigital/dis-wagtail/pull/808/commits/3813f7efff56c0cd5fbb381b5c47723ac863610d)

Note that I kept the "retries" concept rather than changing it to "attempts". Let me know if you prefer the latter.

cms/datavis/clients/chart_exporter.py
MaciekBaron: Fair, I added it here: [845c0a7](https://github.com/ONSdigital/dis-wagtail/pull/808/commits/845c0a754e52bc8a1923d8efe520ee93c146c00f)

Might need to add it to other API integrations for consistency (in a separate PR)

cms/datavis/clients/chart_exporter.py
```diff
@@ -0,0 +1,132 @@
...
+        self.session = requests.Session()
+        self.is_enabled = settings.CMS_CHART_EXPORTER_API_ENABLED
+        self.timeout = settings.CMS_CHART_EXPORTER_API_TIMEOUT_SECONDS
+        self.max_retries = settings.CMS_CHART_EXPORTER_API_MAX_RETRIES
+        self.session.headers.update({"Content-Type": "application/json", "Accept": "application/json"})
+
+    def create_chart(
+        self, chart_config: dict[str, Any], *, language: str = "en", device: str = "desktop"
```
kevinhowbrook, line 61: Looking at the chart explorer (as suggested in how to review) - `language` and `device` are `Literals` https://github.com/ONSdigital/design-system-chart-exporter/blob/c8c7dae30df7156aadc23ba93ace85eaf29d6f98/app/schemas/requests.py#L22-L23 is that right?

kevinhowbrook (commented):
(commenting whilst reviewing others in the stack)

cms/datavis/clients/chart_exporter.py
MaciekBaron: Yep that's right, currently those are the only options acceptable by the API.

However I thought to make it a bit more "future proofed" here by allowing to override it instead of hardcoding it in the request.

But this is a good point, happy to hear what other people think and change this if needed.

kevinhowbrook: That's fine - jest double checking it was that intentional :)

PR #809, opened 2026-09-04
fix(deps): downgrade dis-authorisation-client-js to ~1.1.0

MebinAbraham:

Downgrades `dis-authorisation-client-js` from `^1.2.0` to `~1.1.0` to avoid a bug introduced in 1.2. The tilde range allows 1.1.x patch releases while excluding 1.2.

### How to review

Ensure changes make sense

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

- Extends Cognito tests to cover the cases that are currently missed. The bug is a bit obscure, hence the complexity of adding a test

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.1.0",
       "license": "MIT",
       "dependencies": {
-        "dis-authorisation-client-js": "^1.2.0"
+        "dis-authorisation-client-js": "~1.1.0"
       },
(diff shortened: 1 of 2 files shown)


joepercival (approved):

kevinhowbrook (approved):

sanjeevz3009 (commented):
Reviewed!

PR #810, opened 2026-09-04
feat(chart-api): add RenderedChartImageChooserBlock for chart exporter API

MaciekBaron:

This PR is part of a stack and addresses CMS-1286. It adds a `RenderedChartImageChooserBlock` which is similar to previously implemented custom blocks which have hidden elements / tamper protection.

The chooser block exists so Wagtail's `ReferenceIndex` sees the page to image relationship.

`_publish_media` ([cms/private_media/signal_handlers.py](https://github.com/ONSdigital/dis-wagtail/blob/eaa1bb95de81eb3cf12df0892b0f1a3b76e03c98/cms/private_media/signal_handlers.py#L21)) calls `ReferenceIndex.get_references_for_object(page)`, filters to the model's content type, and runs `bulk_make_public` on what it finds.

### How to review

* Ensure that all tests pass
* Confirm that data cannot be modified by the user, and tamper protection works

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

Merge stack

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

---
Ticket: [CMS-1286](https://officefornationalstatistics.atlassian.net/browse/CMS-1286)

[CMS-1286]: https://officefornationalstatistics.atlassian.net/browse/CMS-1286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cms/articles/forms.py
+++ b/cms/articles/forms.py
@@ -4,7 +4,12 @@ from typing import Any
 from wagtail.blocks.stream_block import StreamValue
 from wagtail.models import Page
 
-from cms.core.forms import DeduplicateTopicsAdminForm, PageWithCorrectionsAdminForm, PageWithEquationsAdminForm
+from cms.core.forms import (
+    DeduplicateTopicsAdminForm,
+    PageWithCorrectionsAdminForm,
+    PageWithEquationsAdminForm,
+    PageWithProtectedChartImagesAdminForm,
+)
 
 
 class PageWithHeadlineFiguresAdminForm(DeduplicateTopicsAdminForm):
@@ -42,5 +47,9 @@ class PageWithHeadlineFiguresAdminForm(DeduplicateTopicsAdminForm):
 
 
 class StatisticalArticlePageAdminForm(
-    PageWithHeadlineFiguresAdminForm, PageWithCorrectionsAdminForm, PageWithEquationsAdminForm
-): ...
+    PageWithHeadlineFiguresAdminForm,
+    PageWithCorrectionsAdminForm,
+    PageWithProtectedChartImagesAdminForm,
+    PageWithEquationsAdminForm,
+):
+    protected_chart_image_fields = ("content", "featured_chart")
--- a/cms/core/forms.py
+++ b/cms/core/forms.py
@@ -242,7 +242,7 @@ class PageWithProtectedChartImagesAdminForm(WagtailAdminPageForm):
             raise ImproperlyConfigured(f"{cls.__name__} must define protected_chart_image_fields")
 
     def clean(self) -> dict[str, Any] | None:
-        cleaned_data = super().clean()
+        cleaned_data: dict[str, Any] | None = super().clean()
 
         for field_name in self.protected_chart_image_fields:
             if field_name not in self.cleaned_data:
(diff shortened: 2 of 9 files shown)


cms/articles/forms.py
```diff
@@ -42,5 +47,9 @@ def clean_headline_figures(self) -> StreamValue:
 
 
 class StatisticalArticlePageAdminForm(
-    PageWithHeadlineFiguresAdminForm, PageWithCorrectionsAdminForm, PageWithEquationsAdminForm
-): ...
+    PageWithHeadlineFiguresAdminForm,
+    PageWithCorrectionsAdminForm,
+    PageWithProtectedChartImagesAdminForm,
```
BJacksonONS, line 52: Methodology pages can have charts too, they'll need this also.

MaciekBaron: Thanks for spotting, added here: [04ac026](https://github.com/ONSdigital/dis-wagtail/pull/810/commits/04ac02615ce23881fa9e5ef9e5785d6bf4025934)

cms/core/forms.py
```diff
@@ -207,6 +213,51 @@ def clean_content(self) -> StreamValue:
         return content
 
 
+def _iter_chart_blocks(value: StreamValue | None) -> Iterator[StreamChild]:
+    """Recursively yield chart blocks from a StreamValue, including those nested in sections."""
```
kevinhowbrook, line 217: **nitpick** 

Probably worth mentioning the elif and why on line 223?

MaciekBaron: Could you explain a bit more how I can improve the existing comment? It mentions that it yields chart blocks, which is what happens on line 223.

Should I add a comment around line 222 to say "the block is section, iterate recursively"?

kevinhowbrook: I was more thinking about the condition on line 223.

I can see in the next pr in the stack you've removed this method though so guess it's moot

kevinhowbrook: ah not removed... renamed... my bad

I'm getting at  _why_ a section block is treated different;y from the others... is it just because we need the `content`?

MaciekBaron: Oh right - a section block contains other blocks, so we need to recursively call the function on its content.

Image

In practice, in most places all blocks at the top level will be section blocks, and section blocks can't be nested within section blocks. But this handles all scenarios including places where top level blocks are not sections.

Does that make sense? With this in mind, how should I modify the comment to make it clearer? I might be taking too many things for granted so a fresh pair of eyes will help!

kevinhowbrook: ah yea thanks - that helps... perhaps - I understand the latter part of the docstring now (including those nested in sections) :+1:

kevinhowbrook (approved):

PR #813, opened 2026-09-07
fix(release-calendar): keep release calendar pinned after filters are cleared

charlotteromaniuk:

Defect raised during testing of [CMS-1216](https://officefornationalstatistics.atlassian.net/browse/CMS-1216)

The `pin_release_calendar_page` hook now applies to both the homepage explorer view and the homepage explorer results view. It checks the request for active non-empty query values so the Release Calendar Index page is only pinned on the default explorer listing.

That means:
- the Release Calendar Index page is pinned by default
- it is not pinned when a user is filtering, searching, or sorting
- once a filter/search is cleared and the results request only contains empty values, pinning is applied again

Note: creating a custom viewset using [PageViewSet](https://docs.wagtail.org/en/stable/advanced_topics/customization/custom_page_listings.html#custom-page-listings) was considered here to replace the `construct_explorer_page_queryset` hook, but it felt like a heavier solution than this change really needs.

### How to review

1. In the home page explorer view, apply a filter or search for a page
2. Ensure the default ordering is applied (Release Calendar Index page is not pinned)
3. Remove the filter or search text
4. Check that the Release Calendar Index page is pinned
5. Ensure that when a user selects a specific order, the Release Calendar Index page is not pinned

### Deployment Safety

Please select one:

- [X] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

N/A

---
Ticket: [CMS-1245](https://officefornationalstatistics.atlassian.net/browse/CMS-1245)

[CMS-1216]: https://officefornationalstatistics.atlassian.net/browse/CMS-1216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CMS-1245]: https://officefornationalstatistics.atlassian.net/browse/CMS-1245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cms/release_calendar/tests/test_wagtail_hooks.py
+++ b/cms/release_calendar/tests/test_wagtail_hooks.py
@@ -85,11 +85,21 @@ class ReleaseCalendarHooksTestCase(WagtailTestUtils, TestCase):
         """Explorer results with an active filter should use Wagtail's default ordering."""
         response = self.client.get(
             reverse("wagtailadmin_explore_results", args=[self.home_page.id]),
-            {"content_type": str(self.older_topic_page.content_type_id)},
+            {
+                "content_type": [
+                    str(self.release_calendar_index.content_type_id),
+                    str(self.older_topic_page.content_type_id),
+                ]
+            },
         )
         pages = list(response.context["pages"])
 
-        self.assertNotEqual(pages[0], self.release_calendar_index, "Filtered results should not be pinned")
+        # Expect pages to be orderded by most recently updated
+        expected_page_order = [self.newer_topic_page, self.older_topic_page, self.release_calendar_index]
+
+        self.assertEqual(
+            pages, expected_page_order, "Release calendar index page should not be pinned in filtered results"
+        )
 
     def test_release_calendar_index_is_first_in_explorer_results_after_filter_is_cleared(self):
         """Explorer results with only blank filter values should pin the release calendar again."""
--- a/cms/release_calendar/tests/test_wagtail_hooks.py
+++ b/cms/release_calendar/tests/test_wagtail_hooks.py
@@ -1,14 +1,43 @@
 from types import SimpleNamespace
 
-from django.test import RequestFactory, TestCase
+from django.test import RequestFactory, SimpleTestCase, TestCase
 from django.urls import reverse
 from wagtail.models import Page
 from wagtail.test.utils import WagtailTestUtils
 
 from cms.home.models import HomePage
 from cms.release_calendar.models import ReleaseCalendarIndex
-from cms.release_calendar.wagtail_hooks import pin_release_calendar_page
+from cms.release_calendar.wagtail_hooks import _explorer_has_active_query, pin_release_calendar_page
 from cms.topics.tests.factories import TopicPageFactory
+
+# Mirrors the empty query params Wagtail keeps on the results view after filters are cleared.
+CLEARED_FILTER_QUERY = {
+    "q": "",
+    "latest_revision_created_at_from": "",
+    "latest_revision_created_at_to": "",
+    "has_child_pages": "",
+    "locale": "",
+}
+
+
+class ExplorerHasActiveQueryTests(SimpleTestCase):
+    def test_explorer_has_active_query_returns_false_for_blank_values(self):
+        """Blank explorer query values should not count as active query state."""
+        request = RequestFactory().get("/", CLEARED_FILTER_QUERY)
+
+        self.assertFalse(_explorer_has_active_query(request))
+
+    def test_explorer_has_active_query_ignores_fragment_refresh_and_pagination(self):
+        """Wagtail fragment refresh and pagination params should not count as active query state."""
+        request = RequestFactory().get("/", {"_w_filter_fragment": "1", "p": "2"})
+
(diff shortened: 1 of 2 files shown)


cms/release_calendar/tests/test_wagtail_hooks.py
```diff
@@ -1,15 +1,44 @@
...
 from cms.topics.tests.factories import TopicPageFactory
 
+# Mirrors the empty query params Wagtail keeps on the results view after filters are cleared.
+CLEARED_FILTER_QUERY = {
+    "q": "",
+    "latest_revision_created_at_from": "",
+    "latest_revision_created_at_to": "",
+    "has_child_pages": "",
+    "locale": "",
+}
```
MaciekBaron, line 20: This is only used in one of the tests - I'd move it there so it's less confusing. We can probably even inline it in the function call as the comment explains what's happening.

cms/release_calendar/wagtail_hooks.py
```diff
@@ -56,22 +56,38 @@ def hide_release_date_text_field_for_non_provisional_release_pages() -> str:
     return format_html('<script src="{}"></script>', static("js/hide-date-text-on-non-provisional-releases.js"))
 
 
+def _explorer_has_active_query(request: HttpRequest) -> bool:
+    """Return whether the explorer request has any non-empty user query values."""
+    # Ignore pagination and internal Wagtail fragment-refresh params which are left after a filter is removed.
+    ignored_query_keys = {"_w_filter_fragment", "p"}
```
BJacksonONS: Nit: this should probably be a const outside of the function.

cms/release_calendar/wagtail_hooks.py
```diff
@@ -56,22 +56,38 @@ def hide_release_date_text_field_for_non_provisional_release_pages() -> str:
...
-
+    """Pin the Release Calendar index to the top of the homepage explorer page."""
     # Only apply to the homepage Explorer view.
     resolver_match = getattr(request, "resolver_match", None)
-    is_homepage_explorer = getattr(resolver_match, "view_name", "") == "wagtailadmin_explore" and isinstance(
-        parent_page.specific_deferred, HomePage
-    )
+    is_homepage_explorer = getattr(resolver_match, "view_name", "") in {
+        "wagtailadmin_explore",
+        "wagtailadmin_explore_results",
```
MaciekBaron: This, on the other hand, I'd move to a named constant

MaciekBaron (commented):
Overall LGTM, just two minor stylistic comments, feel free to disagree with me

cms/release_calendar/tests/test_wagtail_hooks.py
```diff
@@ -52,6 +81,40 @@ def test_release_calendar_index_is_first_in_explorer_page(self):
...
 
+    def test_release_calendar_index_is_not_first_in_explorer_results_with_active_filter(self):
+        """Explorer results with an active filter should use Wagtail's default ordering."""
+        response = self.client.get(
+            reverse("wagtailadmin_explore_results", args=[self.home_page.id]),
+            {"content_type": str(self.older_topic_page.content_type_id)},
+        )
+        pages = list(response.context["pages"])
+
+        self.assertNotEqual(pages[0], self.release_calendar_index, "Filtered results should not be pinned")
```
BJacksonONS: Filtering on content type of topic page removes the release calendar page from the results entirely, rather than just unpinning, so this wouldn't fail even if we removed the hook changes.

I think a better test would be to include both topic page and release calendar in the filter, check it still exists in the result set but after the topic pages.

BJacksonONS (changes requested):

cms/release_calendar/tests/test_wagtail_hooks.py
charlotteromaniuk: Thanks for spotting, fixed here [d7a6138](https://github.com/ONSdigital/dis-wagtail/pull/813/commits/d7a61387db0369dc12a266b6becf082615e4b943)

BJacksonONS (approved):
Looks good, thanks @charlotteromaniuk

cms/release_calendar/tests/test_wagtail_hooks.py
charlotteromaniuk: This is used in two different tests so I thought it’d probably be best to separate out to avoid duplication

cms/release_calendar/wagtail_hooks.py
charlotteromaniuk: [c0b217b](https://github.com/ONSdigital/dis-wagtail/pull/813/commits/c0b217b271bb58ad0ea11d59f95b74f854cc42ea)

joepercival (approved):
LGTM other than existing comments :)

MaciekBaron (approved):

sanjeevz3009 (approved):
LGTM!

PR #815, opened 2026-09-08
fix(a11y): address release calendar accessibility audit

joepercival:

An audit of the Release Calendar page was carried out using Axe, WAVE and Lighthouse, alongside a visual review against the agreed designs.

The audit identified a mixture of technical, accessibility, visual and responsive implementation issues that should be addressed. This PR addresses the issues raised in the audit where possible. The audit can be found here: [Accessibility audit](https://officenationalstatistics-my.sharepoint.com/:x:/r/personal/joseph_nation_ons_gov_uk/Documents/RC_audit_UX.xlsx?d=w8b9fe08d5b8f425ab19b63bc4a86a2b8&csf=1&web=1&e=WgcYbI)

Ticket: [CMS-1396](https://officefornationalstatistics.atlassian.net/browse/CMS-1396?atlOrigin=eyJpIjoiNDY1MjBjZmFmOWYwNDYxZjhmYjk5ZGJhMmIzMWE3YjkiLCJwIjoiaiJ9)

### How to review

- Table of Contents anchor links
Confirm the 'Data' anchor link functions correctly in the table of contents.

-  Incorrect ordered list structure
An unexpected spacing `<div>` was inserted between the `<ol>` and its `<li>` elements:
`<ol><div class="spacing"><li>`
List items are now direct children of the ordered list:
`<ol><li>`
Confirm that this is the case for the 'changes to release date' blocks.

- Accreditation blocks – visual implementation
Confirm the following: 
Removal of the utility `ons-u-p-s class` from the `<div class="accreditations__block ">` block
`min-height: 200px;` added to the CSS class `"accreditations__block”`

- Accreditation blocks – mobile responsiveness
The accreditation blocks were not stacking appropriately at smaller/mobile viewport sizes.
In `<div class="accreditations">`, confirm that the following attributes are now added and function as expected:
`flex-direction:row;` for desktop and `flex-direction:column;` for mobile
`margin-bottom: 1em;`

- Design System consideration
The accreditation blocks used on the Release Calendar page do not have an equivalent component or documented pattern within the Design System. The implementation should follow existing Design System and UCD principles and be developed in a reusable way.
The resulting component can then be raised with the Design System team for consideration and potential inclusion where there is evidence of a wider reusable need.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

---
Ticket: [CMS-1396](https://officefornationalstatistics.atlassian.net/browse/CMS-1396)

[CMS-1396]: https://officefornationalstatistics.atlassian.net/browse/CMS-1396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cms/jinja2/templates/components/accredited/accredited-information.html
+++ b/cms/jinja2/templates/components/accredited/accredited-information.html
@@ -1,5 +1,5 @@
 {# appears in the body of the release page #}
-<div class="accreditations__block {% if not page.is_census %}accreditations__block--full-width{% endif %}">
+<div class="accreditations__block {% if not page.is_census %}accreditations__block--full-width{% endif %} ons-u-p-m">
     {% with accredited_logo_classes="accreditations__accredited-logo" %}
         {% include "templates/components/accredited/accredited-logo.html" %}
     {% endwith %}
--- a/cms/jinja2/templates/components/accredited/accredited-information.html
+++ b/cms/jinja2/templates/components/accredited/accredited-information.html
@@ -1,5 +1,5 @@
 {# appears in the body of the release page #}
-<div class="accreditations__block {% if not page.is_census %}accreditations__block--full-width{% endif %} ons-u-mb-s ons-u-p-s">
+<div class="accreditations__block {% if not page.is_census %}accreditations__block--full-width{% endif %} ons-u-mb-s">
     {% with accredited_logo_classes="accreditations__accredited-logo" %}
         {% include "templates/components/accredited/accredited-logo.html" %}
     {% endwith %}
--- a/cms/jinja2/templates/components/accredited/accredited-information.html
+++ b/cms/jinja2/templates/components/accredited/accredited-information.html
@@ -1,5 +1,5 @@
 {# appears in the body of the release page #}
-<div class="accreditations__block {% if not page.is_census %}accreditations__block--full-width{% endif %} ons-u-mb-s">
+<div class="accreditations__block {% if not page.is_census %}accreditations__block--full-width{% endif %}">
     {% with accredited_logo_classes="accreditations__accredited-logo" %}
         {% include "templates/components/accredited/accredited-logo.html" %}
     {% endwith %}
--- a/cms/jinja2/templates/pages/release_calendar/release_calendar_page--cancelled.html
+++ b/cms/jinja2/templates/pages/release_calendar/release_calendar_page--cancelled.html
@@ -45,7 +45,9 @@
             <h2>{{ _("Changes to this release date") }}</h2>
 
             <ol class="ons-list ons-list--bare">
-                {% include_block page.changes_to_release_date %}
+                {% for block in page.changes_to_release_date %}
+                    {% include_block block %}
+                {% endfor %}
             </ol>
         </section>
     {% endif %}
--- a/cms/jinja2/templates/pages/release_calendar/release_calendar_page--cancelled.html
+++ b/cms/jinja2/templates/pages/release_calendar/release_calendar_page--cancelled.html
@@ -44,7 +44,7 @@
         <section id="changes-to-release-date">
             <h2>{{ _("Changes to this release date") }}</h2>
 
-            <ol class="ons-list ons-list--bare spacing">
+            <ol class="ons-list ons-list--bare">
                 {% for block in page.changes_to_release_date %}
                     {% include_block block %}
                 {% endfor %}
--- a/cms/jinja2/templates/pages/release_calendar/release_calendar_page--cancelled.html
+++ b/cms/jinja2/templates/pages/release_calendar/release_calendar_page--cancelled.html
@@ -44,7 +44,7 @@
         <section id="changes-to-release-date">
             <h2>{{ _("Changes to this release date") }}</h2>
 
-            <ol class="ons-list ons-list--bare">
+            <ol class="ons-list ons-list--bare spacing">
                 {% for block in page.changes_to_release_date %}
                     {% include_block block %}
                 {% endfor %}
--- a/cms/jinja2/templates/pages/release_calendar/release_calendar_page--confirmed.html
+++ b/cms/jinja2/templates/pages/release_calendar/release_calendar_page--confirmed.html
@@ -42,7 +42,7 @@
         <section id="changes-to-release-date">
             <h2>{{ _("Changes to this release date") }}</h2>
 
-            <ol class="ons-list ons-list--bare spacing">
+            <ol class="ons-list ons-list--bare">
                 {% for block in page.changes_to_release_date %}
                     {% include_block block %}
                 {% endfor %}
--- a/cms/jinja2/templates/pages/release_calendar/release_calendar_page--confirmed.html
+++ b/cms/jinja2/templates/pages/release_calendar/release_calendar_page--confirmed.html
@@ -43,7 +43,9 @@
             <h2>{{ _("Changes to this release date") }}</h2>
 
             <ol class="ons-list ons-list--bare">
-                {% include_block page.changes_to_release_date %}
+                {% for block in page.changes_to_release_date %}
+                    {% include_block block %}
+                {% endfor %}
             </ol>
         </section>
     {% endif %}
--- a/cms/jinja2/templates/pages/release_calendar/release_calendar_page--confirmed.html
+++ b/cms/jinja2/templates/pages/release_calendar/release_calendar_page--confirmed.html
@@ -42,7 +42,7 @@
         <section id="changes-to-release-date">
             <h2>{{ _("Changes to this release date") }}</h2>
 
-            <ol class="ons-list ons-list--bare">
+            <ol class="ons-list ons-list--bare spacing">
                 {% for block in page.changes_to_release_date %}
                     {% include_block block %}
                 {% endfor %}
--- a/cms/jinja2/templates/pages/release_calendar/release_calendar_page.html
+++ b/cms/jinja2/templates/pages/release_calendar/release_calendar_page.html
@@ -107,7 +107,7 @@
                     <section id="changes-to-release-date" class="spacing">
                         <h2>{{ _("Changes to this release date") }}</h2>
 
-                        <ol class="ons-list ons-list--bare spacing">
+                        <ol class="ons-list ons-list--bare">
                             {% for block in page.changes_to_release_date %}
                                 {% include_block block %}
                             {% endfor %}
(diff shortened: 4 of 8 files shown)


cms/jinja2/templates/pages/release_calendar/release_calendar_page--cancelled.html
```diff
@@ -44,8 +44,10 @@
         <section id="changes-to-release-date">
             <h2>{{ _("Changes to this release date") }}</h2>
 
-            <ol class="ons-list ons-list--bare">
-                {% include_block page.changes_to_release_date %}
+            <ol class="ons-list ons-list--bare spacing">
```
helenb: The spacing class isn't working as expected here. It has a `last:child` rule which sets the bottom spacing to 0. But the spacing class is applied to the `ol` rather than the `li`. There's just one `ol` so the bottom spacing is always 0. There's also no gap between the `li` items. If you can move it to the `li` it should work. I'll add some screenshots of a page where I have more than one change to a release calendar page to make it clear.

Image
Image

cms/jinja2/templates/pages/release_calendar/release_calendar_page--confirmed.html
```diff
@@ -42,8 +42,10 @@
         <section id="changes-to-release-date">
             <h2>{{ _("Changes to this release date") }}</h2>
 
-            <ol class="ons-list ons-list--bare">
-                {% include_block page.changes_to_release_date %}
+            <ol class="ons-list ons-list--bare spacing">
```
helenb: I think this will have the same issue

cms/jinja2/templates/pages/release_calendar/release_calendar_page.html
```diff
@@ -107,8 +107,10 @@ <h2>{{ _("Data") }}</h2>
                     <section id="changes-to-release-date" class="spacing">
                         <h2>{{ _("Changes to this release date") }}</h2>
 
-                        <ol class="ons-list ons-list--bare">
-                            {% include_block page.changes_to_release_date %}
+                        <ol class="ons-list ons-list--bare spacing">
```
helenb: Same here

cms/static_src/sass/components/_accreditations.scss
```diff
@@ -4,6 +4,13 @@
     width: 100%;
     display: flex;
     gap: rem-sizing(24);
+    flex-direction: column;
```
helenb, line 7: This seems to be overridden when I test it by a rule in `accreditations__block--full-width` - might be worth checking when and why that override is applied.

Image

cms/static_src/sass/components/_accreditations.scss
```diff
@@ -12,6 +19,7 @@
         flex-direction: column;
         align-items: center;
         justify-content: center;
+        min-height: 200px;
```
helenb: Out of interest what was the reason for this minimum height? It might be better to use a value in `rems` here so that it adjusts to different zoom levels in the browser.

helenb (changes requested):
Hi @joepercival thanks for this. I've had a test and added a few comments on things that need changing.

cms/jinja2/templates/pages/release_calendar/release_calendar_page--cancelled.html
joepercival: Thanks, 350036e

cms/jinja2/templates/pages/release_calendar/release_calendar_page--confirmed.html
joepercival: 350036e

cms/jinja2/templates/pages/release_calendar/release_calendar_page.html
joepercival: 350036e

cms/static_src/sass/components/_accreditations.scss
joepercival: Good spot, should be fixed here 554d582

cms/static_src/sass/components/_accreditations.scss
joepercival: This was just implementing the request from the ticket: 

> Also need to add  min-height: 200px; to the CSS class "accreditations__block”

Happy to switch to `rems` if it's better though

cms/static_src/sass/components/_accreditations.scss
joepercival: Actually, should this always be flex-direction: column or always flex-direction: row?
 
Right now for the --full-width rule its contents have flex-direction: row for desktop and column for smaller devices

helenb: I've refreshed my memory about this. The accreditations block gets the `--full-width` class modifier when there is no census information displaying. If both are displayed then the census info and the accreditation logo display side by side, and the `--full-width` modifier is not applied.

### Options in the page editor for selecting accreditation and census options
Screenshot 2026-09-10 at 11 01 16

### Screenshot with both options (from `main`)
Screenshot 2026-09-10 at 11 03 19

The flex-direction: row should only be applied to the `--full-width` variant - the version without it never grows wide enough to need it and should always be column.

 The minimum height rule is introducing some oddities for the `--full-width` display at mobile. It can probably be resolved by ensuring the `justify-content: flex-start` rule only gets applied at medium.

Screenshot 2026-09-10 at 11 12 09

cms/static_src/sass/components/_accreditations.scss
helenb: I've double checked and the DS uses `rems` everywhere for `min-height` so we should definitely use `rems` here too.

joepercival: d95bbf1

cms/static_src/sass/components/_accreditations.scss
joepercival: b74d157

cms/static_src/sass/components/_accreditations.scss
```diff
@@ -12,12 +19,17 @@
         flex-direction: column;
         align-items: center;
         justify-content: center;
+        min-height: 12.5rem;
```
helenb: Super-nitpick: you could use the `rem-sizing` function here I think so we can see the px value in the code.

helenb (approved):
Thanks @joepercival all LGTM now apart from one minor nitpick

cms/static_src/sass/components/_accreditations.scss
joepercival: e239569

BJacksonONS: @helenb is that a design system thing? On all my old projects we always used to set it so 1rem = 10px to make the maths easy

helenb: I think the DS have a similar one but this is custom in the CMS. `rem` values are relative to the base font size, which is 16px unless it gets overridden - in most projects I've worked on it's kept at the base value to avoid other complications with font sizing.

helenb (approved):
Re-approving

cms/static_src/sass/components/_accreditations.scss
BJacksonONS: for a moment I thought I'd been missing some core CSS functionality for years 😂

kevinhowbrook (approved):

cms/static_src/sass/components/_accreditations.scss
```diff
@@ -4,6 +4,13 @@
     width: 100%;
     display: flex;
     gap: rem-sizing(24);
+    flex-direction: column;
+    margin-bottom: 1em;
+
+    @include media-query('m') {
+        // desktop only
```
MaciekBaron: Nit-pick: comment seems redundant really as it's just below the media query
```suggestion
```

MaciekBaron (approved):
LGTM, just one nit-pick

cms/jinja2/templates/components/accredited/accredited-information.html
```diff
@@ -1,5 +1,5 @@
 {# appears in the body of the release page #}
-<div class="accreditations__block {% if not page.is_census %}accreditations__block--full-width{% endif %} ons-u-mb-s ons-u-p-s">
+<div class="accreditations__block {% if not page.is_census %}accreditations__block--full-width{% endif %} ons-u-mb-s">
```
charlotteromaniuk: Compared to the attached design doc, the accreditation blocks don't align at the bottom. I think removing `ons-u-mb-s` here fixes it.

Image

joepercival: Thanks, e135991

cms/static_src/sass/components/_accreditations.scss
joepercival: True, I included it as I thought it wasn't immediately obvious that `m` was a breakpoint (i.e. it might look like sizes larger than `m` aren't included). But you're right it probably isn't needed: 2705793

helenb:
Screenshot 2026-09-11 at 10 40 45

 I just spotted another issue - ~it looks like the census block needs a similar fix we gave to the accreditation block when it has the `--full-width` variant - `justify-content: flex-start` is only needed up to the medium breakpoint.~ Edit: it might actually be a padding issue. The full width accreditation block also looks like the content close to the left edge at certain breakpoints.

Screenshot 2026-09-11 at 10 45 58

helenb (changes requested):
Changing review to request changes as there are still some styling issues

cms/jinja2/templates/components/accredited/accredited-information.html
```diff
@@ -1,5 +1,5 @@
 {# appears in the body of the release page #}
-<div class="accreditations__block {% if not page.is_census %}accreditations__block--full-width{% endif %} ons-u-mb-s ons-u-p-s">
+<div class="accreditations__block {% if not page.is_census %}accreditations__block--full-width{% endif %}">
```
helenb:
```suggestion
<div class="accreditations__block {% if not page.is_census %}accreditations__block--full-width{% endif %} ons-u-p-m">
```

helenb: This would also need applying to `census-information.html` It adds some padding both at the `--full-width` variant and the base style - from further testing I think both are needed.

helenb (changes requested):

cms/jinja2/templates/components/accredited/accredited-information.html
joepercival: Have added it here - 470c1af looks much better

helenb (approved):
Approving again :)

charlotteromaniuk (approved):
LGTM

sanjeevz3009 (approved):
LGTM!

PR #816, opened 2026-09-09
fix(taxonomy): skip already-removed topics in sync_topics

MebinAbraham:

`_check_for_removed_topics` in the `sync_topics` management command compared the topics returned by the API against **every** topic in the database, including those already marked as `removed=True`. As a result, every already-removed topic was re-saved (via `_set_topic_as_removed`) and re-logged as "removed" on every sync cycle, even though nothing had changed.

This PR makes the removal step idempotent by comparing the API response against only the **active** topics (`removed=False`):

### How to review

Ensure the changes make sense

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

---

### DIFF
--- a/cms/taxonomy/tests/test_management_commands.py
+++ b/cms/taxonomy/tests/test_management_commands.py
@@ -509,18 +509,34 @@ class SyncTopicsTests(TestCase):
         # When, then raises
         self.assertRaises(RuntimeError, sync_topics.Command().handle)
 
-    def test_already_removed_topic_not_reprocessed(self):
-        """A topic already marked as removed, and still absent from the API, should not be re-saved."""
-        # Given an already-removed topic that the API still does not return
-        TopicFactory(id="9999", title="Gone", removed=True, slug="gone")
-        self.mock_requests.get.return_value = mock_successful_json_response([])
+    def test_only_missing_topics_are_marked_as_removed(self):
+        """A topic should only be marked as removed when it is no longer returned by the API.
+
+        This uses three topics to cover:
+        - a topic that the API still returns -> should not be marked as removed,
+        - a topic that the API no longer returns -> should be marked as removed (once),
+        - a topic that was already removed and is still not returned -> should be left alone (not saved again).
+        """
+        # Given a topic that the API still returns (should be not be marked as removed)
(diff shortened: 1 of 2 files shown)


BJacksonONS (approved):
LGTM

cms/taxonomy/tests/test_management_commands.py
```diff
@@ -509,6 +509,19 @@ def test_duplicate_topic_id_in_responses(self):
...
+        # Given an already-removed topic that the API still does not return
+        TopicFactory(id="9999", title="Gone", removed=True, slug="gone")
+        self.mock_requests.get.return_value = mock_successful_json_response([])
+
+        # When
+        with patch.object(sync_topics, "_set_topic_as_removed") as mock_set_removed:
+            call_command("sync_topics")
+
+        # Then it should not be re-marked/re-saved
+        mock_set_removed.assert_not_called()
```
MaciekBaron: I don't think we have a test to check if this gets called when a topic is not removed. Perhaps we could add/update the test so that it uses a mix of removed and not removed topics, and then assert how many times it was called?

BJacksonONS: Do you mean if it is removed? `test_removed_topic` verifies that a topic is marked as removed if it wasn't already, and IMO probably better than mocking internal methods.

MaciekBaron: Unless I'm misreading the code, if `removed=False` then it gets picked up by `_get_active_topic_ids`. `_check_for_removed_topics` receives `[]` from the mocked request, `removed_topics` contains the item which has `removed=False` and `_set_topic_as_removed` gets called on it. So just wanted to ensure that if, say, we have two items A, B. A has `removed=True` and B has `removed=False`, `_set_topic_as_removed` gets called once on B.

MebinAbraham: Updated this test to cover a couple of scenarios. Tbh, as a result, it is a bit more complicated than the previous test and does have some overlap with existing tests. See what you think. I can always split it into multiple tests.

MaciekBaron (approved):

BJacksonONS (approved):

sanjeevz3009 (approved):
LGTM!

PR #817, opened 2026-09-09
feat(taxonomy): support internal env topic shape in sync_topics

MebinAbraham:

The Topic API returns topics in two different shapes depending on the environment, and the internal environment additionally requires authentication. This PR updates the `sync_topics` management command to support both, in a backward-compatible way.

**1. Support the internal-env response shape (`current` wrapping)**
- External env: each item in `items` **is** the topic, with fields at the top level.
- Internal env: each item wraps the topic under a `current` key (alongside `id` and `next`).

**2. Optional `SERVICE_AUTH_TOKEN` authentication**
- New `_get_auth_headers` helper reads `SERVICE_AUTH_TOKEN` from settings.
- Auth is **optional**: when the token is unset/empty (external env), no auth  headers are sent and requests remain unauthenticated (existing behaviour).
- When set (internal env), both the standard `Authorization: Bearer <token>` and the legacy `X-Florence-Token: Bearer <token>` headers are sent, mirroring the `sync_teams` command.

### How to review
 
Ensure the changes make sense.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

- Create temporary management to clean up "removed" topics and clean up lingering FK references.

---

### DIFF
--- a/cms/taxonomy/management/commands/sync_topics.py
+++ b/cms/taxonomy/management/commands/sync_topics.py
@@ -6,7 +6,9 @@ import requests
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.core.management import BaseCommand
+from django.core.management.base import CommandError
 
+from cms.auth.utils import get_service_auth_headers
 from cms.taxonomy.models import Topic
 
 logger = logging.getLogger(__name__)
@@ -15,9 +17,19 @@ logger = logging.getLogger(__name__)
 class Command(BaseCommand):
     """Topic Sync management command."""
 
+    service_auth_token: str | None = None
+    auth_enabled: bool = False
+
     def handle(self, *args: Any, **options: Any) -> None:
+        # When auth is enabled, a service token is mandatory
+        # When disabled, requests are made unauthenticated and no token is required.
+        self.auth_enabled = settings.CMS_TOPIC_SYNC_AUTH_ENABLED
+        self.service_auth_token = getattr(settings, "SERVICE_AUTH_TOKEN", None)
+        if self.auth_enabled and not self.service_auth_token:
+            raise CommandError("SERVICE_AUTH_TOKEN must be set when CMS_TOPIC_SYNC_AUTH_ENABLED is True")
+
         logger.info("Fetching topics from API...")
-        fetched_topics = _fetch_all_topics()
+        fetched_topics = self.fetch_all_topics()
         logger.info("Fetched topics", extra={"count": len(fetched_topics)})
 
         logger.info("Syncing topics...")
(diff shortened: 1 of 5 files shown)


cms/taxonomy/management/commands/sync_topics.py
BJacksonONS: Is this worth thinking about more now we're also sending credentials?

cms/taxonomy/management/commands/sync_topics.py
```diff
@@ -76,11 +76,47 @@ def _fetch_all_topics() -> list[dict[str, str]]:
...
+
+
+def _normalise_topic_item(item: Mapping[str, Any]) -> dict[str, Any]:
+    """Return the topic representation for an item.
+
+    Internal environments wrap the topic under a `current` key; external put the topic
+    fields at the top level. Unwrap `current` when present, otherwise use the item as-is.
+    """
+    current = item.get("current")
+    if isinstance(current, Mapping):
```
BJacksonONS, line 110: I'm not sure this will handle drafts correctly. It won't have a current field, and then will raise an error in `_validate_required_topic_fields`.

cms/taxonomy/management/commands/sync_topics.py
```diff
@@ -76,11 +76,47 @@ def _fetch_all_topics() -> list[dict[str, str]]:
...
+
+def _get_auth_headers() -> dict[str, str]:
+    """Get request headers for the topic API.
+
+    Authentication via `SERVICE_AUTH_TOKEN` is optional: when the setting is unset or empty,
+    no auth headers are sent and requests are made unauthenticated (external env).
+    When set (internal env), send both the standard `Authorization` header and the legacy
+    `X-Florence-Token` header, mirroring the teams sync.
+    """
+    service_auth_token = getattr(settings, "SERVICE_AUTH_TOKEN", None)
```
BJacksonONS: If we are confident the config will always be set up this way then I'm happy, but there is no meaningful guard against a misconfig here and could possibly end up sending to the public API if it was wrong.

cms/taxonomy/management/commands/sync_topics.py
MebinAbraham: This is being fixed in a follow up PR shortly as this assumption currently breaks sync in Staging.

cms/taxonomy/management/commands/sync_topics.py
MebinAbraham: `current` will be published topics, the `next` key holds "drafts" which we won't be using.

BJacksonONS: aah okay

BJacksonONS: For lack of better places to put it another Q here

Looks like the private API is using topics from the `next` key rather than current, is that going to be an issue for us here? What if a topic is removed from `next` but still used on `current` in the public API? Might not have dug into it in enough detail to fully understand now though.

https://github.com/ONSdigital/dp-topic-api/blob/3f66986a5bc759d06139bd96404645e335c83e86/api/topic_private.go#L92

cms/taxonomy/management/commands/sync_topics.py
MebinAbraham: https://github.com/ONSdigital/dis-wagtail/pull/820

cms/taxonomy/management/commands/sync_topics.py
MebinAbraham: Fair point but isn't that already true for all places we send a token? i.e. if we "accidentally" point to a public api / url, we risk sending auth to places we didn't plan to?. How are you suggesting we guard?. Maybe I'm missing something obvious.

BJacksonONS: I was thinking a gate to also check that the api url has been modified from the default which currently is prod public, but yeah like you say there's still technically nothing to stop someone adding the wrong one in their so it'll probably be down to review either way.

BJacksonONS: There is also the point I raised in the other comment that the public and private API don't have equivalent behaviour on `/topics/{id}/subtopics`, and we may want to toggle between the two based on more than the presence of the token which is needed for other syncs.

MebinAbraham: Will add a gate. I was thinking we could omit the service auth token, but our set-up means it is always configured (in helm), even if using the prod value; hence, in theory, it would send the token to our prod public URL.

cms/taxonomy/management/commands/sync_topics.py
MebinAbraham: Each topic document itself holds both current and next on the same record. "current" is what the public view always returns. But private response shows both keys.

Note: there's no active topic management right now; it's dev-led, and Wagtail is expected to take over as the topic API in future. Atm next/current should always match, right now we only care about "published" topic hence "current".

BJacksonONS: If we know there isn't any topic management happening atm then the point is moot, but I also think we both might be talking about different steps in the sync process.

MebinAbraham: Let's have a quick catch up to ensure we are aligned and there are no gaps.

cms/taxonomy/tests/test_management_commands.py
```diff
@@ -531,6 +531,81 @@ def test_duplicate_topic_id_in_responses(self):
...
+        topic = create_topic("1234")
+        self.mock_requests.get.return_value = mock_successful_json_response([build_topic_api_json(topic)])
+
+        # When
+        call_command("sync_topics")
+
+        # Then
+        self.mock_requests.get.assert_called_once()
+        _, kwargs = self.mock_requests.get.call_args
+        self.assertEqual(kwargs.get("headers"), {})
```
MaciekBaron: Nit: the same thing can be achieved with

```python
from unittest.mock import ANY
...
self.mock_requests.get.assert_called_once_with(ANY, headers={})
```

This works a bit better as it can assert both positional and keyword arguments.

MaciekBaron (approved):
Apart from the comments raised, and one nit from me, LGTM

cms/taxonomy/management/commands/sync_topics.py
MebinAbraham: [`0a2f3d7` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/817/changes/0a2f3d7e674bf1b0747d1a1725e7f12ca6c5cb9e)

cms/taxonomy/tests/test_management_commands.py
MebinAbraham: Good shout [`d2f5a84` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/817/changes/d2f5a84cf607fbc40eac249a6226320d1e80e63f)

cms/taxonomy/management/commands/sync_topics.py
BJacksonONS: Had a catchup, this is a non-blocker but @MebinAbraham has some checks to do in sandbox/staging to verify whether there might be inconsistencies in next/current.

MaciekBaron (approved):

BJacksonONS (approved):

BJacksonONS (approved):

sanjeevz3009 (approved):
LGTM!

PR #818, opened 2026-09-09
fix(csp): allow google analytics and video embed sources

charlotteromaniuk:

Addresses [CMS-1364](https://officefornationalstatistics.atlassian.net/browse/CMS-1364) and [CMS-1378
](https://officefornationalstatistics.atlassian.net/browse/CMS-1378)

- Adds CSP domains for GA and GTM preview mode as per https://developers.google.com/tag-platform/security/guides/csp
   - Preview mode is controlled by the feature flag `CMS_GTM_PREVIEW_MODE_ENABLED`. It is enabled by default in the internal env
- Adds CSP allowances for YouTube and Vimeo embeds
- Updates `Permissions-Policy` so encrypted-media and fullscreen are allowed for YouTube and Vimeo embeds, enabling normal playback behaviour and fullscreen support

### How to review

**Google Analytics and GTM:**
1. Confirm the CSP includes the required Google Analytics and GTM preview mode domains, excluding advertising-related domains: https://developers.google.com/tag-platform/security/guides/csp
2. Check that GTM preview mode sources are only added when `CMS_GTM_PREVIEW_MODE_ENABLED` is enabled, and are not always present for the external site.

**Video embeds:**
1. Add YouTube and Vimeo video embeds to a page and publish it
2. Ensure the embeds load correctly on the page
3. Verify that the embeds can play normally and enter fullscreen
4. Ensure there are no browser console errors when loading the page or when playing the embedded videos

### Deployment Safety

Please select one:

- [X] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

Needs to be back ported to v1.6.0-rc.1, v1.7.0-rc.1 and v1.8.0-rc.1 once merged

---
Ticket: [CMS-1364](https://officefornationalstatistics.atlassian.net/browse/CMS-1364)

[CMS-1364]: https://officefornationalstatistics.atlassian.net/browse/CMS-1364?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cms/settings/base.py
+++ b/cms/settings/base.py
@@ -1201,6 +1201,10 @@ WAGTAIL_POLYMATH = {
     "mathjax_sri": "sha512-M36RUChWzAh1veeenRZFql7HydLEnkYmoloiCvVrhz402UZgKI93qkV7SsaxtVKdN95Wzajh39ysrXCq34NTsg==",
 }
 
+GOOGLE_TAG_MANAGER_PREVIEW_MODE_ENABLED = (
+    not IS_EXTERNAL_ENV or env.get("GOOGLE_TAG_MANAGER_PREVIEW_MODE_ENABLED", "false").lower() == "true"
+)
+
 # Content Security policy settings
 # https://docs.djangoproject.com/en/6.0/ref/csp/
 static_sources = [ONS_CDN_URL]
@@ -1209,13 +1213,15 @@ SECURE_CSP: dict[str, list] = {
     "frame-src": [CSP.SELF, *IFRAME_VISUALISATION_CSP_SOURCES],
     # UNSAFE_INLINE is required by mathjax
     "style-src": [CSP.SELF, *static_sources, CSP.UNSAFE_INLINE, "*.hotjar.com"],
-    "img-src": [CSP.SELF, ONS_CDN_URL, "www.googletagmanager.com", "*.hotjar.com"],
+    "img-src": [CSP.SELF, ONS_CDN_URL, "www.googletagmanager.com", "*.google-analytics.com", "*.hotjar.com"],
     # UNSAFE_INLINE is required by hotjar
     "script-src": [CSP.SELF, *static_sources, "*.hotjar.com", "www.googletagmanager.com", CSP.UNSAFE_INLINE],
     "font-src": [CSP.SELF, *static_sources, "*.hotjar.com"],
     "connect-src": [
         CSP.SELF,
         "www.googletagmanager.com",
+        "*.google-analytics.com",
+        "*.analytics.google.com",
         "www.google.com",
         "*.hotjar.com",
         "*.hotjar.io",
@@ -1224,12 +1230,21 @@ SECURE_CSP: dict[str, list] = {
     "manifest-src": [CSP.SELF, ONS_CDN_URL],
     "frame-ancestors": [CSP.NONE if IS_EXTERNAL_ENV else CSP.SELF],
 }
-# Google Fonts are only needed for the Wagtail admin.
-# The external site loads fonts directly from the ONS CDN, so these CSP sources
-# should not be allowed there.
-if not IS_EXTERNAL_ENV:
+
+allow_google_fonts = not IS_EXTERNAL_ENV or GOOGLE_TAG_MANAGER_PREVIEW_MODE_ENABLED
+
+# Google Fonts are needed for the Wagtail admin and GTM Preview Mode.
+# The external site otherwise loads fonts directly from the ONS CDN, so these
+# sources should only be allowed when one of those cases applies.
+if allow_google_fonts:
     SECURE_CSP["style-src"].append("fonts.googleapis.com")
     SECURE_CSP["font-src"].append("fonts.gstatic.com")
+
+if GOOGLE_TAG_MANAGER_PREVIEW_MODE_ENABLED:
+    SECURE_CSP["script-src"].append("tagmanager.google.com")
+    SECURE_CSP["style-src"].extend(["www.googletagmanager.com", "tagmanager.google.com"])
+    SECURE_CSP["img-src"].extend(["ssl.gstatic.com", "www.gstatic.com"])
+    SECURE_CSP["font-src"].append("data:")
 
 if s3_custom_domain := env.get("AWS_S3_CUSTOM_DOMAIN"):
     SECURE_CSP["img-src"].append(f"https://{s3_custom_domain}")
--- a/cms/settings/base.py
+++ b/cms/settings/base.py
@@ -1205,12 +1205,14 @@ GOOGLE_TAG_MANAGER_PREVIEW_MODE_ENABLED = (
     not IS_EXTERNAL_ENV or env.get("GOOGLE_TAG_MANAGER_PREVIEW_MODE_ENABLED", "false").lower() == "true"
 )
 
+VIDEO_EMBED_CSP_SOURCES = ["www.youtube.com", "player.vimeo.com"]
+
 # Content Security policy settings
 # https://docs.djangoproject.com/en/6.0/ref/csp/
 static_sources = [ONS_CDN_URL]
 SECURE_CSP: dict[str, list] = {
     "default-src": [CSP.SELF],
-    "frame-src": [CSP.SELF, *IFRAME_VISUALISATION_CSP_SOURCES],
+    "frame-src": [CSP.SELF, *IFRAME_VISUALISATION_CSP_SOURCES, *VIDEO_EMBED_CSP_SOURCES],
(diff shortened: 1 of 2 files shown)


cms/settings/tests/test_settings.py
```diff
@@ -103,6 +135,31 @@ def test_gtm_csp(self):
...
                 self.assertIn("www.googletagmanager.com", self._get_csp_expressions(csp, "script-src"))
                 self.assertIn("www.google.com", self._get_csp_expressions(csp, "connect-src"))
+                self.assertIn("*.google-analytics.com", self._get_csp_expressions(csp, "img-src"))
+                self.assertIn("*.google-analytics.com", self._get_csp_expressions(csp, "connect-src"))
+                self.assertIn("*.analytics.google.com", self._get_csp_expressions(csp, "connect-src"))
+
+    def test_gtm_preview_mode_csp(self):
+        self.addCleanup(importlib.reload, base)
+
+        with mock.patch.dict(os.environ, {"GOOGLE_TAG_MANAGER_PREVIEW_MODE_ENABLED": "true"}, clear=False):
```
BJacksonONS: Would be good to explicitly set `IS_EXTERNAL_ENV` true in this case so we're definitely testing that the new env var is doing it's job here.

cms/settings/tests/test_settings.py
```diff
@@ -40,19 +40,42 @@ def test_internal_env_csp_settings(self):
...
             self.assertIn("fonts.googleapis.com", reloaded_base.SECURE_CSP["style-src"])
             self.assertIn("fonts.gstatic.com", reloaded_base.SECURE_CSP["font-src"])
+            self.assertIn("tagmanager.google.com", reloaded_base.SECURE_CSP["script-src"])
+            self.assertIn("data:", reloaded_base.SECURE_CSP["font-src"])
             self.assertEqual(reloaded_base.SECURE_CSP["frame-ancestors"], [CSP.SELF])
 
     def test_external_env_csp_settings(self):
         self.addCleanup(importlib.reload, base)
 
         with mock.patch.dict(os.environ, {"IS_EXTERNAL_ENV": "true"}, clear=False):
```
BJacksonONS: Here would be good to also patch the new env var to an explicit value so the test isn't affected by any changes to the default in the test environment.

BJacksonONS (commented):
Only a couple minor comments, otherwise looks good

cms/settings/tests/test_settings.py
charlotteromaniuk: [f3984a2](https://github.com/ONSdigital/dis-wagtail/pull/818/commits/f3984a2160d25bbfd820ccb441ec76ebf70a5c79)

cms/settings/tests/test_settings.py
charlotteromaniuk: [f3984a2](https://github.com/ONSdigital/dis-wagtail/pull/818/commits/f3984a2160d25bbfd820ccb441ec76ebf70a5c79)

BJacksonONS (approved):
LGTM, thanks @charlotteromaniuk

cms/settings/base.py
```diff
@@ -1201,21 +1201,30 @@ def _iam_database_config(host: str) -> dj_database_url.DBConfig:
     "mathjax_sri": "sha512-M36RUChWzAh1veeenRZFql7HydLEnkYmoloiCvVrhz402UZgKI93qkV7SsaxtVKdN95Wzajh39ysrXCq34NTsg==",
 }
 
+GOOGLE_TAG_MANAGER_PREVIEW_MODE_ENABLED = (
+    not IS_EXTERNAL_ENV or env.get("GOOGLE_TAG_MANAGER_PREVIEW_MODE_ENABLED", "false").lower() == "true"
+)
```
MebinAbraham: Will update the card. We should namespace the envs and also let's shorten it. We can accept the inconsistency as debt as we want to rename a lot of env vars at somepoint.

Can you update to `CMS_GTM_PREVIEW_MODE_ENABLED`

charlotteromaniuk: [88dcfac](https://github.com/ONSdigital/dis-wagtail/pull/818/commits/88dcfaca7b7dd37aab8968f08ffd4ad0ae764cfe)

sanjeevz3009 (approved):
Thanks @charlotteromaniuk

kevinhowbrook (approved):

cms/settings/base.py
```diff
@@ -1242,8 +1258,8 @@ def _iam_database_config(host: str) -> dj_database_url.DBConfig:
     "autoplay": [],
     "camera": [],
     "display-capture": [],
-    "encrypted-media": [],
-    "fullscreen": [],
+    "encrypted-media": VIDEO_EMBED_PERMISSIONS_POLICY_SOURCES,
```
MebinAbraham, line 1263: Q: Where has this come from?. Doesn't look like Firefox/Safari supports this: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy/encrypted-media#browser_compatibility

Have you tested the behaviour of this in the unsupported browsers?. Is it graceful?
If graceful, we can keep, worth sense checking.

MebinAbraham: Thinking about it, looks like we already specified it, hence "already" a problem / proven graceful; you just added an allowlist now.

Worth checking just incase.

cms/settings/base.py
```diff
@@ -1201,21 +1201,28 @@ def _iam_database_config(host: str) -> dj_database_url.DBConfig:
     "mathjax_sri": "sha512-M36RUChWzAh1veeenRZFql7HydLEnkYmoloiCvVrhz402UZgKI93qkV7SsaxtVKdN95Wzajh39ysrXCq34NTsg==",
 }
 
+CMS_GTM_PREVIEW_MODE_ENABLED = not IS_EXTERNAL_ENV or env.get("CMS_GTM_PREVIEW_MODE_ENABLED", "false").lower() == "true"
+
+VIDEO_EMBED_CSP_SOURCES = ["www.youtube.com", "player.vimeo.com"]
```
MebinAbraham, line 1208: It is worth adding a comment that, although we allow different variants such as `youtu.be` and `youtube.com` without www, we normalise the URL when calling `get_embed_url`.

cms/settings/base.py
```diff
@@ -1224,13 +1231,22 @@ def _iam_database_config(host: str) -> dj_database_url.DBConfig:
...
+
+# Google Fonts are needed for the Wagtail admin and GTM Preview Mode.
+# The external site otherwise loads fonts directly from the ONS CDN, so these
+# sources should only be allowed when one of those cases applies.
+if allow_google_fonts:
     SECURE_CSP["style-src"].append("fonts.googleapis.com")
     SECURE_CSP["font-src"].append("fonts.gstatic.com")
 
+if CMS_GTM_PREVIEW_MODE_ENABLED:
+    SECURE_CSP["script-src"].append("tagmanager.google.com")
```
MebinAbraham, line 1247: Any reason we are not using `script-src-elem` as per the docs?

MebinAbraham (approved):
Functionally looks good 👍

joepercival (approved):

cms/settings/base.py
charlotteromaniuk: Yes, I added this because DRM-protected Vimeo videos error and aren’t able to be played on Chrome and Edge without allowing `encrypted-media`. I’ve checked on Firefox and Safari now and it’s handled gracefully.

cms/settings/base.py
charlotteromaniuk: I added it to `script-src` because the policy doesn’t have a `script-src-elem` rule already, so script tags would fall back to `script-src` anyway. I thought keeping it in `script-src` would be a bit simpler and avoids introducing a second script rule. I'm happy to switch to `script-src-elem` though to stick with the docs.

cms/settings/base.py
charlotteromaniuk: [72afe8b](https://github.com/ONSdigital/dis-wagtail/pull/818/commits/72afe8b38918c9fcc30f6f2fd0fdacb60b21b5b7)

cms/settings/base.py
MebinAbraham: Great stuff 👍

cms/settings/base.py
MebinAbraham: It makes scripts a bit stricter, but since we didn't introduce it already, it's not a blocker for this, as otherwise you have to change existing behaviour.

MebinAbraham (approved):

PR #819, opened 2026-09-10
feat(taxonomy): exclude configured topic slugs from sync

MebinAbraham:

We want a way to keep certain top-level topics, and everything under them, out of the topic sync, starting with `census`. There is duplication (2 different ones in prod) and issues with the census topic structure, which is why it is being excluded for now. It may be included in future once product decisions have been made on its future.

This PR adds a `CMS_TOPIC_SYNC_EXCLUDED_SLUGS` setting (comma-separated env var, defaults to `census`). In `_fetch_all_topics` we drop any topic whose slug is excluded and skip queuing its subtopics, so the whole branch is left out. Set it to an empty string to disable.

`_check_for_removed_topics` was left as is, so an excluded topic that already exists in the DB is treated as missing and marked removed.

### How to review

- Ensure changes make sense
- Run `sync_topics` with different values for `CMS_TOPIC_SYNC_EXCLUDED_SLUGS ` and ensure the behaviour is as expected.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

---

### DIFF
--- a/cms/taxonomy/tests/test_management_commands.py
+++ b/cms/taxonomy/tests/test_management_commands.py
@@ -4,7 +4,7 @@ from unittest.mock import Mock, patch
 
 import requests
 from django.core.management import call_command
-from django.test import TestCase
+from django.test import TestCase, override_settings
 from requests import HTTPError
 
 from cms.taxonomy.management.commands import sync_topics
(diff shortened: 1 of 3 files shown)


cms/taxonomy/tests/test_management_commands.py
```diff
@@ -509,6 +509,79 @@ def test_duplicate_topic_id_in_responses(self):
         # When, then raises
         self.assertRaises(RuntimeError, sync_topics.Command().handle)
 
+    @override_settings(CMS_TOPIC_SYNC_EXCLUDED_SLUGS=["census"])
```
BJacksonONS: tiny nit: we pass a list here instead of a set

BJacksonONS (approved):

kevinhowbrook (approved):
(made a few edits to the PR description, management command and env var)

cms/taxonomy/tests/test_management_commands.py
MebinAbraham: [`e492af8` (this PR)](https://github.com/ONSdigital/dis-wagtail/pull/819/commits/e492af8c6351bb56ea0e31de5d232c1b4ab06641)

MaciekBaron (approved):

BJacksonONS (approved):

sanjeevz3009 (approved):
LGTM!

PR #821, opened 2026-09-11
build(deps): add npm package ecosystem to dependabot

joepercival:

We want Dependabot to scan our JavaScript packages in the same way that we are currently scanning our Python packages so that we can more easily and regularly keep them upgraded to their latest suitable versions.

Successfully forked and triggered a Dependabot scan from this branch: 
Screenshot 2026-09-15 at 14 06 03

### How to review

Check that the Dependabot configuration looks right. See [Configuring Dependabot version updates - GitHub Docs](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/configure-version-updates)

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

---
Ticket: [CMS-1358](https://officefornationalstatistics.atlassian.net/browse/CMS-1358)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -50,3 +50,29 @@ updates:
           - '*'
     cooldown:
       default-days: 7
+
+    # Enable security updates for npm packages
+  - package-ecosystem: npm
+
+    # When set to 0, version updates are disabled.
+    open-pull-requests-limit: 0
+
+    # Look for `package.json` and `lock` files in the `root` directory
+    directory: '/'
+    # Every weekday
+    schedule:
+      interval: daily
+
+    groups:
+      # Group updates into fewer pull requests
+      npm-security-updates:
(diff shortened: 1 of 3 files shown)


.github/dependabot.yml
```diff
@@ -50,3 +50,29 @@ updates:
           - '*'
     cooldown:
       default-days: 7
+
+    # Enable security updates for npm packages
+  - package-ecosystem: npm
+
+    # When set to 0, version updates are disabled.
+    open-pull-requests-limit: 0
```
MebinAbraham: We already get security updates, so let's enable this, maybe trial just limit of 1 or 2 for now.

We also want to exclude the major version for `dis-authorisation-client-js` / have it be a separate group.

PR #822, opened 2026-09-11
fix: check pages have unpublished changes before allowing bundle to save

joepercival:

If a user selects a page with unpublished changes in the bundle chooser, then that page is published before the bundle is saved, it still saves the bundle without an error.

This page adds validation for ~both the release calendar chooser and~ the page chooser in the bundle editor to ensure that all pages have unpublished changes befor the bundle can be saved. 

### How to review

- Create a bundle
- Create a page
- Go to the bundle editing page in one tab, select and confirm the page in the modal but don’t hit “Save Draft” or put the bundle to preview
- Go to the page, submit for moderation and move through the workflow stages to publish
- Return to the bundle and save
- Confirm that the bundle does not save and has validation errors
Confirm this works with both regular pages ~and release calendar pages (scheduling)~

**Updated**: This PR is no longer adding validation for the release calendar picker in bundles: we want to be able to use a published release calendar page without any unpublished content to be used for scheduling.

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

n/a

---
Ticket: [CMS-1268](https://officefornationalstatistics.atlassian.net/browse/CMS-1268)

[CMS-1268]: https://officefornationalstatistics.atlassian.net/browse/CMS-1268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cms/bundles/forms.py
+++ b/cms/bundles/forms.py
@@ -381,14 +381,6 @@ class BundleAdminForm(DeduplicateInlinePanelAdminForm):
                 if page.live and not page.has_unpublished_changes:
                     form.add_error("page", "This page has no unpublished changes")
 
-    def _validate_release_calendar_page_no_unpublished_changes(self) -> None:
-        release_calendar_page = self.cleaned_data.get("release_calendar_page")
-        if not release_calendar_page:
-            return
-
-        if release_calendar_page.live and not release_calendar_page.has_unpublished_changes:
-            self.add_error("release_calendar_page", "This page has no unpublished changes")
-
     def _validate_release_calendar_page_status(self) -> None:
         release_calendar_page = self.cleaned_data["release_calendar_page"]
         if not release_calendar_page:
@@ -496,7 +488,6 @@ class BundleAdminForm(DeduplicateInlinePanelAdminForm):
 
         # pages must have unpublished changes
         self._validate_bundled_pages_no_unpublished_changes()
-        self._validate_release_calendar_page_no_unpublished_changes()
 
         self._validate_bundled_pages()
 
--- a/cms/bundles/forms.py
+++ b/cms/bundles/forms.py
@@ -366,6 +366,29 @@ class BundleAdminForm(DeduplicateInlinePanelAdminForm):
                 f"page{pluralize(num_pages_not_ready)} not ready to be published."
             )
 
+    def _validate_bundled_pages_no_unpublished_changes(self) -> None:
+
+        formset = self.formsets.get("bundled_pages")
+        if not formset:
+            return
+
+        for form in formset.forms:
+            if not form.is_valid() or form.cleaned_data.get("DELETE"):
+                continue
+
+            if page := form.clean().get("page"):
+                page = page.specific
+                if page.live and not page.has_unpublished_changes:
+                    form.add_error("page", "This page has no unpublished changes")
+
+    def _validate_release_calendar_page_no_unpublished_changes(self) -> None:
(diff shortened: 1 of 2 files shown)


cms/bundles/tests/test_forms.py
```diff
@@ -79,6 +79,12 @@ def raw_form_data(self) -> dict[str, Any]:
             "bundled_datasets": inline_formset([]),
         }
 
+    def raw_form_data_with_unpublished_page_changes(self) -> dict[str, Any]:
+        """Returns form data with a bundled page that has an unpublished revision."""
```
MaciekBaron, line 83: Minor: I'm slightly worried about the readability of this because this method makes changes to the page, which might not be obvious. Perhaps the docstring could mention that explicitly for readability?

E.g. "Makes unpublished changes to the page and returns the resulting form data"?

MaciekBaron (approved):
LGTM, just one minor comment

cms/bundles/forms.py
```diff
@@ -366,6 +366,21 @@ def _validate_bundled_pages_status(self) -> None:
...
+            return
+
+        for form in formset.forms:
+            if not form.is_valid() or form.cleaned_data.get("DELETE"):
+                continue
+
+            if page := form.clean().get("page"):
+                page = page.specific
+                if page.live and not page.has_unpublished_changes:
+                    form.add_error("page", "This page has no unpublished changes")
```
MebinAbraham, line 382: Nit: how about we just extend `_validate_bundled_pages` instead, given we already have a method to check "conditions" on save?. Less code needed; precedent already set.

cms/bundles/tests/test_forms.py
joepercival: b24c627

cms/bundles/forms.py
joepercival: That's much better, 84d012c

cms/bundles/forms.py
```diff
@@ -338,6 +338,8 @@ def _validate_bundled_pages(self) -> None:
                     raise ValidationError(f"'{page}' is already in an active bundle ({page.active_bundle})")
                 if self.cleaned_data.get("release_calendar_page") == page:
                     raise ValidationError(f"'{page}' is already set as the Release Calendar page for this bundle.")
+                if page.live and not page.has_unpublished_changes:
```
charlotteromaniuk, line 341: I think the field level validation makes sense here but it’s now a bit inconsistent with the other validation errors in this method. I’m wondering if the above errors should also be field level errors?

charlotteromaniuk (approved):
Looks good to me, just one minor comment

cms/bundles/forms.py
joepercival: I agree it's better to keep them consistent, and using ValidationError is blocking, whereas with the field-level validation we can have multiple errors raised at the same time - 4050ba4 Screenshot 2026-09-23 at 10 29 03

charlotteromaniuk (approved):

BJacksonONS (approved):
LGTM, I like the field error changes.

This reminded me of this issue where the "Add to bundle" button appears on pages that we don't really want in the bundle. Should be a follow up, not this ticket but in the same area. Might raise as possibly something else to tackle this sprint.

Image

PR #826, opened 2026-09-14
fix(email): prevent sending bundle notifications if transaction rolls back

BJacksonONS:

Currently emails for bundle updates are triggered from an `on_save` signal handler. This means if any errors prevent the transaction from committing, the email is still sent.

This PR fixes this, such that the notification only goes out if the transaction commits and the bundle status is updated by adding the email send as a transaction commit callback rather than executing immediately.

### How to review

Introduce an artificial error mid transaction by adding this to the bottom of `cms/bundles/signal_handlers.py` and compare behaviour on main and this branch.

```python
@receiver(post_save, sender=Bundle)
def artificial_fail_after_email_hook(instance: Bundle, created: bool, **kwargs: Any) -> None:
    if instance.status == BundleStatus.IN_REVIEW:
        raise RuntimeError("Artificial failure after sending bundle in review email for testing purposes.")
```

Ensure the mail trap is running and you can see emails from the CMS.

1. Save a draft bundle with a page
2. Add a team and attempt to submit to review
3. On main you should see an email in the mail trap, on this branch you should not
4. Remove/comment out the snippet above and you should get the email on a successful update

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

There are some other bugs noted around errors occuring on this form that will be addressed in follow ups, such as form state being lost if an error is raised after `clean()`.

---
Ticket: [CMS-833](https://officefornationalstatistics.atlassian.net/browse/CMS-833)

[CMS-833]: https://officefornationalstatistics.atlassian.net/browse/CMS-833?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cms/bundles/tests/test_signal_handlers.py
```diff
@@ -151,10 +163,161 @@ def test_notification_sent_only_on_first_change_to_in_preview(self):
...
+            raise RuntimeError("Simulated error after bundle was saved")
+
+        return patch.object(BundleEditView, "save_instance", save_instance_then_fail)
+
+    def move_bundle_to_in_review(self):
+        with self.captureOnCommitCallbacks(execute=True):
+            self.bundle.status = BundleStatus.IN_REVIEW
+            self.bundle.save(update_fields=["status"])
+
+        mail.outbox = []
```
MaciekBaron: Minor - this is a side effect that isn't obvious from reading the method's name. Perhaps we could either rename the method or inline the clearing of the outbox in the test method?

MaciekBaron (changes requested):

cms/bundles/notifications/email.py
```diff
@@ -93,6 +95,16 @@ def _send_bundle_email(bundle: Bundle, team: Team, subject: str, email_template_
...
 
 
+def _send_bundle_email_on_commit(bundle: Bundle, team: Team, subject: str, email_template_name: str) -> None:
+    """Queue bundle email to go out after surrounding transaction commits successfully.
+
+    These emails can be triggered by signals fired mid-transaction, so we want to guard against
+    the emails sending if the transaction rolls back.
+    Outside a transaction this runs the callback immediately.
+    """
+    transaction.on_commit(partial(_send_bundle_email, bundle, team, subject, email_template_name))
```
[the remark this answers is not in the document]
BJacksonONS: I think the risk here is pretty low, as all the bundle info is coming from in-memory objects. There's a small risk of getting stale team membership if it changes within the replica lag, but added the decorator to `_send_bundle_email` https://github.com/ONSdigital/dis-wagtail/pull/826/commits/02b76511bb80fbb8d99d41bee1bc313f336638a3.

cms/bundles/tests/test_signal_handlers.py
BJacksonONS: Good point, we can just avoid running the callbacks in the first place so clearing the mailbox is redundant.

https://github.com/ONSdigital/dis-wagtail/pull/826/commits/caf183ef9f3715956d204150fac2cd30e61ad02e

kevinhowbrook (approved):

charlotteromaniuk (approved):

joepercival (approved):
Tested functionally, thanks for the great how to review steps too 👍

PR #834, opened 2026-09-15
feat: upgrade DS to 74.0.0-next.12

helenb:

Upgrades the DS to 74.0.0-next.12

- Apply changes necessary to ensure all text meets the reading width restrictions as detailed in the migration guide
- Remove some unused imports spotted when following the migration guide
- Ensure the search toggle aria label is translated - I happened to spot it wasn't when testing the DS changes. There is a [ticket in the backlog](https://officefornationalstatistics.atlassian.net/browse/CMS-1171) that includes further changes to the text here, but it seems like a win to ensure it is translated now even if we change the text down the line (and the changes in the ticket would need to be made on the DS side).

### How to review

- There should be no immediate visible changes to the text width in the CMS. If you want to see the impact of the change, you can temporarily edit [page_config_tags.py](https://github.com/ONSdigital/dis-wagtail/blob/main/cms/core/templatetags/page_config_tags.py#L222) and add `"wide": True,` to the `base_page_config` dict. You'll then see that text elements like paragraph text and headings are automatically capped at the reading width, whilst elements like images and charts fill the full width of the page.
- Check that the label for the search input is now translated correctly when on a Welsh page
- Check that the `aria-label` for the search toggle button is translated correctly when on a Welsh page

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions

---
Ticket: [CMS-1433](https://officefornationalstatistics.atlassian.net/browse/CMS-1433)

[CMS-1433]: https://officefornationalstatistics.atlassian.net/browse/CMS-1433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cms/locale/cy/LC_MESSAGES/django.po
```diff
@@ -173,18 +172,24 @@ msgstr "Dewislen"
 msgid "Toggle menu"
 msgstr "Togl dewislen"
 
-#: cms/core/templatetags/page_config_tags.py:240
+#: cms/core/templatetags/page_config_tags.py:237
+#, fuzzy
+#| msgid "Toggle menu"
+msgid "Toggle search"
+msgstr "Togl dewislen"
```
MaciekBaron: Is this the correct translation? As it says the same thing for "toggle menu".

cms/locale/cy/LC_MESSAGES/django.po
```diff
@@ -144,8 +144,7 @@ msgstr "Thema"
 msgid "Released"
 msgstr "Rhyddhawyd"
 
-#: cms/core/models/base.py:201
-#: megalinter-reports/updated_sources/cms/core/models/base.py:226
```
MaciekBaron, line 148: Oops can't believe we didn't spot this before 🙈

cms/locale/cy/LC_MESSAGES/django.po
helenb: These changes were automatic when I ran `makemessages` - maybe not accurate?

helenb: Google translate says it should be "Toglo'r chwiliad"

helenb: Now updated in d2dbee32b57b59c6aab02a29e22dd2340ef8bc2d

cms/locale/cy/LC_MESSAGES/django.po
```diff
@@ -173,18 +172,22 @@ msgstr "Dewislen"
 msgid "Toggle menu"
 msgstr "Togl dewislen"
 
-#: cms/core/templatetags/page_config_tags.py:240
+#: cms/core/templatetags/page_config_tags.py:237
+msgid "Toggle search"
+msgstr "Toglo'r chwiliad"
```
AWitcherONS, line 177: from lines 173-174 
msgid "Toggle menu"
msgstr "Togl dewislen"

I think it should be "Togl chwiliad"

AWitcherONS (commented):
Just a comment about the translation in the po
all else works ok

cms/locale/cy/LC_MESSAGES/django.po
BJacksonONS: Isn't "chwiliad" search? why would we update the "Toggle menu" text to that?

cms/core/tests/test_tags.py
```diff
@@ -368,7 +368,7 @@ def test_welsh_page_config(self):
 
         self.assertEqual(
             config["header"]["search"],
-            {"id": "search", "form": {"action": "/search", "inputName": "q"}},
+            {"id": "search", "toggleAriaLabel": "Toggle search", "form": {"action": "/search", "inputName": "q"}},
```
BJacksonONS, line 371: Should this test have `with translation.override('cy')` and assert on the welsh translation of "Toggle search"?

cms/core/templatetags/page_config_tags.py
BJacksonONS, line 59: This should be updated to include the new `toggleAriaLabel` key.

Maybe there's a follow up ticket here to ensure this would be flagged by linting tools in future.

BJacksonONS (changes requested):

cms/locale/cy/LC_MESSAGES/django.po
helenb: Ah maybe we should use "Togl chwiliad" here instead of "Toglo'r chwiliad" in order to be consistent with the "Toggle menu" translation - is that what you meant @AWitcherONS ?

helenb: Awaiting confirmation on this from a Welsh speaker.

cms/core/tests/test_tags.py
helenb: Addressed in 2699bedcb62694d09b53c6bc583cf51567776417

cms/core/templatetags/page_config_tags.py
helenb: Addressed in 2699bedcb62694d09b53c6bc583cf51567776417

sanjeevz3009 (approved):
Apart from the comments already made. LGTM!

cms/locale/cy/LC_MESSAGES/django.po
sanjeevz3009: Resolving on behalf!

PR #835, opened 2026-09-15
chore(makefile): update load-topics to include search index update

BJacksonONS:

Updates the search index after loading topics in make command, as `loaddata` bypasses this.

### How to review

1. Start docker stack with fresh database volumes
2. Run `make dev-init`
3. Login to admin panel
4. Try to create a topic page, and add a search when selecting a topic
5. Should return results

### Deployment Safety

Please select one:

- [x] Safe to auto-deploy
- [ ] Not safe to auto-deploy

### Follow-up Actions


AWitcherONS (approved):
build ok lint/format ok tests ok manual test amended the topic page local and choose another taxonomy Search Worked
LGTM

Makefile
```diff
@@ -114,6 +114,8 @@ load-design-system-templates:  ## Load the design system templates
 .PHONY: load-topics
 load-topics:  ## Load our fixture of taxonomy topics
 	poetry run python ./manage.py loaddata cms/taxonomy/fixtures/topics.json
+	poetry run python ./manage.py shell -c "from wagtail.search.backends import get_search_backend; from cms.taxonomy.models import Topic; \
+	get_search_backend().add_bulk(Topic, Topic.objects.all())"
```
MebinAbraham: Nit: `add_bulk` won't handle removed topics, i.e if we edited topics.json. Any reason not to just call update index?
```sh
poetry run python ./manage.py update_index
```
Yes, it does more than topics, but for a dev command, does it really matter?. ~Worth also adding to `delete-topics`.~

BJacksonONS: Fair, I wasn't sure if anyone would be annoyed if they had a lot of data locally and it took a while to update pages etc. also, but this is much simpler.

https://github.com/ONSdigital/dis-wagtail/pull/835/commits/95313c7bbb23f2fc18a70a4b98806afaf0263fc0

kevinhowbrook (approved):

Makefile
sanjeevz3009: I am happy with this @BJacksonONS

sanjeevz3009: Could revisit if devs feel otherwise.

sanjeevz3009 (approved):
Thanks @BJacksonONS Good stuff!

